PR #11, opened 2017-05-24
Add builder creation doc

### LINKED ISSUE #9: Document creating a new builder
A rough sketch to start a doc for creating a new builder (cc @willingc):

# So you want to create a new builder

*Creating a new builder for jupyter-repo2docker*

Repo2Docker uses [s2i][] to turn git repositories into docker images that:

1. can be used with jupyter notebooks, including via JupyterHub
2. include the contents of the repo

This is handled via a collection of s2i **builders**. There are two pieces to a builder:

1. the s2i builder image, which turns GitHub repos into docker images
2. the BuildPack object in repo2docker/detectors.py, which detects which builder to use based on the contents, and invokes the appropriate builder image

[s2i]: https://github.com/openshift/source-to-image/blob/master/docs/builder_image.md

## Making the s2i builder

First: pick a name. Use `mine`

To get started creating a builder, start with `s2i create`. In `s2i-builders`, run:

    s2i create jupyterhub/repo2docker-mine mine
    
To build the image, in the newly created `s2i-builders/mine`:

    make

Now, you can work on your Dockerfile as you would any other, and fill out the docker build, as you want. There are two pieces here:

1. the Dockerfile for the builder itself
2. the `s2i/bin/assemble` script

The Dockerfile is used to create the base builder image (e.g. installing the base runtime environment)

The `s2i/bin/assemble` script is run when creating each new image from a given repository (e.g. pulling in repo-specific dependencies and repo contents)

You can view the existing builders for examples to follow.

## Adding the BuildPack, so that your builder is used on the appropriate repos

Once your builder image is finished, you need to add a BuildPack to run your builds.

- define your BuildPack class in `detectors.py`. In most cases, you only need to implement two things if you subclass `S2IBuildPack`:
  -  the `detect` method to determine whether a repo should choose your BuildPack
  - setting the `build_image`  attribute, either via config or during the `.detect()` method

Finally, to get the builder application to use your BuildPack, add it to the `buildpacks` list in `app.py`.

Once everything above is done, you should be able to build a repo with:

    python -m repo2docker https://github.com/yourname/your-repo

willingc:
Closes #9 

I'm not completely happy with the text here, but I think it's good enough to iterate on.

### DIFF
--- a/builder.md
+++ b/builder.md
@@ -79,4 +79,6 @@ repos.
 3. Once everything above is done, you should be able to build a repo by
    specifying a source repo URL with this command:
 
-    python -m repo2docker https://github.com/yourname/your-repo
+   ```bash
+   python -m repo2docker https://github.com/yourname/your-repo
+   ```
--- a/builder.md
+++ b/builder.md
@@ -1,0 +1,82 @@
+# Creating a New Builder for jupyter-repo2docker
+
+## Overview
+
+Repo2Docker uses [s2i][] to turn a git repository into a Docker image. The
+image:
+
+    - may be used with Jupyter notebooks, including via JupyterHub
+    - includes the contents of the source git repo
+
+This process is handled via an [s2i][] **builder**. A builder contains two
+components:
+
+    1. the **s2i builder image**, which turns GitHub repos into docker images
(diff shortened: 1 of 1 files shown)


minrk:
Looks great, thanks!

PR #12, opened 2017-05-24
bundle s2i in wheels

minrk:
- downloads s2i binary on linux & darwin
- puts it in repo2docker/s2i
- prepends repo2docker dir to PATH when launching s2i (append so existing s2i on PATH gets higher priority?)

### DIFF
--- a/repo2docker/detectors.py
+++ b/repo2docker/detectors.py
@@ -75,7 +75,9 @@ class S2IBuildPack(BuildPack):
             output_image_spec,
         ]
         env = os.environ.copy()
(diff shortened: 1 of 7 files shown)


repo2docker/detectors.py
```diff
@@ -73,8 +74,10 @@ def s2i_build(self, workdir, ref, output_image_spec, build_image):
             build_image,
             output_image_spec,
         ]
+        env = os.environ.copy()
+        env['PATH'] = os.pathsep.join([here, env.get('PATH') or os.defpath])
```
yuvipanda: add a comment here about why we're doing this maybe?

PR #16, opened 2017-05-25
add LegacyBinderDockerfile buildpack

minrk:
only run on dockerfiles with `FROM andrewosh/binder-base`

- installs required versions of ipykernel, jupyterhub, notebook
- registers kernelspecs explicitly
- launches notebook server with python3 env
- removes nb_conda_kernels, which was causing trouble when not launched from the root env
- sets JUPYTER_PATH so kernelspecs and extensions that were being installed into the python 2 env should still be picked up

### DIFF
--- a/repo2docker/detectors.py
+++ b/repo2docker/detectors.py
@@ -45,7 +45,12 @@ class DockerBuildPack(BuildPack):
                 decode=True
         ):
             if 'stream' in progress:
-                self.log.info(progress['stream'], extra=dict(phase='building'))
+                if self.capture:
+                    self.log.info(progress['stream'], extra=dict(phase='building'))
+                else:
+                    sys.stdout.write(progress['stream'])
+
+_legacybinder_docker_appendix = """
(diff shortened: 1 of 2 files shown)


repo2docker/detectors.py
```diff
@@ -37,14 +38,56 @@ def detect(self, workdir):
...
                 decode=True
         ):
             if 'stream' in progress:
-                self.log.info(progress['stream'], extra=dict(phase='building'))
+                if self.capture:
+                    self.log.info(progress['stream'], extra=dict(phase='building'))
+                else:
+                    sys.stdout.write(progress['stream'])
+
+_legacy_binder_docker_appendix = """
```
yuvipanda: can you move this to be a traitlet under the Legacy Builder? keeps it contained and out of the global namespace of the file, + possibly allows overrides if needed...

minrk: good call, done.

PR #19, opened 2017-05-28
Switch LICENSE to match rest of Jupyter projects

yuvipanda:
The following people should say 'OK' or something like that here before we can merge this.

@minrk @willingc @choldgraf @yuvipanda


yuvipanda:
Ok!

willingc:
Thanks @yuvipanda

PR #26, opened 2017-06-02
Add initial sphinx docs

willingc:


willingc:
@choldgraf I've merged this but RTD's site is being unresponsive. Not sure if it is currently down.

PR #29, opened 2017-06-15
Add support for Julia buildpack

### LINKED ISSUE #23: Julia builds
We should support dependencies for Julia.

Major to-do items
---
- [x] Allow users to specify requirements with a `REQUIRE` file
- [x] Allow users to pre-compile packages before building the docker image

Julia handles dependencies in a text file similar to how Python does this. 

Thoughts on that? @yuvipanda

yuvipanda:
- Looks for REQUIRE file in a repo, assumes is Julia
- Appends content of REQUIRE file in repo to current existing
  REQUIRE file and then runs `Pkg.resolve()`
- Version pinned to v0.5.2 now - in the future, pick one
  based on presence of Julia spec in REQUIRE

Fixes #23


tkelman:
Julia versions in REQUIRE are lower bounds, either having multiple versions installed with a kernel dropdown in the notebook, or allowing users to choose when creating the container would be most useful

PR #47, opened 2017-07-30
Add a little more info to README

yuvipanda:


yuvipanda:
I'm going to self merge this now since it's better than what exists, but /ccing @choldgraf and @willingc! Should work on fleshing these out into actual documentation!

PR #51, opened 2017-08-01
adding doc generation from the tests

choldgraf:
This is a quick first-stab at auto-generating some documentation from our tests. It requires that there is a `README.rst` file inside any of the folders that we want to document. If it finds this file, then it adds a section that includes whatever is inside `README.rst` for that folder, and then shows the contents of the other files in that folder.

I added the readme file for a few sample repos so we can see how it looks. For some reason github isn't serving my website now so it's hard for me to show you, but here's a screenshot of what it looks like:

image


yuvipanda:
This is awesome!

PR #52, opened 2017-08-01
Make readmes required for tests

yuvipanda:
And add some

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,6 +8,7 @@ return a non-zero exit code for the test to be considered a
 success.
 """
 import pytest
+import os
 import subprocess
 import yaml
 import shlex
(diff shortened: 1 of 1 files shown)


choldgraf:
ah shit, I just realized that you added these :-/

choldgraf:
wanna take a look at my PR and pick which version of each you like more, and I can remove as necessary?

yuvipanda:
@choldgraf I rebased, and only added the commit that makes READMEs mandatory :D All my text was shit anyway :D

tests/conftest.py
```diff
@@ -28,6 +29,9 @@ def __init__(self, name, parent, path):
         self.path = path
 
     def runtest(self):
+        readme_path = os.path.join(self.path.dirname, 'README.rst')
```
minrk: Probably best to require `README*`, not .rst. People don't tend to write .rst READMEs these days:

```python
readmes = glob.glob(os.path.join(self.path.dirname, 'README*'))
if not readmes:
    raise ...

choldgraf: though the readme *will* need to actually be in rST syntax. We're using them to generate some documentation from these tests (for now anyway)

http://repo2docker.readthedocs.io/en/latest/

minrk: Fair enough. Though sphinx also works with .md files via [recommonmark](https://github.com/jupyterhub/jupyterhub/blob/0.7.2/docs/source/conf.py#L55). I'm always hesitant to make people write new .rst documents.

PR #53, opened 2017-08-01
updating autogen examples

choldgraf:
Round two! @yuvipanda


choldgraf:
also can you make me an admin of this repo so that I can do integrations w/ RTD etc?

PR #65, opened 2017-08-26
Add license - Issue #64

sumalaika:
License was missing

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2017, Project Jupyter Contributors
+All rights reserved.
+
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thank you for adding it! I've removed the AppVeyor integration since we don't have any tests there yet, and Travis passed - so going to merge it now!

yuvipanda:
Whoops, wrong button!

yuvipanda:
Actually, I just realized we already have a license specified as `COPYING.md` - https://github.com/jupyter/repo2docker/blob/master/COPYING.md which looks like the proper license too? @jzf2101 do you know what's the right thing to do?

PR #68, opened 2017-08-27
Add --debug & --no-build options

yuvipanda:
If --debug is passed, it'll print the Dockerfile generated. 

Note that  the Dockefile thus produced is *not* in any way usefu outside of debuggingl -
we do a bunch of path munging and adding scripts in here (such as install-miniconda).
If it isn't useful as is now, I don't think we should encourage people to use this to print 
Dockerfiles for basing their Dockerfiles on from here, and should just write that as docs
instead.

Either way, this is still very useful for Debugging.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -145,6 +145,12 @@ class Repo2Docker(Application):
         )
 
         argparser.add_argument(
+            '--print-dockerfile',
+            help="Print dockerfile contents to stdout",
+            action='store_true',
+        )
+
+        argparser.add_argument(
             '--no-run',
             dest='run',
             action='store_false',
@@ -156,6 +162,13 @@ class Repo2Docker(Application):
             dest='clean',
             action='store_false',
             help="Don't clean up remote checkouts after we are done"
+        )
+
+        argparser.add_argument(
+            '--no-build',
+            dest='build',
+            action='store_false',
+            help="Do not actually build the image. Useful in conjugation with --print-dockerfile."
         )
 
         argparser.add_argument(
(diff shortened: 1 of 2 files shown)


choldgraf:
lgtm - if you use it w/ no-build will it stop relatively quickly after printing the dockerfile? (or will it get buried quickly w/ a wall of log messages)

yuvipanda:
well, it'll build however long it takes to build if you use it *without* --no-build. If you use it with --no-build then the build step is skipped :)

yuvipanda:
I'm now owndering if we should even have --print-dockerfile here. None of these dockerfiles are actually going to build outside of repo2docker - we inject scripts, move the actual checkout into a 'src' directory, etc. We'll probably continue doing this kinda munging in the future :D 

I might just make it print Dockerfile output if loglevel is DEBUG, and leave --no-build in there. I am convinced you shouldn't be using repo2docker to 'bootstrap' a copy of your Dockerfile

choldgraf:
sounds good - you don't imagine that the dockerfile contents will be useful for _anything_ outside of debugging?

arokem:
If repo2docker is called with a Dockerfile as input, should this generate an identical output? I would expect this to have some calls to jupyter to start up the notebook and so on. But that's not what I've gotten when I tried running this.

yuvipanda:
Yup - with a Dockerfile we don't munge it at all. I'll work with Chris to
document what our expectations of Dockerfiles are tomorrow!

arokem:
Thanks! In particular, it's not clear to me right now what starts jupyter here? Is repo2docker itself starting jupyter? How does this then integrate with z2jh? Do we need to install repo2docker on all the machines in the cluster?? Or would kubernetes run jupyterhub for us in the right configuration?

yuvipanda:
@arokem kuberentes specifies the correct cmdline to start jupyter - it just requires that jupyterhub package is installed in the container image. There is no direct relation between z2jh and repo2docker - repo2docker is just an easy way to build a Docker image for use with z2jh *if* you don't have a Dockerfile. If you already have a Dockerfile, there is 0 advantage to using repo2docker with z2jh.

PR #74, opened 2017-09-07
don't install requirements.txt if environment.yml is present

minrk:
The most likely cause of both being present is equivalent dependencies for pip and conda users, not a single multi-stage conda environment.

If deps should be installed in conda with pip, they should be in `environment.yml` under `pip`.

### DIFF
--- a/repo2docker/detectors.py
+++ b/repo2docker/detectors.py
@@ -554,7 +554,7 @@ class CondaBuildPack(BuildPack):
                 conda clean -tipsy
                 """
             ))
-        if os.path.exists('requirements.txt'):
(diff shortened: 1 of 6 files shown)


repo2docker/detectors.py
```diff
@@ -554,7 +554,7 @@ def setup_assembly(self):
                 conda clean -tipsy
                 """
             ))
-        if os.path.exists('requirements.txt'):
+        elif os.path.exists('requirements.txt'):
```
yuvipanda: This would never be triggered I think?

minrk: It would not, good catch.

PR #76, opened 2017-09-07
make JUPYTERHUB_VERSION a build arg

minrk:
and respect it in all buildpacks

We probably want to eventually do the same for every other package with a hardcoded version (notebook, ipython, etc.).

In testing, I also noticed that Legacy Dockerfile isn't working at all right now because the dockerfile_appendix is ignored.

The new tests here revealed this and ought to now be fixed.

### DIFF
--- a/repo2docker/detectors.py
+++ b/repo2docker/detectors.py
@@ -149,8 +149,12 @@ class BuildPack(LoggingConfigurable):
         and the Hub itself should have the same version number.
         """
     )
+    @default('jupyterhub_version')
+    def _jupyterhub_version_default(self):
+        """Allow setting JUPYTERHUB_VERSION via env"""
+        return os.environ.get('JUPYTERHUB_VERSION') or '0.7.2'
+
     packages = Set(
-        set(),
         help="""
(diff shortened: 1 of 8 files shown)


repo2docker/detectors.py
```diff
@@ -136,6 +139,16 @@ class BuildPack(LoggingConfigurable):
     and there are *some* general guarantees of ordering.
 
     """
+
+    jupyterhub_version = Unicode(
+        '0.7.2',
+        config=True,
+        help="""JupyterHub version to install.
+
+        In general, the JupyterHub version in the image
```
choldgraf: Isn't this more than "in general"? I thought this was a hard requirement?

minrk: That depends (and depends on the role of repo2docker to some degree). Right now, repo2docker can be used to build and run standalone images locally, in which case jupyterhub version doesn't matter.

choldgraf: sounds good - just wanna make sure we're not leading people into the danger zone ;-)

PR #77, opened 2017-09-11
More forgiving rmtree

natbusa:
rmtree throws exceptions on a git repo checked out on a Windows filesystem. I believe that ignoring errors is ok here as app.py fully controls the checkout dir.


yuvipanda:
Thank you very much for the patch!

choldgraf:
yes! thanks so much! fixes for windows are especially appreciated since most of us don't run windows and we easily miss these bugs :-)

Issue #73, opened 2017-09-07
Provide a mechanism for ignoring environment files


minrk:
Since we are following mixed conventions of installation patterns (package.json, requirements.txt, etc.), we need to make sure that it's possible for repos to tell binder *not* to use some of these files, even if they are present.

Repos may have multiple of these files that should be considered mutually exclusive (e.g. requirements.txt and environment.yml to serve both pip and conda users, or `package.json` that's only part of a build step, and binder should ignore it, or a Dockerfile for standalone running/testing that shouldn't be used on binder). We should make sure that there is a way for users to tell binder what to do, regardless of the presence of top-level dependency files that binder should ignore.

This is related to #72, where the presence of a `/binder` directory or `binder.yml` file could prevent looking at any top-level files (for instance).

choldgraf:
I think we can close this now that we support `binder` folders, yeah?

willingc:
I'm going to close as binder-examples and #165 have done a bunch to improve this issue. @minrk please reopen if there is a next step that I've missed here. Thanks.

PR #83, opened 2017-10-02
Look in .binder directory for files

### LINKED ISSUE #72: Single binder directory support
We should support specifying binder environment information in a single location that is not used by anything else, a la `.travis.yml` (could be a file or directory), rather than requiring users to fill their repos with potentially several binder-specific files at the top-level that may conflict with files already there. Allowing these to be in top-level `binder.yml` or `binder/` directory would be a big improvement.

Packages may have requirements.txt and/or environment.yml, etc. that describe certain dependencies. To use the same repo on binder, the user may want to have *different* dependencies than the most basic required for the project. As it is now, we are forcing people to make binder-specific *repos* in order to be able to make these choices. It would be nice if this could be done with a single file and/or directory in an existing repo, to avoid conflicting with existing files.

minrk:
If .binder exists, top-level files are ignored. This allows for top-level Dockerfiles for other uses, but binder-specific environment.yml, etc.

closes #72, #73

### DIFF
--- a/repo2docker/detectors.py
+++ b/repo2docker/detectors.py
@@ -330,6 +330,13 @@ class BuildPack(LoggingConfigurable):
         result.components = (self, ) + self.components + (other, ) + other.components
         return result
 
+    def binder_path(self, path):
+        """Locate a file"""
+        if os.path.exists('.binder'):
+            return os.path.join('.binder', path)
+        else:
+            return path
+
     def detect(self):
         return all([p.detect() for p in self.components])
 
@@ -457,9 +464,10 @@ class BaseImage(BuildPack):
 
     @default('post_build_scripts')
     def setup_post_build_scripts(self):
-        if os.path.exists('postBuild'):
-            if stat.S_IXUSR & os.stat('postBuild')[stat.ST_MODE]:
-                return ['postBuild']
+        post_build = self.binder_path('postBuild')
+        if os.path.exists(post_build):
+            if stat.S_IXUSR & os.stat(post_build)[stat.ST_MODE]:
+                return [post_build]
         return []
 
 class PythonBuildPack(BuildPack):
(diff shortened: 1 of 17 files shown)


minrk:
Forgot to open this PR a while ago.

yuvipanda:
@minrk I think this is great! Can you add a few tests for it? Can merge after that!

choldgraf:
LMK when this is merged and I can put together a repo that uses this for an example

minrk:
Tested.

hydrosquall:
This is great, I needed to override `Dockerfile` with `requirements.txt` in mybinder.org. This solution works perfectly!

repo2docker/detectors.py
```diff
@@ -520,18 +528,18 @@ def setup_assembly(self):
...
         # surprise than requiring python2 to be requirements2.txt tho.
         try:
-            with open('runtime.txt') as f:
+            with open(self.binder_path('runtime.txt')) as f:
                 runtime = f.read().strip()
         except FileNotFoundError:
             runtime = 'python-3.5'
         if runtime == 'python-2.7':
-            requirements_file = 'requirements3.txt'
+            requirements_file = self.binder_path('requirements3.txt')
```
hydrosquall: This looks like a typo, python-2.7 leading to reading requirements3.txt ?

choldgraf:
@hydrosquall try it again with just `binder/` instead of `.binder/`! I think we decided not to make the folder hidden (though that is totally unclear from this PR)

repo2docker/detectors.py
betatim: Could you open an issue to discuss this? I also think this is a bit puzzling/should have a comment explain the thinking

hydrosquall:
@choldgraf : ah, excellent! Yup, changing the folder name straightened out the issue. 

https://mybinder.org/v2/gh/hydrosquall/pandas-cookbook/8794bf41efe3279703d5a3bd280dd69facba3f5e

choldgraf:
yahoo!

PR #85, opened 2017-10-04
raise on non-executable postBuild

minrk:
If the permissions are wrong the build will fail, which is more informative than silently ignoring the file.

cf https://github.com/jupyterhub/binderhub/issues/155 which this would have caught more promptly.


choldgraf:
why not just raise an error here that says `postBuild` needs to be executable?

minrk:
Much better!

PR #86, opened 2017-10-04
make binder directory visible

minrk:
use 'binder' instead of hidden '.binder'

cf https://github.com/jupyterhub/binderhub/issues/157


yuvipanda:
woohooo (shall merge once CI passes)

PR #90, opened 2017-10-06
Updating postBuild docs + doc build script

choldgraf:
This does 2 things:

1. Fixes a bug in the doc build script where it would build two versions of some folders.
2. Adds instructions for making `postBuild` executable and adds a link to the docs in the error message that comes up when this happens. The reason for this is that making a file executable with windows is pretty idiosyncratic because you have to do it with git...this preempts some confusion by giving people a path forward.
3. Skips the `binder` folder examples in the auto-generated page, and instead added a note at the top explaining what will happen if a `binder` folder exists in the repo. That way this page is just about looking at the syntax of the build files.

LMKWYT


tests/venv/postBuild/README.rst
```diff
@@ -5,3 +5,10 @@ It is possible to run scripts after you've built the environment specified in
 your other files. This could be used to, for example, download data or run
 some configuration scripts. For example, this will download and install a
 Jupyter extension.
+
+.. note::
+
+   This file needs to be executable in order to work with ``repo2docker``. The
+   easiest way to do this is to run the following command with ``git``::
+
+       git update-index --chmod=+x postBuild
```
yuvipanda: Is... this the easiest way to do it? Isn't that just 'chmod +x postBuild'? What does update-index do? I've never encountered it before...

choldgraf: Def not the easiest, but I don't think chmod exists in windows, so this is a command that (I think) should work on any platform. I could expand the text to say this

yuvipanda: Please do! I'd prefer we just said 'on OS X and Linux, do `chmod +x postBuild` and on Windows do git <blah>' rather than just have one blanket unexplained command.

choldgraf: Sounds good. I guess I just think of this command as the chmod version of what "git mv" is to "mv"

choldgraf:
@yuvipanda how's that?

tests/venv/postBuild/README.rst
```diff
@@ -5,3 +5,15 @@ It is possible to run scripts after you've built the environment specified in
 your other files. This could be used to, for example, download data or run
 some configuration scripts. For example, this will download and install a
 Jupyter extension.
+
+.. note::
+
+   This file needs to be executable in order to work with ``repo2docker``. If
+   you're on a UNIX operating system, run::
```
yuvipanda: Can we explicitly say 'macOS or Linux'? A lot of people running macOS probably don't think of their computers as Unix... Sorry to be nitpicky!

choldgraf: how bout now?

choldgraf:
ugh god damnit, yeah...hold on lemme fix that

choldgraf:
ok that work?

yuvipanda:
Yup!

Issue #91, opened 2017-10-06
When passed --json-logs, all output *must* be JSON


yuvipanda:
Currently, this is not the case - exceptions get thrown as normal, bypassing the JSON mechanism. We should capture all exceptions that aren't being caught otherwise and log them with phase failed.

yuvipanda:
Fixed in #97

PR #97, opened 2017-10-17
avoid producing non-JSON output on errors

minrk:
register a custom excepthook that is called instead of the default traceback formatter. This produces a phase: failed message with the error.

This fixes the primary, controllable cause of #91

It's not entirely possible to rigorously ensure that *all* output is JSON from within the process itself, though.
Warnings etc. or debug messages that might come from other libraries will not be captured.
The only way to ensure truly rigorous log output is at one level higher than the repo2docker process itself.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -179,6 +179,17 @@ class Repo2Docker(Application):
 
         return argparser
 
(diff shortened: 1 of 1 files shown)


yuvipanda:
ty!

Issue #98, opened 2017-10-17
Upgrade to jupyterlab 0.28.0


gnestor:
If I define `jupyterlab>=0.28.0` in the `requirements.txt` of my own repo, will it install 0.28.0 or the version defined in repo2docker?

choldgraf:
juts to answer your question - it should update the jupyterlab version if you put it in requirements.txt

PR #99, opened 2017-10-17
Upgrade to jupyterlab 0.28.0

gnestor:
Closes https://github.com/jupyter/repo2docker/issues/98


gnestor:
Thanks @willingc!

choldgraf:
This is starting to be the primary way that I learn about new jupyterlab releases haha

I believe that the beta deployment is now upgraded so I'll merge!

PR #101, opened 2017-10-19
Remove intermediate containers in failed builds

yuvipanda:
Fixes  https://github.com/jupyterhub/binderhub/issues/200


betatim:
Should we also remove intermediates for the other `build()` call?

yuvipanda:
@betatim good catch! we should. do you wanna make a PR?

PR #104, opened 2017-10-20
Only parse `REQUIRE` file in `binder` (if exists)

darwindarak:
If the `binder` folder is in a package directory, then the existing setup will call `using` on the package requirements instead of `binder/REQUIRE`.  This can run into some issues if the package requires a specific version of Julia:

```
ERROR: ArgumentError: Module julia not found in current path.
Run `Pkg.add("julia")` to install the julia package.
```


yuvipanda:
This is awesome! ty :)

yuvipanda:
Congratulations on your first patch getting merged! Hopefully one of many more to come to help us solidify our Julia support (and elsewhere too :D)

PR #106, opened 2017-10-23
Replace theme with alabaster

willingc:
For better display on RTD, let's change the theme to the default, generic alabaster theme instead of the jupyter theme. cc/ @Carreau @choldgraf

### DIFF
--- a/docs/doc-requirements.txt
+++ b/docs/doc-requirements.txt
@@ -1,3 +1,2 @@
 sphinx>=1.4, !=1.5.4
-jupyter_alabaster_theme
 recommonmark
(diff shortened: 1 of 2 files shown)


choldgraf:
agree this is a good decision!

as it's design and that affects the whole site, I'll let somebody else (maybe @yuvipanda or @Carreau weigh in before merging)

yuvipanda:
I approve!

willingc:
Going to merge since I don't see @Carreau objecting ;-)

Carreau:
I object to not merging :-) I should be able to start pitching in from a
computer later today.

PR #107, opened 2017-10-23
reorganizing arguments

choldgraf:
This reorganizes a couple of the docstring parameters so that they more naturally follow linear time. LMK if you think there is a more intuitive ordering here. I saw that @ctb was running `--no-run` assuming it was `--no-build`, so hopefully this will catch some of those usage patterns.

This assumes the order of operations is:

build -> cmd -> run -> push

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -152,7 +152,7 @@ class Repo2Docker(Application):
 
         argparser.add_argument(
             'cmd',
-            action='append',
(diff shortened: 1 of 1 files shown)


betatim:
LGTM

yuvipanda:
@betatim I invited you to the org so you can merge stuff :D

betatim:
More like:
http://i.imgur.com/loW46yj.gif

PR #109, opened 2017-10-23
updating index and api page

choldgraf:
This updates the index page, and also adds an API page. I'm not sure how to document a traitlets API automatically...anyone have an idea for how to do this with sphinx? @yuvipanda or @minrk or @willingc ?

(the `samples.rst` is also moved from CRLF to LF, *hides head in shame*)

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -1,0 +1,10 @@
+repo2docker API
+---------------
+
+The following arguments may be given in the call to::
+
+  jupyter-repo2docker <path-or-url-to-git-repo>
+
+.. autotraitletsdocumentation::
+
+   XXX how to do this?
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -5,6 +5,9 @@ The following arguments may be given in the call to::
 
   jupyter-repo2docker <path-or-url-to-git-repo>
 
-.. autotraitletsdocumentation::
 
-   XXX how to do this?
+repo2docker arguments
+=====================
+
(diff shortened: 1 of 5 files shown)


willingc:
@choldgraf Are you looking to do something similar to [this section of the jupyterhub docs](http://jupyterhub.readthedocs.io/en/latest/api/auth.html#authenticator)?

If so, you would do something similar to [this source](https://raw.githubusercontent.com/jupyterhub/jupyterhub/master/docs/source/api/auth.rst) for the doc page. You would also add https://github.com/jupyterhub/jupyterhub/blob/master/docs/source/conf.py#L23 to `conf.py` and add the [`sphinxext` directory](https://github.com/jupyterhub/jupyterhub/tree/master/docs/sphinxext) too.

choldgraf:
ah perfect, yes I think `autodoc_traits` is what I was looking for :-) I'll give it a shot

choldgraf:
hmmm, OK latest push has autodoc_traits installed and building, though I can't get the docs quite how I want them. I'm hoping to list the arguments that are defined in `get_argparser` for the `Repo2Docker` class...but it isn't showing up with `autoconfigurable`. Any idea how to get that working?

Here's what it looks like now:

http://predictablynoisy.com/repo2docker/api.html

choldgraf:
I'm gonna close this one in favor of #165 which covers the same ground and is less stale. We can figure out API stuff in a separate PR but it seems like it'll be non-trivial for me to tack on

PR #111, opened 2017-10-23
Remove jupyterhub from default installs

### LINKED ISSUE #92: Stop installing JupyterHub by default with repo2docker
It's no longer required by binder, and we should remove it here too to simplify our codebase.

People using this for their own jupyterhub installs can / should specify jupyterhub version explicitly
in their requirements.txt file instead.

willingc:
Closes #92 

Remove jupyterhub from dockerfiles, build args, and tests.

Left jupyterhub code in legacy dockerfiles and tests.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -1,6 +1,7 @@
 """repo2docker: convert git repositories into jupyter-suitable docker images
 
-Images produced by repo2docker can be used with Jupyter notebooks standalone or via JupyterHub.
+Images produced by repo2docker can be used with Jupyter notebooks standalone,
+with BinderHub, or with JupyterHub.
 
 Usage:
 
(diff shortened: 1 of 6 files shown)


repo2docker/app.py
```diff
@@ -1,6 +1,7 @@
 """repo2docker: convert git repositories into jupyter-suitable docker images
 
-Images produced by repo2docker can be used with Jupyter notebooks standalone or via JupyterHub.
+Images produced by repo2docker can be used with Jupyter notebooks standalone,
```
yuvipanda: This is no longer true if we remove jupyterhub from the default install, however.

yuvipanda:
I'm pro this, but I'd like to wait for @minrk's opinion.

yuvipanda:
From https://travis-ci.org/jupyter/repo2docker/jobs/292457816 it looks like we missed removing jupyterhub== from the legacy dockerfile implementation.

PR #112, opened 2017-10-23
adding jupyter logos

### LINKED ISSUE #110: Adding a logo to the docs
Now that we've back on `alabaster`, we've got an empty hole in the top left where the logo usually goes. Do we want to:

1. Use the jupyter logo
2. Use the binder logo
3. Make a logo for repo2docker (maybe just something like below?)

image

?

choldgraf:
Adds a logo in the top left as well as a favicon

closes #110 

does this work @willingc ?


willingc:
LGTM. Thanks @choldgraf

PR #117, opened 2017-10-24
Add a basic FAQ doc + 1 entry

yuvipanda:
We get lots of questions that can be answered here!

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -1,0 +1,19 @@
+# Frequently Asked Questions (FAQ)
+
+A collection of frequently asked questions with answers!
+
+If you have a question & have found an answer, send a PR to add it here!
+
+## Can I use repo2docker to bootstrap my own Dockerfile?
+
+If you pass the `--debug` flag to `repo2docker`, it outputs the intermediate
+Dockerfile it is using to build the docker image. While it is tempting to copy
+this as a base for your own Dockerfile, that is not supported & in most cases
+will not work. The `--debug` output is just our intermediate generated
+Dockerfile, and is meant to be built in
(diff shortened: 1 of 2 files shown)


betatim:
Any idea why travis would be unhappy about more docs?

docs/source/faq.md
```diff
@@ -0,0 +1,19 @@
+# Frequently Asked Questions (FAQ)
+
+A collection of frequently asked questions with answers!
+
+If you have a question & have found an answer, send a PR to add it here!
+
+## Can I use repo2docker to bootstrap my own Dockerfile?
+
+If you pass the `--debug` flag to `repo2docker`, it outputs the intermediate
```
betatim: Lead with the answer: "No you can't."

betatim: Should we be consistent and use `jupyter-repo2docker` when referring to the command line program?

docs/source/faq.md
```diff
@@ -0,0 +1,19 @@
...
+Dockerfile it is using to build the docker image. While it is tempting to copy
```
betatim: "Dockerfile it is using" -> "Dockerfile that is used to build"?

docs/source/faq.md
yuvipanda: @betatim https://github.com/jupyter/repo2docker/pull/121 is the fix for the jupyter-repo2docker / repo2docker inconsistentcy I think.

yuvipanda:
@betatim just seem to be travis / docker issues, unrelated to our code

choldgraf:
+1 on the page. The subsections will be placed to the left so it should be easy for people to scan the page if it ever gets big...though at some point in the future I'd be +1 on having collapsible Q/A sections

betatim:
LGTM

Issue #102, opened 2017-10-20
Adding OSF support to repo2docker


yuvipanda:
We want to be able to support the immutable frozen relases on OSF as an input to repo2docker.

So we should introduce the concept of content providers. Their job is to check out the given URL to  the current local directory. They won't do any caching or whatever, and would use traitlets to be pluggable.

We decided against having autodetection, since there is no way to do that reliably for everything, and doing that only for *some* providers seems complex. 

So new commandline would be like:

```
jupyter-repo2docker --provider=osf https://<some-osf-url>
```

And the provider class should be like:

```
class OSFProvider(Provider):
     name = 'osf'
     
     description = 'some description of this provider'

     @gen.coroutine
     def provide(self, identifier):
     """
     This method should fetch the contents of the given identifier to the current working directory
     """
     pass
```

This means we only support a single identifier (such as file path, git url including ref, or osf url) to each provider. This makes implementation simpler, but we can modify this in the future if we need.

yuvipanda:
This will also need a supplemental issue in binderhub to support this on binder.

betatim:
Think that looks good.

If we wanted to support more than a URL we could also consider that the provider gets to eat all the arguments, so:
```
jupyter-repo2docker --provider=osf https://<some-osf-url> blah foo bar
```
we would pass the whole `sys.argv`s as arguments to `provide`??

yuvipanda:
@betatim cool!

I think let's just go with providing a single identifier for now, and we can extend it when we run into a use case that explicitly requires it.

betatim:
Related jupyterhub/binderhub#216

The schema.org JSON-LD can be used [get a list of files associated with a Zenodo DOI w](https://github.com/whole-tale/whole-tale/issues/35#issuecomment-427411937). This is what Google use for their dataset search so there is hope that this will become widely adopted because of their marketing power.

betatim:
The problem with the JSON-LD is that it only gives you a list of files for a dataset, not a software archive.

However you can ask for information in JSON format (what ever that might be) and then you get a list of files both for datasets and for software:

* `curl -s -H "Accept: application/json" https://zenodo.org/api/records/1408168 | jq .files`
* `curl -s -H "Accept: application/json" https://zenodo.org/api/records/1419226 | jq .files`

betatim:
[...]

PR #118, opened 2017-10-24
[MRG] Use stdlib tempfile module

betatim:
The `tempfile` module can take care of finding a good temporary directory
for us on platforms where `/tmp` does not exist.

Should we change the name from `git_workdir` to `workdir`? Do you care that this creates one extra level of nesting for the temporary directories?

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -55,6 +55,7 @@ class Repo2Docker(Application):
     git_workdir = Unicode(
         None,
         config=True,
+        allow_none=True,
         help="""
         Working directory to check out git repositories to.
 
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -14,6 +14,7 @@ import logging
 import uuid
 import shutil
 import argparse
+import tempfile
 from pythonjsonlogger import jsonlogger
 import escapism
 
@@ -53,10 +54,10 @@ class Repo2Docker(Application):
         return logging.INFO
 
     git_workdir = Unicode(
-        "/tmp",
+        tempfile.mkdtemp(prefix='repo2docker'),
         config=True,
         help="""
-        The directory to use to check out git repositories into.
+        Working directory to check out git repositories to.
 
         Should be somewhere ephemeral, such as /tmp
         """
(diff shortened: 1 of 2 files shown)


repo2docker/app.py
```diff
@@ -53,10 +54,10 @@ def _default_log_level(self):
         return logging.INFO
 
     git_workdir = Unicode(
-        "/tmp",
+        tempfile.mkdtemp(prefix='repo2docker'),
```
yuvipanda: we're creating this unconditionally, even if the repo is just already locally existing. I think this should be made conditional

yuvipanda:
I left a comment, but I also think this whole set of code needs to change a bit to accomodate #102...

repo2docker/app.py
betatim: True that. We could use `tempfile.gettempdir()` to get the tmp directory but then we still have to somehow generate a directory inside of it. I'd prefer using something from the stdlib instead of doing our own thing (a bit "never roll your own crypto"). However there is no way (I think) of reserving a temporary filename without also creating it.

👍 on this changing for #101 this PR mainly started as a way to get my feet wet 🌊 ...creating a temporary name in a secure way will still be part of that though.

yuvipanda: How about making the default be 'None' and using tempfile when git_workdir is None?

betatim:
@yuvipanda what do you think of this? I added a `maybe_cleanup` context manager to deal with the case of exceptions making us exit early and not cleaning up. Using `None` as you suggested. Also did the thing you shouldn't: cleanup some `pep8` 🚓 while adding new code.

repo2docker/app.py
```diff
@@ -53,12 +53,13 @@ def _default_log_level(self):
         return logging.INFO
 
     git_workdir = Unicode(
-        "/tmp",
+        None,
         config=True,
```
yuvipanda: this needs allow_none=True here, tests are failing because of that.

yuvipanda:
Code looks good, I'll merge as soon as the restarted build (I think the error is a flake) finishes

yuvipanda:
Tests pass! \o/

betatim:
🎈🎈🎈🎈

PR #120, opened 2017-10-24
[MRG] Add limit on git clone depth

betatim:
Only checkout last 50 commits from a remote repository. This helps speed
up the cloning for large repositories.

50 is the same depth to which travis clones by default. This does have the potential to fail though because there is no reason for the ref we might want to checkout to be in the last 50 commits.

WDYT?

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -104,7 +104,7 @@ class Repo2Docker(Application):
             except subprocess.CalledProcessError:
                 self.log.error('Failed to clone repository!',
                                extra=dict(phase='failed'))
-                sys.exit(1)
+                raise RuntimeError("Failed to clone %s." % url)
 
         def _unshallow():
             try:
@@ -115,7 +115,8 @@ class Repo2Docker(Application):
             except subprocess.CalledProcessError:
                 self.log.error('Failed to unshallow repository!',
                                extra=dict(phase='failed'))
-                sys.exit(1)
+                raise RuntimeError("Failed to create a full clone of"
+                                   " %s." % url)
 
         def _contains(ref):
             try:
@@ -137,15 +138,16 @@ class Repo2Docker(Application):
             except subprocess.CalledProcessError:
                 self.log.error('Failed to check out ref %s', ref,
                                extra=dict(phase='failed'))
-                sys.exit(1)
+                raise RuntimeError("Failed to checkout reference %s for"
+                                   " %s." % (ref, url))
 
         # create a shallow clone first
         _clone(depth=50)
-        if ref:
-            if not _contains(ref):
-                # have to create a full clone
-                _unshallow()
-            _checkout(ref)
+
+        if not _contains(ref):
+            # have to create a full clone
+            _unshallow()
+        _checkout(ref)
 
     def get_argparser(self):
         argparser = argparse.ArgumentParser()
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -92,7 +92,8 @@ class Repo2Docker(Application):
 
     def fetch(self, url, ref, checkout_path):
         try:
-            for line in execute_cmd(['git', 'clone', url, checkout_path],
+            for line in execute_cmd(['git', 'clone', '--depth', '50',
+                                     url, checkout_path],
                                     capture=self.json_logs):
                 self.log.info(line, extra=dict(phase='fetching'))
         except subprocess.CalledProcessError:
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -91,23 +91,61 @@ class Repo2Docker(Application):
     )
 
     def fetch(self, url, ref, checkout_path):
-        try:
-            for line in execute_cmd(['git', 'clone', '--depth', '50',
-                                     url, checkout_path],
-                                    capture=self.json_logs):
-                self.log.info(line, extra=dict(phase='fetching'))
-        except subprocess.CalledProcessError:
-            self.log.error('Failed to clone repository!', extra=dict(phase='failed'))
-            sys.exit(1)
+        def _clone(depth=None):
+            if depth is not None:
+                command = ['git', 'clone', '--depth', str(depth),
+                           url, checkout_path]
+            else:
+                command = ['git', 'clone', url, checkout_path]
 
-        if ref:
             try:
-                for line in execute_cmd(['git', 'reset', '--hard', ref], cwd=checkout_path,
+                for line in execute_cmd(command, capture=self.json_logs):
+                    self.log.info(line, extra=dict(phase='fetching'))
+            except subprocess.CalledProcessError:
+                self.log.error('Failed to clone repository!',
+                               extra=dict(phase='failed'))
+                sys.exit(1)
+
+        def _unshallow():
+            try:
+                for line in execute_cmd(['git', 'fetch', '--unshallow'],
+                                        capture=self.json_logs,
+                                        cwd=checkout_path):
+                    self.log.info(line, extra=dict(phase='fetching'))
+            except subprocess.CalledProcessError:
+                self.log.error('Failed to unshallow repository!',
+                               extra=dict(phase='failed'))
+                sys.exit(1)
+
+        def _contains(ref):
+            try:
+                for line in execute_cmd(['git', 'cat-file', '-t', ref],
+                                        capture=self.json_logs,
+                                        cwd=checkout_path):
+                    self.log.debug(line, extra=dict(phase='fetching'))
+            except subprocess.CalledProcessError:
+                return False
+
+            return True
+
+        def _checkout(ref):
+            try:
+                for line in execute_cmd(['git', 'reset', '--hard', ref],
+                                        cwd=checkout_path,
                                         capture=self.json_logs):
                     self.log.info(line, extra=dict(phase='fetching'))
             except subprocess.CalledProcessError:
-                self.log.error('Failed to check out ref %s', ref, extra=dict(phase='failed'))
+                self.log.error('Failed to check out ref %s', ref,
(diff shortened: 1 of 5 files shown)


choldgraf:
This seems like a clever idea to me - I'd actually wonder why not make it even more shallow than 50, since we re-build the whole process for every single commit change. Can anyone think of a reason that Binder would need to care about any of a repository's history?

yuvipanda:
This is a great performance optimization!

However, if the commit we wanna build is over 50 commits old on master, then our 'git checkout' later will fail to find it and fail. So we should add a try / catch to the git checkout, which does a full fetch if git checkout fails with a 'can not find commit' error, and tries again.

choldgraf:
+1 to

```
try:
    shallow pull
    checkout
except CheckoutFails:
    deep pull
    checkout
```

betatim:
They call me Tim "too clever for his own good" Head ;)

Added a bit of code to check if the ref exists and if not clone the whole repo. What is the best way to build a test case for this? For now I will try and find a ref in https://github.com/jakevdp/PythonDataScienceHandbook that is already more than 50 commits behind master and add that as a test.

betatim:
Test added but it fails because the `requirements.txt` from back then doesn't work :-/

choldgraf:
We can create a dummy repo with more than 50 commits in the CI org?

betatim:
👍  though we might have to understand a bit more about this max depth thing. Pretty sure the commit I had to go back to is more than 50 commits from HEAD. So I'm puzzled.

betatim:
I created https://github.com/betatim/repo2docker-ci-clone-depth with 100 commits in it. Where should I transfer the repo to?

If the travises are happy I think this can be merged.

choldgraf:
https://github.com/orgs/binderhub-ci-repos/dashboard ?
--

betatim:
Will transfer there.

Just noticed I pushed a commit here that doesn't need to go here. This cleans up the [left over juypterhub reference](https://github.com/jupyter/repo2docker/pull/111). Keep it here or separate PR?

willingc:
@betatim I think it's fine to keep the clean up of the jupyterhub reference in the legacy builder here.

repo2docker/app.py
```diff
@@ -91,23 +91,62 @@ def _default_log_level(self):
...
 
-        if ref:
             try:
-                for line in execute_cmd(['git', 'reset', '--hard', ref], cwd=checkout_path,
+                for line in execute_cmd(command, capture=self.json_logs):
+                    self.log.info(line, extra=dict(phase='fetching'))
+            except subprocess.CalledProcessError:
+                self.log.error('Failed to clone repository!',
+                               extra=dict(phase='failed'))
+                sys.exit(1)
```
yuvipanda: This should just re-raise, IMO. Calling sys.exit makes it hard for this to be used as a library, for example.

repo2docker/app.py
```diff
@@ -91,23 +91,62 @@ def _default_log_level(self):
...
+        def _unshallow():
+            try:
+                for line in execute_cmd(['git', 'fetch', '--unshallow'],
+                                        capture=self.json_logs,
+                                        cwd=checkout_path):
+                    self.log.info(line, extra=dict(phase='fetching'))
+            except subprocess.CalledProcessError:
+                self.log.error('Failed to unshallow repository!',
+                               extra=dict(phase='failed'))
+                sys.exit(1)
```
yuvipanda: Same as above.

repo2docker/app.py
```diff
@@ -91,23 +91,62 @@ def _default_log_level(self):
...
                     self.log.info(line, extra=dict(phase='fetching'))
             except subprocess.CalledProcessError:
-                self.log.error('Failed to check out ref %s', ref, extra=dict(phase='failed'))
+                self.log.error('Failed to check out ref %s', ref,
+                               extra=dict(phase='failed'))
                 sys.exit(1)
 
+        # create a shallow clone first
+        _clone(depth=50)
+        if ref:
```
yuvipanda: can ref be falsy at all?

repo2docker/app.py
```diff
@@ -91,23 +91,62 @@ def _default_log_level(self):
...
+            try:
+                for line in execute_cmd(['git', 'reset', '--hard', ref],
+                                        cwd=checkout_path,
                                         capture=self.json_logs):
                     self.log.info(line, extra=dict(phase='fetching'))
             except subprocess.CalledProcessError:
-                self.log.error('Failed to check out ref %s', ref, extra=dict(phase='failed'))
+                self.log.error('Failed to check out ref %s', ref,
+                               extra=dict(phase='failed'))
                 sys.exit(1)
```
yuvipanda: Same as above

repo2docker/app.py
yuvipanda: On second thought, I wonder if this code can be simplified by not having them be functions. 

I generally feel that functions that are being called only from one place and without arguments shouldn't be functions - we can always abstract them when we realize we need to call them from two places. This means less mental jumping when reading the code, which is always good. 

In this context, there is also not much use for catching the CalledProcessError manually each time. execute_cmd will already output stderr, so there should already be a more specific error message anyway. So the generic 'failed to clone repository' can be a generic error handler for the entire fetch function, rather than for each mini-function.

So to summarize:

1. Remove the _ functions, just inline the code into fetch
2. Remove the individual try / excepts
3. ???
4. PROFIT

repo2docker/app.py
betatim: I don't know, just blindly copied from what was here before.

betatim:
Currently you can't use `repo2docker` as a library:
```python
import repo2docker.app
app = repo2docker.app.Repo2Docker()
app.initialize()
```
gives you the somewhat confusing:
```
usage: ipython [-h] [--config CONFIG] [--json-logs] [--image-name IMAGE_NAME]
               [--ref REF] [--debug] [--no-build] [--no-run] [--no-clean]
               [--push]
               repo ...
ipython: error: the following arguments are required: repo, cmd
An exception has occurred, use %tb to see the full traceback.
```
which just makes me think "ipython????" :) What magic sauce do we need to add to deal with the traitlets?

repo2docker/app.py
betatim: There will be something on the screen, but I think it makes sense to have a "summary" error message. My guess would be that most users of `repo2docker` don't realise that the below message means that `fetch()` failed:
```
Cloning into '/tmp/5de0d21f-605f-403d-bc7e-6b0955f0b78b'...
ssh: Could not resolve hostname github.com: nodename nor servname provided, or not known
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
  File "/Users/thead/anaconda/envs/bhub/bin/jupyter-repo2docker", line 11, in <module>
    load_entry_point('jupyter-repo2docker', 'console_scripts', 'jupyter-repo2docker')()
  File "/Users/thead/git/jupyterhub/repo2docker/repo2docker/__main__.py", line 6, in main
    f.start()
  File "/Users/thead/git/jupyterhub/repo2docker/repo2docker/app.py", line 366, in start
    checkout_path
  File "/Users/thead/git/jupyterhub/repo2docker/repo2docker/app.py", line 146, in fetch
    for line in execute_cmd(command, capture=self.json_logs):
  File "/Users/thead/git/jupyterhub/repo2docker/repo2docker/utils.py", line 18, in execute_cmd
    raise subprocess.CalledProcessError(ret, cmd)
subprocess.CalledProcessError: Command '['git', 'clone', '--depth', '50', 'git@github.com:binder-examples/requirements.git', '/tmp/4f9641fa-516c-4d5d-90e8-e56d9f1b7cd1']' returned non-zero exit status 128.
```
So I propose to catch the `CalledProcessError` and raise `RuntimeError('Cloning repository (%s) failed.' % repo)` which we can either catch and nicely format or just show as traceback in the main app.

willingc:
@betatim Let's take your last comment and make it it's own issue re: library.

As for the PR, what happens with the `git clone depth` when I have had a binder badge out for a year (say for a publication) but the repo has had many changes (> 50) since that time? I think you would need to do a full clone at that point and not a shallow clone.

repo2docker/app.py
yuvipanda: Fair enough! Do you still wanna keep the split up functions though?

yuvipanda:
I filed https://github.com/jupyter/repo2docker/issues/126 for the library issue! Thanks for responding to the rest, @betatim

betatim:
I find it easier to figure out what is going on with the functions: "shallow clone, not deep enough? -> full clone!" because each of them contains all the lines around running a command, logging, etc. It is true that you need to scroll past that wall of functions to get to the real part which is not ideal.

betatim:
Added a test that checks that if the ref is within the last 50 commits we do not unshallow the clone.

repo2docker/app.py
```diff
@@ -91,22 +91,63 @@ def _default_log_level(self):
...
-                sys.exit(1)
+                self.log.error('Failed to check out ref %s', ref,
+                               extra=dict(phase='failed'))
+                raise RuntimeError("Failed to checkout reference %s for"
+                                   " %s." % (ref, url))
+
+        # create a shallow clone first
+        _clone(depth=50)
+
+        if not _contains(ref):
```
betatim: @willingc this is the magic line that takes care of your case "ref too far in the past"

yuvipanda:
Thanks for the tests and the patience, @betatim! I've added a commit that fixed up the last failing test and am going to merge now!

repo2docker/app.py
willingc: Thanks @betatim 👍

betatim:
Whoop! Better code through thorough review :) 🍻

PR #124, opened 2017-10-25
adding repo name parsing

choldgraf:
this is an attempt at adding more parseable image names to the pushed image. I'm not sure if this is the right location to do this (maybe better in binderhub?) but LMK thoughts


choldgraf:
Closing this and we'll figure it out another way!

PR #128, opened 2017-11-01
Add handling for no ref being provided

betatim:
Fixes the bug noticed (and reverted) here: jupyterhub/mybinder.org-deploy#112

We do not need to checkout a specific ref if none is provided. Added a test to check the behaviour.

The big question is: when does binderhub call `repo2docker` without a ref as commandline argument?

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -144,10 +144,13 @@ class Repo2Docker(Application):
         # create a shallow clone first
         _clone(depth=50)
 
-        if not _contains(ref):
-            # have to create a full clone
-            _unshallow()
-        _checkout(ref)
+        # ref == None means we want to use HEAD so no need to checkout a
+        # specific revision
+        if ref is not None:
+            if not _contains(ref):
+                # have to create a full clone
+                _unshallow()
+            _checkout(ref)
 
     def get_argparser(self):
         argparser = argparse.ArgumentParser()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -51,9 +51,8 @@ class RemoteRepoTest(pytest.Item):
         self.verify = verify
 
     def runtest(self):
-        subprocess.check_call([
-            'jupyter-repo2docker',
-            '--ref', self.ref,
-            self.url,
-            '--',
-        ] + shlex.split(self.verify))
+        if self.ref is not None:
+            cmd = ['jupyter-repo2docker', '--ref', self.ref, self.url, '--']
+        else:
+            cmd = ['jupyter-repo2docker', self.url, '--']
+        subprocess.check_call(cmd + shlex.split(self.verify))
--- a/tests/external/shallow-clone.repos.yaml
+++ b/tests/external/shallow-clone.repos.yaml
@@ -11,3 +11,9 @@
   # we checkout the second to last commit, hence the log has 49 entries at a
   # max clone depth of 50
   verify: /bin/bash -c '[ $(git log --oneline | wc -l) == "49" ]'
+
+- name: A dummy repo with 100 commits - live at HEAD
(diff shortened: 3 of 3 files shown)


repo2docker/app.py
```diff
@@ -144,10 +144,13 @@ def _checkout(ref):
         # create a shallow clone first
         _clone(depth=50)
 
-        if not _contains(ref):
-            # have to create a full clone
-            _unshallow()
-        _checkout(ref)
+        # ref == None means we want to use HEAD so no need to checkout a
+        # specific revision
+        if ref is not None:
```
betatim: this fixes the bug

tests/conftest.py
```diff
@@ -51,9 +51,8 @@ def __init__(self, name, parent, url, ref, verify):
...
-        subprocess.check_call([
-            'jupyter-repo2docker',
-            '--ref', self.ref,
-            self.url,
-            '--',
-        ] + shlex.split(self.verify))
+        if self.ref is not None:
+            cmd = ['jupyter-repo2docker', '--ref', self.ref, self.url, '--']
+        else:
+            cmd = ['jupyter-repo2docker', self.url, '--']
```
betatim: Needed to support the new test case that has no ref.

yuvipanda:
Currently it seems to fail even when you do give it a ref.

```
 jupyter-repo2docker https://github.com/binder-examples/requirements --ref master
```

Fails on master for me.

betatim:
`jupyter-repo2docker https://github.com/binder-examples/requirements --ref master` also fails on master for me. Inspecting `ref` just before calling `_contains` gives `None`.

yuvipanda:
Yup. Looks like it's related to argparse issue somewhere, most likely in the 'cmd' arg.

The following works:

```
jupyter-repo2docker --ref  d7f774a1af892319cd1483fa2255c24511876375 https://github.com/binder-examples/requirements
```

The following does not:

```
jupyter-repo2docker https://github.com/binder-examples/requirements --ref  d7f774a1af892319cd1483fa2255c24511876375
```

This is on this branch, not master.

yuvipanda:
If anything, it looks like we'd just been building master for a while instead of whatever ref was passed in. Your PR removed the superfluous 'if not ref' line, causing the problem to actually show :)

choldgraf:
> If anything, it looks like we'd just been building master for a while instead of whatever ref was passed in. 

👌 nice ;-)

betatim:
Agree, if you pass `--ref` after the URL then `args.ref is None` right after parsing the arguments. After that everything "works", except we lost the ref.

tests/conftest.py
yuvipanda: This is also why tests passed, since --ref was before self.url!

yuvipanda:
I made a patch to binderhub: https://github.com/jupyterhub/binderhub/pull/234

yuvipanda:
Ok, this can be deployed again, but https://github.com/jupyterhub/binderhub/pull/234 should probably also be deployed alongside this to make sure we're actually building the versions we wanna.

betatim:
I think we should declare passing an `--arg` after the URL as undefined behaviour. If you do pass something after the URL (`jupyter-repo2docker https://github.com/binder-examples/requirements --ref master`) you get:
```
Namespace(build=True, clean=True, cmd=['--ref', 'master'],
config='repo2docker_config.py', debug=False, image_name=None,
json_logs=False, push=False, ref=None,
repo='https://github.com/binder-examples/requirements', run=True)
```
from `parse_args()`. We could inspect `cmd` to see if it contains an argument that is also a `repo2docker` argument and then exit. However I think that is a bit too much heuristics (you could legitimately have a `--ref` in the command you want to execute).

PR #130, opened 2017-11-01
Revert "Add handling for no ref being provided"

yuvipanda:
Reverts jupyter/repo2docker#128

It looks like shallow clones are a lot more hairy than we realized.
We discovered the following:

1. We currently have no way to figure out if a clone is shallow or not.
    An option has been added to rev-parse (why rev-parse?!) but that's 
    too new and not in any packaged git.
2. If your repo has under 50 commits, it will not be shallow!
3. If you try to run git fetch --unshallow on a non-shallow repo it is a fatal error

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -144,13 +144,10 @@ class Repo2Docker(Application):
         # create a shallow clone first
         _clone(depth=50)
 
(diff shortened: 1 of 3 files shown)


yuvipanda:
Second part of the revert will be part of https://github.com/jupyter/repo2docker/pull/120

PR #134, opened 2017-11-09
[WIP] Use `repository/` subdirectory to store contents

### LINKED ISSUE #129: Copy repo contents to a subdirectory
Is there a reason to copy the repo into `/home/jovyan` or could we equally well copy it to `/home/jovyan/repo`?

I'd like to be able to mount a local directory into the container instead of copying the repository but currently that also wipes out all the hidden files in the home directory.

betatim:
Fixes #129 

Using a subdirectory allows us to mount the contents of the repository
from a volume instead of copying it.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -126,6 +126,12 @@ class Repo2Docker(Application):
         )
 
         argparser.add_argument(
+            '--mount',
+            default='/',
+            help='Mount local repository into container instead of copying it'
+        )
+
+        argparser.add_argument(
             'repo',
             help='Path to repository that should be built. Could be local path or a git URL.'
         )
--- a/repo2docker/detectors.py
+++ b/repo2docker/detectors.py
@@ -700,7 +700,7 @@ class JuliaBuildPack(BuildPack):
                 eval(:(using $(Symbol(pkg)))) \
                end \
             '
-            """ % { "require" : require }
+            """ % { "require" : self.repo_path(require) }
         )]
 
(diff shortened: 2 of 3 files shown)


betatim:
Ja I was wondering what to do on binder(hub). Indeed people will expect the file viewer to show the directory containing the repo. Same when they open a terminal. Not sure I think (2) is a feature or a bug :-/

I like the idea of a command line argument. I started this as a way to find out how much stuff breaks ;)

yuvipanda:
:D I agree that any hardcoding of $HOME (or ${HOME}) in the code should have been parameterized from the beginning. Would've made this easier...

betatim:
Seems to mostly be replacing `binder_path(fname)` with `repo_path(fname)`

choldgraf:
@betatim where are we on this one? It's unclear to me if there's a path forward on the PR or if we're waiting for something else to change in binderhub?

betatim:
The way forward is to implement this as a command-line option for `repo2docker`. binderhub won't change. I will try and get this to "merge me please" stage this week.

yuvipanda:
@betatim do you think this is still needed now that we support mounting volumes in arbitrary locations?

betatim:
Closing this for now, I think this PR is abandoned.

Not sure mounting arbitrary volumes will help with my original use case? I want a rw mount of my local repo inside the container. If you use the existing mount support you will still lose whatever is in all the other subdirectories of the home directory.

PR #137, opened 2017-11-11
RFC: Ignore "julia" in REQUIRE file when precompiling

darwindarak:
Otherwise builds will throw an error if a specific version of Julia is listed in `REQUIRE`.


choldgraf:
This seems reasonable to me! I wonder what is the best way to make tests for non-python buildpacks. @betatim do you have any thoughts?

That said, I think this could/should be merged if this is a fairly common bug.

yuvipanda:
Thanks for the patch, @darwindarak! This is totally a bug I think.

Now someone needs to deploy this...

PR #140, opened 2017-11-12
Fix binder/apt.txt detection

oschuett:
The same issue also exists in LegacyBinderDockerBuildPack(), but maybe on purpose?

### DIFF
--- a/repo2docker/detectors.py
+++ b/repo2docker/detectors.py
@@ -421,7 +421,7 @@ class BaseImage(BuildPack):
     def setup_assembly(self):
         assemble_scripts = []
         try:
-            with open('apt.txt') as f:
+            with open(self.binder_path('apt.txt')) as f:
                 extra_apt_packages = [l.strip() for l in f]
             # Validate that this is, indeed, just a list of packages
             # We're doing shell injection around here, gotta be careful.


choldgraf:
Nice find! My intuition is that we can leave the legacy build pack the way it is, since that's mostly for people who built binders way back when and have left them stale. If they're going to update them, they should update them to adhere to new binder dockerfile spec :-)

betatim:
restarted travis build.

choldgraf:
I already tried that and I think it failed again... Not sure what's going
on there

betatim:
Green now ... though maybe I'll use that as an opportunity to ask @oschuett if we could get a test to stop the bug from coming back ;)

choldgraf:
@betatim it sounds like we'd need this test to happen after a rebase once the other PR is merged...I'd be fine merging this in and opening an issue to add a test for this in a separate PR. WDYT?

betatim:
Created an issue ... merging.

betatim:
Congratulations @oschuett for your first contribution to repo2docker! 🍻

oschuett:
Cool! Thanks for merging. I'll try to look into #142 this week.

choldgraf:
Yahoo!!! Thanks @oschuett

PR #143, opened 2017-11-13
Apply PEP8 style roller

betatim:
Some changes to make my linter happy.


choldgraf:
Travis is unhappy!

I think this is a good addition, and I am +1 on adopting PEP8 conventions for the repo. Here's a question: is it worth taking a "good enough" approach to PEP8, where a brave person such as @betatim fixes up long lines etc every now and then, or do we put a PEP8 check into travis so that it won't build if there's an error?  (I'm partial to the "good enough" approach, as I've found requiring perfect PEP to be really demoralizing to new contributors)

betatim:
I love PEP8 but I don't think robots are good enough yet at applying PEP8 with common sense -> -0 on having the build go red when there are pep8 violations.

On scikit-optimize we (by that I mean Tim) recently added https://github.com/OrkoHunter/pep8speaks to see if that works without being too noisy. It tells you what is not PEP8, but leaves it up to you to fix stuff. My hope is that it nudges people to fix up the most ugly violations but they don't feel compelled to fix every 81 character line.

yuvipanda:
Is this limiting us to 80char limits? I don't want us to limit ourselves to 80char limits. Otherwise am +1 :)

betatim:
You can tell pep8speaks how long you think your lines should be or even disable that check. This PR doesn't add pep8speaks though.

In this PR I did shorten some very long lines but there are still a few left that are just a few characters too long.

choldgraf:
@betatim since this touches several lines of code can you rebase and we can get this merged? I'm +1 on style changes and these seem reasonable. The longer we don't merge the more conflicting changes will be introduced...

betatim:
Rebased.

repo2docker/app.py
```diff
@@ -243,14 +247,20 @@ def initialize(self):
             # remove the additional newline from the stream handler
             self.log.handlers[0].terminator = ''
             # We don't want a [Repo2Docker] on all messages
-            self.log.handlers[0].formatter = logging.Formatter(fmt='%(message)s')
+            self.log.handlers[0].formatter = logging.Formatter(
+                fmt='%(message)s'
+                )
```
yuvipanda: shouldn't this closing paren be flush left?

repo2docker/app.py
```diff
@@ -243,14 +247,20 @@ def initialize(self):
...
             self.output_image_spec = args.image_name
         else:
             # Attempt to set a sane default!
             # HACK: Provide something more descriptive?
-            self.output_image_spec = 'r2d' + escapism.escape(self.repo, escape_char='-').lower() + str(int(time.time()))
+            self.output_image_spec = (
+                'r2d' +
+                escapism.escape(self.repo, escape_char='-').lower() +
+                str(int(time.time()))
+                )
```
yuvipanda: Here too - looks like all closing parens are in line with args, rather than flush

yuvipanda:
let's merge this when the tests pass, and take another pass :D

repo2docker/app.py
betatim: /shrug

I just put it where my editor put it and the linter didn't complain.

Issue #142, opened 2017-11-12
Add test for `binder/apt.txt` detection


betatim:
Add a test that makes sure #140 keeps working in the future.

The Beyonce strategy: 'Cause if you liked it, then you should have put a test on it  ;)

oschuett:
Done: #145

betatim:
Thank you and closing :)

PR #145, opened 2017-11-17
Add tests for binder/apt.txt and binder/postBuild

oschuett:
Fixes issue #142.

### DIFF
--- a/tests/venv/binder-dir/apt.txt
+++ b/tests/venv/binder-dir/apt.txt
@@ -1,0 +1,1 @@
+thiswontwork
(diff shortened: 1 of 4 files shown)


choldgraf:
will `which` raise an error if the thing doesn't exist so that `verify` will catch it? If so, then this LGTM

oschuett:
Yes, `which` will return with exit code 1 in that case. Furthermore, the [existing test](https://github.com/jupyter/repo2docker/blob/master/tests/venv/apt-packages/verify) for apt does it the same way.

choldgraf:
thanks for the contrib @oschuett!!!

PR #147, opened 2017-11-21
Make all bash verify scripts fail on error

yuvipanda:
Bash scripts by default continue on executing even if any
command errors. That's not the behavior we want for our scripts,
so let's explicitly get them to fail if:

1. Any command exits with non 0 exit code
2. An undefined variable is referenced


choldgraf:
seems good to me, merging

(for folks who are curious about this, here's a useful resource on this command http://redsymbol.net/articles/unofficial-bash-strict-mode/)

Issue #146, opened 2017-11-21
runtime.txt won't install other python 3 versions


choldgraf:
Surfaced by @foozmeat in the gitter channel, if you put something like `python-3.6` then it doesn't look like the runtime gets updated. I don't see anything here:

https://github.com/jupyter/repo2docker/blob/master/repo2docker/detectors.py#L505

that would install other versions of python 3...what would be the best way to go about this?

yuvipanda:
Yep, we don't support python3.6 and should.

This should be a modification in the Python build pack that decides which packages to include.

choldgraf:
do you think this should replace the 3.5 kernel with 3.6, or install it alongside 3.5 similar to how 2.7 behaves?

yuvipanda:
definitely replace. Don't think we can / should support multiple 3.x kernels at the same time for now.

choldgraf:
hmmm - so it looks like the default in apt is python3.5 yeah? would we need to pull it from a different repo?

choldgraf:
then again, we could always..................handle this with minicoda _runs away_

yuvipanda:
there's a python3.6 package in zesty so that should work. we can also switch to artful from zesty (new versions of everything!) but perhaps we should decouple these two into different PRs.

yuvipanda:
@choldgraf you could, by using environment.yml :) Also let's keep the (good natured) conda/venv related trolling to a minimum please :)

choldgraf:
Actually now that I think about it I'm a bit confused...the person that
noticed this originally mentioned that they tried using environment.yml to
define the environment but that it then skipped requirements.txt . To my
knowledge that should have worked fine no? I feel like there's another
issue that addresses this but I'm on a plane/on my phone so can't
investigate

foozmeat:
Hi, it was me that ran into the issue. I first tried using a `runtime.txt` and that got me python 3.5 even though I specified 3.6. When I tried an `environment.yml` with just the runtime specified then my `requirements.txt` was skipped. I finally added my core packages to the `environment.yml` and now its working fine. I added test code to the top of my notebook to check for python 3.6 and it succeeds

https://hub.mybinder.org/user/panicinc-speed_mystery-eyd96te7/notebooks/main.ipynb

willingc:
It would be nice to move to Python 3.6 since it supports f-strings which folks will likely be using as they are a big improvement over the other string formatting options.

choldgraf:
[...]

PR #148, opened 2017-11-22
[WIP] Allow specifying base distro in apt.yaml

yuvipanda:
We should probably routinely bump this every month after ubuntu
new release perhaps?

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -10,7 +10,7 @@ import re
 import docker
 
 TEMPLATE = r"""
-FROM buildpack-deps:artful
+FROM {{ base_image }}
 
 # Set up locales properly
 RUN apt-get update && \
@@ -140,6 +140,15 @@ class BuildPack(LoggingConfigurable):
 
         Versions are not specified, and ordering is not guaranteed. These
         are usually installed as apt packages.
+        """
+    )
+
+    base_image = Unicode(
+        "buildpack-deps:artful",
+        help="""
+        Base image to use.
+
+        Should be debian or ubuntu based image.
         """
     )
 
@@ -357,6 +366,7 @@ class BuildPack(LoggingConfigurable):
             build_script_directives=build_script_directives,
             assemble_script_directives=assemble_script_directives,
             build_script_files=self.build_script_files,
+            base_image=self.base_image,
             base_packages=sorted(self.base_packages),
             post_build_scripts=self.post_build_scripts,
         )
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -10,7 +10,7 @@ import re
 import docker
 
 TEMPLATE = r"""
-FROM buildpack-deps:zesty
+FROM buildpack-deps:artful
 
 # Set up locales properly
 RUN apt-get update && \
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -1,6 +1,7 @@
 import textwrap
 from traitlets.config import LoggingConfigurable
 from traitlets import Unicode, Set, List, Dict, Tuple, default
+from ruamel.yaml import YAML
 import jinja2
 import tarfile
 import io
@@ -144,7 +145,7 @@ class BuildPack(LoggingConfigurable):
     )
 
     base_image = Unicode(
-        "buildpack-deps:artful",
+        "buildpack-deps:zesty",
         help="""
         Base image to use.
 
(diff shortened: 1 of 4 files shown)


yuvipanda:
@choldgraf this bumps base image to Artful, which also defaults to Python3.6 rather than 3.5. I think we can do this for now for #146, and support python-3.x only when someone specifically needs to stick to an older version of python3 (rather than newer).

choldgraf:
seems reasonable - if it's stable enough for ubuntu then it's stable enough for default Binder behavior IMO

betatim:
I'm not sure about bumping this every ~month. It will invariably break people's containers "overnight". My guess would be that most people who create binders don't follow repo2docker at all so it will surprise them/lead to lots of dead links. How do we help them?

One idea: in release X deprecate zesty, start emitting a warning if building with arty fails and rebuild with zesty. In release X+n remove zesty, emit error saying "zesty has been removed, we are now using arty. Common fixes on https://fixmybinder.org". With n chosen to give several months of transition period.

I think even a small fraction of users who get the impression "this binder thing is about reproducibility but it keeps breaking" is something worth avoiding.

It even seems to have broken some of our tests :-/

minrk:
Bumping the default base image is a major breaking change, so we need to be very careful about it and should do it infrequently. We also need to allow users to pin their distros if we want to claim a goal of reproducibility. Otherwise, we are forcing *all* reproducibility-minded users to custom Dockerfiles.

We should probably enable rolling updates of the base image, along the lines of what travis did with the update to trusty:

1. make new dist available, opt-in
2. after a period, bump default dist
3. after another period, drop old dists as unsupported (only after it becomes untenable, ideally never)

That doesn't mean we shouldn't do *this* update now, but for our own maintenance, limiting ourselves to 'current' LTS releases might be a good idea.

choldgraf:
so what if we added a CLI argument `--dist zesty` that could be either `artful` or `zesty` and would chose the respective docker image.

yuvipanda:
Our options I think are:

1. Add an apt.yaml that lets you specify distro and repo. We might extend this in the future to allow people to add custom apt repos
2. Add a 'distro' single-text file

We shouldn't do it as a commandline arg since it is the repo that should specify this, rather than the repo2docker invocation. We can allow users to pin to something if they want, and then after a few months flip the default.

choldgraf:
+1 to `apt.yaml`, we've already had a few people ask for custom apt repositories

minrk:
+1 to apt.yaml (or going all the way to binder.yaml). I also think specifying the distro should end up pinning the Dockerfile template (i.e. we freeze the zesty template when we move to artful). It ought to be only conservative users who are pinning distros, and it would probably be appropriate for these contexts to not get updated default notebook/ipython packages as well.

I think we have a hefty challenge ahead of us with trying to keep repo2docker updated vs our stated goal of reproducible environments. One level of this is to ensure truly reproducible environments means that we are probably going to have to eventually allow users to pin repo2docker version itself on Binder, or ship old versions of the builders along with updated repo2docker (as we do with nbformat).

choldgraf:
As I think we have agreement, can we change the topic of this issue to "add `apt.yaml` build pack" and label it an enhancement?

minrk:
BTW, I think it's okay to do this upgrade now (since we are in beta phase), and then probably stop upgrading the default once we get to the next LTS release (18.04). After that, we should probably only support LTS releases, at least for defaults.

choldgraf:
+1 after a rebase

yuvipanda:
I've repurposed this to provide an 'apt.yaml' that currently lets you set the base distro  out of a whitelist. Once it passes, I'll set the default to artful. This provides a path for people who must pin to zesty.

repo2docker/buildpacks/base.py
```diff
@@ -143,6 +144,15 @@ class BuildPack(LoggingConfigurable):
         """
     )
 
+    base_image = Unicode(
+        "buildpack-deps:zesty",
```
choldgraf: as the topic of the original PR was to bump the base image to artful, shouldn't this be artful?

repo2docker/buildpacks/base.py
```diff
@@ -424,9 +435,36 @@ class BaseImage(BuildPack):
...
+                        return self.base_image_map[base]
+                    else:
+                        raise ValueError(
+                            "Unknown base {} found in apt.yaml, available bases are {}".format(
+                                base, ', '.join(self.base_image_map.keys())
+                            )
+                        )
+        except FileNotFoundError:
+            pass
+        return 'buildpack-deps:zesty'
```
choldgraf: and here as well?

repo2docker/buildpacks/base.py
yuvipanda: See https://github.com/jupyter/repo2docker/pull/148#issuecomment-356143965 :)

choldgraf: oh I see what you mean

yuvipanda:
I'm currently trying to solve the related but different problem of 'decouple python versions from distros' in https://github.com/jupyter/repo2docker/pull/190, and making a lot of progress :)

repo2docker/buildpacks/base.py
```diff
@@ -307,6 +326,8 @@ def compose_with(self, other):
         build_script_files.update(other.build_script_files)
         result.build_script_files = build_script_files
 
+        # FIXME: Validate that they all have same base image!
+        result.base_distro = self.base_distro
```
minrk: Do we need to validate that they are the same? If they should all be the same, we can use a class attribute instead of allowing subclasses to deviate if they shouldn't be allowed to do so.

tests/venv/apt-base/apt.yaml
```diff
@@ -0,0 +1 @@
+base: artful
```
minrk: Does apt.yaml replace apt.txt? Or is it an additional file with only one supported key? What's the plan for transition here?

yuvipanda: Yeah, I think it should replace apt.txt. I was trying to get it to work with just this one key now.

repo2docker/buildpacks/base.py
yuvipanda: I'm thinking of #191 before getting this PR merged maybe

yuvipanda:
We decided against doing this.

PR #153, opened 2017-11-29
Make conda env update be verbose

choldgraf:
If you don't turn on verbose updates, then I'm pretty sure the log will look
 like it's hanging while it runs all the install commands under the hood. 
This adds `-v` in the call to `conda env update` which is triggered when 
`environment.yml` is found.

### DIFF
--- a/repo2docker/detectors.py
+++ b/repo2docker/detectors.py
@@ -562,7 +562,7 @@ class CondaBuildPack(BuildPack):
             assembly_scripts.append((
                 '${NB_USER}',
                 r"""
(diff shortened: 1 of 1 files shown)


yuvipanda:
I think GitHub puts the PR description into git commit messages, so let's make sure to rewrite it to make sense in a commit log before merging :)

choldgraf:
updated comment

yuvipanda:
I've updated it to have line-breaks so git doesn't freak out. Thanks, @choldgraf!

PR #155, opened 2017-11-30
Split detector.py into multiple files

yuvipanda:
Was getting too big, and did not want to mix formatting changes and code changes.

PR #156, opened 2017-11-30
Use a frozen requirements.txt for base install in venv buildpacks

### LINKED ISSUE #154: Pin jupyterlab to 0.28.*
We recently ran into an issue where Jupyterlab and the jupyterlab_launch became incompatible. I think one reason this happened is that a patch was made in jupyterlab (0.28.9) whereas jupyterlab was pinned to (0.28) and this caused the bug to surface. What if we baked in `jupyterlab==0.28.*` instead of just 0.28 so that we get patch updates?

yuvipanda:
Fixes #154 for venv.

We should then automate checking if the versions here are up to date,
and make PRs if they aren't!


yuvipanda:
Includes #155, so once that gets merged the diffs will be more useful.

repo2docker/buildpacks/python/requirements.frozen.txt
```diff
@@ -0,0 +1,41 @@
...
+ipython==6.2.1
+ipython-genutils==0.2.0
+ipywidgets==6.0.0
+jedi==0.11.0
+Jinja2==2.10
+jsonschema==2.6.0
+jupyter-client==5.1.0
+jupyter-core==4.4.0
+jupyterlab==0.28.0
+jupyterlab-launcher==0.6.0
```
jasongrout: This is problematic :). It should be `<0.6.0` if you're using jupyterlab 0.28.0.

jasongrout: (See #154)

yuvipanda: @jasongrout if you look at 'freeze.bash' you can see the script that's generating these frozen files. Can you tell me what's wrong with it?

yuvipanda: @jasongrout I've updated to JupyterLab 0.29.2 and maybe the requirements look better now?

minrk:
All green with JupyterLab 0.29.2

choldgraf:
thanks @yuvipanda !! Yep, the problem was that jlab .28 was only compatible with jlab_launch .5, so upgrading to .29 will fix that problem

repo2docker/buildpacks/python/requirements.frozen.txt
jasongrout: yes, that would be better.

PR #157, opened 2017-11-30
don't run tests in parallel on travis

minrk:
parallel tests might have been the cause of spurious travis failures

plus, not running concurrently should result in reusing the cache for the initial stages, which might even end up saving time

wip: waiting to see what this does on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,10 @@ python:
 - 3.4
 install:
 - pip install -e .
-- pip install pytest-xdist pyyaml
+- pip install pyyaml
 
 script:
-  - pytest -s -v -n4 tests/${REPO_TYPE}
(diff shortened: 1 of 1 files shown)


minrk:
This doesn't seem to have eliminated the failures, but it may have improved them a bit: I had to restart external, but no others. It does seem to result in faster tests for the most parallel case: venv has gone from 6:37 to 4:21. This makes some sense, since we only have one core on Travis, so only parallel IO is going to be helped and only if they aren't fighting over bandwidth. This seems to be overcome by being able to share the initial build cach. The rest seem to get little to no change, or a few seconds slower. I think we can also end up getting better test outputs in the case of failure by going this route.

yuvipanda:
Good idea, and glad it improves performance!

PR #158, opened 2017-11-30
update venv freeze and conda versions

minrk:
follow-up to #156

- sync conda versions with venv pins
- update ipywidgets to 6.0.1
- fix freeze.bash so it runs on macOS (freeze happens in docker)

Specifically not included yet: freezing conda. I'd like to do that in another PR.


yuvipanda:
Tested on linux and works.  Feel free to merge once tests pass!

minrk:
Thanks!

PR #159, opened 2017-12-01
Set limits for how much memory docker build can use

yuvipanda:
This prevents builds from gobbling all the RAM on a node!

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -163,7 +163,7 @@ class Repo2Docker(Application):
         )
 
         argparser.add_argument(
-            '---build-memory-limit',
+            '--build-memory-limit',
             help='Total Memory that can be used by the docker build process'
         )
 
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -29,8 +29,7 @@ from .buildpacks import (
     PythonBuildPack, DockerBuildPack, LegacyBinderDockerBuildPack,
     CondaBuildPack, JuliaBuildPack, Python2BuildPack, BaseImage
 )
-from .utils import execute_cmd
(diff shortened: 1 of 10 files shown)


tests/memlimit/dockerfile/postBuild
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
+"""
+Simplest program that tries to allocate a large amount of RAM.
+
+malloc lies on Linux by default, so we use memset to force the
+kernel to actually give us real memory.
```
betatim: sneaky kernel!

repo2docker/app.py
```diff
@@ -347,7 +364,7 @@ def start(self):
             if self.build:
                 self.log.info('Using %s builder\n', bp.name,
                               extra=dict(phase='building'))
-                for l in picked_buildpack.build(self.output_image_spec):
+                for l in picked_buildpack.build(self.output_image_spec, self.build_memory_limit):
```
betatim: should we make this a named argument? Just thinking we might want to add a CPU limit or other limits in the future and then I won't be able to remember in which order args go.

yuvipanda: Docker doesn't support CPU limits for builds yet, so IMO it's ok right now for this to be just a param that must be specified. We can always just screw ordering and say memory_limit=<whatever> when calling it regardless of it having a default value or not when we have too many of them I think?

betatim:
Merged!

betatim:
https://travis-ci.org/jupyter/repo2docker/jobs/309908787 this is the build for the merge. Deploy failed. :-/

Issue #160, opened 2017-12-01
Alpine Linux, pip/python & symlinks together break our build


yuvipanda:
As part of #159, there's a symlink now in the test suite, since postBuild scripts for both the memorylimit tests are the same. However, this seems to break shutil on alpine. There's more discussion here: https://github.com/docker-library/python/issues/155 and https://github.com/python/cpython/pull/4267. Most likely because of musl vs libc issues.

This breaks the 'deploy' step, since our docker image no longer builds.

betatim:
To get deploys working again I'd propose to duplicate the `postBuild` script. Once we have this deployed and are happy with the memlimiting we can take our time to decide if we want to change based image or what.

yuvipanda:
I agree! am working on a PR now.

willingc:
@yuvipanda Does #161 close this?

yuvipanda:
It works around it for now, the underlying problem is still unfixed tho

yuvipanda:
We moved away from alpine linux completely in 379c52bac3ab796cd945b3457295b3b4c3f9c6b6, so we can close this now.

PR #161, opened 2017-12-01
Remove symlink & just copy postBuild file instead

yuvipanda:
Temp fix for https://github.com/jupyter/repo2docker/issues/160.
Also includes a test to prevent drift

### DIFF
--- a/tests/memlimit.py
+++ b/tests/memlimit.py
@@ -96,4 +96,5 @@ def test_memlimit_same_postbuild():
     for fp in filepaths:
         with open(fp) as f:
             file_contents.append(f.read())
-    assert all(c == file_contents[0] for c in file_contents)
+    # Make sure they're all the same
+    assert len(set(file_contents)) == 1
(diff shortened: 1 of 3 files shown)


tests/memlimit.py
```diff
@@ -79,3 +79,22 @@ def test_memlimit_dockerfile_fail():
         512,
         256
     )
+
+
+def test_memlimit_same_postbuild():
```
betatim: 👍

tests/memlimit/dockerfile/postBuild
```diff
@@ -4,6 +4,11 @@ Simplest program that tries to allocate a large amount of RAM.
 
 malloc lies on Linux by default, so we use memset to force the
 kernel to actually give us real memory.
+
+NOTE: This file has to be duplicated & present in all the following locations:
+ - tests/memlimit/dockerfile/postBuild
+ - tests/memlimit/dockerfile/postBuild
```
minrk: dockerfile and non-dockerfile

PR #163, opened 2017-12-07
Freeze conda environment as well

minrk:
Like #156, but for conda

Script is Python because conda has [a bug](https://github.com/conda/conda/pull/6391) in env export that needs a little fixing.

Updates conda itself to 4.3.30, as well.


choldgraf:
LGTM - only one note: I didn't realize we were using conda-forge to install stuff. I'm personally ok with it, but are there any downsides to doing so? The only thing I can think of is that some users might assume the `anaconda` repo would be the default and this would violate that assumption.

minrk:
If we want recent packages, we have to use conda-forge, since anaconda updates can take longer and we can't influence the process (they still don't have the current `notebook` package. But in general, I think conda-forge is the more logical default channel for community-maintained projects like Binder and repo2docker.

The `defaults` channel is still present, it's just lower priority. And people can specify it in their `environment.yml` if they want those packages.

yuvipanda:
Thank you!

choldgraf:
Sounds good to me! I use forge as my default as well, just making sure we're being intentional about the decision!

PR #168, opened 2017-12-12
adding runtime to preparing + improving config section

choldgraf:
Updates `preparing` so that it includes all the possible build packs, and adds a concise list of them at the beginning of the section. This is semi-linked to https://github.com/jupyterhub/binder/pull/47, which uses `config_files.txt` to create the same section in the binder docs.


docs/source/config_files.txt
```diff
@@ -0,0 +1,115 @@
...
+````````````````
+::
+
+   python-2.7
+
+``Dockerfile``
+^^^^^^^^^^^^^^
+
+This will be treated as a regular Dockerfile and a regular Docker build will be performed. The presence
+of a Dockerfile will cause all other building behavior to not be triggered.
```
betatim: How about: ...of a Dockerfile prevents all other build behaviour.

betatim:
Had a quick read and LGTM except for that small reformulation.

choldgraf:
edit made!

choldgraf:
anything else on this one?

yuvipanda:
Thanks, @choldgraf!

PR #169, opened 2017-12-16
Upgrade base packages

yuvipanda:
- Bump JupyterLab to latest
- Run freeze.py (conda) and freeze.bash (venv) to get new dependent
  packages, primarily newer ipykernel and ipywidgets
- Remove some over-specification in environment.yml

Fixes https://github.com/jupyterlab/jupyterlab/issues/3240


blink1073:
@yuvipanda, I can make a note in the JupyterLab release notes, but it would help if there were some upgrade notes here to point to for which files to update and which scripts to run to make a PR.

minrk:
Deployment has begun. I don't plan to bump the image prefix for this, which means any repos built before the deploy will still be considered valid with the previous version. This should be true in general: if a repo wants to ensure a particular version of jupyterlab/ipython/etc., it must specify that in its dependencies.

PR #172, opened 2017-12-19
Allow mounting arbitrary volumes into the repo2docker container

yuvipanda:
Tries to mimick the docker run '-v' syntax, but only supports
rw volumes for now. Source and destination can be relative paths.

You can then use '-v .:.' to mount current dir into $HOME. Also support
`--user-id` and `--user-name` cmdline parameters to set username and 
uid of the user created inside the container. This is required for volumes
to work seamlessly.

Note that this also switches the user id and username of the primary user
in the container image to be the same as the calling user, which is a pretty
big breaking change (but useful for users running this manually). Requires
modifications to binderhub before this can be deployed.


repo2docker/app.py
```diff
@@ -272,6 +296,15 @@ def initialize(self):
             self.run = False
             self.push = False
 
+        if args.volumes and not args.run:
+            # Can't mount if we aren't running
+            print("Can not mount volumes when not running container")
```
betatim: can we turn this into `To mount ("-v") directories you need to also run the container ("--run").` or some such?

repo2docker/app.py
```diff
@@ -301,6 +334,8 @@ def push_image(self):
 
     def run_image(self):
         client = docker.from_env(version='auto')
+        api_client = docker.APIClient(version='auto',
+                                  **docker.utils.kwargs_from_env())
```
betatim: more spaces needed

repo2docker/app.py
yuvipanda: Done. Default is to run, so I've modified it.

repo2docker/app.py
yuvipanda: done

AaronWatters:
FYI,  I tested this PR with
```
jupyter-repo2docker -v ~/misc/kiley_graim/kgraim:data ~/repos/canine_bc/
```
And the mount worked and was read/write.

betatim:
I read the doc string for `user_name` and `user_id` and ended up more confused than I was before as to what they should be set to. Can we make a general recommendation what to set them to for the average user?

repo2docker/app.py
```diff
@@ -200,6 +239,27 @@ def get_argparser(self):
...
+            '--volume', '-v',
+            dest='volumes',
+            action='append',
+            help='Volumes to mount inside the container, in form src:dest',
+            default=[]
+        )
+
+        argparser.add_argument(
+            '--user-id',
+            help='User id the primary user in the image',
```
betatim: -> "User ID _of_ the primary.."

repo2docker/app.py
```diff
@@ -200,6 +239,27 @@ def get_argparser(self):
...
+            default=1000,
+            type=int
+        )
+
+        argparser.add_argument(
+            '--user-name',
+            help='User name of primary user in the image',
```
betatim: -> "Username of the primary..."

repo2docker/app.py
```diff
@@ -272,8 +332,20 @@ def initialize(self):
...
+            print("To Mount volumes with -v, you also need to run the container")
+            sys.exit(1)
+
+        for v in args.volumes:
+            src, dest = v.split(':')
+            self.volumes[src] = dest
+
         self.run_cmd = args.cmd
 
+        self.user_id = int(args.user_id)
```
betatim: the conversion isn't taking care of by L254?

yuvipanda:
@betatim I changed the defaults so they run as your current userid /username, which is what users running this manually would most likely want. This means we've to explicitly set that to 1000 and jovyan in binderhub when launching tho.

betatim:
Cool! Yeah more complicated for binderhub but more convenient for CLI users is a trade off I like.

I will look at this again (and merge) when I get off the plane in Europe.

betatim:
Merging this now, but not deploying because I think we need https://github.com/jupyterhub/binderhub/pull/378 first.

PR #176, opened 2017-12-21
Add default command to base Dockerfile template

AaronWatters:
This specifies in the generated Dockerfile that the jupyter server should be the default application when the instance starts.  I tested this with `https://github.com/jakevdp/PythonDataScienceHandbook` and verified that the default app was bash before and jupyter after.

Full disclosure: I don't know fully understand the `buildpack` interrelationships and whether this change works for all the buildpacks.


betatim:
Nice! Does anyone have opinions on using exec vs shell style [`CMD`](https://docs.docker.com/engine/reference/builder/#cmd)?

There is also the legacy docker image, not sure if we should also edit that to have a default command or not. On the one hand it is legacy (don't edit it) on the other hand having different behaviour might confuse users. I'd be +0 to also modify the legacy docker file.

yuvipanda:
This is great! I think it should be the exec style variant (so `CMD ["jupyter", "notebook", etc]`) than the current one because sh doesn't do a great job of passing signals through to things it starts sometimes, causing TERM and KILL signals to take a while to pass through.

+1 to not modifying the legacy dockerfile, it should probably remain frozen and we should remove support for it at some point.

yuvipanda:
And congratulations on your first PR, @AaronWatters!

yuvipanda:
Friendly bump :)

choldgraf:
Second friendly bump :-)

What @yuvipanda is talking about by "exec style" is documented here: https://docs.docker.com/engine/reference/builder/#cmd

choldgraf:
ok made the exec-style change and resolved conflicts, LMK if that looks OK @yuvipanda

choldgraf:
ping - I think this is ready for merge, but I won't do it since I've now touched the PR. cc @betatim @yuvipanda who have commented on this PR earlier

betatim:
Congrats @AaronWatters, your first PR got merged!

PR #179, opened 2017-12-25
Fixed run argument check for mounting volumes. 

### LINKED ISSUE #178: Fix volume argument check when --no-build is given
When --no-build argument is given along with -v argument, repo2docker must exit with error instead,
it accepts the volume definition

Need to check against the instance attribute run instead of args.run

mukundans91:
Refactored argument validation test.
Closes #178


yuvipanda:
Ty for the patch, lgtm!

PR #180, opened 2017-12-25
add ability for port mapping to work with non jupyter-workflow

### LINKED ISSUE #100: Make port mapping work with non-jupyter workflows
If the image being built is a Dockerfile, it might expose arbitrary ports. Right now we don't do any port mappings for them.

instead, inspect the image that has been built, figure out what ports it says it needs & expose them!

This came up when working on getting R Studio to work here with @cboettig

mukundans91:
One question is I have removed the get random port method from repo2docker as docker-py supports assigning random ports if None is passed in the dict 
https://github.com/docker/docker-py/blob/190d95c61cc0440108de8a0c14027e72c56b6772/docker/models/containers.py#L596
Do we need to have the default run_cmd jupyter notebook started in the same port inside the container as outside? Is there are necessity for that?
Closes #100

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -30,7 +30,7 @@ from .buildpacks import (
     PythonBuildPack, DockerBuildPack, LegacyBinderDockerBuildPack,
     CondaBuildPack, JuliaBuildPack, Python2BuildPack, BaseImage
 )
-from .utils import execute_cmd, ByteSpecification, maybe_cleanup, is_valid_docker_image_name
+from .utils import execute_cmd, ByteSpecification, maybe_cleanup, is_valid_docker_image_name, validate_and_generate_port_mapping
 from . import __version__
 
 
@@ -265,6 +265,20 @@ class Repo2Docker(Application):
         )
 
         argparser.add_argument(
+            '--publish', '-p',
+            dest='ports',
+            action='append',
+            help='Specify port mappings for the image. Needs a command to run in the container.'
+        )
+
+        argparser.add_argument(
+            '--publish-all', '-P',
+            dest='all_ports',
+            action='store_true',
+            help='Publish all exposed ports to random host ports.'
+        )
+
+        argparser.add_argument(
             '--no-clean',
             dest='clean',
             action='store_false',
@@ -389,6 +403,21 @@ class Repo2Docker(Application):
 
         self.run_cmd = args.cmd
 
+        if args.all_ports and not self.run:
+            print('To publish user defined port mappings, the container must also be run')
+            sys.exit(1)
+
+        if args.ports and not self.run:
+            print('To publish user defined port mappings, the container must also be run')
+            sys.exit(1)
+
+        if args.ports and not self.run_cmd:
+            print('To publish user defined port mapping, user must specify the command to run in the container')
+            sys.exit(1)
+
+        self.ports = validate_and_generate_port_mapping(args.ports)
+        self.all_ports = args.all_ports
+
         if args.user_id:
             self.user_id = args.user_id
         if args.user_name:
@@ -427,15 +456,19 @@ class Repo2Docker(Application):
 
     def run_image(self):
         client = docker.from_env(version='auto')
-        port = self._get_free_port()
         if not self.run_cmd:
             port = str(self._get_free_port())
+
             run_cmd = ['jupyter', 'notebook', '--ip', '0.0.0.0',
                        '--port', port]
             ports = {'%s/tcp' % port: port}
         else:
+            # run_cmd given by user, if port is also given then pass it on
             run_cmd = self.run_cmd
-            ports = {}
+            if self.ports:
+                ports = self.ports
+            else:
+                ports = {}
         container_volumes = {}
         if self.volumes:
             api_client = docker.APIClient(
@@ -453,6 +486,7 @@ class Repo2Docker(Application):
 
         container = client.containers.run(
             self.output_image_spec,
+            publish_all_ports=self.all_ports,
             ports=ports,
             detach=True,
             command=run_cmd,
@@ -487,6 +521,7 @@ class Repo2Docker(Application):
         port = s.getsockname()[1]
         s.close()
         return port
+
 
     def start(self):
         if self.repo_type == 'local':
--- a/repo2docker/utils.py
+++ b/repo2docker/utils.py
@@ -3,6 +3,7 @@ from functools import partial
 import shutil
 import subprocess
 import re
+import sys
 
 from traitlets import Integer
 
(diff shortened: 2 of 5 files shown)


yuvipanda:
@mukundans91 yeah, we need to try to get the same port inside and outside since the notebook prints a URL at the end saying 'go to this URL to access your notebook', including a secret token in the URL. This only works if it's listening on the same port inside and outside...

mukundans91:
@yuvipanda thanks for the response. Will revert that part of the change...

mukundans91:
What is the normal way to resolve conflicts in PRs? Do you want me to rebase the changes or merge the changes from master to my branch?

betatim:
My preference is rebasing over merging. Don't think we have an official policy.

repo2docker/app.py
```diff
@@ -381,6 +395,21 @@ def initialize(self):
...
+
+        if args.port and not self.run:
+            print('To publish user defined port mappings, the container must also be run')
+            sys.exit(1)
+
+        if args.port and not self.run_cmd:
+            print('To publish user defined port mapping, user must specify the command to run in the container')
+            sys.exit(1)
+
+        self.port = validate_and_generate_port_mapping(args.port)
```
yuvipanda: This should be self.ports rather than self.port, no?

repo2docker/app.py
```diff
@@ -264,6 +264,20 @@ def get_argparser(self):
             help='Do not run container after it has been built'
         )
 
+        argparser.add_argument(
+            '--publish',
```
yuvipanda: Does this automatically make -p work or does that need to be explicitly added?

repo2docker/app.py
```diff
@@ -264,6 +264,20 @@ def get_argparser(self):
...
+            dest='port',
```
yuvipanda: Should be 'ports' since it's a list

mukundans91: my bad ... will change

repo2docker/app.py
mukundans91: my bad ... will change

repo2docker/app.py
mukundans91: -p must be explicitly added and will add it now and also rebase the code.

mukundans91:
@yuvipanda One part that I am doing in the tests that I would personally like to change is the time.sleep(30) waiting for the docker image to build and run ... Is there some other way that we can get a notification that the image has been built and is running without the use of time.sleep() ... some kind of notification?

mukundans91:
The tests failed because of the time.sleep() call ... it is not sufficient for the first time the tests is run ... 
later it becomes sufficient ... so one test failed ...

mukundans91:
my current plan for working around this, since I need the image to run to test if the port mapping is correct, is mount a volume and write a file to it with an echo command some thing like 
`repo2docker -p 52000:8000 -v /tmp/test:/home/test-user --user-id 501 --user-name test-user . /bin/bash -c "echo -n 'hi' > /home/mukundans/ts" && python -m http.server 8000`

Wait for the file to be created before checking if the port mapping works ...

repo2docker/utils.py
```diff
@@ -57,6 +58,76 @@ def maybe_cleanup(path, cleanup=False):
...
+                                    (?:6553[0-5]|655[0-2][0-9]|65[0-4](\d){2}|6[0-4](\d){3}|[1-5](\d){4}|(\d){0,4})
+                                    (?:/udp|/tcp)?
+                               )$''', re.VERBOSE)
+    ports = {}
+    if not port_mapping:
+        return None
+    for p in port_mapping:
+        if reg_regex.match(p) is None:
+            print('Invalid port mapping ' + str(p))
+            sys.exit(1)
```
yuvipanda: In general, I don't think functions should call sys.exit - that makes them really hard to re-use in other contexts. IMO, this should raise an exception that the caller should catch - you can't really catch sys.exit.

tests/ports.py
```diff
@@ -0,0 +1,119 @@
...
+                                 '/bin/bash', '-c', 'echo \'hi\' > /home/ts && python -m http.server 8000'],
+                                 cwd=builddir + "/../",
+                                 stderr=subprocess.STDOUT)
+        try:
+            while not os.path.exists(os.path.join(tmpdir, 'ts')):
+                pass
+            # Wait till docker builds image and starts up
+            # Sleep to wait for python http server to start
+            time.sleep(20)
+            http = urllib3.PoolManager()
```
yuvipanda: We already have requests library available. Is there a reason You're using urllib3 directly instead of using requests?

tests/ports.py
```diff
@@ -0,0 +1,119 @@
...
+    host_port = port
+    if host:
+        host_port = host + ':' + port
+    else:
+        host = 'localhost'
+    with tempfile.TemporaryDirectory() as tmpdir:
+        username = os.getlogin()
+
+        proc = subprocess.Popen(['repo2docker',
+                                 '-p',
```
yuvipanda: The other option here would be to figure out a way to specify the container name in repo2docker, but I think for now this is sufficient. Perhaps add a comment explaining why this is here?

repo2docker/utils.py
mukundans91: Changed to raise Exception

tests/ports.py
mukundans91: Could you explain a bit more on what I need to explain? Is it the part about the volumes?

tests/ports.py
mukundans91: Both requests and urllib3 are not used anywhere in the project and not specified in the setup.py requirements.

yuvipanda: Ah, I think they're brought in as dependencies of some other library we use (docker maybe?)

tests/ports.py
yuvipanda: Just a general comment explaining what is going on here :) We're doing a bunch of things - using a volume to signal when something is done building, keeping a child process running (and killing it after, etc).

mukundans91: Ah ok ...

mukundans91:
the tests are failing intermittently ...

yuvipanda:
I restarted the conda one a few times, failed twice. Let's see what happens third time.

mukundans91:
+ [[ -f /tmp/kernel-environment.yml ]]
+ /srv/conda/bin/conda env create -n kernel -f /tmp/kernel-environment.yml
Fetching package metadata .............
ResolvePackageNotFound: 
  - pyzmq 16.0.2 py27_2

yuvipanda:
Yeah, seems completely unrelated to your patch. https://github.com/jupyter/repo2docker/pull/193 should hopefully fix it - although I've no idea how such breakage can happen in conda, since we explicitly pin versions.

mukundans91:
rebased after #193 merge ... tests pass now .. :)

yuvipanda:
w00t, ty for your patience!

PR #181, opened 2017-12-31
Fix indentation causing only last line of apt.txt to be parsed

nmih:
Parsing apt.txt was only appending the last package to the list of packages to install. Fixed indentation to append all packages.

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -443,7 +443,7 @@ class BaseImage(BuildPack):
                     if not re.match(r"^[a-z0-9.+-]+", package):
                         raise ValueError("Found invalid package name {} in "
                                          "apt.txt".format(package))
-                extra_apt_packages.append(package)
+                    extra_apt_packages.append(package)
 
             assemble_scripts.append((
                 'root',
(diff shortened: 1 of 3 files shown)


betatim:
Whoops! Thanks for making this your first contribution to repo2docker :)!

Could you also add a test to make sure this bug doesn't come back? Maybe copy [this test](https://github.com/jupyter/repo2docker/blob/f7096caccddf333b84abcc0e308f45e2299aa146/tests/venv/apt-packages/apt.txt) and use an apt.txt with two or more lines, then check that both bot installed.

nmih:
Sure thing, done!

betatim:
`rolldice` and `sl` get installed in `/usr/games` which doesn't seem to be on the `PATH` so the test fails. Does someone with more ubuntu experience know if this is on purpose/how to deal with this?

Otherwise we could pick two other packages to install maybe.

nmih:
Ah my bad, didn't realize that was the case. I changed it to two other small packages that install to /usr/bin instead for now.

betatim:
LGTM! Merging.

betatim:
Welcome to another first time contributor!

PR #182, opened 2018-01-03
conda: separate base envs by Python version

### LINKED ISSUE #28: handle Python 2 in environment.yml
conda can specify the Python package version in environment.yml. Right now, if the user specifies a requirement for Python 2, this will just break.

We should detect if Python 2 is in environment.yml and install to a new env for the kernel, rather than updating the root env. In fact, it might be preferable to do this if the Python version is specified as anything other than the already-installed version.

minrk:
downgrading the root env from Python 3.6 to 2.7 (or even 3.5 in some cases) is always very expensive and can fail due to the complicated transaction.

This fixes issues where builds fail when Python 2 is requested.

<del>
What I'd like to do eventually, and we will *need* to do once certain packages start dropping py2 support, is to handle Python 2 differently by putting it in a different env, like we do for pip. But for now, it's simpler to handle Python 2 the same as Python 3.
</del>

Python 2 is handled specially by putting it in a different env, just like with requirements.txt.

closes #28


yuvipanda:
Thanks @minrk!

PR #183, opened 2018-01-03
Stop using alpine base images

yuvipanda:
- It is currently breaking the build, since alpine does not
  support binary wheels & we introduced ruamel.yaml which has
  binary dependencies
- pip / python on alpine is broken wrt symlinks, making local
  testing harder

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM python:3.6.3-alpine3.6
+FROM python:3.6.3
 
 # Git is required for repo2docker to work
-RUN apk add --no-cache git
+#RUN apk add --no-cache git
 
 RUN mkdir /tmp/src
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
-FROM python:3.6.3-alpine3.6
+FROM python:3.6.3
 
 # Git is required for repo2docker to work
-RUN apk add --no-cache git
+#RUN apk add --no-cache git
```
willingc: any reason that we wouldn't remove this line (and the one above) vs. commenting it out. Are we going to reenable after the merge?

yuvipanda: @willingc good catch, I've removed those!

willingc:
Thanks @yuvipanda

Issue #32, opened 2017-06-27
Support custom environment variables


choldgraf:
We should let users define environment variables w/ their repositories.

Maybe `env.txt` ? Each line should have `<KEY>=<VAL>`?

There are two components to this:

- [x] Make it possible to invoke repo2docker with environment variables (#186 )
- [x] Document the above CLI functionality (#295)
- [ ] Add support to perform this action with a config file (e.g. `env.txt`)

yuvipanda:
In general I think we should find whatever heroku does and copy it.

yuvipanda:
Heroku uses a `.env` file.

sbinet:
apologies in advance for bumping this issue, but I ran into this issue as well while converting my old `Dockerfile`-based repo to the new recommanded way (_ie:_ `repo2docker`)

My repo uses Go and its Jupyterkernel, so I had to install both.
using `postBuild` and `apt.txt` worked fine during the docker image creation.

but when actually running the notebook, I also need to have a few Go related environment variables available (`$GOPATH` and a couple of entries in `$PATH`)

it would be valuable to just have a simple declarative way to tell the build system what are the variables to be inherited for the notebook.

choldgraf:
so per @yuvipanda 's note above, and from https://devcenter.heroku.com/articles/heroku-local#look-at-the-contents-of-your-env-file, 

we could support a file that was just `.env` and had a structure like:

```
S3_KEY=mykey
S3_SECRET=mysecret
```

and it would get inserted at the beginning of the dockerfile

choldgraf:
another place where this could be useful: I'm apt-installing some things that get installed to a folder that's not in the typical path. I haven't figured out how to update a path via `postBuild` so perhaps this could be used for that as well

rprimet:
Before ruling on a file format, maybe a switch could be added to the `repo2docker` command, e.g. `-e` ?

choldgraf:
I propose that we just go with supporting a file called `.env` that has `{key}={value}` pairs, and that will be inserted into the Dockerfile before anything else. It doesn't seem like any other major alternatives have popped up.

betatim:
I think there are two use cases:
* using `repo2docker` as a CLI tool I would like to be able to insert AUTH tokens at launch time (like `docker run -e`)
* add permanent ENV variables in a binderhub use case (eg. connection string for a DB)

So I'd champion a command-line option and some file based solution like `.env`/`binder.yaml`.

[...]

PR #186, opened 2018-01-07
Provide a flag to pass environment variables at runtime

rprimet:
Submitted in partial fulfillment for #32 

I'm wet-behind-the-ears when it comes to Python; I'd appreciate constructive criticism :-)

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -300,7 +300,7 @@ class Repo2Docker(Application):
         argparser.add_argument(
             '--env', '-e',
             dest='environment',
-            nargs='+',
+            action='append',
             help='Environment variables to define at container run time',
             default=[]
         )
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -297,6 +297,14 @@ class Repo2Docker(Application):
             help='Username of the primary user in the image',
         )
 
+        argparser.add_argument(
+            '--env', '-e',
+            dest='environment',
+            nargs='+',
+            help='Environment variables to define at container run time',
+            default=[]
+        )
+
         return argparser
 
     def json_excepthook(self, etype, evalue, traceback):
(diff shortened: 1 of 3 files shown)


betatim:
Yay! Just had a quick look at your PR and I think you are lowering expectations with your "wet-behind-the-ears" comment. Looks pretty solid to me! I will take a closer look tomorrow morning/when travis is done.

One thing off the top of my head: we had some problem with the repo path and the command-to-run getting in the way of each other. Basically argparse parsing malarkey :-/ So yeah I'd poke around there a bit. Otherwise this looks solid.

rprimet:
So... CI fails and I'm not sure why. Any help? Otherwise I'll just close the PR.

betatim:
To find a test that is failing (in the huge amount of output in the log) I search for "FAILED". In this case it gives me https://travis-ci.org/jupyter/repo2docker/jobs/326160395#L1192 looks like the failure isn't your fault. It mentions that it can't resolve a package called wheel, with a certain version. 

I get similar error locally when trying to run `conda install "wheel=0.30.0"`, seems the latest version conda wants to install is 0.29.0. However conda-forge has 0.30.0.

betatim:
I'd go investigate the `repo2docker/buildpacks/conda/environment.py-3.6.frozen.yml` and friends to understand why wheel 0.30.0 is unresolved despite existing on conda-forge and that channel being selected. @minrk recently did some work on that so you could also ask him for advice.

repo2docker/app.py
```diff
@@ -297,6 +297,14 @@ def get_argparser(self):
             help='Username of the primary user in the image',
         )
 
+        argparser.add_argument(
+            '--env', '-e',
+            dest='environment',
+            nargs='+',
```
yuvipanda: action='append' is probably better here than nargs='+'

rprimet:
@yuvipanda : changed in d5d9b89 but there is a new CI failure -- I can't tell if it's spurious or not.

rprimet:
The change to `action='append'` is done

rprimet:
Cool, thanks for the merge!

betatim:
🎉 !

PR #187, opened 2018-01-08
refreeze conda environments

minrk:
gets a few updates, but fixes issue with noarch wheel package that has been identified as broken.

ref: https://github.com/conda-forge/wheel-feedstock/pull/9


minrk:
I think conda environments are fully broken right now, so deploying this promptly is probably important.

minrk:
This should fix the remaining failures in #186

minrk:
Additionally, this explicitly allows conda-forge packages that we've frozen to be installed even if they've been marked as broken, so this shouldn't be able to happen again. This only applies to those in the `environment.frozen.yml`.

PR #190, opened 2018-01-10
[WIP] Use pyenv to manage base python versions

### LINKED ISSUE #146: runtime.txt won't install other python 3 versions

yuvipanda:
pyenv allows us to install arbitrary python versions without
having to be pinned to the distro. This separates the reproducibility
concern of packages from the concerns of python versions. As a bonus,
it also expands vastly the things that can be in runtime.txt - there
is test cases included for pypy, and 3.6! You can even install HEAD.

This now has the clear concept of a base install, which is just
notebook setup, and then the user's install, which is whatever
version is specified in runtime.txt. `python` will point to the
version specified in runtime.txt.

Since pyenv continues to be installed, users can install new versions of python
easily at runtime too.  

Fixes #146 

This is [WIP] and a lot of questions need answering, such as:

- Do we always need a separate 'base' install? Or only for specific values of runtime.txt?

  We implement a whitelist now, and if those runtimes are used then we do not need a separate base install!

- [ ] What's the performance impact of this?

  A little slower, but caching should make up for it.

- [ ] Are there security impacts here we need to be aware of?
- [ ] Will this break user's current code / repos?


yuvipanda:
Biggest test failure seems to be caused by bumping python version to 3.6.4 - that is causing the Data Science Handbook to fail, due to https://github.com/jakevdp/PythonDataScienceHandbook/issues/108. I might revert the python version bump for now maybe.

yuvipanda:
Questions to answer:

1. Making the CondaBuildPack inherit from the Python one. The benefits seem minimal, so I've left it as is. See https://github.com/jupyter/repo2docker/pull/190/commits/e4356bb3bf3a7652930630cd93ea0d3000eda0b8 for a possible way to do this (I've reverted that commit since)
2. Mixing Traitlets and Inheritance seems to be a recipe for disaster. I filed https://github.com/jupyter/repo2docker/issues/191 to consider getting rid of traitlets. I'd like to do that outside of this commit, though
3. Should we blacklist things like Jython, IronPython, Stackless Python, etc? Are folks using them?
4. Should we switch the default to python3.6 even if that means popular things like Jake's textbook will break?

betatim:
I think for bumping to 3.6 by default and breaking the DS handbook: it means one of us needs to update the handbook to make it work on 3.6 or contribute the required `runtime.txt` to pin it to 3.5 for the moment. Same really goes for other repositories we end up breaking by doing this :-/

IMHO what this shows is that we need a policy and technical means for deprecating stuff. My take on our reproducibility goals is "stuff will eventually start breaking, the only way to keep it working in perpetuity is to keep updating/maintaining things". This is motivated by the fact that we/repo2docker/binderhub needs to have a way to fix mistakes made in the past by deprecating things. As well as external reasons for changing behaviour (security vulnerabilities).

My proposal: if we make backwards incompatible/breaking changes we support the old behaviour for 6-9months while printing a warning stating "this will go away on X, read Y to update". This warning has to be more prominent than the build log though??

yuvipanda:
I think we might need a custom start script here, and I think I'll maybe use this opportunity to do #196 too.

repo2docker/buildpacks/base.py
```diff
@@ -358,6 +387,8 @@ def render(self):
             build_script_files=self.build_script_files,
             base_packages=sorted(self.base_packages),
             post_build_scripts=self.post_build_scripts,
+            entrypoint_json=json.dumps(self.entrypoint),
+            command_json=json.dumps(self.command)
```
minrk: minor style nit: including trailing commas on lists/dicts gives a better diff when adding/removing items on the end of things.

repo2docker/buildpacks/conda/__init__.py
```diff
@@ -26,6 +26,9 @@ class CondaBuildPack(BuildPack):
 
     path = ['${CONDA_DIR}/bin']
 
+    entrypoint = ['/srv/conda/bin/jupyter']
+    command = ['notebook', '--ip', '0.0.0.0']
```
minrk: Splitting one command across entrypoint and command seems unusual. What's the benefit?

yuvipanda:
Going to close this per #185

PR #200, opened 2018-01-31
Add nteract jupyter extension to requirements.txt based setups

yuvipanda:
Didn't add to conda, since there is currently no way to add
pip packages to our conda frozen setup (due to
https://github.com/conda/conda/pull/6438).

This means nteract will be available only when used with
requirement.txt, and not when used with environment.yml files.


minrk:
It's actually [our code](https://github.com/jupyter/repo2docker/blob/bb309da2d3e4647209c942f3f28d0592a92e6e9f/repo2docker/buildpacks/conda/freeze.py#L42) that's preventing this due to being the simplest workaround for that conda bug, since we didn't have any pip dependencies at the time. I can fix it in a subsequent PR.

minrk:
conflicts with the buildpack refactor

yuvipanda:
Rebased!

PR #204, opened 2018-02-01
Remove outdated builder info from docs

yuvipanda:
These are 100% wrong and misleading, so we should remove them


choldgraf:
holy moly - you are right, I didn't realize the docs were _that_ far out of date

PR #206, opened 2018-02-01
Refactor (most) traitlets out of buildpacks

yuvipanda:
This is a 'pure' refactor in the traditional sense of the word. We will
change the internals of how buildpacks are done from Traitlets
to using simple methods in classes + (possibly) inheritance,
but will not touch our tests - this should give us a high degree
of confidence that it won't change behavior. This will thus
be a straight port of current behavior from traitlets to classes,
with no new added functionality or ancillary cleanup.

See #191 for more discussion.

The two traitlets left are `name` and `components`. The former is 
used for logging and the latter for chaining `detect` methods together.
After this PR is merged, I'll make another one that replaces our
composing with simple linear inheritance. At that point, we can get rid
of both these traitlets. I wanted to make this PR be purely 'mechanical'
refactors, to make it easier to review.

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -183,12 +183,11 @@ class BuildPack(LoggingConfigurable):
         """
         return []
 
-    labels = Dict(
-        {},
-        help="""
+    def get_labels(self):
+        """
         Docker labels to set on the built image.
         """
-    )
+        return {}
 
     build_script_files = Dict(
         {},
@@ -283,15 +282,12 @@ class BuildPack(LoggingConfigurable):
         for resolving them.
         """
         result = BuildPack(parent=self)
-        labels = {}
-        labels.update(self.labels)
-        labels.update(other.labels)
-        result.labels = labels
         # FIXME: Temporary hack so we can refactor this piece by piece instead of all at once!
         result.get_packages = lambda: self.get_packages().union(other.get_packages())
         result.get_base_packages = lambda: self.get_base_packages().union(other.get_base_packages())
         result.get_path = lambda: self.get_path() + other.get_path()
         result.get_env = lambda: self.get_env() + other.get_env()
+        result.get_labels = lambda: self.get_labels() + other.get_labels()
 
         result.build_scripts = self.build_scripts + other.build_scripts
         result.assemble_scripts = (self.assemble_scripts +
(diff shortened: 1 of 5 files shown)


yuvipanda:
All tests are green!

repo2docker/buildpacks/base.py
```diff
@@ -286,25 +278,24 @@ def compose_with(self, other):
...
-        result.build_script_files = build_script_files
+        # FIXME: Temporary hack so we can refactor this piece by piece instead of all at once!
+
+        def _merge_dicts(d1, d2):
+            md = {}
+            md.update(d1)
+            md.update(d2)
+            return md
+
+        result.get_packages = lambda: self.get_packages().union(other.get_packages())
```
minrk: Do you think it's appropriate to define a `ComposedBuildPacks` rather than redefining it at runtime with lambdas here?

repo2docker/buildpacks/conda/__init__.py
```diff
@@ -68,10 +70,9 @@ def setup_build_script_files(self):
         files['conda/' + frozen_name] = '/tmp/environment.yml'
         return files
 
-    python_version = Unicode()
-    @default('python_version')
-    def detect_python_version(self):
-        """Detect the Python version for a given environment.yml
+    def get_python_version(self):
```
minrk: We benefitted from the fact that traitlets cached results (like properties, but only running once). We should be careful to do the same here. For instance, the Python version involves opening and reading a YAML file. That shouldn't happen more than once, so we should make sure that this only runs once, and the computed value is used thereafter.

repo2docker/buildpacks/base.py
yuvipanda: Nah, I think we should just wait and move directly to inheritance in another PR (pretty soon after this)

repo2docker/buildpacks/conda/__init__.py
yuvipanda: Yeah, I tried to do that but sortof failed (since I was being very mechanical about it), and reverted my changes. I'll give this another shot!

yuvipanda: Done!

repo2docker/buildpacks/base.py
minrk: ok

PR #207, opened 2018-02-01
Use inheritance to define composed buildpacks

yuvipanda:
- This is much cleaner to understand than traitlets
- Remove 'name' and 'component' traitlets

Depends on #206 

After this, only thing left to do is figure out what to do about LoggingConfigurable.

### DIFF
--- a/repo2docker/buildpacks/python/__init__.py
+++ b/repo2docker/buildpacks/python/__init__.py
@@ -1,7 +1,6 @@
 """
 Generates a variety of Dockerfiles based on an input matrix
 """
-from traitlets import default
 import os
 from ..base import BaseImage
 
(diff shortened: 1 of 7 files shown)


minrk:
> only thing left to do is figure out what to do about LoggingConfigurable.

That shouldn't be too much. We can pass the logger in the constructors (`BuildPack(log=log)`), or use a global named logger:

```python
log = logging.getLogger('repo2docker')
```

repo2docker/buildpacks/python/__init__.py
```diff
@@ -154,5 +152,5 @@ def detect(self):
             with open(runtime_txt) as f:
                 runtime = f.read().strip()
             if runtime == 'python-2.7':
-                return True
+                return True and super().detect()
```
minrk: `True and`?

yuvipanda: Done!

yuvipanda:
Also fixed the logger, I think.

minrk:
Everything's working but Julia

PR #208, opened 2018-02-01
add nteract-on-jupyter to conda

minrk:
via pip

Removes workaround for conda env export bug by updating conda to 4.4 in the freeze container.

equivalent of #200 for conda

### DIFF
--- a/repo2docker/buildpacks/conda/install-miniconda.bash
+++ b/repo2docker/buildpacks/conda/install-miniconda.bash
@@ -30,6 +30,9 @@ ${CONDA_DIR}/bin/conda config --system --set auto_update_conda false
 echo 'update_dependencies: false' >> ${CONDA_DIR}/.condarc
 ${CONDA_DIR}/bin/conda config --system --set show_channel_urls true
 
+# disable pip cache for conda-triggered pip installs
+export PIP_CACHE_DIR=0
+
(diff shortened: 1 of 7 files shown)


minrk:
This grew as a found a few issues. In particular, I found https://github.com/conda/conda/issues/6824, which prevents the pip step from working after switching Pythons in the same go.

I also did a bit of tidying of the conda install while at it.

repo2docker/buildpacks/conda/install-miniconda.bash
```diff
@@ -20,26 +20,43 @@ if ! echo "${MD5SUM}  ${INSTALLER_PATH}" | md5sum  --quiet -c -; then
...
 echo 'update_dependencies: false' >> ${CONDA_DIR}/.condarc
-${CONDA_DIR}/bin/conda config --system --set show_channel_urls true
 
-${CONDA_DIR}/bin/conda env update -n root -f /tmp/environment.yml
+echo "installing root env:"
+cat /tmp/environment.yml
+conda env update -f /tmp/environment.yml
+
+# enable nteract-on-jupyter, which was installed with pip
+jupyter serverextension enable nteract_on_jupyter --sys-prefix
```
yuvipanda: Tests are failing with 'jupyter command not found'. Perhaps the path needs to be explicit?

yuvipanda:
Thanks, @minrk!

PR #210, opened 2018-02-02
Add native R + IRKernel + RStudio support

### LINKED ISSUE #24: Support `install.R` files
We should support R functionality. R handles dependencies differently from both Python and Julia, here are some thoughts from Carl:

Current proposal
---
* Use `runtime.txt` to specify the R version as a date.
  * We'll expect a line with the format `r-YYYY-MM-DD`. E.g. `2017-01-21`. 
  * If that line is found, it will trigger and R build in r2d. This sets up a version of R that was current at that date, and sets the MRAN snapshot for that date as the default repo.
  * We'll also set up an R kernel for jupyter to discover. This R will also make sure we are installing in a user-owned library path so stuff can be installed there at any point.
* Use `install.R` (must be executable) to execute with the installed R kernel (in the point above). This primarily lets people write R code that installs packages.
* If `install.R` is given without a `r-YYYY-MM-DD` line in `runtime.txt` then we raise an error.

Main to-do items
---
- [x] How do users specify what servers contain what packages (e.g. rstudio vs. bioconductor)
- [ ] ~What about non-CRAN packages?~ outside of the standard R repositories, we can't do much else other then tell people to clone a github repo of preference and install it with `install.R` or `postBuild` or something.
- [x] How can we ensure versions are frozen, since R doesn't easily handle specific version installs?
- [ ] Someone needs to actually implement this!

Notes
---
> In an R package, the DESCRIPTION file plays the role of a `requirements.txt` in stating the dependencies, minimal version needed, and where get them (e.g. CRAN or additional cran-type repo like bioconductor). 
 >
> This approach does not accommodate installing something that is not the most recent version of a package.  (CRAN archives old sources, but because, unlike python or ruby gems distribution, CRAN is designed to provide binaries & you can't guarantee binaries build for an old /archived source, the default install does not immediately support installing archived packages).
>
> If you just have a list of packages you want, I recommend something along the lines of what we do with rocker, e.g.
>
>    install2.r `cat deps.txt`
>
> Where deps.txt is just a list of package names you want to install.  If these come from multiple repos (cran & bioconductor), just list those as arguments to -r:
>
>    install2.r -r "https://cran.rstudio.com" -r "https://bioconductor.org/pagkages/release" `cat deps.txt`
>
> If you want to install the same version each time, just use an MRAN snapshot of the appropriate date.

yuvipanda:
This sets up R + RStudio + IRKernel for a repository that contains:

    1. A `runtime.txt` file with the text:

        r-<year>-<month>-<date>

        Where 'year', 'month' and 'date' refer to a specific
        date snapshot of https://mran.microsoft.com/timemachine
        from which libraries are to be installed.

    2. An optional `install.R` file that will be executed at build time,
        and can be used for installing packages from both MRAN and GitHub.

It currently sets up R from the ubuntu repository being used. This
is unideal, and we should investigate other solutions!

Thanks to @cboettig and @taylorreiter for helping nail down the format!

Fixes #24


yuvipanda:
Unfortunately, this PR depends on #207, so the diff should be clearer once that gets merged!

yuvipanda:
Rebased for cleaner diffs!

yuvipanda:
And tests pass now!

repo2docker/buildpacks/r.py
```diff
@@ -0,0 +1,156 @@
...
+        r-<year>-<month>-<date>
+
+        Where 'year', 'month' and 'date' refer to a specific
+        date snapshot of https://mran.microsoft.com/timemachine
+        from which libraries are to be installed.
+
+     2. An optional `install.R` file that will be executed at build time,
+        and can be used for installing packages from both MRAN and GitHub.
+
+    It currently sets up R from the ubuntu repository being used. This
```
willingc: R from the ubuntu repository being used. I'm not sure what that means.

repo2docker/buildpacks/r.py
```diff
@@ -0,0 +1,156 @@
...
+            else:
+                self._checkpoint_date = datetime.date(*[int(s) for s in match.groups()])
+
+        return self._checkpoint_date
+
+    def detect(self):
+        """
+        Check if current repo should be built with the R Build pack
+
+        Note that we explicitly do *not* check if a requirements.txt
```
willingc: I'm confused as I'm not sure where the super().detect() call is. I'm assuming it is called by the parent `PythonBuildPack`. Would be nice to clarify in the docstring.

If I am a user that uses conda, I'm assuming the `CondaBuildPack` can set up R.

yuvipanda: I'll try to clarify in another commit. The point I was trying to make is that we explcitly do *not* call super().detect() here, since that will fail unless there is a requirements.txt file. It's not very clear here, will redo!

And yes, conda has R support with environment.yml so no extra steps are needed. Mixing conda's R and other R probably causes problems.

repo2docker/buildpacks/r.py
yuvipanda: Will clarify (it means R is installed via 'apt-get install r-base', rather than any of the methods available from https://cran.r-project.org/)

willingc: OK that makes sense. Thanks @yuvipanda

repo2docker/buildpacks/r.py
willingc: I got that we were not calling super().detect() explicitly. I didn't get why ;-). 

Perhaps something like:

Note: super().detect() is not called in this function. Calling it here
would fail unless a `requirements.txt` file exists, and that is not
a guaranteed case.

> Mixing conda's R and other R probably causes problems.

I would imagine that it would.

yuvipanda:
@willingc I cleaned up and wrote more detailed docstrings (finding a bug in the process!), hope things are clearer now! I'll slowly add this level of docstrings to all our BuildPacks.

repo2docker/buildpacks/r.py
```diff
@@ -59,23 +59,48 @@ def detect(self):
         """
         Check if current repo should be built with the R Build pack
 
-        Note that we explicitly do *not* check if a requirements.txt
-        is present here (by calling super().detect()).
+        super().detect() is not called in this function - it would return false
```
willingc: Outstanding!

brooksambrose:
Awesome! I’m assuming it will be the latest version of R, or can that be specified?

yuvipanda:
@brooksambrose I think I'll need to add a follow up PR that looks at the date specified in runtime.txt, and picks the version of R that was latest at that time. Currently you are tied to R 3.4.2.

willingc:
> I'll slowly add this level of docstrings to all our BuildPacks.

I think this is something that the team can divide and conquer as well. Possibly more effective to have the team write any needed docstrings and you verify that the understanding was correct. That gives 2 benefits: extra review of code; helps us get more understanding of the internals.

yuvipanda:
@willingc good idea! I filed https://github.com/jupyter/repo2docker/issues/211 for that.

betatim:
Can you add a short section to https://repo2docker.readthedocs.io/en/latest/usage.html#supported-configuration-files for this?

repo2docker/buildpacks/r.py
```diff
@@ -0,0 +1,211 @@
...
+        return self._runtime
+
+    @property
+    def checkpoint_date(self):
+        """
+        Return the date of MRAN checkpoint to use for this repo
+
+        Returns '' if no date is specified
+        """
+        if not hasattr(self, '_checkpoint_date'):
```
betatim: If you wanted to get rid of the repetitive `hasattr()` tests in all the properties you could use [`@lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache):
```python
from functools import lru_cache as cache

@property
@cache(1)
def checkpoint_date(self):
    match = ...
```

There might be a neat way to turn `@cache(1)` into just `@cache` via `cache = functools.partial(lru_cache, 1)`, but maybe that is too much magic :-/

choldgraf: IMO using @cache etc is pretty opaque compared with just using `hasattr`

I just pushed a PR to add a short README to the R test, which should make it show up in the `supported configuration files` section when the docs are built.

I'm +1 on merge - I think we should get this into the wild so we can start spot-checking improvements

choldgraf:
I just pushed a PR to add a short README to the R test, which should make it show up in the `supported configuration files` section when the docs are built.

I'm +1 on merge - I think we should get this into the wild so we can start spot-checking improvements

tests/r/simple/README.rst
```diff
@@ -0,0 +1,10 @@
+R environment - install.R
+-------------------------
+
+You can install an R environment with the following two files:
+
+* ``install.R``: a script that will be run from an R installation. This is
```
willingc: ~~from an R installation~~ using R to install and set up specified packages.

willingc:
Let's do a quick update to the README @choldgraf and then merge. We can explore @betatim's suggestion in a future issue/PR 👍

choldgraf:
@willingc updated! LMK

choldgraf:
yay! will do so when travis is happy

betatim:
Opaque ... tzz, you just don't know the python stdlib inside out do you ;)

Issue #211, opened 2018-02-03
Write detailed docstrings for all current BuildPacks


yuvipanda:
Currently, the following BuildPacks could use better docstrings:

- [x] PythonBuildPack
- [x] Python2BuildPack
- [x] CondaBuildPack
- [x] JuliaBuildPack PR #213
- [x] DockerBuildPack
- [x] LegacyDockerBuildPack

We should look at each of them, and validate that the docstrings are at least as detailed as in the PR #210

yuvipanda:
Quoting @willingc from #210:

> I think this is something that the team can divide and conquer as well. Possibly more effective to have the team write any needed docstrings and you verify that the understanding was correct. That gives 2 benefits: extra review of code; helps us get more understanding of the internals.

Am happy to review them if people wanna take a shot!

PR #213, opened 2018-02-05
Add docstrings and minor style fixes for application files and JuliaBuildPack

willingc:
Partially addresses #211

Covers repo2docker application files:
- `app.py`
- `utils.py`
- `__main__.py`

And `JuliaBuildPack`

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -46,7 +46,7 @@ class Repo2Docker(Application):
 
     @default('log_level')
     def _default_log_level(self):
-        """The system's default log level"""
+        """The application's default log level"""
         return logging.INFO
 
     git_workdir = Unicode(
@@ -484,7 +484,7 @@ class Repo2Docker(Application):
                 last_emit_time = time.time()
 
     def run_image(self):
-        """Run docker image"""
+        """Run docker container from built image"""
         client = docker.from_env(version='auto')
         if not self.run_cmd:
             port = str(self._get_free_port())
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -23,7 +23,7 @@ from docker.errors import DockerException
 import escapism
 from pythonjsonlogger import jsonlogger
 
-from traitlets import Unicode, List, default, Any, Dict, Int
+from traitlets import Any, Dict, Int,  List, Unicode, default
 from traitlets.config import Application
 
 from . import __version__
@@ -39,12 +39,14 @@ from .utils import (
 
 
 class Repo2Docker(Application):
+    """An application for converting git repositories to docker images"""
     name = 'jupyter-repo2docker'
     version = __version__
     description = __doc__
 
     @default('log_level')
     def _default_log_level(self):
+        """The system's default log level"""
         return logging.INFO
 
     git_workdir = Unicode(
@@ -52,7 +54,7 @@ class Repo2Docker(Application):
         config=True,
         allow_none=True,
         help="""
-        Working directory to check out git repositories to.
+        Working directory to use for check out of git repositories.
 
         The default is to use the system's temporary directory. Should be
         somewhere ephemeral, such as /tmp.
@@ -71,7 +73,7 @@ class Repo2Docker(Application):
         ],
         config=True,
         help="""
-        Ordered list of BuildPacks to try to use to build a git repository.
+        Ordered list of BuildPacks to try when building a git repository.
         """
     )
 
@@ -79,7 +81,7 @@ class Repo2Docker(Application):
         PythonBuildPack,
         config=True,
         help="""
-        The build pack to use when no buildpacks are found
+        The default build pack to use when no other buildpacks are found.
         """
     )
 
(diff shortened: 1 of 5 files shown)


repo2docker/app.py
```diff
@@ -444,6 +463,7 @@ def push_image(self):
                 last_emit_time = time.time()
 
     def run_image(self):
+        """Run docker image"""
```
yuvipanda: "Run docker container from built image" perhaps?

repo2docker/buildpacks/julia.py
```diff
@@ -1,18 +1,27 @@
...
+    since Julia does not work with Python virtual environments.
+    See https://github.com/JuliaPy/PyCall.jl/issues/410
 
-    Julia does not work with Virtual Envs,
-    see https://github.com/JuliaPy/PyCall.jl/issues/410
     """
     def get_env(self):
+        """Get additional environment settings for Julia and Jupyter
+
+        Returns:
```
yuvipanda: I think the docstrings here should explain what exactly is being returned, since the generic docstring for what get_env is returning is already in the base BuildPack. I think explaining each of the env variable set here and why we do it would be the useful thing here.

- JULIA_PATH - base path where all Julia Binaries and libraries will be installed
- JULIA_HOME - path where all Julia Binaries will be installed
- JULIA_PKGDIR - path where all Julia libraries will be installed
- JULIA_VERSION - default version of julia to be installed
- JUPYTER - environment variable required by IJulia to point to `jupyter` executable

repo2docker/buildpacks/julia.py
```diff
@@ -22,9 +31,30 @@ def get_env(self):
         ]
 
     def get_path(self):
+        """Get path for Julia executables
+
+         Returns:
```
yuvipanda: Adds path to Julia binaries to user's PATH.

Also we can simplify this code to just be:

```
return super().get_path() + ['${JULIA_HOME}']
```

but also ok to not do that!

repo2docker/buildpacks/julia.py
```diff
@@ -22,9 +31,30 @@ def get_env(self):
...
     def get_build_scripts(self):
+        """
+        Return series of build-steps common to "ALL" Julia repositories
+
+        All scripts found here should be independent of contents of a
+        particular repository.
+
+        This sets up:
+
+        - a directory for the specified Julia version and path
```
yuvipanda: It installs Julia itself, and also installs IJulia (`Pkg.add` sounds like you are adding to something, but it really is installing the package)

repo2docker/buildpacks/julia.py
```diff
@@ -52,11 +82,19 @@ def get_build_scripts(self):
         ]
 
     def get_assemble_scripts(self):
+        """
+        Return series of build-steps specific to "this" Julia repository
+
+        Precompile all Julia libraries found in the repository's REQUIRE
```
yuvipanda: +1!

repo2docker/buildpacks/julia.py
```diff
@@ -69,4 +107,10 @@ def get_assemble_scripts(self):
         )]
 
     def detect(self):
+        """
+        Check if current repo should be built with the Julia Build pack
+
+        super().detect() is called if Julia Build pack should be used.
```
yuvipanda: I think similar to the R build pack, we should actually *not* call super().detect() here, for the same reason as on the R build pack (we want to use Julia even if there is no environment.yml file). Whenever I'm doing doc fixes I'm also happy to fix bugs so documenting them is easier, but happy to deal with that in a separate PR too!

repo2docker/utils.py
```diff
@@ -53,55 +54,65 @@ def flush():
 
 @contextmanager
 def maybe_cleanup(path, cleanup=False):
+    """Execute cleanup of the passed path when cleanup flag is True."""
```
yuvipanda: "Delete the directory at passed path if cleanup flag is True"

repo2docker/buildpacks/julia.py
```diff
@@ -22,9 +39,29 @@ def get_env(self):
...
+        """
+        Return series of build-steps common to "ALL" Julia repositories
+
+        All scripts found here should be independent of contents of a
+        particular repository.
+
+        This installs:
+
+        - the specified Julia version and path
+        - Julia packages with ownership granted to the notebook user
```
yuvipanda: It only creates a directory with permissions for installing julia packages (from get_assemble_scripts), does not actually install them here

willingc: Thanks. I will update. Still working on the corrections from yesterday.

repo2docker/buildpacks/julia.py
willingc: added 👍

willingc:
@yuvipanda I think that I've covered all your review comments.

minrk:
Great!

willingc:
Thanks @minrk for the merge. Going to try to finish all the docstrings today.

PR #220, opened 2018-02-07
Switch freeze guidance to not be a directive

rgbkrk:
I got tripped up by this and thought **I** had to somehow run it in an image rather than knowing the _script would do it for me_.


yuvipanda:
Thanks Kyle!

PR #222, opened 2018-02-07
Install from wheel when testing in travis

yuvipanda:
This helps us catch issues with MANIFEST.in and friends that
might cause problems with files not being included in the final
built package, removing a step from the release process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,16 @@ services:
 python:
 - 3.4
 install:
+# Make a wheel and install it to test to catch possible
+# issues with MANIFEST.in
 - pip install --no-cache-dir pyyaml pytest wheel
 - python setup.py bdist_wheel
 - pip install dist/*.whl
 
 script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -5,11 +5,16 @@ services:
...
+# issues with MANIFEST.in
+- pip install --no-cache-dir pyyaml pytest wheel
+- python setup.py bdist_wheel
+- pip install dist/*.whl
 
 script:
-  - pytest -s -v tests/${REPO_TYPE}
+  # cd into tests so CWD being repo2docker does not hide
+  # possible issues with MANIFEST.in
+  - cd tests && pytest -s -v ${REPO_TYPE}
```
betatim: is there some magic we can use to only show stdout/err when a test fails? As is you get a lot of output which is a bit annoying when trying to find the bits related to the error :-/

yuvipanda: I agree! We should probably do that in a different PR / Issue though!

Issue #188, opened 2018-01-08
document and/or automate updating frozen environments


minrk:
Right now, to update the base environments, we need to:

1. (optional) update the base environment files:
    - buildpacks/conda/environment.yml
    - buildpacks/conda/environment.py-2.7.yml
    - buildpacks/py/requirements.txt
    - buildpacks/py/requirements2.txt
2. regenerate the frozen environments
    - `cd buildpacks/py && bash freeze.bash`
    - `cd buildpacks/conda && python freeze.py`

This requires:

- Python 3.6 (for the conda freeze script)
- Docker (for running the freezes)

We should at least document the process for contributors who want to bump the base env, and potentially automate it somehow if we can.

choldgraf:
now documented in https://github.com/jupyter/repo2docker/blob/master/CONTRIBUTING.md#updating-libraries-installed-for-all-repos

PR #224, opened 2018-02-08
Add a CONTRIBUTING.md

### LINKED ISSUE #215: Document how to cut a new release
We should be releasing r2d more often than once every 8 months :-)

One reason this hasn't happened is that we haven't got proper documentation for how to do this (and also haven't discovered some blockers that need to be un-blocked such as giving people `pip` access). We should document the steps to release a new r2d!

yuvipanda:
Contains local dev, adding new libraries + freezing and release instructions.

Thanks to @willingc for editing the release process notes

Fixes #215 and #188

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ install:
 - pip install dist/*.whl
 
 script:
-  # cd into tests so CWD being repo2docker does not hide 
+  # cd into tests so CWD being repo2docker does not hide
   # possible issues with MANIFEST.in
   - cd tests && pytest -s -v ${REPO_TYPE}
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,12 @@ python:
 install:
 # Make a wheel and install it to test to catch possible
 # issues with MANIFEST.in
-- pip install --no-cache-dir pyyaml pytest wheel
+- pip install --no-cache-dir -r dev-requirements.txt
 - python setup.py bdist_wheel
 - pip install dist/*.whl
 
 script:
-  # cd into tests so CWD being repo2docker does not hide
+  # cd into tests so CWD being repo2docker does not hide 
   # possible issues with MANIFEST.in
   - cd tests && pytest -s -v ${REPO_TYPE}
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,139 @@
+# Local Development
+
+To develop & test repo2docker locally, you need:
+
+1. Familiarity with using a command line terminal
+2. A computer running macOS / Linux
+3. Some knowledge of git
+4. At least python 3.4
+5. Your favorite text editor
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -7,12 +7,12 @@ python:
...
 # Make a wheel and install it to test to catch possible
 # issues with MANIFEST.in
-- pip install --no-cache-dir pyyaml pytest wheel
+- pip install --no-cache-dir -r dev-requirements.txt
 - python setup.py bdist_wheel
 - pip install dist/*.whl
 
 script:
-  # cd into tests so CWD being repo2docker does not hide
+  # cd into tests so CWD being repo2docker does not hide
```
betatim: spurious space at EOL

CONTRIBUTING.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Running tests
+
+We have a lot of tests for various cases supported by repo2docker in the `tests/`
+subdirectory. These use [py.test](https://docs.pytest.org/).
+
+You can run all the tests with:
+
+```bash
+py.test -s tests/*
```
betatim: You can run `pytest` with no args from the top level directory. Simpler because less args people might worry about what they mean?

yuvipanda: That doesn't work for me, it tries to run tests for all libraries installed in myvirtual environment, with hundreds of errors like:

```
ImportError while importing test module '/home/yuvipanda/code/repo2docker/lib64/python3.6/site-packages/ipython_genutils/tests/test_text.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
lib/python3.6/site-packages/ipython_genutils/tests/test_text.py:13: in <module>
    import nose.tools as nt
E   ModuleNotFoundError: No module named 'nose'
```

.travis.yml
yuvipanda: Good catch! Fixed.

CONTRIBUTING.md
betatim: Interesting. Magic. Let's keep it like this then.

CONTRIBUTING.md
```diff
@@ -0,0 +1,211 @@
...
+# Updating libraries installed for all repos
+
+For both the conda and virtualenv base environments, we install specific
+pinned versions of all dependencies. We explicitly list the dependencies
+we want, then *freeze* them at commit time to explicitly list all the
+transitive dependencies at current versions. This way, we know that
+all dependencies will have the exact same version installed at all times.
+
+If you update a library in the base environment, you need to update it
+for both the virtualenv and conda environments before submitting a PR.
```
betatim: 🙄

betatim:
Cool beans!

CONTRIBUTING.md
mpacer: That's odd, if you're at the repo2docker package level it shouldn't run it on all of your packages. Could you mean something different from "top level directory"?

PR #226, opened 2018-02-09
add appendix

minrk:
supports adding arbitrary build steps to the end of Dockerfile

this enables some discussed features like Binder-deployment-specific appendices to images (e.g. enabling the nbmelt extension to shutdown notebooks that don't get used after a shorter timeout than the existing 10 minute idle timeout).


yuvipanda:
@minrk not sure why this needs a rebase / conflict fix now, but LGTM after that.

minrk:
Rebased

PR #228, opened 2018-02-09
do not reuse BuildPack instances

minrk:
rather than instantiating all buildpacks ahead of time, only instantiate as needed.

Makes it a little clearer that a given BuildPack should only perform a single build and a new BuildPack should be created for a subsequent build.


betatim:
Merged, because it was an easy win. Rebase #229

PR #233, opened 2018-02-12
clone recursively

minrk:
ensures submodules, if any, are included

closes https://github.com/jupyterhub/binderhub/issues/447

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -151,7 +151,7 @@ class Repo2Docker(Application):
 
     def fetch(self, url, ref, checkout_path):
         try:
-            for line in execute_cmd(['git', 'clone', url, checkout_path],
+            for line in execute_cmd(['git', 'clone', '--recursive', url, checkout_path],
                                     capture=self.json_logs):
                 self.log.info(line, extra=dict(phase='fetching'))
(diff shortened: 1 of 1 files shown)


choldgraf:
🎉

slayoo:
@minrk @choldgraf I've ended up here Googling for how the submodules are handled by repo2docker. The recursive clone is indeed started, but in our case the submodules are specified by `url = git@github.com:...git` entries, which require `ssh` to be present for cloning. As a result, we get the following cryptic message: 

```
...
Submodule 'tests/devops_tests' (git@github.com:open-atmos/devops_tests.git) registered for path 'tests/devops_tests'
Cloning into '/tmp/repo2dockerjt4w764u/tests/devops_tests'...
error: cannot run ssh: No such file or directory
fatal: unable to fork
fatal: clone of 'git@github.com:open-atmos/devops_tests.git' into submodule path '/tmp/repo2dockerjt4w764u/tests/devops_tests' failed
```
Would it make sense to make `ssh-client` installed by default on these images so that recursive cloning using ssh would work out of the box?

minrk:
Having ssh-client so git always works makes sense. Recursive clone via ssh is going to fail in that situation without credentials, too, right?

PR #235, opened 2018-02-12
Add architecture document

### LINKED ISSUE #214: Documenting "guiding principles" and core functionality of repo2docker
re: core functionality, we came up with:

repo2docker does four things:

* BUILD a docker image from a git repository using configuration files
* COPY repository contents to HOME within the image so you can interact with them
* PUSH a docker image to the cloud
* RUN that image locally so you can debug whether the environment is built properly

Describe that different use-cases might want different combinations of the above things. We came up with three main ones:

* building an image for a jupyterhub (BUILD, PUSH)
* building an image that you want to run locally (so BUILD, COPY, RUN)
* running a binderhub service (BUILD COPY PUSH) 

re: guiding principles

Was chatting with @mpacer this AM about the repo2docker documentation, and they had a good idea that we should define what are the core guiding principles or things we want repo2docker to do.

I took a look at our current documentation and realized that this is already partially addressed here: http://repo2docker.readthedocs.io/en/latest/design.html ...however perhaps there's room for improvement + making it more discoverable. @mpacer do you have any ideas on that page / ways to improve it?

yuvipanda:
Fixes #214

### DIFF
--- a/docs/source/architecture.md
+++ b/docs/source/architecture.md
@@ -1,0 +1,106 @@
+# Architecture
+
+This is a living document talking about the architecture of repo2docker
+from various perspectives.
+
+## Buildpack
+
+The **buildpack** concept comes from [Heroku](https://devcenter.heroku.com/articles/buildpacks)
+and Ruby on Rails' [Convention over Configuration](http://rubyonrails.org/doctrine/#convention-over-configuration)
+doctrine. 
+
+Instead of the user specifying a complete specification of exactly how they want
+their environment to be, they can focus only on how their environment differs from a conventional
+environment. This means instead of deciding 'should I get Python from Apt or pyenv or ?', user
+can just specify 'I want python-3.6'. Usually, specifying a **runtime** and list of **libraries**
+with explicit **versions** is all that is needed.
+
+In repo2docker, a Buildpack does the following things:
+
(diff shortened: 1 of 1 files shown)


yuvipanda:
There's a hackmd at https://hackmd.io/EYMwDAnMCMCsEFoDsAOApgNgQFg7aCKwATInAMwCG2saRSsGQA==# - feel free to edit that, and I can incorporate it back to here with suggestions.

If something is unclear (things probably are! I've a lot of blind spots wrt this!), feel free to also leave comments in the PR (since that's probably easier than in HackMD), and I'll try to clarify. Same if I've gotten something wrong, but you're welcome to just directly edit hackmd too :)

/cc @choldgraf @willingc @minrk @betatim

willingc:
Nice @yuvipanda.

Love this:

 - Detect if it can handle a given repository
- Build a base language environment in the docker image
- Copy the contents of the repository into the docker image
- Assemble a specific environment in the docker image based on repository contents
- Push the built docker image to a specific docker registry (optional)
- Run the build docker image as a docker container (optional)

willingc:
Going to merge as is (since this is way better than anything we have). I think it will be easier for all to iterate any changes in follow up PRs.

PR #238, opened 2018-02-14
bumping ipywidgets version

choldgraf:
This bumps `ipywidgets` to 7.1.1 (so a major version bump)

Also attempts at improving the version bump docs a bit while I'm at it :-)

closes #238


yuvipanda:
@choldgraf do you want to deploy this so you can try out the new deploy process? Should be very similar to the old one, just with less steps.

choldgraf:
Yes!

PR #239, opened 2018-02-14
Add shiny-server.

ryanlovett:
This assumes apps are in ~/ShinyApps/. For now this can be seeded by
postBuild, but a better convention is to specify remote app locations in
something like a shinyapps.txt file.

### DIFF
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -128,11 +128,17 @@ class RBuildPack(PythonBuildPack):
         # This is MD5, because that is what RStudio download page provides!
         rstudio_checksum = '24cd11f0405d8372b4168fc9956e0386'
 
+        shiny_url = 'https://download3.rstudio.org/ubuntu-12.04/x86_64/shiny-server-1.5.6.875-amd64.deb'
+        shiny_checksum = 'de039a6f9e055de693524e6cb44f6a03'
+
         # Version of MRAN to pull devtools from.
         devtools_version = '2018-02-01'
 
         # IRKernel version - specified as a tag in the IRKernel repository
         irkernel_version = '0.8.11'
+
+        nbrsessionproxy_pip_spec = 'nbrsessionproxy==0.6.1'
+        nbrsessionproxy_pip_spec = 'git+https://github.com/ryanlovett/nbrsessionproxy.git@432262c'
 
         return super().get_build_scripts() + [
             (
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks for the work, Ryan!

Can we instead make nbrsessionproxy work such that the path to the shiny application is present in the URL, as a runtime configuration rather than a build time configuration? That would give us a lot more flexibility, and not add an additional  repo2docker/binder-specific spec file!

ryanlovett:
Sure, thanks for reviewing!

The path to the shiny apps is in the URL, but all deployed shiny apps need to be put in some specific location. I chose ${HOME}/ShinyApps where "ShinyApps" is a shiny convention. A url of /shiny/ will show an index listing of apps, and /shiny/:app will run the app in ${HOME}/ShinyApps/:app/.

Are you saying that you'd like the shiny apps assets to be specified as a remote URL, e.g. /shiny/?app=https://github.com/...&app=https://www..., rather than provided in the binder?

repo2docker/buildpacks/r.py
```diff
@@ -128,12 +128,18 @@ def get_build_scripts(self):
...
+        shiny_checksum = 'de039a6f9e055de693524e6cb44f6a03'
+
         # Version of MRAN to pull devtools from.
         devtools_version = '2018-02-01'
 
         # IRKernel version - specified as a tag in the IRKernel repository
         irkernel_version = '0.8.11'
 
+        nbrsessionproxy_pip_spec = 'nbrsessionproxy==0.6.1'
+        nbrsessionproxy_pip_spec = 'git+https://github.com/ryanlovett/nbrsessionproxy.git@432262c'
```
betatim: Is this duplication on purpose or a debugging left over? My preference is that we don't reference "private" forks of tools and instead use something from PyPi.

betatim:
Could you add a test for this as well? We are trying to make deployments to mybinder.org more and more automatic which means we need tests that cover "a lot of stuff" :)

repo2docker/buildpacks/r.py
ryanlovett: This was only for the duration of the PR since it depended on not-yet-reviewed functionality in nbrsessionproxy. I totally agree that personal forks shouldn't be referenced in production.

I chatted with @yuvipanda a few days ago about this PR and I'm actually going to rework it. I'll probably get the new code merged to nbrsessionproxy before I complete this PR.

ryanlovett:
@yuvipanda I've modified jupyterhub/nbrsessionproxy#28 to do as you suggest.

willingc:
I've fixed the merge conflict. Thanks @ryanlovett 👍

yuvipanda:
I had a long conversation with @ryanlovett a few weeks ago, and think we should not introduce a new binder-speicfic path for this (in this case, ~/ShinyApp). I think he reworked nbrsessionproxy to take that into account, but I don't think this PR has been modified to work without requiring things in ~/ShinyApp.

I apologize for not not reporting our conversation here! I'd like to revert this PR now until it can be reworked to not need the ~/ShinyApps folder. Does that sound ok to you, @willingc?

In the future, I"ll be more diligent about reporting offline conversations back to public channels!

ryanlovett:
Yes, nbrsessionproxy now needs less of this patch, and [another part](https://github.com/jupyter/repo2docker/pull/239/commits/ef330192ab1b9f211b85acd1135ae97f463ebdff?diff=unified#diff-26c1d7020998ead6f3513548c91a56b7R227) of it was merged in #254.

I can create a new PR.

willingc:
Thanks @ryanlovett. Sorry for the false alarm ;-)

PR #241, opened 2018-02-16
Do not require postBuild to be executable

### LINKED ISSUE #240: make postBuild executable
While running a workshop, the only step we couldn't do with the GitHub web UI was to make postBuild executable. Making files executable is also cumbersome on Windows. Since we require that postBuild is executable, we could *ensure* it is executable in the Dockerfile rather than *assert* that it is, so we make it executable rather than failing if it isn't. I think this would be a simpler, friendlier experience, especially for novices.

yuvipanda:
This is hard for windows users and from the GitHub UI,
and provides dubious benefits. So let's remove the requirement!

Fixes #240

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -3,7 +3,6 @@ import jinja2
 import tarfile
 import io
 import os
-import stat
 import re
 import logging
 import docker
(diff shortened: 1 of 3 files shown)


betatim:
Restarted travis. Otherwise looks good to me.

betatim:
I think this test failure is real. But I don't understand why. The test suggests that the build should not succeed, and it doesn't. So the test should pass?

minrk:
The failing test had a strict check for the postBuild error message, which needs to be updated now that the postBuild line is slightly changed. I added a commit that should make it a bit more relaxed.

PR #236, opened 2018-02-12
[WIP] base contentproviders commit

yuvipanda:
Very much a WIP, but some thougts on how we can support more
than just git.

PR #242, opened 2018-02-18
[MRG] Add Content Providers Infrastructure

betatim:
Supersedes #236 

The goal is to have a basic content provider for remote git repositories.

Maybe also remove the special casing of the "local directory" by adding a provider for that.


repo2docker/app.py
```diff
@@ -393,20 +405,25 @@ def initialize(self, argv=None):
...
         else:
-            self.repo_type = 'remote'
-            self.repo = args.repo
-            self.ref = args.ref
             self.cleanup_checkout = args.clean
 
+        # user wants to mount a local directory into the container for
+        # editing
+        if args.editable:
+            if os.path.isdir(args.repo):
```
betatim: checking explicitly for it to be a directory not just a path because we might have content providers that can provide content from a local `something.zip` file, which we couldn't mount in editable mode

repo2docker/contentproviders/base.py
```diff
@@ -0,0 +1,58 @@
...
+        self.log = logger
+
+    def detect(self, repo, ref=None, extra_args=None):
+        """Determine compatibility between source and this provider.
+
+        If the provider knows how to fetch this source it will return a
+        `spec` that can be passed to `fetch`. The arguments are the `repo`
+        string passed on the command-line, the value of the --ref parameter,
+        if provided and any provider specific arguments provided on the
+        command-line.
```
betatim: This `--provider-extra-args` flag doesn't exist yet. My thinking is that future providers might require wildly different kinds of argument which we can't all explicitly support so instead we provide this as a provider specific/internal escape hatch.

betatim:
This is ready for some reviewing.

Once this is merged I'd like to add a local and remote ZIP file provider to help get us to supporting Zenodo.

repo2docker/app.py
```diff
@@ -85,6 +85,23 @@ def _default_log_level(self):
         """
     )
 
+    # Git is our content provider of last resort. This is to maintain the
```
willingc: :heart:

repo2docker/app.py
```diff
@@ -85,6 +85,23 @@ def _default_log_level(self):
...
+    # detecting if something will successfully `git clone` is very hard if all
+    # you can do is look at the path/URL to it.
+    content_providers = List(
+        [
+            contentproviders.Local,
+            contentproviders.Git,
+        ],
+        config=True,
+        help="""
+        Ordered list of ContentProviders to try in turn to fetch the contents
```
willingc: Ordered list by priority of ContentProviders to try

repo2docker/app.py
```diff
@@ -166,29 +183,24 @@ def _user_name_default(self):
...
-            if hash is None:
-                self.log.error('Failed to check out ref %s', ref,
-                               extra=dict(phase='failed'))
-                sys.exit(1)
-            # If the hash is resolved above, we should be able to reset to it
-            for line in execute_cmd(['git', 'reset', '--hard', hash],
-                                    cwd=checkout_path,
-                                    capture=self.json_logs):
-                self.log.info(line, extra=dict(phase='fetching'))
+        # Pick a content provider based on URL
```
willingc: Perhaps this comment is better in the docstring. Or modify to something like: Iterate through possible content providers until a valid provider, based on URL, is found.

repo2docker/app.py
willingc: Your comment seems worthy of a brief comment on the `isdir` check

repo2docker/app.py
```diff
@@ -642,7 +659,9 @@ def start(self):
                     raise e
                 sys.exit(1)
 
-        if self.repo_type == 'local':
+        # if the source is a directory we will keep using it, assuming that
```
willingc: @betatim How do I determine whether the user wants to use the directory or not?

repo2docker/app.py
```diff
@@ -166,29 +183,24 @@ def _user_name_default(self):
...
-                sys.exit(1)
-            # If the hash is resolved above, we should be able to reset to it
-            for line in execute_cmd(['git', 'reset', '--hard', hash],
-                                    cwd=checkout_path,
-                                    capture=self.json_logs):
-                self.log.info(line, extra=dict(phase='fetching'))
+        # Pick a content provider based on URL
+        picked_content_provider = None
+        for CP in self.content_providers:
+            cp = CP()
```
willingc: Just to clarify: Since content_providers is a Traitlets List, the individual list element (`CP`) which refers to a list item (i.e. contentproviders.Local) which begins with an uppercase letter can be used to instantiate an object of class `CP`. For example, if CP=contentproviders.Git, then `cp = Git()`  [traitlets doc note](https://traitlets.readthedocs.io/en/stable/config.html#python-configuration-files)

repo2docker/app.py
willingc: Got it. Perhaps: remove the wording re: assumption.

Maybe `# If the source to be executed is a directory, continue using the directory. In the case of a local directory, it is used as both the source and target`

willingc: Thanks!

repo2docker/app.py
willingc: I think that `self.content_providers` is a Traitlets list. I believe that it is the traitlets functionality that allows the instantiation in `cp = CP()`. To create an instance, one would call `Git(contentproviders)` or in this case using the shorthand `Git()`. Perhaps I'm missing how this is working under the hood, since I don't think one can instantiate a list element directly. 🤷‍♀️

betatim: If traitlets help this is an accident, all I had in mind when I wrote this code is:
```python
In [1]: class A:
   ...:     def __init__(self):
   ...:         print("A all day")
   ...:

In [2]: class B:
   ...:     def __init__(self):
   ...:         print("Because")
   ...:

In [3]: classes = [A, B]

In [4]: classes[0]()
A all day
Out[4]: <__main__.A at 0x103dbde10>
```

PR #244, opened 2018-02-22
[WIP] Specify plans for stability in repo2docker dependency

betatim:
What do you think? What did I misrepresent or express too vaguely?


betatim:
Once this is merged we should link to it from places like https://mybinder.readthedocs.io/en/latest/introduction.html#what-can-i-do-with-binder

yuvipanda:
I like it.

PR #250, opened 2018-03-07
omit appendix on custom Dockerfiles

minrk:
using a custom Dockerfile means opting out of *all* repo2docker logic

prompted by discussion in https://github.com/jupyterhub/binderhub/pull/459


betatim:
I think I'm Ok with this. If you are using your own `Dockerfile` you shouldn't have need for an appendix anyway and it means we won't get confused users who are surprised by errors coming from an appendix interacting badly with their docker file.

PR #252, opened 2018-03-07
upgrade nteract_on_jupyter to 1.5.0

rgbkrk:
* Upgraded `nteract_on_jupyter`
* Pinned tornado to `4.5.3`

### DIFF
--- a/repo2docker/buildpacks/python/requirements.frozen.txt
+++ b/repo2docker/buildpacks/python/requirements.frozen.txt
@@ -1,10 +1,10 @@
 # AUTO GENERATED FROM requirements.txt, DO NOT MANUALLY MODIFY
-# Frozen on Wed Feb 14 16:40:31 UTC 2018
-bleach==2.1.2
+# Frozen on Wed Mar  7 21:53:43 UTC 2018
+bleach==2.1.3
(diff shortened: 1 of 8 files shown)


repo2docker/buildpacks/python/requirements.frozen.txt
```diff
@@ -30,13 +30,13 @@ ptyprocess==0.5.2
...
 python-dateutil==2.6.1
 pyzmq==17.0.0
-Send2Trash==1.4.2
+Send2Trash==1.5.0
 simplegeneric==0.8.1
 six==1.11.0
 terminado==0.8.1
 testpath==0.3.1
-tornado==4.5.3
+tornado==5.0
```
betatim: This looks like we should briefly think about it. Especially as the conda environments will remain at 4.5. This makes me think this is not what should be happening -> we should remain on 4.5 here as well. @minrk knows about tornado 5 vs 4.

rgbkrk: I could force pin tornado at the top level too.

rgbkrk:
Pinned tornado so we have a way ahead until the notebook server is upgraded.

PR #254, opened 2018-03-11
[MRG] Setup R libraries path for RStudio

### LINKED ISSUE #237: R libraries aren't linked to RStudio
Was just trying out R functionality in *both* Jupyter and RStudio, and I don't think we're linking libraries correctly in RStudio.

In the [R Demo repo](https://github.com/binder-examples/r) I've got `ggplot2` installing with `install.R`. In the resulting Binder session, `library(ggplot2)` works fine from Jupyter, but does not work from within the RStudio session (resulting in `Error in library(ggplot2) : there is no package called ‘ggplot2’`)

We should figure out how the path linking works with R vs. RStudio, maybe they're not the same thing

betatim:
Fixes #237 

This makes sure our base R install (used by the notebook kernel) and
RStudio share the same packages.

Any ideas how we could test that this worked? I would copy the existing R test but don't know how to run the `verify` command in RStudio.

### DIFF
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -156,6 +156,16 @@ class RBuildPack(PythonBuildPack):
                 )
             ),
             (
+                "root",
+                # Set paths so that RStudio shares libraries with base R
+                # install. This first comments out any R_LIBS_USER that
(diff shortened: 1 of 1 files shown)


choldgraf:
very nice! thanks 👏 👏 👏

PR #255, opened 2018-03-11
make sure there's a newline before each 'File: ' heading

ctb:
Currently, the file `tests/venv/postBuild/README.rst` is missing a newline at the end of the last line, and so [samples.html](https://repo2docker.readthedocs.io/en/latest/samples.html) in the generated docs fails to properly format the `File: postBuild` heading. This fixes the code in `docs/source/conf.py` to add an extra newline Just In Case.


ctb:
(This seemed better than just fixing README.rst, which is also an OK fix; but this fixes any future such problem.)

betatim:
Thanks for this! Restarted the one failing travis job, so waiting on that. After that: LGTM!

PR #260, opened 2018-03-15
refreeze environments

### LINKED ISSUE #257: update ipykernel to work with tornado 5 so Python 2 kernels work again
Presently, the Python 2 kernels in both  https://github.com/binder-examples/python2_runtime and https://github.com/binder-examples/python2_with_3 , go dead before any code can be run.

Min RK @minrk noted:
>Min RK @minrk 13:36 @fomightez indeed, tornado was allowed to be updated to 5.0, but ipykernel is pinned to an earlier version that doesn't work with tornado 5. This needs to be fixed in repo2docker.

minrk:
- bump ipykernel to 4.8.2
- pin tornado to 4.5.3

closes #257

### DIFF
--- a/repo2docker/buildpacks/conda/environment.frozen.yml
+++ b/repo2docker/buildpacks/conda/environment.frozen.yml
@@ -1,5 +1,5 @@
 # AUTO GENERATED FROM environment.py-3.6.yml, DO NOT MANUALLY MODIFY
-# Frozen on 2018-03-07 21:59:06 UTC
+# Frozen on 2018-03-13 17:43:53 UTC
 name: r2d
 channels:
   - conda-forge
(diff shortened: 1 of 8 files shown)


minrk:
I'd like to deploy this as part of https://github.com/jupyterhub/mybinder.org-deploy/pull/502 which also needs to bump the image prefix

willingc:
@minrk Merging. Thanks. Is there a specific conflict re: tornado to pin to 4.5.3.?

minrk:
Just to be consistent, since tornado 5 hasn't landed on conda-forge and there have been a few edge cases where tornado 5 doesn't quite work. I think we can be a bit conservative and pick up tornado 5 after a bit more stabilizing elsewhere.

willingc:
Makes complete sense. Thanks @minrk.

PR #265, opened 2018-03-16
Add docstrings to python3 and python2 buildpacks

willingc:
Addresses #211.

### DIFF
--- a/repo2docker/buildpacks/python/__init__.py
+++ b/repo2docker/buildpacks/python/__init__.py
@@ -1,12 +1,13 @@
-"""
-Generates a variety of Dockerfiles based on an input matrix
-"""
+"""Generates Dockerfiles based on an input matrix based on Python."""
 import os
+
 from ..base import BaseImage
 
 
 class PythonBuildPack(BaseImage):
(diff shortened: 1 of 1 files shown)


repo2docker/buildpacks/python/__init__.py
```diff
@@ -1,39 +1,80 @@
...
+"""Generates Dockerfiles based on an input matrix based on Python."""
 import os
+
 from ..base import BaseImage
 
 
 class PythonBuildPack(BaseImage):
+    """Setup Python 3 for use with a repository."""
     def get_packages(self):
+        """Return list of Python3 packages to be installed."""
```
betatim: "Return list of system-level Python3 dependencies" (I don't like that sentence that much). Something to clarify that these are the apt-get packages we install, not python packages (like say numpy).

betatim:
Looks good! One comment about system packages vs python packages.

repo2docker/buildpacks/python/__init__.py
willingc: Good distinction @betatim. Will update.

willingc:
Thanks folks.

PR #269, opened 2018-03-18
Addresses #268 : local builds should respect .gitignore

### LINKED ISSUE #268: local builds should respect .gitignore ?
I somehow expected that when building a local repository, `repo2docker` will respect `.gitignore` and not package the ignored files/directories in the image itself.

The use case where this was inconvenient was, when I had a `data/` folder in my source code, which was ignored by `git` based on the pathspecs in `.gitignore`. 
The folder in my dev environment contained about 16gb of data in the `data/` folder, and `repo2docker` ended up copying over all that data into the image (and also because of [this issue](https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/base.py#L87), the actual image size ended up being around 35GB 😢  ).

spMohanty:
Fixes #268 (added by @betatim)

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -4,6 +4,7 @@ import tarfile
 import io
 import os
 import re
+import pathspec
 import logging
 import docker
 
@@ -349,7 +350,21 @@ class BuildPack:
             src_path = os.path.join(os.path.dirname(__file__), *src_parts)
             tar.add(src_path, src, filter=_filter_tar)
 
-        tar.add('.', 'src/', filter=_filter_tar)
+        _exclude_tar = None
+        if os.path.exists(".gitignore"):
+            gitignore_fh = open(".gitignore")
+            ignorespec = pathspec.PathSpec.from_lines('gitignore', gitignore_fh)
+
+            def _exclude_tar(filepath):
+                # Conditionally exclude files based on the pathspecs mentioned
+                # in the `.gitignore` file.
+                # Note that, the behaviour of this function is not **exactly**
+                # same to the way `git` excludes files based on `.gitignore`.
+                #
+                # https://github.com/cpburnz/python-path-specification/issues/19
+                return ignorespec.match_file(filepath)
+
+        tar.add('.', 'src/', exclude=_exclude_tar, filter=_filter_tar)
 
         tar.close()
         tarf.seek(0)
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -171,7 +171,7 @@ class BuildPack:
             # FIXME: Use npm from nodesource!
             # Everything seems to depend on npm these days, unfortunately.
             "npm",
-            
+
             "unzip",
         }
 
@@ -363,6 +363,7 @@ class BuildPack:
                     # based on `.gitignore`.
                     #
                     # https://github.com/cpburnz/python-path-specification/issues/19
+                    filepath = os.path.relpath(filepath, "./")
                     return ignorespec.match_file(filepath)
 
         tar.add('.', 'src/', exclude=_exclude_tar, filter=_filter_tar)
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -352,17 +352,18 @@ class BuildPack:
 
         _exclude_tar = None
         if os.path.exists(".gitignore"):
-            gitignore_fh = open(".gitignore")
-            ignorespec = pathspec.PathSpec.from_lines('gitignore', gitignore_fh)
-
-            def _exclude_tar(filepath):
-                # Conditionally exclude files based on the pathspecs mentioned
-                # in the `.gitignore` file.
-                # Note that, the behaviour of this function is not **exactly**
-                # same to the way `git` excludes files based on `.gitignore`.
-                #
-                # https://github.com/cpburnz/python-path-specification/issues/19
-                return ignorespec.match_file(filepath)
+            with open(".gitignore") as gitignore_fh:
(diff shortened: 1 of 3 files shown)


repo2docker/buildpacks/base.py
```diff
@@ -349,7 +350,21 @@ def _filter_tar(tar):
             src_path = os.path.join(os.path.dirname(__file__), *src_parts)
             tar.add(src_path, src, filter=_filter_tar)
 
-        tar.add('.', 'src/', filter=_filter_tar)
+        _exclude_tar = None
+        if os.path.exists(".gitignore"):
+            gitignore_fh = open(".gitignore")
```
betatim: can you use a context manager here so that the file gets closed?

betatim:
Could you add a test to check that a file that should get added gets added and one that shouldn't shouldn't?

Do the others have any thoughts on adding a new dependency?

Otherwise LGTM.

repo2docker/buildpacks/base.py
minrk: Typical docker builds use a separate `.dockerignore` and ignore `.gitignore`. Should we follow that pattern? I'd rather be as close to exact for what `docker build` would normally do as we can.

repo2docker/buildpacks/base.py
```diff
@@ -349,7 +350,21 @@ def _filter_tar(tar):
...
+            ignorespec = pathspec.PathSpec.from_lines('gitignore', gitignore_fh)
+
+            def _exclude_tar(filepath):
+                # Conditionally exclude files based on the pathspecs mentioned
+                # in the `.gitignore` file.
+                # Note that, the behaviour of this function is not **exactly**
+                # same to the way `git` excludes files based on `.gitignore`.
+                #
+                # https://github.com/cpburnz/python-path-specification/issues/19
+                return ignorespec.match_file(filepath)
```
minrk: It concerns me a bit that this isn't exact. Is there any possibility that we could end up excluding files that shouldn't be excluded, or does the bug you link exclusively mean that we will include some files we maybe shouldn't?

repo2docker/buildpacks/base.py
spMohanty: @betatim : sure, I will push in another commit with a bit of refactoring, and also the tests.

@minrk : Thats more of a usability question. If `repo2docker` wants to be transparent to the user then, adding a separate `.dockerignore` might add to the confusion. You could add conditionals to check if both `.dockerignore` and `.gitignore` exist, and prioritize the `.dockerignore`, but if the repository just has a `.gitignore` then you use that instead.

repo2docker/buildpacks/base.py
spMohanty: @minrk : `pathspecs` was the only pure python implementation I could find. Also, I think it does implement the `pathspecs` correctly, but the difference comes from how `git` traverses the directory structure when matching them with the `pathspecs` (eg: if a directory is ignored, then it doesnot traverse it, on the other hand in this use case, `tarfile` does traverse the whole directory and tries to match with all its children too).

In the issue I reported, and in most of the use cases, this will only mean we will _forget_ to exclude some files which were meant to be excluded. 

The only use case when a file that is not meant to be excluded is excluded, is when `.gitignore` specifically ignores a particular file, but the user had force-pushed the file to the repository.

But in the both these use cases, the `.gitignore` should ideally be modified to include the exceptions. For example, when you force-push say a `*.csv` file which was ignored by `.gitignore`, then you are expected to also remember to add a `!<path_to_file>` to `.gitignore`.

betatim:
Hoisting @minrk's inline comment to here:

> Typical docker builds use a separate `.dockerignore` and ignore `.gitignore`. Should we follow that pattern? I'd rather be as close to exact for what `docker build` would normally do as we can.

I hadn't thought of this and it is a very good point. The pragmatist in me would like to merge the git and docker ignore files to make it "just work". But it probably brings more problems than it solves.

On a binderhub we wouldn't ever see this as files in `.gitignore` would not be in the repo. So that maybe points towards using `.dockerignore` as it allows people to exclude things later.

spMohanty:
@betatim : Well true, focusing on the binderhub use case, the `.dockerignore` solution actually also solves the other problems (like force-pushed data will not be excluded).

Going by the discussion [here](https://github.com/docker/docker-py/pull/721), it seems that while they considered `pathspec` at point, they moved away from it. But the good thing is, they do package some of the code internally, so `dockerignore` can be supported without adding an extra dependency, and will always have the same functionality as `docker-py`.

They seem to seem to handle it [here](https://github.com/docker/docker-py/blob/a4e642b015c50d9c628413341ed00c89599f66be/docker/api/build.py#L141), where the call the `docker.utils.tar` function [here](https://github.com/docker/docker-py/blob/a4e642b015c50d9c628413341ed00c89599f66be/docker/utils/build.py#L10), which uses the `create_archive` function [here](https://github.com/docker/docker-py/blob/a4e642b015c50d9c628413341ed00c89599f66be/docker/utils/utils.py#L91).

The only major risk is that, their `tar` + `create_archive` wrapper does not use the `filter` argument, which seems to be used to deal with the caching issue mentioned [here](https://github.com/jupyter/repo2docker/blob/81a31525279507923289aec0ecf68fcb6c70fc86/repo2docker/buildpacks/base.py#L336)

choldgraf:
I'm +1 on respecting `.dockerignore`. repo2docker should try to follow as much pre-existing patterns as possible, and I think it's more intuitive to put pre-existing behavior of `.dockerignore` into r2d rather than building a different kind of behavior into `.gitignore`

betatim:
Closing this as it has been inactive for over a year. We can make a new PR if the need still exists.

PR #270, opened 2018-03-19
add docstrings for legacy buildpack

willingc:
Partially addresses #211

### DIFF
--- a/repo2docker/buildpacks/legacy/__init__.py
+++ b/repo2docker/buildpacks/legacy/__init__.py
@@ -1,5 +1,10 @@
-"""
-Generates a variety of Dockerfiles based on an input matrix
+"""Generates Dockerfiles from the legacy Binder Dockerfiles
+based on `andrewosh/binder-base`.
+
+The Dockerfile is amended to add the contents of the repository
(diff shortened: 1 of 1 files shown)


repo2docker/buildpacks/legacy.py
```diff
@@ -1,11 +1,17 @@
-"""
-Generates a variety of Dockerfiles based on an input matrix
+"""Generates legacy Dockerfiles based on an input matrix.
```
minrk: The 'input matrix' is leftover from a copy-paste when all the buildpacks were in one file. It doesn't apply here.

How about:

> Generates Dockerfiles from the legacy Binder Dockerfiles based on andrewosh/binder-base.
> The Dockerfile is amended to add the contents of the repository to the image and install a supported version of the notebook and IPython kernel.

willingc: Sounds much better. Thanks @minrk.

willingc:
@minrk I'll update this again after #276 lands.

willingc:
Updated docstrings based on changes in PR #276. Should be ready for re-review @minrk

PR #275, opened 2018-03-20
Edit CONTRIBUTING for freezing dependencies

willingc:
Update freezing since the scripts now will use docker containers so there is no need for local versions of each Python release.


betatim:
LGTM

minrk:
Great!

PR #276, opened 2018-03-20
Freeze legacy environment 

### LINKED ISSUE #256: freeze conda environment for legacy buildpack
we have a loose conda install in the legacy buildpack, which will allow dependencies to be upgraded out from under requested packages, which will eventually break.

We should use a frozen environment for the installation in this case like we do for the conda, so that the dependencies are not allowed to float. Unlike the conda buildpack, once frozen we should not update this environment except when absolutely needed.

GladysNalvarte:
Adapt freeze.py from conda, create frozen environments  and use those in legacy buildpack.
Additionally we pinned the legacy base image to specific version and perform the frozen install before the user docker file instead of after to avoid clobbering user installs which was a less of a risk when we were only installing one package.

closes #256


repo2docker/buildpacks/legacy/freeze.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+"""
+Freeze the conda environment.yml
+
+It runs the freeze in a continuumio/miniconda3 image to ensure portability
+
+Usage:
+
+python freeze.py [3.5]
```
minrk: Update this docstring that's copied from conda:

> Freeze the conda environment.yml for legacy dockerfiles
> It runs the freeze in the andrewosh/binder-base image used for legacy dockerfiles

repo2docker/buildpacks/legacy/freeze.py
```diff
@@ -0,0 +1,66 @@
...
+            'conda config --system --set auto_update_conda false',
+            f"conda env update -f /r2d/{env_file} -n {env_name}",
+            # exclude conda packages because we don't want to pin them
+            f"conda env export -n {env_name} | grep -v conda >> /r2d/{frozen_file}",
+        ])
+    ])
+
+
+if __name__ == '__main__':
+    # allow specifying which Pythons to update on argv
```
minrk: "allow specifying which env(s) to update..."

minrk:
Thanks @GladysNalvarte!

willingc:
Thanks @GladysNalvarte. Nice improvement :-)

PR #277, opened 2018-03-20
fixing python2 search

choldgraf:
closes https://github.com/jupyterhub/binder/issues/84 and closes https://github.com/jupyterhub/binder/issues/82

cc @betatim

### DIFF
--- a/repo2docker/buildpacks/python/__init__.py
+++ b/repo2docker/buildpacks/python/__init__.py
@@ -12,7 +12,7 @@ class PythonBuildPack(BaseImage):
 
            Note: The packages specified here are for the core Python3 language.
            Third party libraries are specified in other configuration files.
-           
+
         """
         return super().get_packages().union({
             'python3',
@@ -37,7 +37,7 @@ class PythonBuildPack(BaseImage):
     def get_path(self):
         """Return paths (including virtual environment path) to be added to
         the PATH environment variable.
-        
+
         """
         return super().get_path() + [
             "${VENV_PATH}/bin"
@@ -148,7 +148,7 @@ class Python2BuildPack(PythonBuildPack):
 
            Note: The packages specified here are for the core Python2 language.
            Third party libraries are specified in other configuration files.
-           
+
         """
         return super().get_packages().union({
             'python',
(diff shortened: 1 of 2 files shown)


betatim:
Can you help me grok why this fixes the bug? It looks like this does fix a bug, but one that should manifest by a runtime failure trying to run a `requirements.txt` that (possibly) doesn't exist.

choldgraf:
so that *was* the error I was getting when trying to run the repo originally mentioned in that issue...a file in `binder/` would be missed only if the python 2.7 runtime was used (because the location of the `requirements.txt` file was hard-coded)

repo2docker/buildpacks/python/__init__.py
```diff
@@ -238,10 +238,11 @@ def get_build_scripts(self):
     def get_assemble_scripts(self):
         """Return series of build-steps specific to this repository.
         """
+        path_requirements = self.binder_path('requirements.txt')
```
willingc: Perhaps it would be better to use the consistent naming, such as here: https://github.com/jupyter/repo2docker/pull/277/files#diff-76daac66b17bfc1d62a1e222b0099232R128

`requirements_file` instead of `path_requirements`

choldgraf: done!

betatim:
Thanks! I thought we were trying to track down why files were being ignored in favour of a `Dockerfile`.

choldgraf:
yeah - I don't have an example of a repo where this problem exists, so a bit hard to debug, but looking through the code I couldn't find a reason why a Dockerfile would be "preferred" over anything in a `binder/`  folder :-/

willingc:
In general, we may need to up our game with some new tests now that we're using the buildpacks.

choldgraf:
> In general, we may need to up our game with some new tests now that we're using the buildpacks.

yeah I think that's fair - think it's worth creating an issue with a list of build pack combinations we'd like a test for? It's a bit of a combinatorial problem but I agree we should have a test for most common build pack combos

willingc:
I was actually thinking about lower level tests. How long do tests take to run locally for you?

choldgraf:
Hmmm, not super long though it's been a while since I've run them locally.
On Travis all builds take roughly 6 minutes. So locally probably like 4
min? (gonna head to bed now btw, if there's other stuff to take care of in
this PR I'll get to it tomorrow AM!)

betatim:
Fixing a legit problem!

PR #279, opened 2018-03-21
Add codecov configuration to enable PR comments

betatim:
Trying to work out why we don't have the codecov bot commenting on PRs. We upload [reports](https://codecov.io/gh/jupyter/repo2docker) but don't have the bot comment. I guess it is because it is disabled for the jupyter team?

This is inspired by the comments in #277: if we want to increase coverage (I think we want to) we should make the number a bit more visible. Having the bot comment works well for me. Even better is having the browser extension installed but my assumption would be that most contributors don't have it (and it doesn't work in Firefox anymore :( )

WDYT?

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,10 @@
+# show coverage in CI status, not as a comment. Never consider it a failure.
+comment: off
+coverage:
+  status:
+    project:
+      default:
+        target: 0%
+    patch:
+      default:
+        target: 0%
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,7 +83,9 @@ Then you are good to go!
 ## Running tests
 
 We have a lot of tests for various cases supported by repo2docker in the `tests/`
-subdirectory. These use [py.test](https://docs.pytest.org/). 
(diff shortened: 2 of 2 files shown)


minrk:
Personally, I super don't like coverage comments. That's part of why it's disabled on the global jupyter config. What I *do* like is making the info visible in the ci status, which is what is done [on jupyterhub](https://codecov.io/account/gh/jupyterhub/yaml):

```yaml
comment: off
# show coverage in CI status, but never consider it a failure
coverage:
  status:
    project:
      default:
        target: 0%
    patch:
      default:
        target: 0%
```

You can see an example [on this PR](https://github.com/jupyterhub/jupyterhub/pull/1739). I like this because it makes the info visible and readily available, but without polluting PRs with a huge amount of noise in the comments.

betatim:
I like it!

I also added a comment to the contributing docs to nudge people towards adding tests. I tried to add a short sentence that explained "go to the CI status, check coverage" as well but couldn't formulate a good sentence that didn't sound super patronising/confusing to new comers.

(I don't know why the codecov comment suddenly appeared now ...)

minrk:
Nice!

PR #283, opened 2018-03-21
Add note about os support and update example repo

### LINKED ISSUE #281: Change example repo in README
The example in the README tells people to try  `jupyter-repo2docker https://github.com/jakevdp/PythonDataScienceHandbook`. I think this repo is "broken" because of something with matplotlib(??).

It also takes a long time to build locally and I can't get it to launch reliably on mybinder.org. Can we find a different repo that builds more quickly and more reliably? I think fast runtime is important for getting started examples. We also need something that isn't totally trivial though so that people are impressed. How about https://github.com/norvig/pytudes which contains lots of interesting stuff and builds quite quickly?

willingc:
Closes #281 

- Update example repo
- Add note about os support and link to windows in docs
- reflow text

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,10 @@
 # jupyter-repo2docker
 
-
 [![Build Status](https://travis-ci.org/jupyter/repo2docker.svg?branch=master)](https://travis-ci.org/jupyter/repo2docker)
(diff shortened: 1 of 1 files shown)


betatim:
Tip top!

PR #284, opened 2018-03-23
Enables Python2 just with runtime.txt file

### LINKED ISSUE #280: Enabling Python 2 requires runtime.txt and requirements.txt
When a repository contains only a `runtime.txt` containing `python-2.7` no Python 2 is available. To get the other version you need to also have a `requirements.txt`.

From reading the docs it isn't clear to me if this is the intended behaviour. I found this behaviour counter intuitive. So even if it currently works as designed I would propose we change repo2docker so that it is enough to have a `runtime.txt` to switch to Python 2.

Launch this on mybinder.org: https://github.com/betatim/binder-widgets-py2/tree/b54041929c7c8353b5f2544cf08fa857e5f9be2c as an example.

GladysNalvarte:
Bulidpacks consider now two options to enable Python 2: runtime.txt with and without requirements.txt and for Python 3 it is only needed requirements.txt, runtime.txt can be included or not.

closes #280

### DIFF
--- a/repo2docker/buildpacks/python/__init__.py
+++ b/repo2docker/buildpacks/python/__init__.py
@@ -136,8 +136,17 @@ class PythonBuildPack(BaseImage):
     def detect(self):
         """Check if current repo should be built with the Python 3 Build pack.
         """
-        return (os.path.exists(self.binder_path('requirements.txt')) and
-                super().detect())
+        requirements_txt = self.binder_path('requirements.txt')
+        runtime_txt = self.binder_path('runtime.txt')
+
+        if os.path.exists(runtime_txt):
+            with open(runtime_txt) as f:
+                runtime = f.read().strip()
+            if runtime.startswith("python"):
+                return True
+            else:
+                return False
+        return os.path.exists(requirements_txt)
 
 
 class Python2BuildPack(PythonBuildPack):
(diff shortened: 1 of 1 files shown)


repo2docker/buildpacks/python/__init__.py
```diff
@@ -238,19 +247,21 @@ def get_build_scripts(self):
...
-            (
+        requirements_txt = self.binder_path('requirements.txt')
+        assemble_scripts = super().get_assemble_scripts()
+        if os.path.exists(requirements_txt):
+            assemble_scripts.append((
                 '${NB_USER}',
-                'pip2 install --no-cache-dir -r requirements.txt'
-            )
-        ]
+                'pip3 install --no-cache-dir -r "{}"'.format(requirements_txt)
```
betatim: I think this should be `pip2` and not `pip3`. If the change is on purpose we should leave a comment because it is confusing to have `pip3` in the Python 2 build pack.

repo2docker/buildpacks/python/__init__.py
```diff
@@ -136,8 +136,17 @@ def get_assemble_scripts(self):
...
         """
-        return (os.path.exists(self.binder_path('requirements.txt')) and
-                super().detect())
+        requirements_txt = self.binder_path('requirements.txt')
+        runtime_txt = self.binder_path('runtime.txt')
+
+        if os.path.exists(runtime_txt):
+            with open(runtime_txt) as f:
+                runtime = f.read().strip()
+            if runtime.startswith("python"):
```
betatim: This will also match `python-2.7`. Should we instead check for `startswith("python-3")`?

betatim:
Looks good to me. Merging.

PR #286, opened 2018-03-26
[WIP] test travis changes to see if we can decrease to build times < 10 min to avoid timeouts

willingc:


.travis.yml
```diff
@@ -10,8 +10,9 @@ install:
   # issues with releases
   - pip install --upgrade setuptools pip
   - pip install -r dev-requirements.txt
-  - python setup.py bdist_wheel
-  - pip install dist/*.whl
+  - pip install -e .
+ # - python setup.py bdist_wheel
```
betatim: IIRC this was added to catch mistakes in `MANIFEST.in` that we'd only noticed after making a release. So if this saves us time in each build we should find a replacement for this check.

betatim:
If we can't make the julia build reliably take <10 then [`travis_wait`](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received) might be worth investigating

.travis.yml
minrk: MANIFEST.in doesn't affect bdist_wheel, only sdists. But this does test that we are installing the files we need, where `pip install -e` does not. A `pip install .` should cover the same, though.

betatim:
Can/should we revive this PR? The Julia build/tests keep being annoyingly flaky so improving on them would be good.

yuvipanda:
I'm gonna close this one for now.

PR #289, opened 2018-03-27
Support setup.py in Python buildpacks

GladysNalvarte:
Implements #249 
We should find repos to test this with to be sure that it works on real packages.
Some possibles examples to test: ipython, notebook, pandas, others?

### DIFF
--- a/repo2docker/buildpacks/python/__init__.py
+++ b/repo2docker/buildpacks/python/__init__.py
@@ -117,7 +117,7 @@ class PythonBuildPack(BaseImage):
         # will be installed in python3 venv. This is less of a
         # surprise than requiring python2 to be requirements2.txt tho.
         assemble_scripts = super().get_assemble_scripts()
-        setup_py = self.binder_path('setup.py')
+        setup_py = 'setup.py'
         try:
             with open(self.binder_path('runtime.txt')) as f:
                 runtime = f.read().strip()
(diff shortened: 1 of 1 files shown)


repo2docker/buildpacks/python/__init__.py
```diff
@@ -117,27 +117,36 @@ def get_assemble_scripts(self):
         # will be installed in python3 venv. This is less of a
         # surprise than requiring python2 to be requirements2.txt tho.
         assemble_scripts = super().get_assemble_scripts()
+        setup_py = self.binder_path('setup.py')
```
betatim: This will find a `setup.py` which is in a `binder/` sub-directory. I think we should not pick that up. Not sure when such a file would exist but with enough repos eventually we will find one that does have a `binder/setup.py`.

betatim:
We should add to the documentation explaining the behaviour for `setup.py` files. (I can't keep it straight in my head and keep having to refer back to https://github.com/jupyter/repo2docker/issues/249#issuecomment-373756016.)

betatim:
Ready to merge this?

repo2docker/buildpacks/python/__init__.py
```diff
@@ -146,6 +155,8 @@ def detect(self):
                 return True
             else:
                 return False
+        if not os.path.exists('binder') and os.path.exists(setup_py):
+            return True
```
choldgraf: Does this mean that if a repo *doesn't* have a `binder/` folder, nor a `requirements.txt`, but *does* have a `setup.py` file then it'll still be built?

minrk: yes, i.e. pointing Binder to a typical, basic Python package.

choldgraf:
thanks a bunch @GladysNalvarte !! I opened https://github.com/jupyter/repo2docker/issues/297 to make sure we remember to document it

PR #291, opened 2018-04-02
update nteract_on_jupyter to 1.6.0

rgbkrk:
Upgraded `nteract_on_jupyter`.

I also took the liberty of gitignoring the conda built outputs that are not tracked (according to https://github.com/jupyter/repo2docker/blob/master/CONTRIBUTING.md#conda-dependencies).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,7 @@ include/
 # Docs
 generated/
 test_file_text.txt
+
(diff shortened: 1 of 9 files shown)


betatim:
LGTM. Waiting for travis.

---

In the longer term we should think about when/if we freeze for good some dependencies. Maybe at the same time as we freeze the base image?

PR #294, opened 2018-04-09
split  post build script command in Docker file

bitnik:
Hi, `repo2docker` throws `./postBuild: Text file busy` error with Docker version 1.13.1 if there is any postBuild file in repo. This problem is docker related, I don't get the same error when I use Docker version 18.03.0-ce. Splitting docker command fixes it.

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -109,7 +109,8 @@ USER ${NB_USER}
 # Make sure that postBuild scripts are marked executable before executing them
 {% if post_build_scripts -%}
 {% for s in post_build_scripts -%}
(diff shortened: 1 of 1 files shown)


betatim:
For my education: how are there two versions of docker with such different version numbers?

Otherwise fine with me and travis is happy.

bitnik:
Firstly thank you, before I didn't think about this difference between versions.

1.13.1 is a version of old Docker Engine and 18.03.0-ce is a version of Docker Community Edition. On March 2017 they changed name of the Docker Engine package and also versioning scheme. The new versioning is time-based (for example 18.03.0-ce is released on 03.2018). Here is the related blog post: https://blog.docker.com/2017/03/docker-enterprise-edition/

PR #298, opened 2018-04-16
Always install Python with conda

### LINKED ISSUE #146: runtime.txt won't install other python 3 versions

### LINKED ISSUE #185: use conda to get Python, even for requirements.txt
Using conda to get Python itself would simplify supporting multiple Python versions (#146) for the pure Python requirements.txt case. It would also simplify things because then requirements.txt and environment.yml would only differ by one line: whether to use `conda env update -f` or `pip install -r` to install the user's files.

This will allow us to easily support many Python versions without relying on what's packaged in the base distro.

minrk:
closes #185 
closes #146 

- expands support for multiple Pythons
- consolidates maintenance of base env
- simplifies Python buildpack
- eliminates Python2 buildpack, as conda buildpack handles Python 2 already
- KERNEL_PYTHON_PREFIX env is used for the kernel prefix


repo2docker/buildpacks/python/__init__.py
```diff
@@ -1,139 +1,66 @@
...
-        if runtime == 'python-2.7':
-            pip = "pip2"
-            requirements_file = self.binder_path('requirements3.txt')
-        else:
-            pip = "pip3"
-            requirements_file = self.binder_path('requirements.txt')
+        pip = 'pip'
+        if self.py2:
+            # using python 2 kernel,
+            # requirements3.txt allows installation in the notebook env
```
betatim: -> ... in the notebook server env. For example to install an additional notebook extension.

I had to read the comment a few times before I realised what it was good for. The 3 at the end of the filename was confusing me "why are we installing python 3 things with python 2 in use?????"

minrk:
This should be good to go.

repo2docker/buildpacks/python/__init__.py
```diff
@@ -1,139 +1,70 @@
...
-                """
-            )
-        ]
+        try:
+            with open(self.binder_path('runtime.txt')) as f:
+                runtime = f.read().strip()
+        except FileNotFoundError:
+            runtime = ''
+
+        if not runtime.startswith('python-'):
```
choldgraf: do we have any future plan to support multiple languages with `runtime.txt`? (e.g. both python and R).

tests/venv/py3/README.rst
```diff
@@ -4,6 +4,5 @@ System - Specifying runtime environments
 You can specify runtime environments (such as Python 2 or 3) with a
 ``runtime.txt`` file.
 
-This is an example that selects Python 3. Currently you can not use
-this to select a sepcific version of Python 3 (e.g. 3.4 vs 3.6). If you
-need this level of control we recommend you use a `environment.yml`.
+This is an example that selects Python 3.
+If you need this level of control we recommend you use a `environment.yml`.
```
choldgraf: Perhaps there's something deleted that shouldn't be? It sounds like this line is suggesting that people just don't ever use `runtime.txt`

tests/venv/py35/README.rst
```diff
@@ -0,0 +1,8 @@
+System - Specifying runtime environments
```
choldgraf: Note that these READMEs are mostly just for building the docs...if a test doesn't have a README file then it just won't be added to the docs, so for tests that duplicate functionality (but are there for test coverage purposes) I'm fine not including READMEs in them

choldgraf:
I think this is a nice simplification! A couple small comments in there. @betatim you wanna have another whack at it?

Just so we're making sure we aren't missing something: what are the potential downsides to using `conda` for all of our python needs? Speed? I would have said "complexity", but I think this PR definitely makes the code more simple :-)

minrk:
> what are the potential downsides to using conda for all of our python needs?

Since we're only using conda to get Python itself, the main downside is that not all potential Pythons are packed with conda (e.g. pypy, etc.). They all *could* be, but it's not the case right now.

There is probably a slight speed and size cost for getting Python from conda on a single build from scratch, but the fact that all images will be sharing the same starting point may well make up for it.

The fact that Python comes from conda shouldn't have a complexity cost, since everything after that only assumes that Python is installed somehow, and should be insensitive to how.

tests/venv/py35/README.rst
minrk: I didn't realize that's what these readmes were for! I'll look at these again to see if they should be changed or just removed.

choldgraf: Also note #300 which updates a bunch of the READMEs

minrk:
Removed the extra READMEs, so should be set now

choldgraf:
@betatim wanna give it a look? or are we ready to merge?

yuvipanda:
Since this is a major change, should probably also bump the image prefix in mybinder.org.

betatim:
We considered bumping the prefix when we made the deploy but decided against it. I have no strong feelings so if there is support/no opposition go ahead and bump it.

yuvipanda:
Ah, ok! I do have strong feelings, but I'll defer to the people actually doing the work now :D <3

PR #300, opened 2018-04-23
Documentation and warning improvements

### LINKED ISSUE #125: Make it clear that windows support is experimental
Windows behaves very differently with lots of the commands that repo2docker uses...I don't think it'll work right now (running it on my machine just makes it hang when in windows).

### LINKED ISSUE #135: --ref does not understand branch names
I had assumed that `--ref` would also take a branch name or a tag, however it fails with the following error:
```
Unpacking objects: 100% (22/22), done.
Checking connectivity... done.
fatal: ambiguous argument 'other-branch': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Failed to check out ref other-branch
```

### LINKED ISSUE #218: Improve the READMEs for the example repositories
The sample repositories section is a bit sparse in terms of explanations / clarifications / etc:

http://repo2docker.readthedocs.io/en/latest/samples.html

these are built from the README file of the various testing repositories, e.g.:

https://github.com/jupyter/repo2docker/tree/master/tests/conda/requirements

those READMEs should be improved to be more explanatory!

choldgraf:
This is a quick fix for two issues:

* closes #125 - this now throws a warning if a `win*` platform is detected
* closes #135 - we now throw a more friendly error if a `remote/refname` pattern isn't found for `--ref`.
* closes #218 - this improves the language for several of the readmes, and removes redundant readmes so that we don't have duplicated information.


repo2docker/app.py
```diff
@@ -176,6 +176,8 @@ def fetch(self, url, ref, checkout_path):
             sys.exit(1)
 
         if ref:
+            if len(ref.split('/')) != 2:
+                raise ValueError('--ref must be of the form remote/ref')
```
betatim: This caused travis to fail.

I would change the error message to `"Expected --ref to be of the form remote/ref, but got %s" % ref` so that users can see what value arrived.

repo2docker/app.py
```diff
@@ -245,7 +247,8 @@ def get_argparser(self):
 
         argparser.add_argument(
             '--ref',
-            help='If building a git url, which ref to check out'
+            help=('If building a git url, which ref to check out. Must be of'
```
betatim: is ref a jargon-y word for reference? I am not sure if ref is the shorthand for reference or not. If it is shorthand we should make the effort to spell it out.

choldgraf:
I'm a bit confused as to why travis is failing...it was my understanding from #135 that we needed `remote/ref` as the pattern to `--ref`, but this is the error:

`fatal: ambiguous argument 'origin/b8259dac9eb': unknown revision or path not in the working tree.`

Is the test that we're using different from how people would be using this in the wild? Perhaps I misinterpreted the `--ref` problem?

minrk:
A ref is one of:

- commit hash
- tag
- local branch name
- origin/branch

The `origin/` prefix is only valid when it's a remote branch, and is redundant for any case that's actually a clone (as opposed to building from an already-checked-out repo).

choldgraf:
ah ok - I was thinking for some reason that we were only operating on a remote basis here but I can see why this is an issue.

hmm, so in this case it sounds like we want to check if we're pointing to a remote repository, and then do the `/` split check. What's the best way to do this? Just look for http/https in the first argument?

betatim:
After re-reading https://github.com/jupyter/repo2docker/issues/135 I think we should make it so that `--ref branchname` "just works" (from a UI perspective). I find it is counter-intuitive that it doesn't work without `origin/`. Can we reliably detect that it is a branch name and add `origin/` for the user?

choldgraf:
I'm a bit lost as to what needs to be done on this. Can we break it down into a simple:

* IF XXX, check for `remote/ref`
* ELSE, don't enforce `remote/`

?

tests/venv/py3/README.rst
```diff
@@ -2,8 +2,19 @@ System - Specifying runtime environments
 ----------------------------------------
 
 You can specify runtime environments (such as Python 2 or 3) with a
```
jzf2101: FWIW, we can also use R runtimes here too

repo2docker/app.py
```diff
@@ -176,6 +176,9 @@ def fetch(self, url, ref, checkout_path):
             sys.exit(1)
 
         if ref:
+            if len(ref.split('/')) != 2:
+                raise ValueError('Expected --ref to be of the form '
+                                 'remote/reference, but got %s' % ref)
```
betatim: `repo2docker --ref e287c39f62f290e3b6927da08c81586f647d5a4c ...` is currently valid, I think it should continue to be valid. This change would make it invalid though or is the explicit commit hash case handled elsewhere?

choldgraf:
I think we can merge if you'd like to address @betatim 's comment above in your re-work PR...what do you think?

choldgraf:
ah no I take it back, I think there's an error making travis unhappy...I can try to find some time to fix this soon

choldgraf:
ok I think the latest commit will fix these problems - instead of throwing an error ir "ref" doesn't include "origin", I'm checking if the repo is a remote one, and if it is then adding "origin/" if it isn't already there.

choldgraf:
now I'm a bit confused - travis seems unhappy because a test is building the LIGO repository *without* `origin/` in the `--ref`, but I was under the impression this was needed in order to work with remote repositories.

https://travis-ci.org/jupyter/repo2docker/jobs/397404948#L826

Any ideas what's going on @betatim ?

choldgraf:
ah nvm, I see it's checking out a commit hash and not a branch name. So we should only add `origin/` if it's a branch name...but how to ensure we know it's a branch name 😕

choldgraf:
ok, latest push should ensure that the ref will resolve properly by using `show-ref`. Let's see if that works...

choldgraf:
I think this is now ready to go!

jzf2101:
I'm ok with this are you @betatim ?

betatim:
Thanks for following up with this old PR :)

PR #301, opened 2018-04-25
Add newline conversion for postBuild scripts

### LINKED ISSUE #263: Always convert line endings in `postBuild` to unix
We should filter/convert the line endings in `postBuild` as that seems to cause problems (see https://github.com/jupyter/repo2docker/issues/237#issuecomment-373446647)

Is there a docker option to have it convert line endings on COPY or do we have to convert it before copying or add a tool like `dos2unix` to the container?

betatim:
Fixes #263

Python normalises line endings to always be "platform appropriate". So by reading in a file that potentially uses `\r\n` (because it was checked out from git on Windows) and writing it out again on a linux platform (inside a docker container) we make sure files only use `\n`.


betatim:
What do people in general think about this?

If it has support I would add some tests with explicit `\r\n` in them to make sure this works.

choldgraf:
I'm +1, I think it's a fair concern that it slightly change the thing that is run on the repo2docker side vs. whatever is on the computer. However, since this is limited to line endings I think it's OK. Otherwise any user that makes things on windows will have to 1. learn about the concept of line endings, and 2. remember to change this, every time the wanna use postBuild.

tl;dr, I can't really think of a situation where we'd think "damn, I wish we had not forced *nix line endings", so I'm +1

betatim:
Does anyone have an idea why suddenly the memlimit tests fail? Doesn't seem like something that would be flaky but also not really related to changes here I'd say.

betatim:
Can we add something that asserts that the original `postBuild` in the test contains CRLFs? So that we find out if that ever stops being the case which would make the test useless.

repo2docker/buildpacks/base.py
```diff
@@ -110,6 +110,8 @@
 {% if post_build_scripts -%}
 {% for s in post_build_scripts -%}
 RUN chmod +x {{ s }}
+# Fix up line endings for users who checked out postBuild scripts on Windows
+RUN python -c "open('{{ s }}', 'w').write(open('{{ s }}').read())"
```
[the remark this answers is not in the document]
betatim: We could add a test checking if `file postBuild` thinks this is a binary but I'd vote to do that once we have someone who actually uses binaries for `postBuild`.

tests/venv/newline-conversion/postBuild
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# I contain CRLF line endings on purpose
```
betatim: I had hoped to see some extra characters/symbol here on GH to verify that this file contains CRLF. Could someone on a not-windows system checkout this branch to confirm my `.gitattribute` worked?

choldgraf: I opened it on my windows machine and it said that it was in CRLF mode, which is what we want. Just FYI, you should be able to open this with Atom or something like this, and it should tell you the line endings present in the file (and if not by default, then there are plugins for this kinda thing for most text editors)

betatim: Locally I see the extra characters but because git can/will mangle the line endings when you checkin/-out I wanted to see if someone somewhere else can also see them after it has gone through git :)

betatim:
Where/how would we get that snippet executed as part of the test? `verify` is run inside the container where we will see the converted version :-/

betatim:
Does someone have an idea why the `test_memlimit_nondockerfile_fail` test fails?

yuvipanda:
What shall we do with this one, @betatim? Do you think there have been enough complaints about this in the last two years for us to push this through?

consideRatio:
I'm triaging some PRs - @betatim do you have a recommendation on what to do with this?

betatim:
I think it would be useful and doesn't add too much complexity/magic. However it seems like there isn't a lot of interest/need. If someone is interested to pick up this stalled PR that would be great.

consideRatio:
I'll go for a close at this point since we are not actively working this PR and since it would introduce a possible issue for binary postBuild executables.

PR #302, opened 2018-04-26
bump base image to ubuntu 18.04

### LINKED ISSUE #198: Bump base distro after every new release until next LTS
The next Ubuntu Release (Bionic) is an LTS release supported for a long time. When it comes out, we should bump the default to it, and then just let it be there until the next LTS gets released. This helps prevent issues like #197

minrk:
Release will be later today, but images are already being published with the betas.

closes #198


choldgraf:
woot!

just a note that I believe 18 is an LTS release, so we should stop bumping major version numbers after this, yes?  in that case I believe this closes #198

PR #303, opened 2018-04-26
remove _nb_ext_conf from legacy env

minrk:
`_nb_ext_conf` adds configuration that requires `nb_conda_kernels`, which has been removed, preventing launch.

This means that currently, legacy images cannot launch.


betatim:
Should we/could we add a test for when the legacy image breaks? Different way of asking this: what do we need to add to our existing legacy image test to catch this?

minrk:
We would need to actually start the notebook server to catch this. In general, our tests do not verify that the image results in a running notebook server. We should probably figure out a way to do that, but I think that should be another issue/pr

betatim:
Merge now and put on the list of changes to deploy tomorrow?

minrk:
Yes, now that tomorrow is not a holiday, unlike last week

PR #308, opened 2018-05-10
Add links to the front page and modify travis

betatim:
It should be easier to find us to ask questions, report bugs and to get
started contributing.

Right now you really need to know where to find us to find us.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,11 @@ install:
 script:
   # cd into tests so CWD being repo2docker does not hide
   # possible issues with MANIFEST.in
-  - cd tests && travis_retry pytest --cov repo2docker -v ${REPO_TYPE}
+  - if [ ${REPO_TYPE} == "r" ];
+    then cd tests && travis_wait pytest --cov repo2docker -v ${REPO_TYPE};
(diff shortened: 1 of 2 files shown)


choldgraf:
hmmm, so the R build has failed twice on this PR now, but I don't know why it would be affected by these changes :-/  https://travis-ci.org/jupyter/repo2docker/jobs/377437418

betatim:
Restarted the build, I suspect the test runs just fast enough that under normal load on travis we are done under 10m but when travis is busy it takes longer?

betatim:
I've been running this test locally to see what is up. Installing R packages just seems to take a very long time as there is a lot of compiling going on.

I suggest adding https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received for the R tests. I think it is a good idea to not give a blanket allowance to all tests to not produce output for a long time.

betatim:
This PR now adds links to the front page and modifies the R test to allow it to not produce output for 20minutes instead of the default 10m. If there is need to discuss the travis stuff I'd remove it from this PR and make a new one. Otherwise merge this?

choldgraf:
travis is happy so :shipit:

PR #309, opened 2018-05-11
stencila support

minrk:
Supports stencila editor for DAR files

Additionally separates `get_build_env` from `get_env` to allow defining environment variables that do not affect build, which would force a rebuild from an earlier stage.

Developed at the eLife Innovation Sprint

TODO:

- [x] enable R runtime
- [x] remove hardcoded numpy/matplotlib installs in favor of real dependencies
- [x] remove hardcode of default archive name
- [x] merge upstream [fixes](http://github.com/jupyterhub/nbserverproxy/pull/34) for nbserverproxy
- [x] tests


choldgraf:
This is awesome!

repo2docker/buildpacks/stencila/stencila.py
```diff
@@ -0,0 +1,135 @@
...
+        Dict of files to be copied to the container image for use in building.
+
+        This is copied before the `build_scripts` & `assemble_scripts` are
+        run, so can be executed from either of them.
+
+        It's a dictionary where the key is the source file path in the host
+        system, and the value is the destination file path inside the
+        container image.
+        """
+        return super().get_build_script_files()
```
betatim: Do we need this if it just calls `super()`?

repo2docker/buildpacks/stencila/stencila.py
```diff
@@ -0,0 +1,135 @@
...
+                r"""
+                mkdir -p ${STENCILA_DIR} && \
+                chown -R ${NB_USER}:${NB_USER} ${STENCILA_DIR}
+                """,
+            ),
+            (
+                "${NB_USER}",
+                # install nbserverproxy
+                r"""
+                pip install --no-cache https://github.com/minrk/nbserverproxy/archive/del-starting.tar.gz
```
betatim: Switch to a tag/commit on the official repo?

betatim:
I was confused after reading the code because I was expecting it to look for dependencies to install or something like that. Does that happen somewhere?

Then I saw https://twitter.com/NokomeBentley/status/994831220512092162 and played with it a bit. Now I am thinking stencila is more like nteract-on-jupyter than a new build pack? /me is confused.com

choldgraf:
@betatim FWIW, apparently nteract now runs on Jupyter as well https://blog.nteract.io/nteract-on-jupyter-53cc2c38290d

betatim:
> @betatim FWIW, apparently nteract now runs on Jupyter as well https://blog.nteract.io/nteract-on-jupyter-53cc2c38290d

That is what I meant with nteract-on-jupyter :) The way we added that to repo2docker was adding a new dependency. Which is why I am confused, because this is a build pack.

repo2docker/buildpacks/stencila/stencila.py
minrk: Not at the moment. It used to have things, and it may again before we are done, so I haven't cleaned it out just yet. I'll remove this if it turns out not to be needed in the end.

repo2docker/buildpacks/stencila/stencila.py
minrk: Yes, now that it's been merged.

choldgraf:
@minrk any reason this was closed abruptly?

minrk:
@choldgraf oops. I think i may have accidentally deleted the branch. Reopened.

I've been working on https://github.com/minrk/nbstencilaproxy consolidating much of this.

In the end, I think the logic for stencila should be:

1. detect/install dependencies as usual for any other buildpack
2. locate DAR archive directory and set STENCILA_ARCHIVE_DIR
3. if found, add/enable `nbstencilaproxy` server extension
4. optionally, set *default* url to `/stencila/` to avoid needing to set `urlpath`

It's only the STENCILA_ARCHIVE_DIR bit that can't be addressed by just adding a package to the dependencies in a regular repo of the other flavors. Maybe there's a way to do that?

Basically, this *should* be treated just like nteract as @betatim mentioned, but optional instead of in every env by default.

choldgraf:
sounds good to me re: the logic!

nuest:
@minrk How can I help to get this merged? I'd be happy to put in some time, but would appreciate your guidance on which tasks I don't need 10 times longer than you do ;-) 

Is the task list above complete? AFAICS "enable R runtime" can be checked.

minrk:
I'm happy to ship it

choldgraf:
then may the merge gods look favorably on your path, PR #309!

PR #311, opened 2018-05-12
To add the user bin folder to the path

aborruso:
Here the thread https://github.com/jupyterhub/binderhub/issues/557

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -72,7 +72,7 @@ ENV {{item[0]}} {{item[1]}}
 {% endif -%}
 
 # So users can move executables here with postBuild, issue #557
-ENV PATH {{ /home/jovyan/{local}/bin:${PATH}}}
+ENV PATH {{ '/home/jovyan/local/bin':${PATH}}}
 
 {% if path -%}
 # Special case PATH
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -71,6 +71,9 @@ ENV {{item[0]}} {{item[1]}}
 {% endfor -%}
 {% endif -%}
 
+# So users can move executables here with postBuild, issue #557
+ENV PATH {{ /home/jovyan/{local}/bin:${PATH}}}
+
 {% if path -%}
 # Special case PATH
 ENV PATH {{ ':'.join(path) }}:${PATH}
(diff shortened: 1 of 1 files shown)


repo2docker/buildpacks/base.py
```diff
@@ -71,6 +71,9 @@
 {% endfor -%}
 {% endif -%}
 
+# So users can move executables here with postBuild, issue #557
+ENV PATH {{ /home/jovyan/{local}/bin:${PATH}}}
```
betatim: I think this needs to be `'/home/jovyan/local/bin'` instead of `/home/jovyan/{local}/bin`.

betatim:
Any opinions on `/home/jovyan/bin` vs `/home/jovyan/local/bin`?

choldgraf:
gauging from https://unix.stackexchange.com/a/8658 it does seem like `/usr/local/bin` is better because it's something that the user would be manually adding things to. WDYT?

choldgraf:
looking through it again, I don't think we need the double-curlies in there (and I think that's why a bunch of tests were failing too). I just patched the PR so let's see if that fixes it

betatim:
The SO article you linked talks about `/usr/local/bin` but you changed it to `~/local/bin`. Is that what you meant to do? I don't have any strong feelings for which path we end up using. Potentially a normal user isn't allowed to write to `/usr/local/bin` hence I was thinking use a path in the home directory.

repo2docker/buildpacks/base.py
```diff
@@ -71,6 +71,9 @@
 {% endfor -%}
 {% endif -%}
 
+# So users can move executables here with postBuild, issue #557
```
betatim: I wouldn't refer to the issue (we will forget which repo etc) and instead include any additional info we do need in the comment here. Personally I think the comment is fine as it is.

betatim:
Add a short test for this and then merge?

(restarted travis as it was the R tests that were failing)

choldgraf:
+1 on merge one way or another

@betatim do you think we should add `~/bin` as well? Or keep it as-is?

choldgraf:
woot, thansk for the patch @aborruso !!!

aborruso:
@choldgraf you are very kind, but it is only thanks to you all.

I can not wait to try it on mybinder :)

betatim:
Should be deployed now.

aborruso:
@betatim it works!!!!!!!

Great

yuvipanda:
Congratulations on the PR, @aborruso! 

I've made a minor change to this and proposed it ad #313. It changes the directory from ~/local to ~/.local to be consistent with where other software installs per-user software by default. I'm hoping to get that merged and deployed sooner than latter so not too many repos start depending on it :) Hope that's not too much of a change for you!

aborruso:
thank you @yuvipanda you are right. I will make some small changes in my code ;)

PR #312, opened 2018-05-16
Add test to check for extra directory on PATH

betatim:
Follow up #311 and add a simple test

### DIFF
--- a/tests/venv/usr-bin/verify
+++ b/tests/venv/usr-bin/verify
@@ -1,0 +1,6 @@
+#!/usr/bin/env python
+# Verify that ~/local/bin is on the PATH
+import os
+
+PATH = os.getenv("PATH")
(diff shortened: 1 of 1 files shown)


yuvipanda:
I've incorporated this into #313

choldgraf:
FWIW I think a better way to handle these situations it to put the word "closes #NNN" in the top-level comment of the PR. That way, whenever the PR is merged GitHub will automatically close the issue corresponding to `NNN`.

betatim:
Yes that is what we normally do. Not sure what you mean in this case though as this was a followup to a PR that had been merged. There was no corresponding issue. Then Yuvi cherry picked the commit from here into a new PR.

choldgraf:
ahh my bad, I had a brain fart and thought this was an issue, not a PR! Sorry for the noise

PR #313, opened 2018-05-19
Use XDG standard ~/.local for user software installs

yuvipanda:
~/.local is also where things like pip install --user
seem to install things, so we should use that instead of ~/local.

### DIFF
--- a/tests/venv/usr-bin/verify
+++ b/tests/venv/usr-bin/verify
@@ -1,0 +1,6 @@
+#!/usr/bin/env python
+# Verify that ~/local/bin is on the PATH
+import os
+
+PATH = os.getenv("PATH")
+assert "/home/jovyan/local/bin" in PATH
--- a/tests/venv/usr-bin/verify
+++ b/tests/venv/usr-bin/verify
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
-# Verify that ~/local/bin is on the PATH
+# Verify that ~/.local/bin is on the PATH
 import os
 
-PATH = os.getenv("PATH")
-assert "/home/jovyan/local/bin" in PATH
+assert "$HOME/.local/bin" in os.getenv("PATH")
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -71,9 +71,6 @@ ENV {{item[0]}} {{item[1]}}
 {% endfor -%}
 {% endif -%}
 
-# So users can move executables here with postBuild, issue #557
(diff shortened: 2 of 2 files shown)


yuvipanda:
see https://github.com/jupyter/repo2docker/pull/311#issuecomment-390368673

choldgraf:
I added in a short FAQ entry for this. If that looks OK to you then I'm +1 on this.

@betatim want to give input quickly? Wanna make sure you're cool w/ it since you were involved in round 1. I believe that @aborruso already gave their blessing (thx for the original patch again!)

docs/source/faq.md
```diff
@@ -9,6 +9,19 @@ If you have a question & have found an answer, send a PR to add it here!
 Currently the best way to do this is by using a conda ``environment.yml``
 file and setting the language to whichever version of Python you like.
 
+## Can I add binary files to the user's PATH?
```
betatim: binary files -> executables?

betatim:
Otherwise LGTM. Travis is unhappy though.

choldgraf:
fixed, let's see if travis becomes happy this time

aborruso:
Than now the folder in the path is  `~/.local/` and no more `/home/jovyan/locale/bin`. Am I right?

choldgraf:
@aborruso yep, it'll be in `~/.local/bin`

However, the travis error seems real as it's failing on the test added in this PR: https://travis-ci.org/jupyter/repo2docker/jobs/381119782#L1145

aborruso:
I have no idea on how to solve it 
I'm sorry

betatim:
@yuvipanda are you planning on working on this some more to make travis happy?

tests/venv/usr-bin/verify
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env python
+# Verify that ~/.local/bin is on the PATH
+import os
+
+assert "$HOME/.local/bin" in os.getenv("PATH")
```
minrk: I think this should be `os.path.expanduser('~/.local/bin')`

choldgraf:
restarting a failing travis job because I think it was a hang

choldgraf:
mergey merge merge, thanks for the patch @yuvipanda

aborruso:
thank you @yuvipanda

aborruso:
Is it just ready to use in mybinder?

Thank you

choldgraf:
I believe that we bumped repo2docker this morning, so give it a shot and see if it works

aborruso:
work like a charm

Thank you

PR #315, opened 2018-05-22
replace jovyan in path

brooksambrose:
Related to #311 #312 #313; but hard coding `jovyan` means passing a different `--user-name` leaves one without a directory in PATH to run executables.


betatim:
Good catch! We should land this after #313 at which point it will need a small change.

brooksambrose:
already fixed in #313, thanks folks!

choldgraf:
woot - thanks for keeping up on the issue!

Issue #223, opened 2018-02-07
repo2docker hangs on local install


psychemedia:
Running latest repo2docker fresh from pypi:

```
FROM buildpack-deps:artful

# Set up locales properly
RUN apt-get update && \
    apt-get install --yes --no-install-recommends locales && \
    apt-get purge && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \
    locale-gen

ENV LC_ALL en_US.UTF-8
ENV LANG en_US.UTF-8
ENV LANGUAGE en_US.UTF-8

# Use bash as default shell, rather than sh
ENV SHELL /bin/bash

# Set up user
ARG NB_USER
ARG NB_UID
ENV USER ${NB_USER}
ENV HOME /home/${NB_USER}

RUN adduser --disabled-password \
    --gecos "Default user" \
    --uid ${NB_UID} \
    ${NB_USER}
WORKDIR ${HOME}

RUN apt-get update && \
    apt-get install --yes --no-install-recommends \
       less \
       npm \
       && apt-get purge && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

RUN apt-get update && \
    apt-get install --yes \
       python3 \
       python3-dev \
       python3-venv \
       && apt-get purge && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*
EXPOSE 8888

# Almost all environment variables
ENV APP_BASE /srv
ENV VENV_PATH ${APP_BASE}/venv
ENV NB_PYTHON_PREFIX ${VENV_PATH}
# Special case PATH
ENV PATH ${VENV_PATH}/bin:${PATH}
# If scripts required during build are present, copy them

COPY python/requirements.frozen.txt /tmp/requirements.frozen.txt
RUN mkdir -p ${VENV_PATH} && \
chown -R ${NB_USER}:${NB_USER} ${VENV_PATH}

USER ${NB_USER}
RUN python3 -m venv ${VENV_PATH}

RUN pip install --no-cache-dir -r /tmp/requirements.frozen.txt && \
jupyter nbextension enable --py widgetsnbextension --sys-prefix && \
jupyter serverextension enable --py jupyterlab --sys-prefix && \
jupyter serverextension enable nteract_on_jupyter --sys-prefix



# Copy and chown stuff. This doubles the size of the repo, because
[...]

PR #323, opened 2018-05-30
Remove ability to set UID of the user added to the generated image

juanesarango:
Related with issue #223.

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -36,7 +36,9 @@ ARG NB_UID
 ENV USER ${NB_USER}
 ENV HOME /home/${NB_USER}
 
-RUN adduser ${NB_USER}
+RUN adduser --disabled-password \
+    --gecos "Default user" \
+    ${NB_USER}
 WORKDIR ${HOME}
 
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


betatim:
Comment with some motivation for this change https://github.com/jupyter/repo2docker/issues/223#issuecomment-393280799

Have you already experimented with removing options to see which ones in particular have to be removed and which ones can stay?

juanesarango:
Hi betatim, 

So, I think both `--disabled-password` and `--gecos` are needed to avoid the command to [ask for prompt input](https://askubuntu.com/questions/94060/run-adduser-non-interactively). Even thought that docker will skip and pass the defaults for those inputs, and that if you don't pass `--password` it will be disabled by default.

I see that the problem is with `--uid` option only. So I made some changes to my PR to only remove `--uid` and keep the other options, as I saw some travis tests failed.

choldgraf:
hey all - just a ping to see where we are on this PR - it seems from @betatim 's latest comment that there's some uncertainty as to whether this can be implemented without breaking other functionality?

betatim:
Removing the ability to set the UID is definitely not something we can or should do. I'd close this PR and start a new one to tackle the problem that users whose accounts are centrally managed run into #223. It isn't clear to me why having your account centrally managed causes this problem. Setting the UID to something different from your users account seems to fix the problem, but doing so has other drawbacks (permissions when mounting directories from the host).

PR #327, opened 2018-06-08
Include LICENSE file in wheels

toddrme2178:
The license requires that all copies of the software include the license.  This makes sure the license is included in the wheels.  See the wheel documentation [here](https://wheel.readthedocs.io/en/stable/#including-the-license-in-the-generated-wheel-file) for more information.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,5 @@
 [wheel]
 universal=1
+
+[metadata]
+license_file = LICENSE


betatim:
In principle Ok with me.

https://github.com/toddrme2178 looks like a bot or otherwise weird account though??

toddrme2178:
No, I am not a bot.  I am just trying to deal with a few common issues that I run into a lot.

choldgraf:
seems reasonable to me!

PR #342, opened 2018-07-02
Setup auto releasing on GitHub tagging

### LINKED ISSUE #337: Versions of Notebook and JLab don't match conda buildpack?
I just built https://github.com/binder-examples/conda to test the latest version of Docker with r2d.

However, the image I built uses Notebook `5.2.2` and JLab `alpha release
version: 0.30.6`.  People would probably feel more comfortable having a beta version since we're saying beta is better for general use.  

At the same time, neither version matches https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/conda/environment.yml

Where are these versions coming from?

Related to #336

### LINKED ISSUE #340: Cut New Release?
Right now we're encouraging people to install from pypi, but this gives users older versions of notebook and JLab etc see #337 

Should we cut language and tell people to install from master or cut a release on conda-forge or pypi?

betatim:
Closes #340, closes #337 

This should create a new release on PyPI when we create a new tag on GitHub.

Does anyone have opinions which one of our jobs we should use for the deployment? It seems like any of them would do


jzf2101:
This is related to #340

minrk:
Seems to work! No real way to test the auto-release stuff without making a release, so let's do an alpha of 0.6 when we're ready to test drive it.

PR #343, opened 2018-07-03
Test that images can start notebook servers in all the builders 

### LINKED ISSUE #305: Test that images can start notebook servers
#303 fixes a bug where the legacy buildpack can't launch a notebook server. We should probably be testing that notebook servers can start in all of our builders.

GladysNalvarte:
Updated  app.py and conftest.py to test in all the builders that images can start notebooks servers.
Closes  #305


minrk:
Excellent, thank you!

PR #344, opened 2018-07-03
Add DockerSpawner command in JupyterHub config

rprimet:
This may be useful to DockerSpawner users who want to create custom images using repo2docker


betatim:
Thanks for helping out with making the docs nicer!

rprimet:
Thanks for your suggestion! I tried to change my patch to be more affirmative

betatim:
Tim ❤️ the docs that ooze confidence :)

PR #346, opened 2018-07-06
Fix some rst weirdness in the uage docs

betatim:
Fix up some syntax errors in the usage document.


choldgraf:
nice catches :-) thanks!

PR #352, opened 2018-07-16
[MRG] Add version argument

### LINKED ISSUE #349: expose version from client
it would be great to be able to do:

```bash
jupyter-repo2docker --version
0.2.0
```
akin to:

```bash
python -c "import repo2docker; print(repo2docker.__version__)"
0.2.0
```

vsoch:
This pull request will add a `--version` option to the jupyter-repo2docker client, so the user can do:

```bash
$ jupyter-repo2docker --version
v0.5+234.g5c263a9.dirty
```
Since the repo and cmd are required arguments (and it would get angry without them) the easiest way to do this is to check for `--version` in the arguments **before** it's parsed by the argparser. The other way I've done is to have subcommands, but that would require a large change to the user interaction (and I don't think necessary). This will close #349

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -367,6 +367,8 @@ class Repo2Docker(Application):
         if argv is None:
             argv = sys.argv[1:]
 
+        # version must be checked before parse, as repo/cmd are required and
+        # will spit out an error if allowed to be parsed first.
         if '--version' in argv:
             print(self.version)
             sys.exit(0)
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -351,7 +351,6 @@ class Repo2Docker(Application):
             help='Print the repo2docker version and exit.'
         )
 
-
         return argparser
 
     def json_excepthook(self, etype, evalue, traceback):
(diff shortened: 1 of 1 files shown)


repo2docker/app.py
```diff
@@ -342,6 +343,15 @@ def get_argparser(self):
...
         )
+
+        argparser.add_argument(
+            '--version',
+            dest='version',
+            action='store_true',
+            help='Print the repo2docker version and exit.'
+        )
+
+
```
betatim: extra blank line

repo2docker/app.py
```diff
@@ -353,10 +363,16 @@ def json_excepthook(self, etype, evalue, traceback):
                        exc_info=(etype, evalue, traceback),
                        extra=dict(phase='failed'))
 
+
```
betatim: single blank lines between methods

repo2docker/app.py
```diff
@@ -342,6 +343,15 @@ def get_argparser(self):
             type=str,
             help=self.traits()['appendix'].help,
         )
+
```
betatim: should we copy your comment from the PR about why we are checking explicitly for `--version` without parsing and yet add this arg here as a code comment so that future generations remember why things are this way?

vsoch: sure!

betatim:
LGTM modulo above small comments.

betatim:
Merged! Thanks for helping out and congratulations on your first merged PR in repo2docker!

Issue #325, opened 2018-06-07
Hostname in link printed to access the notebook server


betatim:
Seems like when we switch to notebook 5.5 the link printed in the output will no longer work for people. See for example https://github.com/jupyter/docker-stacks/issues/642

Two questions:
* what is our policy for bumping versions on things like the notebook?
* should we hardcode the `--hostname` argument to docker in `repo2docker` to be `localhost` or `0.0.0.0` or something else entirely?

choldgraf:
I'm +1 on bumping versions to latest as long as it doesn't break things. I guess in this case we should just make the `localhost` hardcoded in there. Any idea why the notebook isn't using localhost by default anymore?

betatim:
It used to/currently print `http://0.0.0.0:62267/?token=e55a7...`. Don't know why this changed but my guess would be that for most uses cases printing the hostname of the interface is nicer for users than printing an IP address.

Instead of hard-coding things I'd check if we can turn off this behaviour via a flag.

betatim:
I think with the default version of the notebook (v5.4) that repo2docker uses this does not happen. So my recommendation would be to stick with that instead of using the newer notebook v5.5

The way to move this forward is to help get this discussion to a point where there is consensus on how to handle this (configuration in the notebook or hardcoded command line argument or ...).

choldgraf:
is this something that either @takluyver or @minrk have insight on?

takluyver:
There's work underway to make the address it prints configurable: https://github.com/jupyter/notebook/pull/3668

What it printed before was right by coincidence, but with Docker it's a coincidence that comes up quite reliably, so this change has upset a few people.

willingc:
Thanks @takluyver. 

I agree with @AaronWatters that we should strive for a solution that has minimal impact on the end user.

AaronWatters:
Thanks @betatim -- I can confirm that the problem doesn't occur if i use
```
pip3 install --no-cache-dir notebook==5.4.1 
```

betatim:
From https://github.com/jupyter/notebook/pull/3668 it looks like there will be a configuration option for this in notebook in the next release of the notebook.

Currently repo2docker uses notebook v5.4 so we aren't yet impacted by this. I'd suggest we don't update the default version of the notebook to v5.5 and wait till https://github.com/jupyter/notebook/pull/3668 has been merged and released. Then go straight for that version with the appropriate configuration set in the default command repo2docker executes.

minrk:
[...]

PR #354, opened 2018-07-17
bump versions and refreeze

minrk:
gets notebook 5.6, which includes fixes that blocked 5.5 in #325

A subsequent PR (@GladysNalvarte is working on this) can set the display_url so that it's guaranteed to be correct when run in repo2docker, which has never been the case before.

### DIFF
--- a/repo2docker/buildpacks/conda/environment.frozen.yml
+++ b/repo2docker/buildpacks/conda/environment.frozen.yml
@@ -1,29 +1,34 @@
 # AUTO GENERATED FROM environment.py-3.6.yml, DO NOT MANUALLY MODIFY
-# Frozen on 2018-05-08 20:22:20 UTC
+# Frozen on 2018-07-16 23:09:21 UTC
 name: r2d
 channels:
   - conda-forge
   - defaults
   - conda-forge/label/broken
 dependencies:
+  - attrs=18.1.0=py_1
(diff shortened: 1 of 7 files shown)


betatim:
For my education: what is this display_url thing?

Do you know where we picked up twisted and zope.interfaces as new dependencies?

jzf2101:
I'm ok with this tho we'll need to update again as soon as jupyterlab has their next release as well.

minrk:
@betatim #358 is aiming at #325. notebook 5.6 allows setting `NotebookApp.custom_display_url`, which lets repo2docker tell the container exactly what to log for the "copy this URL" bit. We can always get that right, because *repo2docker* knows the right URL (even when using docker machine), while the container can't know without being told.

betatim:
If we are happy with the extra new dependencies: merge!

SylvainCorlay:
@minrk @betatim it appears that this change was just applied to the prod binder and unfortunately, it will break all C++ binders 😒 

the xeus-cling kernel requires a binary-compatible version of `pyzmq`, and we use a build from the QuantStack channel for this, which is in conflict with your pinning of pyzmq.

Would it be possible to revert that change and use the previous environment until we figure out a solution?

jzf2101:
We need to create a new release of r2d - how would you suggest we fix this

PR #355, opened 2018-07-17
Switch from PyDataScienceHandbook to Pytudes

betatim:
The data science handbook build has been broken for a while and users
end up with an error while following our documentation.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -20,15 +20,15 @@ repo2docker is called with this command::
 
 where ``<URL-or-path to repo>`` is a URL or path to the source repository.
 
-For example, use the following to build an image of the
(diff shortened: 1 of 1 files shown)


choldgraf:
I think this is a great idea, too bad that the DataScienceHandbook hasn't been able to resolve the scikit-image/numpy bug :-(

I'm +1 on merge whenever the tests pass, happy for you to merge it @betatim if you notice before I do

choldgraf:
also of note: I'm pretty sure we use the datasciencehandbook in other places in the docs as well..we should try and replace this wherever we can

betatim:
`grep -Ri datasciencehandbook *` doesn't find any more instances in this repo. Merging.

PR #358, opened 2018-07-20
Specify custom_connect_url argument

### LINKED ISSUE #325: Hostname in link printed to access the notebook server

GladysNalvarte:
so that the displayed connection URL is always correct

Closes #325

### DIFF
--- a/tests/test_connect_url.py
+++ b/tests/test_connect_url.py
@@ -1,0 +1,55 @@
+"""
+Test if the environment.yml is empty or it constains other data structure than a dictionary
+"""
+import os
+import sys
+import pytest
+import requests
+import time
+from repo2docker.app import Repo2Docker
+
+def test_env_yml(tmpdir):
+    tmpdir.chdir()
+    #q = tmpdir.join("environment.yml")
+    #q.write("dependencies:\n"
+    #        "  -  notebook==5.6.0rc1")
+    p = tmpdir.join("requirements.txt")
+    p.write("notebook==5.6.0rc1")
+
+    app = Repo2Docker()
+    argv = [str(tmpdir), ]
+    app.initialize(argv)
+    app.run = False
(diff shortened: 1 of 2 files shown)


tests/test_connect_url.py
```diff
@@ -0,0 +1,51 @@
...
+            except Exception as e:
+                print("Error: %s" % e)
+                time.sleep(i * 3)
+            else:
+                print(info)
+                success = True
+                break
+        assert success, "Notebook never started in %s" % container
+    finally:
+        # stop the container
```
minrk: In this test, we will want to make sure that our "custom_display_url" is included in the output. You can see the logs with:

```python
logs = c.logs().decode('utf8', 'replace')
```

So we want to check:

```python
expected_url = "http://{}:{}".format(app.hostname, app.port)
logs = c.logs().decode('utf8', 'replace')
assert expected_url in logs
```

tests/test_connect_url.py
```diff
@@ -0,0 +1,51 @@
...
+import time
+from repo2docker.app import Repo2Docker
+
+def test_env_yml(tmpdir):
+    tmpdir.chdir()
+    #q = tmpdir.join("environment.yml")
+    #q.write("dependencies:\n"
+    #        "  -  notebook==5.6.0rc1")
+    p = tmpdir.join("requirements.txt")
+    p.write("notebook==5.6.0rc1")
```
minrk: This can be just 5.6.0 now that it's been released.

minrk:
Thanks!

PR #363, opened 2018-07-25
Add launch script

### LINKED ISSUE #196: Add a preStart hook
This would be similar to postBuild, but would be run at *runtime*, before execution of the actual notebook. This easily allows things such as:

1. setting environment variables at runtime
2. Minor adjustments that can't be performed at build time 
3. New things we haven't thought of yet!
4. Custom parameters for notebooks / other applications

To prevent people from putting too many things here, it should have a default short timeout (5s?) to start.
 
The requirements for this script should be:

1. Be marked executable, can be in any language
2. Receives as parameters arguments to `exec` that it must perform as a final line to actually start the process we want (we will provide templates for this in various languages)
3. Not run for more than x seconds.

jhamman:
This is a rough draft of the launch script concept. It follows the pattern used by the `postBuild` script, except uses the `ENTRYPOINT` directive. 

closes #196


repo2docker/buildpacks/base.py
```diff
@@ -118,6 +118,14 @@
...
 {% endif -%}
 
+# Run launch script
+{% if launch_scripts -%}
+{% for s in launch_scripts -%}
+RUN chmod +x {{ s }}
+ENTRYPOINT ./{{ s }}
+{% endfor %}
+{% endif -%}
+
```
jhamman: question for @yuvipanda, in https://github.com/jupyter/repo2docker/issues/196#issue-288831988, you mentioned that this script should be time limited:

>Not run for more than x seconds.

Do you have an idea of how to do this? Perhaps prefacing this command with the `timeout` command?

repo2docker/buildpacks/base.py
```diff
@@ -118,6 +118,14 @@
 {% endfor %}
 {% endif -%}
 
+# Run launch script
+{% if launch_scripts -%}
+{% for s in launch_scripts -%}
+RUN chmod +x {{ s }}
+ENTRYPOINT ./{{ s }}
```
betatim: there can only be one ENTRYPOINT so I think we should error out if there is more than one script before we get here or construct things in a way that doesn't allow us to get into the situation where we have more than one launch script.

betatim:
What do you think of implementing the idea from https://github.com/jupyter/repo2docker/issues/196#issuecomment-364078438?

We also need to test this PR with the various ways kubespawner allows people to set the command that is executed. I remember it was a bit hairy last time we discussed this because there are several similar words that all mean slightly different things for docker and kubernetes :-/

jhamman:
@betatim - I've just pushed another commit here. This follows @yuvipanda's suggestion for how this could be implemented. I've tested locally and it actually works. 

> We also need to test this PR with the various ways kubespawner allows people to set the command that is executed.

What is the best way to do this? I haven't quite sorted out the unit test approach in this repo yet.

repo2docker/buildpacks/base.py
```diff
@@ -283,6 +295,21 @@ def get_post_build_scripts(self):
...
+        """
+        An ordered list of executable scripts to be executated at runtime.
+        These scripts are added as an `ENTRYPOINT` to the container.
+
+        Is run as a non-root user, and must be executable. Used for doing
+        things that are currently not supported by other means and need to be
+        applied at runtime (set environment variables).
+
+        The scripts should be as deterministic as possible - running it twice
+        should not produce different results.!
```
minrk: `.!` typo

docs/source/config_files.rst
```diff
@@ -119,6 +119,20 @@ An example usecase of ``postBuild`` file is JupyterLab's demo on mybinder.org.
 It uses a ``postBuild`` file in a folder called ``binder`` to `prepare
 their demo for binder <https://github.com/jupyterlab/jupyterlab-demo/blob/master/binder/postBuild>`_.
 
+.. _start:
+
+``start``
```
choldgraf: How do folks feel about the name "start"? It seems reasonable to me, but I'm wondering if anybody knows of a similar pattern that other communities have used? (similar to how we've used Heroku as inspiration a few times)?

(also either way, can you make sure the number of `^^^^` below matches the length of the title :-) )

choldgraf:
Can you make travis happy? https://travis-ci.org/jupyter/repo2docker/jobs/411604989#L912 :-)

jhamman:
I can. I was waiting until we had come to some agreement on the approach but if there are not high-level concerns, I will wrap this up.

choldgraf:
ah fair enough - if @minrk is happy w/ the approach (maybe @yuvipanda has thoughts too?) we can move forward!

betatim:
I think the general approach is good. However we need to adjust it so the existing tests pass. Seems like at least some of this is due to handling multiple arguments and such.

repo2docker/buildpacks/base.py
```diff
@@ -118,6 +118,18 @@
...
+{% if start_script -%}
+RUN echo '#!/bin/bash \nexec {{ start_script }} $@ \n' > ./launch
+RUN chmod +x {{ start_script }}
+{% else %}
+RUN echo '#!/bin/bash \nexec $@ \n' > ./launch
+{% endif -%}
+RUN chmod +x ./launch
+
+# Run launch script
+ENTRYPOINT ["./launch"]
```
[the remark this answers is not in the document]
jhamman: preferences on where this should land if not in the local/home directory?

jhamman:
Just changing the launch script path to using the home directory seems to have fixed all the failing tests. Nice call @minrk. This is ready for final review.

choldgraf:
@jhamman you're at the pangeo thin in Colorado, yeah? Maybe you can chat w/ @yuvipanda about this while you're both there!

choldgraf:
@minrk do you imagine people doing malicious things with this somehow?

Issue #316, opened 2018-05-23
Install nodejs/npm from nodesource


minrk:
We should be getting nodejs/npm from nodesource following the directions [here](https://github.com/nodesource/distributions#manual-installation) rather than installing from the base repo, which tends to be quite out of date (even now that we are on bionic, which came out this month, npm is already out of date by three years and three major versions).

PR #371, opened 2018-08-05
[MRG] Fix up node install and add test

betatim:
This adds a test and fixes (I think) the node install procedure added in #316


choldgraf:
hmmm, it seems like travis is unhappy, WDYT? https://travis-ci.org/jupyter/repo2docker/jobs/412427427#L1175

betatim:
Travis should be happy now.

choldgraf:
travis is happy!

PR #372, opened 2018-08-06
Fix warnings in the documentation build and add to CI

betatim:
Adds building our docs in "warnings as errors" mode to the CI setup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ script:
 - if [ ${REPO_TYPE} == "r" ]; then cd tests && travis_wait pytest --cov repo2docker
   -v ${REPO_TYPE}; else cd tests && travis_retry pytest --cov repo2docker -v ${REPO_TYPE};
   fi
+- pip install -r docs/doc-requirements.txt
+- cd docs && make html
 after_success:
 - pip install codecov
 - codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,17 @@ install:
 script:
 # cd into tests so CWD being repo2docker does not hide
 # possible issues with MANIFEST.in
-- if [ ${REPO_TYPE} == "r" ]; then cd tests && travis_wait pytest --cov repo2docker
-  -v ${REPO_TYPE}; else cd tests && travis_retry pytest --cov repo2docker -v ${REPO_TYPE};
-  fi
+- pushd tests;
+  if [ ${REPO_TYPE} == "r" ]; then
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -18,6 +18,8 @@ script:
 - if [ ${REPO_TYPE} == "r" ]; then cd tests && travis_wait pytest --cov repo2docker
   -v ${REPO_TYPE}; else cd tests && travis_retry pytest --cov repo2docker -v ${REPO_TYPE};
   fi
+- pip install -r docs/doc-requirements.txt
```
willingc: I'm not sure that these two lines are using the correct directory or if the script is still in the `tests` directory when running. Going to play a bit with my fork and this PR.

willingc:
@betatim @choldgraf I'm not quite sure what the interaction between repo2docker and the binder docs is. What are we trying to test here vs test in the binder docs?

betatim:
This PR doesn't test any interaction, my goal was to make sure we (pedantically) build the docs as part of the CI for each PR to prevent the docs build from breaking and never being fixed.

.travis.yml
betatim: Good point. I was assuming that `cd tests && pytest` would leave me back in the original directory.

willingc:
@betatim Ahh...got it. Currently, we use conda for doc build on RTD. I would prefer to continue builds on RTD using conda instead of pip. 

Either way, it may be best to have RTD provide notification on build https://docs.readthedocs.io/en/latest/guides/build-notifications.html

Even if the docs are tested with Travis, the actual docs state depends on RTD's build.

betatim:
Agree that the RTD build should be the authority and this is what we display in the README badge. What I'd like to add is that PRs can't be merged if they break the docs, can we trigger a build on RTD from a PR like a travis build?

The motivation for building as part of a PR is that most errors that break the docs are small things that are easy to overlook and easy to fix. The end result is that our docs aren't building for days until someone notices and that it then falls to one person to fixup all these small things, which is now a big task.

Does the RTD build via conda use https://github.com/jupyter/repo2docker/blob/master/readthedocs.yml? It seems that is based on the `docs/doc-requirement.txt`. I used that because it means we don't have to install miniconda in travis, no other reason.

willingc:
@betatim Thanks! This looks good. It appears the pop helped with the tests. For this repo, I can't remember why we use the doc-requirement.txt. While in most of the other repos, we use conda environment.yml.

Merging. Thank you :tada: :cake:

PR #373, opened 2018-08-07
remove conda-meta/history

minrk:
this file records transactions and seems to result in effectively turning the initial env install into a restrictive pinning, which we do not intend.

Adds xeus-cling@0.4.5 to external tests, since this repo cannot be built due to this effective pinning, but seems to build after removing the history file. It's very weird, and I'm not sure if it's intended behavior of conda or not.

cc @SylvainCorlay


betatim:
Thanks for this detective work.

SylvainCorlay:
Woot!

egpbos:
Thanks!

PR #375, opened 2018-08-15
Split out tests, tweak docs, and a style change

betatim:
This splits out the test for the launch script from the postBuild script
test. Makes some adjustments to the documentation of the launch script
functionality and applies some pythonista styling.

Follows up #363.

What do you think of these changes @jhamman ?


docs/source/config_files.rst
```diff
@@ -126,12 +126,18 @@ their demo for binder <https://github.com/jupyterlab/jupyterlab-demo/blob/master
 ``start``
 ^^^^^^^^^
 
-A script that can contain arbitrary commands to be run at runtime (as an
+A script that can contain simple commands to be run at runtime (as an
```
betatim: Try and suggest that you can only do simple things here in an attempt to keep people from doing heavy computations, even if we do not enforce it.

choldgraf:
This LGTM. Ready to merge @betatim ?

PR #378, opened 2018-08-19
Update to JupyterLab 0.34

blink1073:


choldgraf:
+1 unless anything comes up per @betatim's comments

blink1073:
Other than installing compatible extensions, should only be backwards compatible improvements for users.

choldgraf:
Then let's 🚢 it! Thanks @blink1073 :-)

PR #385, opened 2018-08-26
Change PyPI deploy account to mybinderteam

betatim:
The current travis setup uses my personal PyPI credentials. This PR switches to using the mybinderteam account.

After this we can retry auto-deploying to PyPI which currently fails because my PyPI password has changed.


yuvipanda:
Thanks, @betatim!

Issue #388, opened 2018-09-06
git-credential-env not working in debian


zymergen-luke:
The dockerfile for repo2docker is based on Debian (as well as the builder image used by binder). Debian replaces `/bin/sh` with `dash` - a shell with a version of the echo builtin that does _not_ support the `-e` flag. When using git-credential-env in this environment, it will produce
```
-e username=foo
password=bar
```
rather than the correct
```
username=foo
password=bar
```
that git expects. This prevents git-credential-env from working when inside the container.

The current contents of git-credential-env are:
```
#!/bin/sh
echo -e $GIT_CREDENTIAL_ENV
```

There are a couple possible workarounds:
1. Use `/bin/bash` instead of `/bin/sh` (adds a dependency on bash)
2. Modify the dockerfile to provide the proper flavor of `sh` (Not likely to work when switching image bases)
3. use `/bin/echo` rather than the `echo` builtin of the shell (though this introduces the opposite problem - e.g. /bin/echo on a mac does _not_ support -e while /bin/echo on debian _does_)
4. simply drop the `-e` (similarly, this will only work on half the systems).

Given that `/bin/echo` is non-uniform and even `/bin/sh` is non uniform, I would suggest we go with option 1 - always specify `/bin/bash` explicitly. This slightly raises the bar for dependencies, but I think bash is ubiquitous enough that it is a worthwhile tradeoff.

yuvipanda:
Fixed with #389

PR #389, opened 2018-09-06
Issue #388 - Use bash echo for git-credential-env

zymergen-luke:
sh is not consistent in its support of the echo builtin's -e flag. On
debian sh is symlinked to dash, which does not support the flag
resulting in bad credential output. Unfortunately /bin/echo is also
nonstandard - e.g. on macs -e is not supported while it is on debian. I
think the best approach is to rely on bash's echo builtin which should
be consistent across systems (at the slight cost of requiring bash as a
dependency)

This solves issue #388

### DIFF
--- a/docker/git-credential-env
+++ b/docker/git-credential-env
@@ -1,2 +1,2 @@
-#!/bin/sh
+#!/bin/bash
 echo -e $GIT_CREDENTIAL_ENV


yuvipanda:
Thanks for the PR! I like it!

Congratulations on getting your first PR merged, @zymergen-luke! Hope to see more :) <3

zymergen-luke:
Thank you, yuvipanda! I'll hopefully have more coming soon. I've got local changes that have enabled us to use Artifactory as a docker repo and access private git repositories.

Issue #392, opened 2018-09-07
Update default Julia version and/or allow customization


NHDaly:
Hi! Currently, repo2docker chooses `julia v0.6.0` by default, but there are several newer versions of julia that would be better choices.

The default is set here:
https://github.com/jupyter/repo2docker/blob/7552361c2b51bea73f3c2fba5e290d6b834b7f36/repo2docker/buildpacks/julia.py#L38

Better default values would be any of:
- v0.6.4 (the last 0.6 release)
- v0.7.0 (the "compatibility" release, which is mostly compatible with code written for 0.6 and code written for 1.0)
- v1.0 (the latest stable release. However lots of code and packages still aren't 1.0 ready yet.)

I personally think setting `v0.7.0` as the default makes the most sense.

-----------------------

As a separate thing, it would be nice if we allowed the user to choose a julia version in the REQUIRE file, as is commonly done in julia REQUIRE files:
https://github.com/JuliaLang/IJulia.jl/blob/master/REQUIRE#L1
https://github.com/JuliaPlots/StatPlots.jl/blob/master/REQUIRE#L1

This way, the user can choose between julia 0.6, 0.7, and 1.0.

If you think these should be two separate Issues, let me know and I'll split them up! :)

davidanthoff:
Maybe 1.0 would be the best default at this point? But I almost would prefer that one _must_ specify a julia version somehow. Whatever default we pick will be outdated soon, but updating a default here will break a lot of stuff everytime we do. So not support a default in the first place might be a good idea...

davidanthoff:
Alright, scrap that :)

I think the right approach is this:

if there is a `Project.toml` (or `JuliaProject.toml`, then generally ignore any `REQUIRE` file. If there is no julia version info inside `Project.toml`, then use the latest released version of julia. If there is julia version info inside `Project.toml`, then use the highest version that is compatible with the constraint that is in the `Project.toml`. 

If there is _no_ `Project.toml` or `JuliaProject.toml`, _and_ there is no julia version info inside `REQUIRE`, use the highest julia 0.6.x version.

betatim:
Sounds like a good plan.

For Python we do have a default version and bump it occasionally. The assumption is that those who don't specify a version don't care beyond "it is Python 3". I think people would be surprised if the default Python was 2.7 hence 3 is the default.

[...]

PR #393, opened 2018-09-07
Julia v1.0 support: Add option to specify julia version in REQUIRE

NHDaly:
This PR adds the option to specify a julia version in the REQUIRE file, and updates the JuliaBuildPack to support julia v0.7 and v1.0.

It also bumps the default julia version from v0.6.0 to v0.6.4 (the latest release for v0.6).

Adds tests for specifying a custom version, both for a `julia 0.6.3` and `julia 1`.

### DIFF
--- a/repo2docker/buildpacks/julia.py
+++ b/repo2docker/buildpacks/julia.py
@@ -22,9 +22,6 @@ class JuliaBuildPack(CondaBuildPack):
         '1': '1.0.0',
     }
 
-    def _short_version(version_str):
-        return '.'.join(version_str.split('.')[:2])
-
     @property
     def julia_version(self):
         require = self.binder_path('REQUIRE')
--- a/repo2docker/buildpacks/julia.py
+++ b/repo2docker/buildpacks/julia.py
@@ -12,6 +12,47 @@ class JuliaBuildPack(CondaBuildPack):
     See https://github.com/JuliaPy/PyCall.jl/issues/410
 
     """
+
+    minor_julias = {
+        '0.6': '0.6.4',
+        '0.7': '0.7.0',
+        '1.0': '1.0.0',
+    }
+    major_julias = {
+        '1': '1.0.0',
+    }
+
+    def _short_version(version_str):
+        return '.'.join(version_str.split('.')[:2])
+
+    @property
+    def julia_version(self):
+        require = self.binder_path('REQUIRE')
+        try:
+            with open(require) as f:
+                julia_version_line = f.readline().strip()  # First line is optionally a julia version
+        except FileNotFoundError:
+             julia_version_line = ''
+
+        if not julia_version_line.startswith('julia '):
+            # not a Julia version line.
+            # use the default Julia.
+            self._julia_version = self.minor_julias['0.6']
+            return self._julia_version
+
+        julia_version_info = julia_version_line.split(' ', 1)[1].split('.')
+        julia_version = ''
+        if len(julia_version_info) == 1:
+            julia_version = self.major_julias[julia_version_info[0]]
+        elif len(julia_version_info) == 2:
+            # get major.minor
+            julia_version = self.minor_julias['.'.join(julia_version_info)]
+        else:
+            # use supplied julia version
+            julia_version = '.'.join(julia_version_info)
+        self._julia_version = julia_version
+        return self._julia_version
+
     def get_build_env(self):
         """Get additional environment settings for Julia and Jupyter
 
@@ -35,7 +76,7 @@ class JuliaBuildPack(CondaBuildPack):
             ('JULIA_PATH', '${APP_BASE}/julia'),
             ('JULIA_HOME', '${JULIA_PATH}/bin'),
             ('JULIA_PKGDIR', '${JULIA_PATH}/pkg'),
-            ('JULIA_VERSION', '0.6.0'),
+            ('JULIA_VERSION', self.julia_version),
             ('JUPYTER', '${NB_PYTHON_PREFIX}/bin/jupyter')
         ]
 
@@ -81,10 +122,15 @@ class JuliaBuildPack(CondaBuildPack):
                 # FIXME: Find way to get it to install under /srv and not $HOME?
                 r"""
                 julia -e 'Pkg.init(); Pkg.add("IJulia"); using IJulia;' && \
-                mv ${HOME}/.local/share/jupyter/kernels/julia-0.6  ${NB_PYTHON_PREFIX}/share/jupyter/kernels/julia-0.6
+                mv ${HOME}/.local/share/jupyter/kernels/julia-${JULIA_VERSION%[.-]*}  ${NB_PYTHON_PREFIX}/share/jupyter/kernels/julia-${JULIA_VERSION%[.-]*}
                 """
             )
         ]
+        require = "/Users/daly/Documents/developer/website/nhdaly.github.io/notebooks/REQUIRE"
+        f = open(require)
+        julia_version_line = list(filter(lambda l: l != "", map(lambda l:l.strip(), f.readlines())))[0]
+        julia_version_line = f.readline().strip()
+
 
     def get_assemble_scripts(self):
         """
@@ -96,6 +142,8 @@ class JuliaBuildPack(CondaBuildPack):
 
         """
         require = self.binder_path('REQUIRE')
+        # TODO: I think this only works or v0.6... How do we do the equivalent
+        # of this for 1.0?
         return super().get_assemble_scripts() + [(
             "${NB_USER}",
             # Pre-compile all libraries if they've opted into it.
--- a/repo2docker/buildpacks/julia.py
+++ b/repo2docker/buildpacks/julia.py
@@ -121,7 +121,7 @@ class JuliaBuildPack(CondaBuildPack):
                 # HACK: Can't seem to tell IJulia to install in sys-prefix
                 # FIXME: Find way to get it to install under /srv and not $HOME?
                 r"""
-                julia -e 'Pkg.init(); Pkg.add("IJulia"); using IJulia;' && \
+                julia -e 'if (VERSION > v"0.7-") using Pkg; else Pkg.init(); end; Pkg.add("IJulia"); using IJulia;' && \
                 mv ${HOME}/.local/share/jupyter/kernels/julia-${JULIA_VERSION%[.-]*}  ${NB_PYTHON_PREFIX}/share/jupyter/kernels/julia-${JULIA_VERSION%[.-]*}
                 """
             )
@@ -142,19 +142,47 @@ class JuliaBuildPack(CondaBuildPack):
 
         """
         require = self.binder_path('REQUIRE')
-        # TODO: I think this only works or v0.6... How do we do the equivalent
-        # of this for 1.0?
         return super().get_assemble_scripts() + [(
             "${NB_USER}",
-            # Pre-compile all libraries if they've opted into it.
-            # `using {libraryname}` does the right thing
+            # Install and pre-compile all libraries if they've opted into it.
+            # In v0.6, Pkg.resolve() installs all the packages, but in v0.7+, we
+            # have to manually Pkg.add() each of them (since the REQUIRES file
+            # format is deprecated).
+            # The precompliation is done via `using {libraryname}`.
             r"""
-            cat "%(require)s" >> ${JULIA_PKGDIR}/v0.6/REQUIRE && \
             julia -e ' \
-               Pkg.resolve(); \
-               for pkg in keys(Pkg.Reqs.parse("%(require)s")) \
-                pkg != "julia" && eval(:(using $(Symbol(pkg)))) \
-               end \
+                require_file = "%(require)s" ;\
+                if VERSION < v"0.7-" ;\
+                  pkg_dir = "$(ENV["JULIA_PKGDIR"])/v$(VERSION.major).$(VERSION.minor)" ;\
+                  open("$pkg_dir/REQUIRE", "a") do io ;\
+                    write(io, read(require_file)) ;\
+                  end ;\
+                  Pkg.resolve(); ;\
+                  Reqs = Pkg.Reqs ;\
+                else ;\
+                  using Pkg ;\
+                  Reqs = Pkg.Pkg2.Reqs ;\
+                  emptyversionlower = v"0.0.0-" ;\
+                  for reqline in Reqs.read(require_file) ;\
+                    if reqline isa Reqs.Requirement ;\
+                        pkg = String(reqline.package) ;\
+                        if pkg == "julia" continue end ;\
+                        version = try; reqline.versions.intervals[1].lower; catch; emptyversionlower; end ;\
+                        if version != emptyversionlower ;\
+                          Pkg.add(PackageSpec(name=pkg, version=version)) ;\
+                        else ;\
+                          Pkg.add(pkg) ;\
+                        end ;\
+                    end ;\
+                  end ;\
+                end ;\
+                # Precompile the packages ;\
+                for reqline in Reqs.read(require_file) ;\
+                  if reqline isa Reqs.Requirement ;\
+                      pkg = reqline.package ;\
+                      pkg != "julia" && eval(:(using $(Symbol(pkg)))) ;\
+                  end ;\
+                end ;\
             '
             """ % { "require" : require }
         )]
--- a/repo2docker/buildpacks/julia/__init__.py
+++ b/repo2docker/buildpacks/julia/__init__.py
@@ -123,10 +123,6 @@ class JuliaBuildPack(CondaBuildPack):
                 """
             )
         ]
-        require = "/Users/daly/Documents/developer/website/nhdaly.github.io/notebooks/REQUIRE"
-        f = open(require)
-        julia_version_line = list(filter(lambda l: l != "", map(lambda l:l.strip(), f.readlines())))[0]
-        julia_version_line = f.readline().strip()
 
     def get_assemble_scripts(self):
         """
@@ -146,9 +142,10 @@ class JuliaBuildPack(CondaBuildPack):
             # format is deprecated).
             # The precompliation is done via `using {libraryname}`.
             r"""
-            julia /tmp/install-repo-dependencies.jl "%(require)s" && \
-            rm /tmp/install-repo-dependencies.jl
+            julia /tmp/install-repo-dependencies.jl "%(require)s"
             """ % { "require" : require }
+            # TODO: For some reason, `rm`ing the file fails with permission denied.
+            # && rm /tmp/install-repo-dependencies.jl
         )]
 
     def get_build_script_files(self):
@@ -169,4 +166,6 @@ class JuliaBuildPack(CondaBuildPack):
         Instead we just check if the path to `REQUIRE` exists
 
         """
+        # TODO(nhdaly): Add support for value in variable:
+            pass Project.toml here as well.
         return os.path.exists(self.binder_path('REQUIRE'))
--- a/repo2docker/buildpacks/julia/__init__.py
+++ b/repo2docker/buildpacks/julia/__init__.py
@@ -71,7 +71,8 @@ class JuliaBuildPack(CondaBuildPack):
         """
         return super().get_build_env() + [
             ('JULIA_PATH', '${APP_BASE}/julia'),
-            ('JULIA_HOME', '${JULIA_PATH}/bin'),
+            ('JULIA_HOME', '${JULIA_PATH}/bin'),  # julia <= 0.6
+            ('JULIA_BINDIR', '${JULIA_HOME}'),  # julia >= 0.7
             ('JULIA_PKGDIR', '${JULIA_PATH}/pkg'),
             ('JULIA_VERSION', self.julia_version),
             ('JUPYTER', '${NB_PYTHON_PREFIX}/bin/jupyter')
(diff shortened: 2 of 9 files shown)


NHDaly:
Fixes #392.

NHDaly:
@KristofferC could you maybe take a look at this to verify that i've handled the REQUIRE file acceptably in julia 0.7/1.0? I think it works, but it seems like there might've been a more elegant way for me to do this?

Thanks so much!! :)

KristofferC:
Would be a lot simpler if you have a Project + Manifest, which in case the whole thing would just be a `Pkg.instantiate()` to download all the packages at the versions specified in the Manifest.

Otherwise doing something like a `Pkg.add("path/to/repo")` on the path of the package itself should install all the required dependencies even if it has a REQUIRE file.

NHDaly:
> Would be a lot simpler if you have a Project + Manifest
I was under the impression that we're really not supposed to be using these yet, until the Great Migration moves everything from REQUIRE to Project+Manifest. Am I getting that wrong? Is it that you're supposed to use Project+Manifest in your local projects (or notebook repo in this case), but still use REQUIRE files in packages that you submit to METADATA?

Thanks, and sorry for my confusion!

In that case, how do you recommend proceeding for this code? Should we simply say we don't support REQUIRE files in your notebook repo if you're using julia v0.7+?

> Otherwise doing something like a `Pkg.add("path/to/repo")` on the path of the package itself should install all the required dependencies even if it has a REQUIRE file.

Sorry, I don't think I understand what you mean. If it helps explain better, the repo being built in this case usually _isn't_ a Package. Rather, it's a Project that contains `.ipynb` files, and the REQUIRE file is used to specify dependencies needed to run those notebooks. Here's an example repo:
https://github.com/binder-examples/julia-python

So I don't think `Pkg.add` is what we want, since there's no package there at all. Something like `Pkg.resolve("path/to/repo")` would be excellent though! :)

betatim:
Nice work and thanks for helping us get better Julia support.

For Python and R we use `runtime.txt` to let users choose versions. So from a uniformity point of view it would be good to use that for Julia as well. However from one of your comments it sounds like the Julia community uses the `REQUIRE` file for this? What do you think is the best option?

KristofferC:
> I was under the impression that we're really not supposed to be using these yet, until the Great Migration moves everything from REQUIRE to Project+Manifest. Am I getting that wrong? Is it that you're supposed to use Project+Manifest in your local projects (or notebook repo in this case), but still use REQUIRE files in packages that you submit to METADATA?

You can use Project + Manifest now, it is just that the way things are registered in METADATA right now is using REQUIRE files. For notebooks, I am not sure what the point of a REQUIRE file would be on 1.0. There's no `Pkg.resolve(".../REQUIRE")` in Pkg3.

> If it helps explain better, the repo being built in this case usually isn't a Package. 

Yeah, I haven't really checked what this repo is doing heh.

NHDaly:
> Slightly related: JuliaLang/IJulia.jl#673.

Oh cool, thanks, yes i think that's a related conversation. Thanks for the pointer!

> Yeah, I haven't really checked what this repo is doing heh.

hehe. :) Np. This repo is used by MyBinder (and probably other parts of the jupyter universe?) to instantiate a docker container for a given directory (repo) that contains .ipynb files. They currently allow you to specify dependencies that should be loaded **while building the docker image -- so that you don't need to re-install them every time you run a notebook**. This is done via a top-level `REQUIRE` for julia, an `environment.yml` for python, etc.

After reading the conversation you linked, I think this process is used to solve some of the issues discussed in that thread!

NHDaly:
> You can use Project + Manifest now, it is just that the way things are registered in METADATA right now is using REQUIRE files. For notebooks, I am not sure what the point of a REQUIRE file would be on 1.0. There's no Pkg.resolve(".../REQUIRE") in Pkg3.

Gotcha. Okay! It seems like a reasonable answer here is just to say: Sorry, if you're using v0.7+ you can't use REQUIRE files and instead have to use Project+Manifest! that seems reasonable to me. Thanks! :)

NHDaly:
But also, the other thing I was hoping to change in this PR is allowing you to specify the julia version in the REQUIRE file. Currently, the way this tool works is it _only_ allows you to pull down `julia 0.6.0`, and so I think allowing the user to specify which version of julia to install would be preferable.

So I think maybe the logic for this PR should instead be:

- If there's a REQUIRE, assume the julia version to be 0.6.4. If it contains a julia version string, read the version out of it. If the listed version is >0.7, error out or something?
  - Then, copy the REQUIRE to ~/.julia/version/and do `Pkg.resolve()`
- If there's a Project.toml, assume the julia version to be 1.0.0. If it contains a `[compat]` section with a julia version string, read the version out of it. If < 0.7, error out or something.
  - Then, copy  Project.toml and possible Manifest.toml to ~/.julia/environments/version/ and do `Pkg.instantiate()`.


Does that sound reasonable? Thanks for your help!

NHDaly:
Although (sorry for the post-spam), I do think it would be reasonable to _also_ support REQUIRE files for 0.7, at least for now, since they're still a part of the julia package ecosystem (for METADATA). So it seems reasonable to me to support both for now?

KristofferC:
The "correct" thing would be to record the julia version in the Manifest but we don't currently do that. So next best thing:

Note that `julia = "1.0"` defines a range of intervals in Pkg3 (`[1.0 - 2.0)`). So I think the way to do it is if there is no `[compat]` for julia use the latest, otherwise use the latest version such that it is within the range (which can be tested as): 

```
julia> v"1.0" in Pkg.Types.semver_spec("1.0")
true

julia> v"1.3" in Pkg.Types.semver_spec("1.0")
true

julia> v"2.0" in Pkg.Types.semver_spec("1.0")
false
```

So if the julia versions `[1.0, 1.2, 1.3, 2.0]` existed and `julia = 1.0` was in `[compat]` then use `1.3`.

With this, the julia version used will change as new versions get released (but that is not any different from what `Pkg.resolve` did on 0.6 when new package version was released though).

NHDaly:
Oh, cool, yeah that makes sense. That should be an easy change. So how does this behavior sound:

- If there's a Project.toml:
  - If it contains a `[compat]` section with a julia version string, read the version out of it. If < 0.7, error out or print or warning or something.
    - Install the latest version of julia that satisfies the provided semver specification. (e.g if julia versions `[1.0, 1.2, 1.3, 2.0]` existed, and `julia = 1.0` in `[compat]`, install `1.3`.)
  - Otherwise, just install the latest version of julia.
  - Then, copy the Project.toml and possible Manifest.toml to `~/.julia/environments/version/`, open julia, and run `Pkg.instantiate()`.
- If there's a REQUIRE file (and no Project.toml):
  -  If it contains a julia version string, install the latest version that matches the prefix. (e.g. `julia 0.6` installs `v0.6.4`, `julia 0.7` installs `v0.7.0` and `julia 0.6.3` installs `v0.6.3`.)
  -  Otherwise, install the latest 0.6 (`v0.6.4`).
  - Then, if version < 0.7, copy the REQUIRE to ~/.julia/version/ and do `Pkg.resolve()`
  - Else (version >0.7), call `Pkg.add()` on each line of the REQUIRE, using the code I've currently implemented in this PR, [here](https://github.com/jupyter/repo2docker/pull/393/files#diff-c45c0c9be667207c01a12f4e83b15fc4R163).


I think this sounds like a good approach that is the most permissive of user behavior. But it's seems reasonable to say it's too permissive, and that we should disallow using REQUIRE if you're going to use julia 0.7. I'm open to your suggestions there! :)

Thanks again for your help!

KristofferC:
> Then, copy the Project.toml and possible Manifest.toml to ~/.julia/environments/version/, open julia, and run Pkg.instantiate().

There is no need to copy. You can set the `JULIA_PROJECT` env variable to the project path, start julia with `julia --project=/path/to/project` or after starting julia run`Pkg.activate("path/to/project")` and then that will be the active project. If the Project is in the `pwd()` then `julia --project` is enough (implies pwd).

NHDaly:
> There is no need to copy. You can set the JULIA_PROJECT env variable to the project path, start julia with julia --project=/path/to/project or after starting julia runPkg.activate("path/to/project") and then that will be the active project. If the Project is in the pwd() then julia --project is enough (implies pwd).

Hmm, yeah I know about that, but I am wondering if that will work correctly with the jupyter notebook kernels... Maybe it will? Basically, the goal is that whenever the user opens a .ipynb and starts the Julia kernel, that it gets an environment with all the right packages installed.

If I start julia with the `JULIA_PROJECT` env variable set when I first install `IJulia`, will it set things up so that the kernel it installs will always point to the right environment? Am I understanding that correctly?

KristofferC:
I don't know what configuration options you have over the kernel file from here but if you can do zero configuration then you would indeed have to put the Project + Manifest in the default place (`environments/...`).

NHDaly:
Okay cool, that makes sense. thanks.

Alright, well, let's leave aside this issue, since it's relatively minor. Would you mind giving me your thoughts on the overall pseudocode behavior above? I appreciate your help!! :) :)

KristofferC:
I would just not bother with 0.6 but if you really want to then that pseudo code seems ok. REQUIRE files however have no implicit upper bounds like the Pkg3 compat has, so `julia 0.6` means `julia [0.6, 99999999...) ` so it is a bit tricky to know what Julia version to install. But as a heuristic, the latest 0.6 version is probably the right thing to do.

yuvipanda:
@betatim we should probably use whatever is canonical in the language for specifying version, falling back to runtime.txt only when there is no canonical way. environment.yml for example has a canonical way.

@NHDaly @KristofferC <3 for working on this! Have you run repo2docker locally to verify that IJulia works as intended?

repo2docker/buildpacks/julia.py
```diff
@@ -98,15 +144,45 @@ def get_assemble_scripts(self):
...
+            # format is deprecated).
+            # The precompliation is done via `using {libraryname}`.
             r"""
-            cat "%(require)s" >> ${JULIA_PKGDIR}/v0.6/REQUIRE && \
             julia -e ' \
-               Pkg.resolve(); \
-               for pkg in keys(Pkg.Reqs.parse("%(require)s")) \
-                pkg != "julia" && eval(:(using $(Symbol(pkg)))) \
-               end \
+                require_file = "%(require)s" ;\
```
yuvipanda: I think this is now complex enough to be moved to a file of its own and then called from here. This lets us stop mixing Julia and Bash. See https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/conda/__init__.py#L83 for how to add a script to the repo. Then this script (which should be in Julia) can be called from here.

NHDaly: Awesome, great idea. I'm pushing up a commit now that just makes this change: splitting it out into its own file! :) I moved them into a new folder, `buildpacks/julia`.

Good suggestion! :)

-----

It seems to be working correctly, except that the line I added to `rm /tmp/install-repo-dependencies.jl` is failing:
```bash
rm: cannot remove '/tmp/install-repo-dependencies.jl': Operation not permitted
The command '/bin/sh -c julia /tmp/install-repo-dependencies.jl "REQUIRE" && rm /tmp/install-repo-dependencies.jl' returned a non-zero code: 1F
```

So now the tests are failing.

Any ideas? It seems like this file has the same permissions as `conda/environment.frozen.yml`, which is deleted successfully by the conda buildpack [here](https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/conda/__init__.py#L73).

I see that the python buildpack [copies its files to /home/main](https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/legacy/__init__.py#L35).. Should I do that instead?

NHDaly:
> @betatim we should probably use whatever is canonical in the language for specifying version, falling back to runtime.txt only when there is no canonical way. environment.yml for example has a canonical way.

Oh yeah, I forgot about `runtime.txt`... So that should be added as a fallback mechanism? Right now, without this PR, is that already a thing? Could I have just specified the Julia version via runtime.txt?

> @NHDaly @KristofferC <3 for working on this! Have you run repo2docker locally to verify that IJulia works as intended?

haha <3 thanks!! :) I'll be happy to get it updated!

Yes, I have run it locally to verify that the tests (including those I added) all pass, and I launched a jupyter notebook in one of the images it created, and it worked correctly with different julia versions!

(That said, the [contributing guide](https://github.com/jupyter/repo2docker/blob/master/CONTRIBUTING.md) doesn't explicitly say how to actually launch a notebook after creating the docker image, it only covers tests. I just did a `docker container ls` to find one of the ones created by the tests, and then launched that one.)

------

I'm going to put in some work in the next couple days to update the logic to what Kristoffer and I discussed [above](https://github.com/jupyter/repo2docker/pull/393#issuecomment-419558127). Do I also need to add in logic to handle falling back to `runtime.txt`, or does that behavior come automatically for free? Thanks!

repo2docker/buildpacks/julia.py
NHDaly: Here's where I do the deleting:
https://github.com/jupyter/repo2docker/pull/393/files#diff-fcf7bb456679b31cc83386be059cb3abR150

NHDaly:
@KristofferC: One more question, should we require people use `JuliaProject.toml` here, since the whole point of this file is to tell `repo2docker` "Hey this repo needs Julia installed! Plz install this version and these deps."? Or should we support either `Project.toml` or `JuliaProject.toml`?

What other languages use `Project.toml`, and is there an easy way to read the file and tell if it's intended for julia or some other language?

yuvipanda:
@NHDaly awesome. I would recommend against adding any runtime.txt logic right now. Let us know when this is ready for review again!

KristofferC:
Pkg and code loading in Julia Base support both names. FWIW, I personally argued that only using JuliaProject.toml was the way to go but oh well. For now, I think it is easiest to use `JuliaProject.toml` if it exists and otherwise use `Project.toml`

NHDaly:
Haha yeah, i mean, I can see why you argued for that. :/ oh well indeed.

Okay great, thanks for the guidance, that makes sense to me as well. :)

repo2docker/buildpacks/julia/__init__.py
```diff
@@ -80,11 +118,15 @@ def get_build_scripts(self):
...
                 # FIXME: Find way to get it to install under /srv and not $HOME?
                 r"""
-                julia -e 'Pkg.init(); Pkg.add("IJulia"); using IJulia;' && \
-                mv ${HOME}/.local/share/jupyter/kernels/julia-0.6  ${NB_PYTHON_PREFIX}/share/jupyter/kernels/julia-0.6
+                julia -e 'if (VERSION > v"0.7-") using Pkg; else Pkg.init(); end; Pkg.add("IJulia"); using IJulia;' && \
+                mv ${HOME}/.local/share/jupyter/kernels/julia-${JULIA_VERSION%[.-]*}  ${NB_PYTHON_PREFIX}/share/jupyter/kernels/julia-${JULIA_VERSION%[.-]*}
                 """
             )
         ]
+        require = "/Users/daly/Documents/developer/website/nhdaly.github.io/notebooks/REQUIRE"
```
minrk: leftover debug lines?

tests/julia/julia_version/README.rst
```diff
@@ -0,0 +1,25 @@
...
+the names of packages you'd like to be installed. For example:
+
+```
+PyPlot
+IJulia
+DataFrames
+```
+
+Each one will be installed but **not** pre-compiled. If you'd like to
+pre-compile your Julia packages, consider using a ``postBuild`` file.
```
betatim: is this comment still true? Asking because of https://github.com/jupyter/repo2docker/pull/393/files#diff-fcf7bb456679b31cc83386be059cb3abR143

betatim:
All tests pass, some small nits, merge after those?

What is the new default Julia version going to be when the user does not specify anything? Should we bump it to 1.0 or leave it at the (now quite old) 0.6 that has been the default?

NHDaly:
> All tests pass, some small nits, merge after those?

Ah, actually, this still needs some pretty major changes: adding support for the Project.toml files per this comment:
https://github.com/jupyter/repo2docker/pull/393#issuecomment-419558127

I started working on it again yesterday, but it's actually been a bit tough. The hardest part was trying to handle the "semver" logic in python that the `.toml` files use for specifying a julia version. (Needs to be done in python, since it's deciding _which_ version of julia to install!)

**But actually**, I think maybe you're right that we could just merge this PR as-is, and I could fork the changes I've been working on into a second PR. That might actually make more sense, just to allow getting something working sooner rather than later! **What do you think? Should I split that into a separate PR?**

> What is the new default Julia version going to be when the user does not specify anything? Should we bump it to 1.0 or leave it at the (now quite old) 0.6 that has been the default?

I had kept it at 0.6 so as not to break anyone currently using the system. I think it is _perfectly_ reasonable (and probably preferable) to bump it, but I assume we'd want to publish that somewhere first? Like, HEY default julia versions are going to bump in a week, please use a REQUIRE file to keep yourself at 0.6 if that's what you really want?

As for which version, no I think we should go to 0.7, since a lot of packages still aren't yet 1.0 compatible. I think we could probably bump to 1.0 in a month or two.

betatim:
> As for which version, no I think we should go to 0.7, since a lot of packages still aren't yet 1.0 compatible. I think we could probably bump to 1.0 in a month or two.

Ok, maybe we stick with the current default and use our "one chance" to annoy people by bumping straight to 1.0.

> What do you think? Should I split that into a separate PR?

So now we can choose the version via the first line in `REQUIRE` and the second PR would add support for `{Julia}Project.toml`? If the answer is yes I'd vote to merge now and make a second PR. (Smaller PRs are the better PRs 😄)

NHDaly:
> Smaller PRs are the better PRs 😄

Haha yes agreed, 100%. :)

> So now we can choose the version via the first line in REQUIRE and the second PR would add support for {Julia}Project.toml? If the answer is yes I'd vote to merge now and make a second PR.

Yes the answer is yes! :)

Okay, then, before you merge, allow me the chance to address these remaining nits. I'll have something pushed up in a few hours! :)


> Ok, maybe we stick with the current default and use our "one chance" to annoy people by bumping straight to 1.0.

Haha okay yeah I think that could make sense.. The one interesting and weird thing to note about this version change is that 0.7 is closer to 1.0 than to 0.6. 0.7 is basically "1.0 with deprecations", or a backwards-compatible 1.0.

So given that, I think I agree that the 1.0 bump definitely should be announced, but actually that we could probably silently bump the default to 0.7 and people would on the whole be glad, not sad. There are fewer and fewer existing use cases where people would prefer 0.6 over 0.7.

So maybe we submit this, send a silent PR to bump to 0.7, and then eventually do an announced PR to bump to 1.0?

tests/julia/julia_version/README.rst
NHDaly: Fixed.

NHDaly:
One more thing to do: I realized while playing with Project.toml that some of the environment variables needed by julia have changed between 0.6 and 0.7/1.0.

Do you think it's safe to just always define all of them as a simple solution? (for example, JULIA_HOME is now renamed to JULIA_BINDIR, so can we just define both?)

betatim:
> Do you think it's safe to just always define all of them as a simple solution? (for example, JULIA_HOME is now renamed to JULIA_BINDIR, so can we just define both?)

From a repo2docker point of view: sure. Though it sounds like it would be more of a question for Julia people? Maybe add specific comments to the ones that we can remove at some point when we remove 0.6 support? As a form of forward looking house keeping.

NHDaly:
> Maybe add specific comments to the ones that we can remove at some point when we remove 0.6 support? As a form of forward looking house keeping.

Done!

Alright, _I think this is all set to be merged!_ I'm sorry for the delay on just that last piece.

This PR adds unit tests for specifying a specific old julia version (v0.6.3) and for specifying using the new julia version, v1.0.0, as well as testing that the environment variables are set correctly for both julia <=0.6 and >=0.7.

I have also verified manually that the julia notebooks run correctly from the docker image created by each of the unit tests! :)

So merge away! :)

betatim:
Merged!

Do you want to deploy these changes to mybinder.org directly? If yes we have [instructions for deploying](https://github.com/jupyterhub/mybinder.org-deploy/#site-reliability-guide-) that should explain how to deploy the new revision of repo2docker.

choldgraf:
wahooooo! Way to go!

NHDaly:
Hooray! Thanks Tim! :)

> Do you want to deploy these changes to mybinder.org directly?
Yes, I think so! Is there anything I should consider as to why I might not want to do that? Thanks!

betatim:
> Yes, I think so! Is there anything I should consider as to why I might not want to do that? Thanks!

I don't think so. We can always revert it when we notice a bug (or fix the bug and deploy the fix). Beware the feeling of responsibility and excitement from deploying a change you made to production where lots of people will use it straight away can be addictive 😄.

I'd suggest you drop by https://gitter.im/jupyterhub/mybinder.org-deploy and give people a short heads up about the deploy when you are ready. We use the chat as a kind of log book and source of "what has been going on". We try and group deploys to Tuesdays and Thursdays, mostly so that people can plan a bit to be available in case follow up is needed (luckily tomorrow is Tuesday!)

NHDaly:
Thanks for the explanation!

> beware: the feeling of responsibility and excitement from deploying a change you made to production where lots of people will use it straight away can be addictive 😄

Haha 😋 I will beware, indeed!

Ah, sorry i missed today! Actually, if you don't mind, would you be able to deploy this? Or just allow it to be deployed the next time you have something to deploy? I'm a bit busy this week. :) No problem if you can't; i'll look at it this weekend!

betatim:
This has been deployed: https://github.com/jupyterhub/mybinder.org-deploy/pull/758

NHDaly:
Thanks a lot Tim!! :)

PR #394, opened 2018-09-09
Add more details to our setup() command

betatim:
This updates some of the metadata fields in our `setup()`.

This is the last thing I'd change before releasing 0.6.0 proper. After this is out we can start doing regular releases with all the bells and whistles (changelogs, planned features, etc).

@yuvipanda you ok with the email change?

@willingc any other PyPI things we should quickly tidy up?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
     },
     # this should be a whitespace separated string of keywords, not a list
     keywords="reproducible science environments docker",
-    description = "Repo2docker: Turn git repositories into Jupyter enabled Docker Images",
+    description = "Repo2docker: Turn code repositories into Jupyter enabled Docker Images",
     long_description = readme,
     long_description_content_type = 'text/markdown',
     license='BSD',
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,8 @@ setup(
         'Source': 'https://github.com/jupyter/repo2docker/',
         'Tracker': 'https://github.com/jupyter/repo2docker/issues',
     },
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -13,9 +17,28 @@
         'ruamel.yaml>=0.15',
     ],
     python_requires='>=3.4',
-    author='Repo2docker contributors',
-    author_email='yuvipanda@gmail.com',
+    author='Project Jupyter Contributors',
+    author_email='jupyter@googlegroups.com',
```
betatim: is this the best email address to use? Copied from JupyterHub.

setup.py
```diff
@@ -13,9 +17,30 @@
...
+    url='https://repo2docker.readthedocs.io/en/latest/',
+    project_urls = {
+        'Documentation': 'https://repo2docker.readthedocs.io',
+        'Funding': 'https://jupyter.org/about',
+        'Source': 'https://github.com/jupyter/repo2docker/',
+        'Tracker': 'https://github.com/jupyter/repo2docker/issues',
+    },
+    # this should be a whitespace separated string of keywords, not a list
+    keywords="reproducible science environments docker",
+    description = "Repo2docker: Turn git repositories into Jupyter enabled Docker Images",
```
yuvipanda: Git isn't required, which makes the repo in repo2docker a misnomer. However, I can't think of any alternative right now, so I'm ok with this :)

willingc: Maybe: "Turn a set of files, like a git code repository, into Jupyter enabled Docker Images.

PR #401, opened 2018-09-14
[MRG] Make travis fail when pytest fails

betatim:
Currently travis will succeed even though a test failed.

The problem is that the `pytest` command fails but then we run `popd` which succeeds and as a result traivs things all is well.

An example run: https://travis-ci.org/jupyter/repo2docker/jobs/428640113

Thanks for helping find this: @giovannipizzi


betatim:
Let's give people till Monday when it is a working day to comment on this. It should get deployed on Tuesday if it has been merged before then.

Issue #396, opened 2018-09-12
Support DESCRIPTION files for R projects


betatim:
https://github.com/benmarwick/rrtools/blob/master/DESCRIPTION

These have been mentioned a few times to me as a way that R users specify dependencies. We should try and find out how popular they are and consider supporting htem in addition to `install.R`.

betatim:
After your investigations: should we still attempt to add support for `DESCRIPTION` files? They could be used in the same way as `setup.py` is currently supported (if there is nothing else but a `setup.py` we assume this is a repository for a package and the user wants us to run `pip install -e.`. This makes it super nice for trying out "random" packages.

gedankenstuecke:
Yep, I think the `DESCRIPTION` support would be valuable and should be pretty easy as you can run the equivalent of `pip install -e.` for `R` to do the install of the package! For `packrat` I've given up for now as it seems to be not mature enough (or I'm too unskilled in R!)

PR #406, opened 2018-09-17
add R DESCRIPTION support

gedankenstuecke:
Addresses #396 and runs the R commands to install a package when a `DESCRIPTION` file is in the repo, hinting that the repo describes a complete R package.

I couldn't test it locally yet as FRA airport WiFi doesn't like me running `repo2docker` / downloading miniconda. 😂 ✈️ 

Will test more extensively once I'm at a more proper WiFi spot.


gedankenstuecke:
Turns out it's even easier as `devtools` is already setup in the R images in any case! Now it just runs `devtools::install_git(".")` if a `DESCRIPTION` is present. 

As for the `install.R` (at least that's how I interpret the docs) it is right now still necessary to have the `runtime.txt` with an `r-YYYY-…` present at the same time. 

I also added a test-repo with a fake-R-package at https://github.com/gedankenstuecke/binder-r-description that successfully builds when running the `repo2docker` version with these changes and added it to the `reproductions.repos.yaml` and some docs for the `DESCRIPTION`.

choldgraf:
really nice! Am I correct in thinking that DESCRIPTION is more geared towards developers?

and is this ready for review?

gedankenstuecke:
Yep on both accounts! 

`DESCRIPTION` is the `R` equivalent of `setup.py`, so it's more or less exclusively used for automated package installs.

repo2docker/buildpacks/r.py
```diff
@@ -257,4 +257,15 @@ def get_assemble_scripts(self):
                 )
             ]
 
+        descriptionR_path = self.binder_path('DESCRIPTION')
+        if os.path.exists(descriptionR_path):
+            assemble_scripts += [
+                (
+                    "${NB_USER}",
+                    r"""
+                    R --quiet -e "devtools::install_git('.')"
```
choldgraf: what if the `DESCRIPTION` is in a `binder/` folder? Should this be `R --quiet -e "devtools::install_git('{}'.format(descriptionR_path)`?

Also it seems there's an extra `"` in there at the end

gedankenstuecke: That's a good point, though it seems a bit unlike (would be like a `setup.py` hiding outside the main folder) but I've added that bit 👍

gedankenstuecke:
@choldgraf I think this would be good to go for a second review round!

repo2docker/buildpacks/r.py
betatim: We only act on the `setup.py` if it is in the top level and there is no `binder/`:

https://github.com/jupyter/repo2docker/blob/46f056a4e281bb84b0c6b6e321ad0f1879e89ddd/repo2docker/buildpacks/python/__init__.py#L69-L74

The discussion when we added it was that we want to enable the opportunistic "just try it out" of packages even if they don't know of binder (yet). However when the package maintainers are aware of binder and want to do something different the signal for that is to create `binder/` subdir to specify exactly how to setup the environment for this package and its examples.

I think we should take the same approach for `DESCRIPTION` files.

tests/external/reproductions.repos.yaml
```diff
@@ -30,3 +30,7 @@
   url: https://github.com/QuantStack/xeus-cling
   ref: 0.4.5
   verify: jupyter kernelspec list
+- name: Test compiling R package
```
betatim: Can we put a test under https://github.com/jupyter/repo2docker/tree/master/tests/r as well?

External repos for tests are (like other external things) one of the reasons we get flaky tests so where ever possible I think we should have a test in this repository. So when the travis network connection is flaky we don't get a red test.

docs/source/config_files.rst
```diff
@@ -109,6 +109,17 @@ While one can specify dependencies in ``setup.py``,
...
 
+``DESCRIPTION``
+~~~~~~~~~~~~~~~
+
+To install your repository like an R package, you may include a
+``DESCRIPTION`` file. repo2docker installs the package and dependencies
+ from the ``DESCRIPTION`` by running ``devtools:install_git(".")``.
+
+ You also need to have a ``runtime.txt`` file that is formatted as
+ ``r-<YYYY>-<MM>-<DD>``, where YYYY-MM-DD is a snapshot at MRAN that will be
```
betatim: "at MRAN" or "of MRAN"? My vote would be the second.

repo2docker/buildpacks/r.py
gedankenstuecke: That makes sense to me. In that case I'd also extend the `detect` method to run the buildpack if only the `DESCRIPTION` is available but no `runtime.txt`. In that case we'd need to decide which snapshot of MRAN to use though. Do you have any suggestions for that? I'd be tempted to just go for the latest one if nothing is specified?

tests/external/reproductions.repos.yaml
gedankenstuecke: Sure, I can put the content of the external repo into the `R` test folder for that case? And do you mean to just replace the external test in that case?

docs/source/config_files.rst
gedankenstuecke: mine too, actually. copy & pasted this I think :joy:

tests/external/reproductions.repos.yaml
betatim: Jupp.

I'd copy what the `simple/` subdirectory is doing and come up with a nice descriptive name :)

gedankenstuecke:
Ok, I followed @betatim's ideas around the `DESCRIPTION` implementation that should be analogous to `setup.py`: 

1. Now it will only run the package build if the `DESCRIPTION` is in the top-level directory
2. It doesn't require a `runtime.txt` any longer. Instead it builds with second-to-latest MRAN as snapshot as long as there's no `runtime.txt` specified.

tests/r/description-file/verify
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env Rscript
+library('binderdescription')
```
betatim: I think the travis failure is because this file isn't marked as executable.

gedankenstuecke: doh! Let's see whether this'll fix it!

betatim:
Seems like now the error from travis is legit. Does the test run locally: `pytest tests/r/start`?

gedankenstuecke:
Locally it fails with:

```
Error in library("binderdescription") :
  there is no package called ‘binderdescription’
Execution halted
``` 

But that is after it prints the install stuff during the test (and the same error occurs locally on my machine for the `simple` test too!) 

Also I'm not 100% sure how the tests are set up, I just tried c&p from the `simple` test, but where is the actual test config done? :-)

betatim:
We have a custom kind of pytest test. This is what it does:

https://github.com/jupyter/repo2docker/blob/79cb2ba2405ce8397325b75c2322c49a3bb0fc2d/tests/conftest.py#L104-L109

The `args` are passed to the `repo2docker` script. My short summary is that we run r2d on the test directory and tell it to execute the `verify` script inside the container once built. If that exists with status code 0 the test passes, else it fails.

Looking at it now to see if I can work out why it fails.

betatim:
>  Could it be that the R executable used in `verify` needs some help with search paths?

In an RStudio session it can also not find the package. So I think something doesn't quite work with the install process :(

gedankenstuecke:
Thanks for the explanation! I think I might run into issues with cached image builds on my end and that's why it doesn't fail if I do it locally? I just removed all the relevant images I had to force a new build (is there an easier way to achieve that?). I'll update once I found out what's going on there.

gedankenstuecke:
Ok, I think this was some weird caching error on my end which was just re-using stuff despite it actually being broken. After re-doing the image properly I could find the bug and I think it should work now, let's see whether Travis agrees/ 👍

gedankenstuecke:
Yep, that seems to have fixed it! 🎉

repo2docker/buildpacks/r.py
```diff
@@ -69,7 +69,14 @@ def detect(self):
         """
         # If no date is found, then self.checkpoint_date will be False
         # Otherwise, it'll be a date object, which will evaluate to True
-        return bool(self.checkpoint_date)
+        if bool(self.checkpoint_date):
```
betatim: You don't need to call `bool(x)` explicitly, the `if` statement will take care of that for you.

repo2docker/buildpacks/r.py
```diff
@@ -69,7 +69,14 @@ def detect(self):
...
+            return True
+        description_R = 'DESCRIPTION'
+        if not os.path.exists('binder') and os.path.exists(description_R):
+            if not self.checkpoint_date:
+                self._checkpoint_date = datetime.date.today() - datetime.timedelta(days=2)
```
betatim: could you add a comment as to why "two days before today" is "the right date" to use? Already I can't remember why so I expect in 3months we won't be able to figure this out again :)

docs/source/config_files.rst
```diff
@@ -95,6 +95,27 @@ version of Ubuntu.
 We use ``apt.txt``, for example, to install LaTeX in our
 `example apt.txt for LaTeX <https://github.com/binder-examples/latex/blob/master/apt.txt>`_.
 
+``setup.py``
```
betatim: How did this get here? Just to make sure this isn't going to cause a merge conflict or overwrite edits  recently made in master

betatim:
LGTM modulo the small comments.

repo2docker/buildpacks/r.py
gedankenstuecke: dah, `True` 😂

docs/source/config_files.rst
gedankenstuecke: whops, my bad - i screwed up the merge conflict and thought this needs to stay.

gedankenstuecke:
ok, should all be fixed 😄

betatim:
Thanks a lot! And congrats on your first PR to repo2docker!

betatim:
In the deed! (as we Germans say)

Do you also want to attempt to deploy your changes to mybinder.org today? There is nothing quite like putting things into production :)

gedankenstuecke:
Yes, what do I have to do for that? 👍

betatim:
We use https://github.com/jupyterhub/mybinder.org-deploy/ for the deployment. The README there contains a link to our "How to do a deployment" guide.

choldgraf:
woooooooo go @gedankenstuecke !!!! 🎉 🎉 🎉 and thanks @betatim for helping guide this one through!

Issue #273, opened 2018-03-20
Can we exploit build cache for faster launch of R build pack?


brooksambrose:
**tldr** for R build pack, any hope of using `postBuild` to add a layer without invalidating cache for `install.R`?

This is a repo2docker question for a binderhub: is there a way to exploit Docker build cache to speed up building images using the r2d R build pack? The use case is a heavy base image with some light user customization on top, e.g. building R with an `install.R` script for tidyverse seems to take about 15 minutes on a beefyish rig. Using binderhub in a course, I want to give students a good data science image and give them the option to layer on top for original student projects, without having to wait for tidyverse to rebuild as well. As far as I can tell there's no way to add a new layer, but want to ask here in case I missed it.

So classic 80/20 problem, possibly same conversation here #115.  Goal is to build a "never variable" heavy layer once, and then give users the option to build a "sometimes variable" lighter layer of  package installs on that, but before the "always variable" data layer.

I'm trying to use the `postBuild` option to add that light layer of R package installs on at the very end, which seems like it would make sense given that the r2d Dockerfile adds a layer just for postBuild as the last line. To the contrary, the behavior seems to be to invalidate the cache ~after using runtime.txt but before using install.R~ before copying anything from the repo, even if most of the repo is the same.

For example, using debug on a local r2d:

```
jupyter-repo2docker --debug --no-build https://github.com/w201rdada/final
```
Gives something like this:

```Dockerfile
#...truncated jupyter build layers!
# Copy and chown stuff. This doubles the size of the repo, because
# you can't actually copy as USER, only as root! Thanks, Docker!
USER root
COPY src/ ${HOME}
RUN chown -R ${NB_USER}:${NB_USER} ${HOME}

# Run assemble scripts! These will actually build the specification
# in the repository into the image.
RUN echo "options(repos = c(CRAN='https://mran.microsoft.com/snapshot/2018-03-19'), download.file.method = 'libcurl')" > /etc/R/Rprofile.site

USER ${NB_USER}
RUN Rscript install.R


# Container image Labels!
# Put these at the end, since we don't want to rebuild everything
# when these change! Did I mention I hate Dockerfile cache semantics?


# We always want containers to run as non-root
USER ${NB_USER}

RUN ./postBuild
```

The behavior seems to be to use cache up until copying the binder files from the repo (line 35), then all cache after is invalidated even if only one file has changed in the repo.

```
Step 31/47 : USER ${NB_USER}
 ---> Using cache
[...]

Issue #164, opened 2017-12-07
use COPY --chown to stage repo (docker 17.09)


minrk:
Right now, we copy and chown the repo in two layers, which doubles the size the repo takes up in the env. This used to be required. Docker 17.09 adds support for `COPY --chown` to do it in one layer. So we can do this when it becomes okay to require docker 17.09 as the base version.

mukundans91:
Can we add the functionality now with a check done to see if the underlying docker version is >= 17.09, if so use COPY --chown else use the older form of the template?

minrk:
Good idea! We can indeed do that.

willingc:
Adding this to the next release as this seems pretty straightforward.

choldgraf:
I'm moving this to milestone v0.7, since we've already released v0.6 and this issue doesn't seem to be in it!

betatim:
What does it mean if something is added to a milestone?

Should we make it to mean that we will wait to release the next version until this is done or do we take it as an indication that it would be a good fit in the next release (it isn't too early to work on this) but release will happen if no one works on it?

choldgraf:
in other projects that I work on, the milestone issues are blocking on a release, though often people decide to punt them to the next release once it gets closer to deadline time

betatim:
I think we should use it like that too (if it is in a milestone then the release waits till it is done). Maybe we can ease into it by only adding things to a milestone a few days before a release as a kind of todo list/making sure we don't forget something. With the current state of contributors I worry that if we add something to a milestone that isn't security or usability critical (and hence has a high likelihood of getting worked on) we will end up punting a lot or slow down our releases even more :-/

Xarthisius:
Just as a note: an implementation of `COPY --chown` that expands variables in options was merged only two weeks ago (see https://github.com/moby/moby/issues/35018#issuecomment-486774627). Now it's going to be possible to actually do:

```
COPY --chown=${NB_USER}:${NB_USER} src/ ${REPO_DIR}
```

yuvipanda:
Me and @jhamman found out that version of docker used by GKE does not support this yet, so the default install of BinderHub can't use it (since it uses the same docker as the underlying cluster)

minrk:
> in other projects that I work on, the milestone issues are blocking on a release, though often people decide to punt them to the next release once it gets closer to deadline time

[...]

PR #407, opened 2018-09-17
[WIP] Feature: Adding nix support for repo2docker

costrouc:
This pull request is a work in progress. I would like feedback from the nix community that it is implemented correctly and has the necessary features.

Demo repo2docker repository: https://gitlab.com/costrouc/nix-binder-example

# Why add nix as a buildpack?

[Nix](https://nixos.org/nix/) would a great addition to reproducible data science. It is a unique package manager. Some notable features:
  - 100% reproducible environments (pin to exact commit in repository)
  - both a source and binary package repository
  - allows customized compilation and version of every package
  - can run identical environment outside of docker (all linux distros + dawin)
  - as of now [45,000+ packages](https://repology.org/repository/nix_unstable)
  - fully declarative environments

Questions needing feedback:

__repo2docker__ questions

1. Should I use a custom Dockerfile not taking advantage of base? 

The reason for this is that nix does not need `apt` or other package managers. So the smaller the base image the better. Nix handles all dependencies.

2. Where should `NixBuildPack` be in ordered list of buildpacks?

As of now `NixBuildPack` only depends on a `default.nix` file being present.

__nix specific questions__

1. Any shorter way to initialize nix in a docker container and launch with a nix shell?

The definition as of now is quite short but I would like to make it shorter. In general things that I was not happy including `nixshell` command and nix-channel initialization. This adds about 25 Mb to the image.

2. Should `nix-shell` be launched with the `--pure` option?

Launching with a pure option would make the environment more reproducible by removing all non included packages. Has the downside that some expected tools will not be available.

3.  Anyway to ensure that `jupyter` is in installed dependencies?

Right now I do not check that jupyter is present.

4. How to prevent `nix-shell` from evaluating `shellHook` twice?

Right now `shellHook` is evaluated in lines:

 - `nix-shell default.nix --command "command -v jupyter"`
 - `ENTRYPOINT ["/usr/bin/nixshell"]` when the docker container is run

5. Should I include the unstable channel in the `NIX_PATH`? 

I have included it for ease of use. However, like in the nix binder example I pin the exact nixpkgs version https://gitlab.com/costrouc/nix-binder-example/blob/master/default.nix#L4-7


betatim:
Nice work!

re: New base image or not? I would keep the shared base image. We try and have our build packs be composable (you can use the R and conda build pack at the same time) which will be hard to do if they don't share a base image. It also means a lot of the standard things (creating the right user, copying the repo, etc0 are taken care of and we only need to maintain one setup.

costrouc:
Definitely. Yeah I see from a maintainer side that sharing the same dockerfile would be a better approach. From the tests I did the difference is about 100 MB in the final uncompressed docker image. I will make the changes.

betatim:
In my experience the images that you end up with are measured in GBs, so I think we don't have to feel too bad about inflating them by 100MB.

For maintenance this is a big win. Also I'd imagine that people who use nix might want to also pull in things from other package managers (or vice versa). Say 90% of my packages comes from nix and the rare ones I directly install with `pip`. Or something like that. This means that maintaining composability is a big thing for users as well.

repo2docker/buildpacks/nix/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+RUN adduser --disabled-password \
+    --gecos "Default user" \
+    --uid 1000 \
+    ${NB_USER}
+WORKDIR /home/${NB_USER}
+
+EXPOSE 8888
+
+# Copy and chown stuff. This doubles the size of the repo, because
+# you can't actually copy as USER, only as root! Thanks, Docker!
```
nlewo: FYI this may be fixed by using https://docs.docker.com/develop/develop-images/multistage-build/#before-multi-stage-builds

betatim: 👍  Would be interesting to explore. Do you want to create a PR to get going on this? Otherwise we should create a new issue to track this.

betatim: #273 would also be relevant for this.

minrk: This is fixed in docker itself. We just need to add a check for the docker version, or bump our base supported docker version to 17.09 and use it: https://github.com/jupyter/repo2docker/issues/164

costrouc:
@betatim I have modified the commit to make it composable with other buildpacks. Looking at the size of the builds the difference with 900 MB with alpine as a base and 1.7 GB for buildpack as the base. That is with installing `numpy`, `scipy`, and `jupyterlab`. The 1.7 GB may not be such a big deal especially since the early layers will match other images. 

I have checked that it worked using the following command locally. Please let me know what additionally needs to be done for testing. I have added a nix test that hopefully is done right.

```
repo2docker https://gitlab.com/costrouc/nix-binder-example
```

repo2docker/buildpacks/nix/__init__.py
```diff
@@ -0,0 +1,70 @@
+"""BuildPack for nixpkgs environments"""
+import os
+
+from ..base import BaseImage
+
+
+class NixBuildPack(BaseImage):
```
betatim: if we inherit from `PythonBuildPack` instead people will be able to use nix together with conda and pip.

This depends a bit on the use-case you see for nix. Should nix be something that people use without anything else (right now) or as an addition to install tools that they can't install with what they already have (a bit like `apt.txt` which lets you install some additional things as well as using your `environment.yml` or `REQUIRE`).

repo2docker/app.py
```diff
@@ -64,6 +64,7 @@ def _default_log_level(self):
 
     buildpacks = List(
         [
+            NixBuildPack,
```
betatim: nix should probably be where the Julia and R build packs are in this list, not at the end here. The Dockerfile build pack is the ultimate escape hatch, so we need to give priority to it over all other build packs.

tests/nix/simple/default.nix
```diff
@@ -0,0 +1,21 @@
+let
+  # Pinning nixpkgs to specific release
+  # To get sha256 use "nix-prefetch-git <url> --rev <commit>"
+  commitRev="5574b6a152b1b3ae5f93ba37c4ffd1981f62bf5a";
```
betatim: should we expose this via `runtime.txt`  like we do for MRAN which gets set via `r-<date>` in `runtime.txt`. I assume everyone who uses nix will want to set this SHA to something so we could move this code from the `default.nix` file? The advantage would be less typing and that this behaviour would be enforced because you have to put the SHA in the `runtime.txt`.

What do you think how that fits with nix user's expectations?

betatim:
Any thoughts from @minrk or @yuvipanda on adding nix support in general and specifics when it comes to composing it with other ways of specifying dependencies? Unfortunately nix isn't a tool I use or know anyone who uses a lot so trying to gently feel my way around for what is best.

tests/nix/simple/default.nix
minrk: I think we should avoid using runtime.txt for anything where a provider has a mechanism to specify something like this. It's exclusively an escape hatch where there's no better way to specify (i.e. only in R, requirements.txt, not needed and shouldn't be supported for Julia, conda, etc.).

minrk:
Re: alpine, I would avoid it here. Lots of things don't really behave as expected (many Python packages don't work right on alpine, and Python wheels can't be installed). It's a great base for well-defined service images, but I wouldn't recommend it for anything interactive.

minrk:
I think it definitely should use the same base image, though I'm not sure about inheriting the default Python env or not. It depends a bit on what is to be installed (i.e. is it the kernel environment). One of the reasons composing works well with the other setups is that the installation mechanisms (conda/pip/julia/r) play nice together (at least if they are done in the right order). Since nix owns its whole install, it's less clear what the right thing to do is:

1. give nix control over the whole env, or
2. inherit the notebook server env, and give nix control of the *kernel env only*

costrouc:
Nix installs all of the dependencies for a package (including things like libc etc.) so technically the base image does not matter to nix. I think that sharing a base image for now is the best path because as of now `nix` does not have every python package and conda does handle some packages better. In the future I certainly see that a smaller base image could be used. 

I have spent some time trying to figure out why the travis test is failing? Is it because of appendix? The issue is that in order to expose all of the nix package I need to create a "virtualenv" by setting the `ENTRYPOINT=nix-shell`. It is similar to running `pipenv shell` or `source activate myenv`.

costrouc:
Ok to summarize the changes that have been made with feedback. The `NixBuildPath` now passes (for some reason julia keeps failing? Looks like this happens in other PRs). I have intentionally kept this PR as the bare minimum to have `nix` fully working with `repo2docker`. This bare minimum should be more than enough for any `nix` based repository. 

 - based on base `BuildPack`. I found that when basing off of `PythonBuildPack` there was no advantage since `nix` can install `conda` if it needs it. Using `PythonBuildPack` increased the docker image size by around `700 MB`.
 - a non pure nix-shell is used for now. This means that the programs installed by the base image are still available.
 - Resulting docker images are around 1.5 GB. With the first 700 MB cachable.
 - Documentation has been added for how to use `nix`

I believe that with these changes it is ready for merging pending additional questions.

costrouc:
Fixed merge conflict and small documentation typo. Ready for merge pending any questions that you may have. I've described in the previous comments exactly what was included in this commit.

choldgraf:
woooo! thanks for all the work @costrouc

betatim:
You could deploy it tomorrow if you want to. Deploys are made via a PR to https://github.com/jupyterhub/mybinder.org-deploy/ with a guide here: https://mybinder-sre.readthedocs.io/en/latest/deployment/how.html#updating-repo2docker

We usually batch deploys on Tuesdays and Thursdays so that people can plan to be around for a few hours after the deploy they made to revert/fix side effects.

While I have your attention maybe you have some thoughts on #448.

costrouc:
Thank you! I looked in the PRs and saw how easy it is to update. 

I commented here with my thoughts earlier https://github.com/jupyter/repo2docker/pull/448#issuecomment-435901093. I don't expect any changes to the base image to have any effect on `nix` so I think the change is perfectly fine. It will only affect the resulting docker image size from my perspective.

PR #409, opened 2018-09-17
updating index, usage, and install docs

choldgraf:
This PR refactors a bit of the index, usage, and install pages. I tried to make the information a bit more discoverable from the main page and more cleanly separated out.

### DIFF
--- a/docs/source/config_files.rst
+++ b/docs/source/config_files.rst
@@ -1,9 +1,8 @@
-.. NOTE: the header characters are different in this file because it is 'included' in another site
-.. see https://raw.githubusercontent.com/jupyterhub/binder/master/doc/using.rst
 .. _config-files:
 
+===================
 Configuration Files
--------------------
+===================
 
 ``repo2docker`` looks for configuration files in the repository being built
 to determine how to build it. In general, ``repo2docker`` uses the same
@@ -13,16 +12,8 @@ rather than creating new custom configuration files.
 A number of ``repo2docker`` configuration files can be combined to compose more
 complex setups.
 
-``repo2docker`` will look for configuration files in either:
-
-* A folder named ``binder/`` in the root of the repository.
-* The root directory of the repository.
-
-If the folder ``binder/`` is located at the top level of the repository,
-  **only configuration files in the** ``binder/`` **folder will be considered**.
-
 The `binder examples <https://github.com/binder-examples>`_ organization on
-GitHub contains a list of sample repositories  for common configurations
+GitHub contains a list of sample repositories for common configurations
 that ``repo2docker`` can build with various configuration files such as
 Python and R installation in a repository.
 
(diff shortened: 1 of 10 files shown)


docs/source/config_files.rst
```diff
@@ -32,64 +23,75 @@ Below is a list of supported configuration files (roughly in the order of build
...
+``setup.py`` - Install Python packages
+======================================
+
+To install your repository like a Python package, you may include a
+``setup.py`` file. repo2docker installs ``setup.py`` files by running
+``pip install -e .``.
+
+While one can specify dependencies in ``setup.py``,
+repo2docker **requires configuration files such as** ``environment.yml`` or
+``requirements.txt`` to install dependencies during the build process.
```
betatim: This sentence was already here so maybe we should leave it along. However do you understand what it is trying to tell us? Dependencies from your `setup.py` will be installed by repo2docker and you don't need to have a `requirements.txt`/`environment.yml` :-/

docs/source/config_files.rst
```diff
@@ -163,3 +156,17 @@ used for installing libraries.
...
+======================================
+
+In the majority of cases, providing your own Dockerfile is not necessary as the base
+images provide core functionality, compact image sizes, and efficient builds. We recommend
+trying the other configuration files before deciding to use your own Dockerfile.
+
+With Dockerfiles, a regular Docker build will be performed.
+**If a Dockerfile is present, all other configuration files will be ignored.**
+
+See the `Binder Documentation <https://mybinder.readthedocs.io/en/latest/dockerfile.html>`_ for
```
betatim: "See the Advanced Binder Documentation ..."

we are already in the Binder docs here after all :)

docs/source/index.rst
```diff
@@ -1,26 +1,37 @@
...
+``jupyter-repo2docker`` is a tool to **build, run, and push Docker
+images from source code repositories** that run via a Jupyter server.
+
+``repo2docker`` fetches a repository
 (e.g., from GitHub or other locations) and builds a container image
 based on the configuration files found in the repository. It can be
 used to explore a repository locally by building and executing the
-constructed image of the repository.
+constructed image of the repository, or as a means of building images that
+are pushed to a Docker registry.
```
betatim: I think we discussed this previously and @yuvipanda had a strong opinion that we do not state that repo2docker is a tool for building images that you then distribute. Let's double check both current thinking and feelings.

docs/source/usage.rst
```diff
@@ -35,54 +68,68 @@ specify the ``branch-name`` or ``commit-hash``::
...
 
-   http(s)://<server:port>/<lab-location>/tree
+   http(s)://<server:port>/tree
 
 To learn more about URLs in JupyterLab and Jupyter Notebook, visit
 `starting JupyterLab <http://jupyterlab.readthedocs.io/en/latest/getting_started/starting.html>`_.
 
-``--debug`` and ``--no-build``
-------------------------------
+RStudio
```
betatim: 🎉

docs/source/index.rst
yuvipanda: I think using repo2docker as a tool for you to distribute is totally fine. I have a strong opinion that repo2docker is not a tool for producing **Dockerfiles** that you distribute.

docs/source/config_files.rst
choldgraf: I.....have no idea what this means lol. Should I just remove this paragraph?

choldgraf:
ok I believe the latest push addresses @betatim's comments! However I've now added another thing: a "howto" section, similar to TLJH. I've found this to be a really nice documentation pattern for TLJH since it focuses on questions people often have, and because it's fairly extensible. I've also written these in a way that we could embed these rST files in the Binder docs too, if we want to duplicate some content.

What do folks think? Maybe @yuvipanda has thoughts on this one, since he's the original architect of the TLJH docs?

choldgraf:
@jzf2101 I'd recommend submitting a pull request for changes you want to see - I'm happy to review!

PR #410, opened 2018-09-18
[WIP] add assemble_subset

minrk:
Marked WIP for discussion. I though I opened this ages ago, but apparently forgot.

This improves the use of build caching by separating 'assemble' into two steps:

1. assemble the environment, using only the files required to do so (e.g. environment.yml)
2. stage the rest of the repo afterward

This enables re-using the build cache from one build to the next, greatly expediting rebuilds in cases where the environment spec doesn't change.

Changes:
- get_assemble_files() returns list of files needed for assembly
- if buildpack.assemble_with_subset is set, only load assemble_files
  prior to running assembly scripts. Load the rest of the repo afterward

conda opts in to this, but currently I think this works for everything *except* requirements.txt. And requirements.txt could as well, as long as we peek in the contents to look for local `-e .` references.

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -93,10 +93,15 @@ COPY {{ src }} {{ dst }}
 {{sd}}
 {% endfor %}
 
-# Copy and chown stuff. This doubles the size of the repo, because
-# you can't actually copy as USER, only as root! Thanks, Docker!
+# FIXME: use COPY --chown with docker 17.09 to avoid copy+chown in two steps
+{% if assemble_from_subset -%}
+{% for f in assemble_files %}
+COPY src/{{ f }} ${HOME}/{{ f }}
+{% endfor %}
+{% else -%}
 USER root
 COPY src/ ${HOME}
+{% endif -%}
 RUN chown -R ${NB_USER}:${NB_USER} ${HOME}
 
 {% if env -%}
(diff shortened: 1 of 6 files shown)


betatim:
Can a `environment.yml` with a pip section contain local references?

What do you think of a build pack being able to have assemble files that are run before copying over the repo and after? This way a build pack could support this but lets it opt out sometimes? Too complex/convoluted?

repo2docker/buildpacks/base.py
```diff
@@ -93,10 +93,15 @@
 {{sd}}
 {% endfor %}
 
-# Copy and chown stuff. This doubles the size of the repo, because
-# you can't actually copy as USER, only as root! Thanks, Docker!
+# FIXME: use COPY --chown with docker 17.09 to avoid copy+chown in two steps
```
betatim: Move this to above L105

repo2docker/buildpacks/julia.py
```diff
@@ -12,6 +12,10 @@ class JuliaBuildPack(CondaBuildPack):
     See https://github.com/JuliaPy/PyCall.jl/issues/410
 
     """
+
+    # REQUIRE can't refer to local files, can it?
+    assemble_from_subset = True
```
betatim: Worth checking with #393 and the new way of specifying dependencies in Julia.

minrk: Good catch! I think it will affect Julia with the project stuff.

minrk:
> What do you think of a build pack being able to have assemble files that are run before copying over the repo and after? This way a build pack could support this but lets it opt out sometimes? Too complex/convoluted?

You can see an example of this in PythonBuildPack, where `assemble_from_subset` is a property, set after a check for local references in requirements.txt.

minrk:
Closing in favor of #743

PR #413, opened 2018-09-26
Subdirectory support

dsludwig:
This adds support for creating a Docker image from a subdirectory of a Git repo, rather than just the root.

I wanted to store my Docker image definition and some other configuration in the same repo, without the configuration getting into my image. This seemed like the most straight-forward way to do that.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -663,7 +663,13 @@ class Repo2Docker(Application):
             if self.repo_type == 'remote':
                 self.fetch(self.repo, self.ref, checkout_path)
 
-            os.chdir(os.path.join(checkout_path, self.subdir).rstrip('/'))
+            if self.subdir:
+                checkout_path = os.path.join(checkout_path, self.subdir).rstrip('/')
+                if not os.path.exists(checkout_path):
+                    self.log.error('Subdirectory %s does not exist', self.subdir, extra=dict(phase='failure'))
+                    sys.exit(1)
+
+            os.chdir(checkout_path)
 
             for BP in self.buildpacks:
                 bp = BP()
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -62,6 +62,16 @@ class Repo2Docker(Application):
         """
     )
 
+    subdir = Unicode(
+        '',
+        config=True,
+        help="""
+        Subdirectory of the git repository to examine.
+
+        Defaults to ''.
+        """
+    )
+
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks for opening this, @dsludwig! Looks good to me. Can you add a test too?

dsludwig:
Thanks @yuvipanda, I added a test for this. I couldn't get the pytest capture to pick up the log, but when I run with `-s`, it does show up.

tests/test_subdir.py
```diff
@@ -0,0 +1,30 @@
+"""
+Test if the explict hostname is supplied correctly to the container
```
betatim: This should be about sub directories :)

betatim:
@dsludwig is there something missing from this PR or can we go ahead and merge it?

tests/test_subdir.py
```diff
@@ -0,0 +1,30 @@
+"""
+Test if the subdirectory is correctly navigated to
+"""
+import logging
+from os.path import abspath, dirname
+
+import pytest
+from repo2docker.app import Repo2Docker
+
+repo_path = dirname(dirname(abspath(__file__)))
```
betatim: Why there `dirname(dirname())`? Maybe add a comment and/or use `..` to signal "one level up"?

tests/test_subdir.py
```diff
@@ -0,0 +1,30 @@
...
+    app = Repo2Docker()
+    argv = ['--subdir', 'tests/conda/invalid', repo_path]
+    app.initialize(argv)
+    app.debug = True
+    app.run = False
+    with pytest.raises(SystemExit):
+        app.start()  # Just build the image and do not run it.
+
+    # Can't get this to record the logs?
+    # assert caplog.text == "Subdirectory tests/conda/invalid does not exist"
```
betatim: You could check for the exit code here instead for the moment?

Not quite sure why this doesn't get captured without digging in :-/

tests/test_subdir.py
dsludwig: I added a comment here - repo_path is the root directory of the repo2docker git repository.

tests/test_subdir.py
dsludwig: Thanks, I added a check for the exit code.

dsludwig:
@betatim I addressed your comments and resolved the merge conflicts, I think it's ready to go!

PR #419, opened 2018-10-01
Refactor Repo2Docker.fetch method to not need be passed its own members as arguments

evertrol:
This is a tiny clean-up: Since `fetch()` is only called inside the
class, and two of the three arguments in that call are attributes of
`self`, there is no need to pass them as extra arguments.

To not break the API, the parameters in the method definition are now
set to a default of `None`; if these are used with their default,
they'll be set to the respective `self` attributes. Note that this
breaks behaviour where the `self` attribute is not `None`,
but `None` is explicitly passed for the respective parameter.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -164,11 +164,16 @@ class Repo2Docker(Application):
         """
     )
 
-    def fetch(self, url, ref, checkout_path):
-        """Check out a repo using url and ref to the checkout_path location"""
+    def fetch(self, checkout_path, url=None, ref=None):
+        """Check out a repository using its url and ref to the checkout_path
+        location"""
+        if url is None:
+            url = self.repo
+        if ref is None:
+            ref = self.ref
         try:
-            for line in execute_cmd(['git', 'clone', '--recursive', url, checkout_path],
-                                    capture=self.json_logs):
+            for line in execute_cmd(['git', 'clone', '--recursive', url,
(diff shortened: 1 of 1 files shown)


evertrol:
Notes and further food for thought:
- I've put this here as a suggestion. The change may not be deemed necessary.
- As the last sentence states, this may break external code. A quick search through related repositories (binder, binderhub) does not reveal any calls to `fetch()`. Most of the `repo2docker` use seems to be through its command line arguments.
- Is there any use for calling `fetch()` with different arguments than the `self` attributes? If not, `fetch()` should probably become a private method `_fetch()` and have its last two parameters completely removed.

consideRatio:
This makes sense, I wish it would have been reviewed and accepted in this projects infancy. At this point, I don't dare go for it as it is just a smaller code cleanup but could be possibly breaking for project external code as you also mention.

> * I've put this here as a suggestion. The change may not be deemed necessary.
> * As the last sentence states, this may break external code. A quick search through related repositories (binder, binderhub) does not reveal any calls to `fetch()`. Most of the `repo2docker` use seems to be through its command line arguments.
> * Is there any use for calling `fetch()` with different arguments than the `self` attributes? If not, `fetch()` should probably become a private method `_fetch()` and have its last two parameters completely removed.

I appreciate the thought you put into this PR, highlighting key parts of whats relevant for a person reviewing this!! I'm glad to see your still active in GitHub and open source! :heart: :tada:

PR #420, opened 2018-10-01
Speed up cloning by using a depth of 1 if there is no refspec

evertrol:
To speed up cloning a repository, a shallow clone containing just the
last commit can be made, but *only* if no refspec is provided.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -167,8 +167,11 @@ class Repo2Docker(Application):
     def fetch(self, url, ref, checkout_path):
         """Check out a repo using url and ref to the checkout_path location"""
         try:
-            for line in execute_cmd(['git', 'clone', '--recursive', url, checkout_path],
-                                    capture=self.json_logs):
+            cmd = ['git', 'clone', '--recursive']
+            if not ref:
+                cmd.extend(['--depth', '1'])
+            cmd.extend([url, checkout_path])
+            for line in execute_cmd(cmd, capture=self.json_logs):
                 self.log.info(line, extra=dict(phase='fetching'))
         except subprocess.CalledProcessError:
             self.log.error('Failed to clone repository!',
--- a/tests/test_clone_depth.py
+++ b/tests/test_clone_depth.py
@@ -1,0 +1,105 @@
+"""
+Test that a clone depth of 1 is used and good enough when no refspec is used
+
+Note: the tests don't actually run the container. Building the
+container requires a specific repository and commit to be checked out,
+and that is the only thing that is tested.
+
+"""
+import subprocess
+import requests
+import time
+from repo2docker.app import Repo2Docker
+
+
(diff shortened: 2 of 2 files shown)


evertrol:
I'm aware of #120 and #131 , but as far as I can tell, the issue there was that the shallow clone was performed for cases where the refspec was either empty or given. Here, the shallow clone is only made when there is no refspec.

Potentially, the same shallow cloning can be used for a branch of tag (but not for a specific commit), but this can't work with the current implementation: the refspec is parsed once the repository is cloned, and could be a branch, tag or commit. It would require a separate option to specify a branch or tag to allow shallow clones for those cases, but this would complicate the user interface.

betatim:
It would be great to add this feature back in! I think as is this PR won't speed up BinderHubs as they always specify a `--ref`, making repo2docker faster for local users would be great though.

One question for my git education: if there are two branches in a remote git repository: `master` and `some-feature` and the last (in time) commit was to `some-feature` we will still get the last commit to `master` with our `--depth 1` checkout?

Could you add some tests of the shallow/not shallow cloning based around a repo like https://github.com/betatim/repo2docker-ci-clone-depth so that this behaviour keeps working also in the future? Once we are actively using https://github.com/betatim/repo2docker-ci-clone-depth we should move it to https://github.com/binderhub-ci-repos (or similar org).

evertrol:
Git will use `HEAD` when cloning, which (for GitHub at least) points to `master` (unless this was explicitly changed by a repository owner):
- https://stackoverflow.com/questions/18726037/what-determines-default-branch-after-git-clone
- https://help.github.com/articles/setting-the-default-branch/

If I try for a test repository, I indeed get `master` when cloning, not matter the last commit and branch, while a local (file) clone results in the current branch.

---

It's probably for the better that an explicit refspec is used (I assume the default is `master`), to avoid surprises; I hadn't realised BinderHub did that.
Of course, if there is a default and that is a branch, there would be the option to use `--branch` in the git command, as suggested above. But this requires some proper thinking how to implement that nicely, if wanted at all.

I'll try and add some tests in one of the next days.

betatim:
[This is the code](https://github.com/jupyterhub/binderhub/blob/230bbde0cac7ffe5da4f445734dab7a90f5bfde0/binderhub/repoproviders.py#L419) that we use to look up explicit SHAs. [This is the code](https://github.com/jupyterhub/binderhub/blob/230bbde0cac7ffe5da4f445734dab7a90f5bfde0/binderhub/build.py#L61-L68) used to run `repo2docker` to build the image.

Looking forward to some tests and thanks for explaining something new about git to me!

betatim:
LGTM. I moved the "clone depth test repo" to https://github.com/binderhub-ci-repos/repo2docker-ci-clone-depth.

I think we should merge this after updating that link in the test. What do you think?

evertrol:
Thanks!
Let's see in time if the real world agrees with this change and doesn't fall over.

Issue #357, opened 2018-07-19
feature request: "editable" mode


neuromusic:
Here's the idea...

``` bash
$ cd ~/Projects/my_repo/
$ repo2docker -e ./
```
This launches my environment, I run my notebooks, then... oh no!! there's a bug!

I fix the bug, save the notebook (running in the repo2docker docker container), then shut down the docker container.

``` bash
$ git status
```
The changes I saved in the docker container are applied locally! Now, I can commit them, certain that they work in the environment built by repo2docker.

betatim:
You can currently do this with `repo2docker -v .:/home/$USER .` which is a bit exotic.

I thought until I recently tried using it that "editable" was the default when using a local repository. Tracking down when and why that changed (or if it was just something I was imagining) would be a welcome contribution. For sure I'd support making it the default (that I thought we always had :) ).

evertrol:
Is my assumption correct that this can changed to be done by default with the following diff:

```
index 73ac12c..d1f5126 100644
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -389,6 +389,7 @@ class Repo2Docker(Application):
             self.repo = args.repo
             self.ref = None
             self.cleanup_checkout = False
+            self.volumes['.'] = '/home/$USER'
         else:
             self.repo_type = 'remote'
             self.repo = args.repo
```

Thus, the `-v .:/home/$USER` is automatically implemented if `os.path.exists(args.repo)` is true. 

---

A grep on `volumes` through the history of `app.py` doesn't give any indication that this behaviour was the default previously. 
In that case, it may be better to add it as an option (as suggested above, `-e` or `--edit`), to prevent confusion (i.e., an `if args.edit:` statement just above the new line in the above diff).

betatim:
That looks about right. Though it shouldn't be `self.volumes['.']`, the key has to be the actual path to the repository that is being built. I think in my example it was just a coincidence that this was `.`.

Unfortunately `-e` is already taken for environment variables so we need to invent a new short hand. Maybe `-E` and `--editable` which kinda mirrors `pip install`?

choldgraf:
`--editable` sounds great toe me!

evertrol:
I think this can be closed, now that #421 is merged.

betatim:
Good catch! I've become super reliant on "Fixes #NNN" being mentioned in the PR's first comment which means GitHub automatically close that issue.

evertrol:
[...]

Issue #376, opened 2018-08-16
Repo2docker v0.6.0


betatim:
Should we make a new release of v0.6.0?

It has been a long time since the last release, some links in error messages that v0.5 prints don't work anymore, so let's cut a release?

I'd tag the current state and be done with it. We run ~master on mybinder.org so we know it works well enough to go straight for a release. It is a shame to not have a ChangeLog, maybe we can start this for the next release by making entries as we go along.

betatim:
When we tried to release v0.6.0rc1 we ran into #385.

I tagged v0.6.0rc2 just now to see how far we get. Slowly catching up with things after a week away and assume no one is objecting to getting 0.6.0 out "as is".

betatim:
🎉  see: https://pypi.org/project/jupyter-repo2docker/#history

PR #421, opened 2018-10-01
Add an edit-mode option

evertrol:
This adds an option to run from a local repository in edit mode, where
changes in a running Docker container (for example, through a
notebook) are reflected in the local repository.

Implements the feature suggested in #357.

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -70,3 +70,31 @@ or similar traditional docker command.
 
 Check out the [binder-examples](http://github.com/binder-examples/) GitHub
 organization for example repositories you can copy & modify for your own use!
+
+## Can I use repo2docker to edit a local repository within a Docker environment?
+
+Yes: use the `--editable` or `-E` flag (don't confuse it with the `-e`
+flag for environment variables), and run repo2docker on a local
+repository: `repo2docker -E my-repository/.`.
+
+This builds a Docker container from the files in that repository
+(using, for example, a `requirements.txt` file or `Dockerfile`), then
+runs that container, while connecting the home directory inside the
+container to the local repository outside the container. For example,
+in case there is a notebook file (`.ipynb`), this will open in a local
+webbrowser, and one can edit it and save it. The resulting notebook is
+updated in both the Docker container and the local repository. Once
+the container is exited, the changed file will still be in the local
+repository.
+
+This allows for easy testing of the container while debugging some
+items, as well as using a fully customizable container to edit, for
+example, notebooks.
+
+**note**
+
+Editable mode is a convenience option that will mount the repository
+to container working directory (usually `/home/$USER`). If you need to
+mount to a different location in the container, use the `--volumes`
+option instead. Similarly, for a fully customized user Dockerfile,
+this option is not guaranteed to work.
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -343,6 +343,13 @@ class Repo2Docker(Application):
         )
 
         argparser.add_argument(
+            '--editable', '-E',
+            dest='editable',
+            action='store_true',
+            help='Use the local repository in edit mode',
+        )
+
+        argparser.add_argument(
             '--appendix',
             type=str,
             help=self.traits()['appendix'].help,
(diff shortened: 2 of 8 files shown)


evertrol:
This may still require some tests, which are probably quite a bit more work than implementing the actual feature (given that changes in both a local directory and a container need to be verified).

betatim:
Test will indeed require a bit of thought. Some ideas below.

What do you think of using the tests from #413 (https://github.com/jupyter/repo2docker/pull/413/files#diff-73b9b926dafec45bd12128b598a0dd45) as a starting point?

I was thinking if we can execute `repo2docker` from the "outside" (with most of the other tests we only get to run code in `verify` inside the container) then we can use a [`start` file](https://repo2docker.readthedocs.io/en/latest/config_files.html#start-run-code-before-the-user-sessions-starts) that is executed inside the container to make a change and check that the local directory was indeed changed.

evertrol:
I've put the documentation as a FAQ (even if it isn't (yet)), since I'm now sure how well it fits as a how-to. And I didn't see documentation on the individual options (other than what argparse produces on the command line, obviously).

Also: I can't tick the check-box :-(

evertrol:
I seem to have hit a little snag: I can't bind an external volume to an internal volume using an internal variable. I'd like to bind to `$HOME` or `~`, but that just get escaped inside `client.containers.run(..., volumes=container_volumes, ...)`. I can get hardcode `/home/jovyan` for tests, but that is obviously not a solution when using user-supplied Docker files.

I've been looking at another way to work around this, or find a configuration option to change this, but no luck so far. Would anyone know how to do this?

(to be clear, I'd like to do:
```
client.containers.run(..., volumes={'/host/directory', {'bind': '$HOME', 'mode': 'rw'}}, ...)
```
)

betatim:
I think we could resolve `$HOME` before making that call?

I'd be Ok with having the docs say "Editable mode is a convenience option that will mount the repository to `/home/$USER`, if you need to mount to a different location in the container please use `--volumes`." if we can't find a solution to this.

betatim:
You could inspect `image['ContainerConfig']['Env']` to find `$HOME`. This would make it a bit harder to reason about which directory inside the container will be used as it would be "Relative to WorkingDir if set, otherwise relative to $HOME if set, if neither is set relative paths aren't supported"

I think it is fine to assume that `WorkingDir` is set and error out if not. The general philosophy for repo2docker is that if you use your own `Dockerfile` you are an expert and know what you are doing. We offer very little support to you when things don't work beyond "do you really need to make your own Dockerfile?". Conversely we try to support the things a lot of users want to do natively so that they do not have to write a `Dockerfile`.

So I would keep the current behaviour and add a test that a `Dockerfile` with WorkingDir set works as expected.

evertrol:
I've added two tests:

- a test that creates a file inside the container, for which it is checked that it can be viewed by the host (with its contents verified)
- a test that creates a file on the host, while the container is running. The existence of that file is then verified inside the container (though not its contents).

----

Additionally, I have converted the FAQ from Markdown to reStructuredText, since I wanted an admonition note, which is not possible with Markdown. Perhaps this change should be taken out of this PR and put into its own PR.

Lastly, I was looking for a CHANGES file to add this change, but there isn't any. Perhaps the repository is considered the CHANGES file, but generally, I find a separate file easier to read and spot (important) changes between versions, or bug-fixes.
Perhaps it's something to consider? Though I notice (sadly) few (notable) Python projects actually have such a file.

betatim:
Having a [ChangeLog would be welcome](https://github.com/jupyter/repo2docker/issues/376#issue-351079554) at least by me. We haven't been very disciplined so far so if you oulc get us started that would be great.

Do you have a preferred format? My suggestion would be to start with a style a bit like https://github.com/scikit-learn/scikit-learn/blob/master/doc/whats_new/v0.19.rst#version-0191.

Changing from MD to rst would be ideal in a separate PR so we can keep reformatting separate from content changes.

betatim:
I'll merge this now to get this cool feature in there.

neuromusic:
sorry for taking so long to say "thanks" but this is awesome! going to test it out ASAP

PR #422, opened 2018-10-03
Fix documentation: avoid clashing markup

evertrol:
Some minor documentation fixes:
- avoid nested inline markup, which is not supported by Sphinx
- remove / change some strong emphasis to a admonition


evertrol:
In case some changes feel a bit unnecessary, I can remove those commits.
Alternatively, I can squash the commits.
The double \`\` inside the bold face sections were the main itch I wanted to scratch; the other changes just followed along.

betatim:
LGTM. If someone has strong feelings on the bold font or not, let's make a new PR.

PR #424, opened 2018-10-04
put appendix into end of custom dockerfiles

bitnik:


repo2docker/buildpacks/docker.py
```diff
@@ -9,18 +9,26 @@ class DockerBuildPack(BuildPack):
     """Docker BuildPack"""
     dockerfile = "Dockerfile"
 
+    @property
+    def dockerfile_path(self):
+        return self.binder_path(self.dockerfile)
```
choldgraf: couldn't this just be `self.binder_path('Dockerfile')`?

minrk:
I'm not sure we should revert this decision. This is excluded deliberately in https://github.com/jupyter/repo2docker/pull/250, see some discussion in https://github.com/jupyterhub/binderhub/pull/459 . We can't reliably assume that the requirements for running the appendix are present when using a custom Dockerfile, so any nontrivial appendix cannot be expected to work. If we do revert this, then we would need to make sure that all authors of an appendix ensure that the appendix runs successfully on every possible base image, not just those created by repo2docker.

There are a number of disadvantages with using the appendix, and see https://github.com/jupyterhub/mybinder.org-deploy/issues/760 for my current plan to remove all use of appendix in mybinder.org, in favor of mounting a volume for custom config.

betatim:
Thanks for the history lesson. Apparently March 2018 Tim thought it wasn't a good idea to do this, so I think I will stick with my prior self.

I'd propose to close this PR then.

bitnik:
Exactly, thanks for the explanation why not to do that! I am closing this PR.

PR #426, opened 2018-10-08
Add a change log

evertrol:
Adds an initial (template) change log, with a description on its usage
in the CONTRIBUTING documentation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,0 +1,56 @@
+Version x.y.z (unreleased)
+==========================
+
+Release date: unknown
+
+New features
+------------
+
+- Editable mode: allows editing a local repository from a live container [#421]
+
+- Change log added [#426]
+
+
+API changes
+-----------
+
+
+Bug fixes
+---------
+
+
+
+Version 0.6
+===========
+
+Released 2018-09-09
+
+
+Version 0.5
+===========
+
+Released 2018-02-07
+
+
+Version 0.4.1
+=============
+
+Released 2018-09-06
+
+
+Version 0.2
+===========
+
+Released 2018-05-25
+
+
+Version 0.1.1
+=============
+
+Released 2017-04-19
+
+
+Version 0.1
+===========
+
+Released 2017-04-14
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,7 @@ This document covers:
 - Setting up for Local Development
 - Running Tests
 - Updating and Freezing BuildPack Dependencies
+- Updating the change log
 - Creating a Release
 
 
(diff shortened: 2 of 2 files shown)


evertrol:
This adds an initial change log, as briefly discussed in https://github.com/jupyter/repo2docker/pull/421#issuecomment-426986597

This change log should probably not be merged (yet), since the three issues shown are not merged yet: these serve as example entries.

A description on suggested usage is given in CONTRIBUTING.md

Ideas were taken from the [scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/master/doc/whats_new/v0.19.rst) and the [Astropy](https://github.com/astropy/astropy/blob/master/CHANGES.rst) projects.

CHANGES.rst
```diff
@@ -0,0 +1,58 @@
+Version 0.7 (unreleased)
```
betatim: We follow semver and I don't think we will have two releases "on the go" in parallel for quite a while. This means that we will keep releasing 0.6.x until we have changes that require us to bump to 0.7.0. And then there won't be any more 0.6.x+1 releases ever again.

So I'd write `Version 0.x (unreleased)` here to have it all together. I think this will make it easy to put "Version 0.6.13" at short notice and release it when we are about to merge a PR that will move us to 0.7. Instead of having to have that PR hang around a long time. WDYT?

betatim:
I like it as it is here. I think we should try and make it as simple as possible both for contributors and for ourselves so simplifying the language and format around adding a change to:

"Find the relevant section Feature/Bug fix/API change for the next release at the top of the file and add one to two sentences ..."

Maybe with a hint/tip to only do it just before the PR is going to be merged as we will hit merge conflicts that need resolving if something else gets merged in the mean time.

choldgraf:
I think this is a really nice addition, +1 when @betatim is happy

evertrol:
Thanks for the response.
I've made a few minor changes (I couldn't get myself to delete all bullet points in the contributing section yet, so I've put that in a "Details" part). The new version number template is now a generic x.y.z; that can serve any purpose.

I'll leave it here for a few days for others to comment. Once there's nothing more, I'll remove the example bullet points that aren't merged yet, so that it is correct for the current status and it can be merged.

betatim:
Needs a bit of coordinating with #427 for potential merge conflicts/garbled sentences.

betatim:
Not sure how many others will comment in the next few days as life here is pretty quiet at the moment. Let's wait for Chris to wake up and if he is happy with it merge this.

evertrol:
I'm happy to wait for #427 to be merged first, and then update this PR with those changes.

choldgraf:
This seems good to me! Perhaps we can tweak over time if there are some changes to be made, but this is a definite improvement! Once you resolve the merge conflicts I'm +1

betatim:
Merged.

We might become a real grown up project after all!

PR #428, opened 2018-10-10
Adjust css for header links

evertrol:
This removes underlines from H2 headers that are links, and feel obnoxiously out of fashion with other headers.

It is mainly a suggestion, to scratch an itch of mine on the [configuration page](https://repo2docker.readthedocs.io/en/latest/config_files.html#id5). 

The `contents::` directive seems a bit out-of-sync with the `toctree::`, since the latter doesn't let headers link back (admittedly, that is not necessary, since that would be to the navigation bar, which is already there).

### DIFF
--- a/docs/source/_static/custom.css
+++ b/docs/source/_static/custom.css
@@ -5,3 +5,7 @@ img.logo {
 div.body p.caption {
     font-size: 1.5em;
 }
+
+h2 a.toc-backref {
+	text-decoration: none;
+}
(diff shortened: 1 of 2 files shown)


choldgraf:
Agreed w/ the underlines, thanks for spotting and fixing that!   (a friendly note on the word "obnoxious"...try to keep it positive, please. While I kinda doubt it, you never know if some other repo2docker dev really loved those underlines ;-) )

PR #429, opened 2018-10-11
[MRG] Only check beginning of name of Python shared library

betatim:
Only uses the beginning of the string to determine which version of
Python we use in Julia.

In recent builds the Julia tests fail because they use the full path to the Python shared library to determine where Python is coming from. Recently that path has become `/srv/conda/lib/libpython3.6m.so.1.0 ` which fails the strict equality test. Instead we now only check for the beginning of the path.

### DIFF
--- a/tests/julia/pyplot/verify
+++ b/tests/julia/pyplot/verify
@@ -2,7 +2,7 @@
 using PyPlot
 
 # Make sure we are re-using the same conda python
-if PyCall.libpython != "/srv/conda/lib/libpython3.6m"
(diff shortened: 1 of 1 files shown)


choldgraf:
Is there any downside to this? Could we imagine *wanting* to know what came after the `6m`?

betatim:
The way I understand the test is that it wants to know that the Python shared library is loaded from the Python conda environment and not from somewhere else/the default location. For that it shouldn't care about the trailing bit which (I think) have to do with the versioning of the shared library itself.

betatim:
Agreed on the Python version and thanks for the hint to use an environment variable instead of hard coding the path.

PR #431, opened 2018-10-11
Add packages required to build docs to dev-requirements.txt

matthewfeickert:
To build the docs with

```
make html
```

requires both `sphinx` and `recommonmark`, so these should be in the `dev-requirements`.


betatim:
Thanks for this. It might be because we re-use the dev environment to also install the docs that we missed this https://github.com/jupyter/repo2docker/blob/a161b0925a3141e55db29b8d7928814ea33aa3c7/.travis.yml#L25 as it seems the docs build without a problem on travis.

matthewfeickert:
Yeah, about 1 minute after making this PR I saw that file. I'm going to just revert this and then add a new line to the `CONTRIBUTING.md` telling contributors to also install that.

betatim:
Nods it seems to me more like a bug in our documentation than in the CI setup (though that could cleaner and less like a maze).

matthewfeickert:
@betatim If you don't mind, since you merge PRs and don't squash and merge them, I might do some rewriting of history so that there doesn't need to be an unnecessary revert in the commit history.

betatim:
Feel free to doctor away at the history in this branch. I frequently use `git rebase -i HEAD~3` (or larger numbers than 3) to reorganise commits in a branch for a PR.

betatim:
Ah! Apparently I need to hard reload this page/pages with badges for them to change colour :-/

Also I think the docs build command should be `make html || exit 1` for travis to actually fail if the docs are broken. So the fact that travis doesn't fail might not tell us all that much :-(

I can't see the read the docs project for repo2docker so we need someone else to add me before I can see the output of the build/error message.

matthewfeickert:
Docs build badge is passing again! Yay!

[Documentation Status](http://repo2docker.readthedocs.io/en/latest/?badge=latest)

matthewfeickert:
Once PR #432 has fixed the same issue that is causing the Julia build to fail here and is merged in I'll rebase this against `master`.

betatim:
Rebase time :)

matthewfeickert:
@choldgraf Oh, it looks like there is a build error in the deployment. https://travis-ci.org/jupyter/repo2docker/builds/440781216

It isn't really clear to me from [the Travis log why either](https://travis-ci.org/jupyter/repo2docker/jobs/440781226#L730-L737). :? Strange.

choldgraf:
hmmm, it passed the checks in this PR, I'll restart and see if it's a travis burp

choldgraf:
here's the error message:

```
Already up to date!
HEAD detached at f133960
nothing to commit, working tree clean
Dropped refs/stash@{0} (507fc3cfeb6cfa4e283f4d2372a6cdf052bf70d8)
Script failed with status 127
failed to deploy
```

that's....weird. @betatim any thoughts?

betatim:
https://github.com/jupyter/repo2docker/pull/436 fixes that. I broke it in #432, nothing to do with this PR.

PR #433, opened 2018-10-12
[WIP] Experiment with just one travis job

betatim:
Want to see what happens if we run just one travis job which shares an image cache


betatim:
Isn't fast enough to run in one job.

PR #436, opened 2018-10-12
[MRG] Change back into main repository directory

betatim:
Follow up on #432 that changes back into the main repository directory so that deploy steps work.


choldgraf:
wow, TIL what `pushd` is haha

PR #441, opened 2018-10-13
be more specific about specifying branches

sneumann:
People are not used to including origin/ when asking for a branch-name,
see also https://github.com/jupyter/repo2docker/issues/135#issuecomment-343556938
so add that to the examples here.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -59,9 +59,10 @@ Building a specific branch / commit / tag
 =========================================
 
 To build a particular branch and commit, use the argument ``--ref`` and
-specify the ``branch-name`` or ``commit-hash``. For example::
+specify the ``branch-name`` (use the full remote name) or ``commit-hash``. For example::
 
   jupyter-repo2docker --ref 9ced85dd9a84859d0767369e58f33912a214a3cf https://github.com/norvig/pytudes
+  jupyter-repo2docker --ref origin/master https://github.com/norvig/pytudes
 
 .. tip::
    For reproducible research, we recommend specifying a commit-hash to


choldgraf:
Thanks @sneumann!!

Didn't we try fixing this a while back, so that you didn't have to use origin? If not, then I agree this is a helpful doc change :-)

betatim:
For me `jupyter-repo2docker --ref master https://github.com/norvig/pytudes` works.

jzf2101:
Yeah I haven't had a problem with not mentioning master.

docs/source/usage.rst
```diff
@@ -59,9 +59,10 @@ Building a specific branch / commit / tag
 =========================================
 
 To build a particular branch and commit, use the argument ``--ref`` and
-specify the ``branch-name`` or ``commit-hash``. For example::
+specify the ``branch-name`` (use the full remote name) or ``commit-hash``. For example::
 
   jupyter-repo2docker --ref 9ced85dd9a84859d0767369e58f33912a214a3cf https://github.com/norvig/pytudes
+  jupyter-repo2docker --ref origin/master https://github.com/norvig/pytudes
```
jzf2101: Can we use another named branch here instead while we figure out what we mean here. Also somehting like 

```
or 

jupyter-repo2docker --ref origin/master <commit-hash> https://github.com/norvig/pytudes 
```

betatim: What do you mean with `jupyter-repo2docker --ref origin/master <commit-hash> ...`? Is it a copy&paste error? Having a branch and a ref isn't something you should specify.

jzf2101: Right - the hashes are unique to the branch.  Scrap that then and just say  add the `or` in there

betatim:
Hey @sneumann let us know what you think we should do next with this.

betatim:
Closing because of inactivity. Feel free to re-open if you come back to this.

PR #443, opened 2018-10-16
[MRG] Switch repository used to test sub-directory support

betatim:
Using a repository that contains invalid instructions in the root of the
repository so that the test for subdirectory support will fail if
repo2docker doesn't actually switch into the requested sub-directory.

Follow up to #413 

This PR grew a bit because traitlets uses `os.getcwd()` when trying to load config files. The test run after `test_subdir` would find itself in a working directory that didn't exist anymore. As a result I added the `chdir()` context manager.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -35,7 +35,7 @@ from .buildpacks import (
 from . import contentproviders
 from .utils import (
     ByteSpecification, maybe_cleanup, is_valid_docker_image_name,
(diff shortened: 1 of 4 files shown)


betatim:
I think the problem with not restoring the working directory did not show up before because usually we only do one thing per process and then shutdown again. However in the tests we instantiate `Repo2Docker` multiple times and reuse the process and its working directory. Cleaning up after ourselves seems like a good idea despite this.

PR #448, opened 2018-10-19
[MRG] Switch to BaseImage for nix build pack

betatim:
This is based on #407 and experimenting with using `BaseImage` vs `BuildPack` as the class to inherit from. With `BaseImage` we get support for `postBuild` and `start` but I don't know what this would break or not.

This switches to the `BaseImage` base class, updates documentation and adds support for the `binder/` sub-directory pattern.


choldgraf:
Heya - what's the status on this one? Are we blocking on something?

betatim:
It started as an experiment and right now blocked by "not sure what to do because I don't know nix well enough". Given that I think we should merge #407 and then discuss if it should support `start` and `postBuild` or not. We only lock ourselves in when we make a release (I think).

costrouc:
Nix should package everything that it needs -- if it doesn't it is a bug with nix and I will report it to [nixpkgs](https://github.com/NixOS/nixpkgs). It should not matter what the base image is to nix. That said the smaller the image it inherits the smaller the overall image should be. Here is the reference [dockerfile for `nix`](https://hub.docker.com/r/nixos/nix/~/dockerfile/). As long as `/nix/*` is writeable to nix, ssl certs are good, and `nix-*` tools exist in `PATH` it will work.

I am more than happy to keep nix updated and working in `repo2docker`. 

Currently I working on another idea that may remove the requirement to have a nix buildpack. The idea is is to allow full virtual environments per notebook via a meta `nix-kernel` (specified within the notebook through magics). Should allow any package in your environment and have setup and teardowns for when the kernel is activated/deactivated.

yuvipanda:
For nix, I'd think we should treat it the same as we do Dockerfiles - it takes full control of the stack and does all the things needed. I *think* we should inherit from BuildPack for this one. Does nix have nix-specific ways to run postBuild? That feels like the only use case for inheriting from BaseImage...

costrouc:
Nix "builds" (really just downloads) everything in the `assemble_script_directives` https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/base.py#L112-L114. The additional `postBuild` step really doesn't have much benefit to nix. Also having support for a start script really doesn't make much sense for `nix` either since the `shellHook` attribute in `default.nix` handles this already.

So looking at it more carefully `start` and `postBuild` really don't add any value to the nix buildpack.

yuvipanda:
yeah, in that case I'd say we should go with inheriting from BuildPack.

betatim:
In repo2docker there are two phases: building the image and running the container. `postBuild` is like adding extra `RUN` lines to your `Dockerfile`. `start` let's you modify the entrypoint.

To double check I got what you were saying:

* if I want processes to be executed when the container is started then I would define the script that does that as a `shellHook`. Example use case: starting a web server and setting environment variables.

* if I want to run a process at the end of the build phase I should also add it to the `shellHook`. Example use case: downloading data files

How does nix know what to run when?

costrouc:
Yes that is exactly correct. While `shellHook` really does not know if it has been invoked within the build or run process, I thought of a work around that would allow both. I have documented this in the [nix-binder-example](https://github.com/costrouc/nix-binder-example).

```bash
    if [ ! -f $HOME/.dockerbuildphase ]; then
      touch $HOME/.dockerbuildphase
      export DOCKER_BUILD_PHASE=true
    fi
    
    if [ "$DOCKER_BUILD_PHASE" = true ]; then
      echo "Do some action in build phase"
    fi
    
    if [ "$DOCKER_BUILD_PHASE" = false ]; then
      echo "Do some action in run phase"
    fi
    
    echo "Do some action in both phases"
```

However, I believe that a nix user would take a different approach. We would download the data files using nix expressions within the `default.nix`. This would allow the download to be an idempotent operation. For starting a web server or database this would be statefull and require the workaround that I have documented above. Thanks for pointing this out or else I would not have thought to document it!

betatim:
How about making that snippet part of the nix build pack and execute `postBuild` and `start` (if they exist)? It would facilitate using the snippet above without users having to discover it first. It could also encode the nix philosophy of "please don't postBuild use a nix expression instead" by checking if a `postBuild` exists and erroring out.

One of the goals of repo2docker is to reduce complexity when it comes to building good images and putting a bit of constraints on users (or guiding them, depending on how you want to look at it :) ). Removing options is a good way to reduce complexity.

One hurdle with this is that we invented these escape hatches called `postBuild` and `start`. So we spent time teaching people about them. But also told people that if they use the `Dockerfile` build pack they are on their own and things are completely different from how repo2docker does stuff. And now we also have to teach them that the nix build pack is different again, a bit like the `Dockerfile` build pack (it controls everything) but it again has its own little escape hatches that are different from the escape hatches we invented earlier.

This is why I keep trying to find a way to reuse as much of the escape hatches we already have as possible.

repo2docker/buildpacks/nix/nix-shell-wrapper
```diff
@@ -8,8 +8,13 @@ _term() {
 
 trap _term SIGTERM
 
-echo "$*"
-nix-shell default.nix --command "$*" &
+echo 'command' "$*"
+
+if [ -f ./start ] then
+   nix-shell default.nix --command ./start "$*" &
```
costrouc: Need to be something like `nix-shell default.nix --command "./start && $*" &`. This is why I have to write a wrapper in the first place. It only looks at the first argument.

tests/nix/simple/default.nix
```diff
@@ -11,8 +11,6 @@ in
 pkgs.mkShell {
   buildInputs = with pkgs; [
     python36Packages.numpy
-    python36Packages.scipy
-    python36Packages.jupyterlab
```
costrouc: `python36Packages.jupyterlab` really should be mandatory. Otherwise `jupyterlab` will not be installed in the nix environment.

repo2docker/buildpacks/nix/nix-shell-wrapper
betatim: Struggling to understand/make this work :-/ What I was trying to achieve (and then had to leave in this half done state for a call) is to pass the arguments on to `start` because `start` scripts [expect to be given arguments that they should `exec` when they are done](https://repo2docker.readthedocs.io/en/latest/config_files.html#start-run-code-before-the-user-sessions-starts).

repo2docker/buildpacks/nix/__init__.py
```diff
@@ -57,7 +57,7 @@ def get_assemble_scripts(self):
             ('${NB_USER}', """
             nix-channel --add https://nixos.org/channels/nixpkgs-unstable nixpkgs && \
             nix-channel --update && \
-            nix-shell default.nix --command "command -v jupyter"
+            nix-shell default.nix
```
betatim: What does this do? Was experimenting here and removed it as I was getting errors that this wasn't being found. So this change is currently empirical and doesn't even fix the problem.

repo2docker/buildpacks/nix/nix-shell-wrapper
```diff
@@ -8,8 +8,13 @@ _term() {
 
 trap _term SIGTERM
 
-echo "$*"
-nix-shell default.nix --command "$*" &
+echo 'command' "$*"
```
betatim: Do you have thoughts on `$@` vs `$*`? In `start` we use `$@`

tests/nix/simple/default.nix
betatim: Nope. A test can do what it wants really because the command that is executed in the container are those listed in the `verify` script.

Was trying to cut down on the things that we install in each test to gain time as the sum of all our tests takes the age of the universe to execute :( so I am currently trying to remove things from them when I pass by.

repo2docker/buildpacks/nix/nix-shell-wrapper
costrouc: Okay so now I understand. How does `nix-shell default.nix --command "./start $*" &` work?

repo2docker/buildpacks/nix/nix-shell-wrapper
costrouc: Nope in fact I believe that `$@` is the recommended way

repo2docker/buildpacks/nix/__init__.py
betatim: Isn't jupyter installed by default by the build pack -> it should always exist?

betatim:
Let's see if this is fixed now.

repo2docker/buildpacks/nix/nix-shell-wrapper
```diff
@@ -8,8 +8,25 @@ _term() {
...
 
-echo "$*"
-nix-shell default.nix --command "$*" &
+# if there is a binder/ sub-directory it takes precedence
+# files outside it are ignored
+if [ -e ./binder ]; then
+  nixpath="./binder/default.nix";
+  if [ -f ./binder/start ]; then
+    chmod u+x ./binder/start
+    nix-shell $nixpath --command "./binder/start $*" &
```
betatim: Using `$@` here which is what the internet recommends leads to errors when the command is something like `jupyter --ip=0.0.0.0 ...` as `nix-shell` picks that up as an argument to it instead of the command.

There are several issues on the nix repos discussing this and adding support for `--` to indicate "all arguments after this are for the command, not nix-shell" but it seems they have stalled/not yet produced an implementation.

Let's use this unless someone knows what deep shell quoting magic we need :-/

betatim:
Updated documentation preview: https://33-90722372-gh.circle-artifacts.com/0/html/config_files.html#default-nix-the-nix-package-manager

This PR ends up doing several things:
* switches to a different base class
* adds support for the binder sub-directory pattern
* updates the documentation to note that nothing is installed for you, not even jupyter

docs/source/config_files.rst
```diff
@@ -164,6 +164,27 @@ used for installing libraries.
...
+repo2docker expects to be installed. If you do not install Jupyter explicitly
+repo2docker will no be able to start your container.
+
+`nix-shell <https://nixos.org/nix/manual/#sec-nix-shell>`_ is used to evaluate
+a ``nix`` expression written in a ``default.nix`` file. Make sure to
+`pin your nixpkgs <https://discourse.nixos.org/t/nixops-pinning-nixpkgs/734>`_
+to produce a reproducible environment.
+
+To see an example repository visit
+`nix binder example <https://gitlab.com/costrouc/nix-binder-example>`_.
```
costrouc: One last thing that is my fault. Could you change the link to github.com/costrouc/nix-binder-example

Otherwise everything looks great to me! Thank you for your work on making this compatible.

costrouc: I'll say that getting `nix` to work in the build pack the first time was not easy. Especially the part `nix-shell default.nix --command ....`. Yeah I would be happy to transfer over the repo. I made two due to https://github.com/costrouc/nix-binder-example/issues/1.

Would you like [nix-binder-example](https://github.com/costrouc/nix-binder-example) or [nix-binder-example-simple](https://github.com/costrouc/nix-binder-example-simple)?

betatim: Let's start with the simple one?

betatim: If you accept the invitation to the binder-examples org you should be able to transfer the repo.

costrouc: Done! Thanks.The repository will most likely need a rename. As far as I can tell I don't have permissions to do that.

costrouc: Also this link will need to be updated. Looks like you have commit access so it would be an easy change once you set the name. 

https://github.com/jupyterhub/binder/blob/master/doc/examples.rst#scientific-research-and-reproducibility

repo2docker/buildpacks/nix/nix-shell-wrapper
yuvipanda: I added this comment as a comment in the script so it's preserved.

yuvipanda:
Thanks for your work, @costrouc and @betatim!

PR #456, opened 2018-10-29
Minor typos

nuest:


.gitignore
```diff
@@ -28,3 +28,5 @@ test_file_text.txt
 # Untracked artifacts from the conda script
 repo2docker/buildpacks/conda/environment.py-3.5.yml
 repo2docker/buildpacks/conda/environment.py-3.6.yml
+
+\.vscode/
```
betatim: For my (never ending) git education: what does the `\` at the start of the line do? To ignore a hidden directory with this name I would have put `.vscode/`.

betatim:
Thanks for taking the time to make things more human readable!

choldgraf:
thanks for the patch @nuest !!

.gitignore
nuest: I though it would escape the `.` as a special character, but it works with `.vscode/` as well. Was added by the git client and I did not question it.

PR #457, opened 2018-10-30
Updates for Stencila

nuest:
- detect required contexts automatically
- install R if needed
- update dep versions
- do not do anything if `binder/` is present
- update docs
- more strictly test Stencila installation in `verify`

Note that the docs contain links to two example repositories (https://github.com/nuest/stencila-py and https://github.com/nuest/stencila-r) who have not been transfered yet to the binder-examples org. If you'd welcome the changes in this PR, we can sort that out.


repo2docker/buildpacks/base.py
```diff
@@ -518,14 +556,28 @@ def get_assemble_scripts(self):
             ))
         except FileNotFoundError:
             pass
+        if self.stencila_manifest_contexts:
+            for context in self.stencila_manifest_contexts:
+                if(context == 'py' or context == 'pyjp'):
```
minrk: Could both of these be true? If so, maybe `if {'py', 'pyjp'}.intersection(self.stencila_manafest_contexts):` to only trigger this once.

nuest: Yes, a document can contain both `py` and `pyjp` code chunks. Neat trick, I must get more used to truthiness in Python :-).

nuest: Regardin triggering: This would have been evaluated only once though because of the internal `self._stencila_manifest_contexts`, right? Or would that be once be class instance, not once per session?

docs/source/config_files.rst
```diff
@@ -107,6 +107,22 @@ You also need to have a ``runtime.txt`` file that is formatted as
...
+===================================
+
+`Stencila <https://stenci.la/>`_ is an open source office suite for reproducible research.
+It is powered by the open file format `Dar <https://github.com/substance/dar>`_.
+
+If your repository contains a Stencila document, repo2docker detects it based on the file ``manifest.xml``.
+The required `execution contexts <https://stenci.la/learn/intro.html>` are extracted from a Dar article (i.e.
+files named ``*.jats.xml``).
+
+You may also have a ``runtime.txt`` and/or an ``install.R`` to manually configure the your R installation.
```
betatim:
```suggestion
You may also have a ``runtime.txt`` and/or an ``install.R`` to manually configure your R installation.
```

repo2docker/buildpacks/r.py
```diff
@@ -64,21 +71,22 @@ def detect(self):
...
         description_R = 'DESCRIPTION'
-        if not os.path.exists('binder') and os.path.exists(description_R):
-            if not self.checkpoint_date:
-                # no R snapshot date set through runtime.txt
-                # set the R runtime to the latest date that is guaranteed to be on MRAN across timezones
-                self._checkpoint_date = datetime.date.today() - datetime.timedelta(days=2)
-                self._runtime = "r-{}".format(str(self._checkpoint_date))
-            return True
+        if not os.path.exists('binder'): 
+            if os.path.exists(description_R) or (self.stencila_manifest_contexts and "r" in self.stencila_manifest_contexts):
```
betatim: Can we make `self.stencila_manifest_contexts` default to an empty list/set instead of `None` and then change this to:

```suggestion
            if os.path.exists(description_R) or "r" in self.stencila_manifest_contexts:
```

repo2docker/buildpacks/r.py
```diff
@@ -64,21 +71,22 @@ def detect(self):
         unless a `requirements.txt` is present and we do not want to require the
         presence of a `requirements.txt` to use R.
 
-        Instead we just check if runtime.txt contains a string of the form
-        `r-<YYYY>-<MM>-<DD>`
+        Instead we check the options described in the class comment above.
```
betatim: I'd delete this line if we don't want to duplicate the contents of the doc string for the class. Ideally we'd be able to come up with a nice short summary about all the various ways to trigger the R buildpack but I currently can't think of one :-(

repo2docker/buildpacks/r.py
```diff
@@ -19,7 +19,14 @@ class RBuildPack(PythonBuildPack):
        date snapshot of https://mran.microsoft.com/timemachine
        from which libraries are to be installed.
 
-    2. An optional `install.R` file that will be executed at build time,
+    2. A `DESCRIPTION` file signaling an R package, or
```
betatim: Maybe we should split this discussion into two things: how we detect that R needs installing and how to specify which libraries should be installed.

A more eloquent version of:
```
Setup R + RStudio if runtime.txt or DESCRIPTION or jats thing

Libraries will be installed if specified in install.R or as dependencies in DESCRIPTION
```

repo2docker/buildpacks/r.py
betatim: The reason that `DESCRIPTION` files are only recognised if there is no `binder/` sub-directory is that we want to allow "opportunistic" building of R packages from their Git repository but allow package maintainers to override that by creating a `binder/` directory which specifies how to install things.

Is this what users of stencila expect? I think of stencila as a different frontend (jupyter, RStudio, nteract, stencila) or a viewer for certain types of files. So I think we still need to let users specify dependencies that should be installed which they might want to keep tidy by putting them in a `binder/` directory. What do you think?

repo2docker/buildpacks/base.py
```diff
@@ -271,16 +271,54 @@ def stencila_manifest_dir(self):
...
                 self.log.info(
-                    "Found stencila manifest.xml in %s",
+                    "Using stencila manifest.xml in %s",
                     self._stencila_manifest_dir,
                 )
                 break
         return self._stencila_manifest_dir
 
+    @property
+    def stencila_manifest_contexts(self):
```
betatim: Could we rename this to `stencila_contexts`? Makes it a bit easier to read I think and helps with line length later on.

Or are there contexts that don't come from manifests?

repo2docker/buildpacks/base.py
```diff
@@ -271,16 +271,54 @@ def stencila_manifest_dir(self):
...
+        for root, dirs, files in os.walk("."):
+            for filename in files:
+                if filename.endswith(".jats.xml"):
+                    self.log.debug("Found a .jats.xml: %s", filename)
+                    self._stencila_manifest_contexts = set()
+
+                    # extract code languages from file
+                    with open(os.path.join(root, filename)) as f:
+                        for line in f:
+                            match = re.match('.*language="(.+?)"', line)
```
betatim: Can we use a XML parser to read in the jats.xml file and find these attributes? Using a regex on a XML file feels like we will get weird breakage in a few months or so.

repo2docker/buildpacks/base.py
```diff
@@ -518,14 +556,27 @@ def get_assemble_scripts(self):
...
+                            "${NB_USER}",
+                            r"""
+                            ${KERNEL_PYTHON_PREFIX}/bin/pip install --no-cache https://github.com/stencila/py/archive/f1260796.tar.gz && \
+                            ${KERNEL_PYTHON_PREFIX}/bin/python -m stencila register
+                            """,
+                        )
+                    ]
+                )
+            #if {'r'}.intersection(self.stencila_manifest_contexts):
+                # handled in RBuildPack
```
betatim: delete these two lines

repo2docker/buildpacks/r.py
```diff
@@ -226,6 +235,22 @@ def get_build_scripts(self):
             ),
         ]
 
+        if self.stencila_manifest_contexts:
```
betatim: If we make the default value an empty set we can skip this level of indention and directly test if R is a detected language.

betatim:
Is it easy to add a test to check that stencila can run the document in the repository? As a way to check all the plumbing is connected properly.

repo2docker/buildpacks/r.py
nuest: Deleted. Your call, really. I guess the "nice short summary" means it can't be automated.

Do you think there should be a developer-facing variant of the config files docs?

repo2docker/buildpacks/r.py
nuest: We sure can! I just learned that empty lists are falsy, and other checks just do `if self.stencila_manifests_contexts`.

nuest: I thought it's better to align the Stencila workflow with other use cases, i.e. if a `/binder` directory is there, no "repo2docker" magic happens. But I must admit I have have misunderstood the use of the `binder/` directory! So you are right, this change is not useful and Stencila users should be able to put their configurations into a `binder/` directory.

As for _Stencila users_, the whole project is aimed at non-coders, so I think it is unlikely that we have an author of a Stencila document who wants to write her own Dockerfile. An install.R script is probably as far as they may want to go.

_Should this also mean we should _not_ stop detecting `manifest.xml` and the executions contexts if there is a `binder/` directory?_

I think so! Because if we stop the detection, a user must install all the Stencila stuff herself, which means that particular user may also define a `Dockerfile`, see new example at https://github.com/nuest/stencila-multi (still checking if that works as advertised).

I think this is more in line with the "regular way" of using binder, and not mimicking the R package use case. Thanks for pointing this out @betatim

repo2docker/buildpacks/base.py
nuest: Well, actually the contexts do not come from the manifests, but from the document files :see_no_evil:  - so good point, renamed.

repo2docker/buildpacks/base.py
nuest: The regex does match, correct. It did not hurt so far, but you're right to keep things cleaner.

I just marked the PR as "WIP", because using an XML parser is no quick fix I'll do today.

repo2docker/buildpacks/r.py
betatim: They are indeed false-y, but you can check membership of an empty list so we don't have to check first if it is empty/false-y or not.

nuest:
@betatim Thanks so much for the thorough review! Here's my to-do list:

- [ ] add a check that Stencila can "run the document" - @nokome, any hint for this? Is there somethink like `knitr:render(.)` on CLI for both `stencily/py` and `stencila/r`
- [x] use XML parser to extract contexts, not a regex
- [x] use XML parser to extract the document `path` instead of looking for `*.jats.xml`

nuest:
The tests fail because building the image for the `stencila/r` test takes over 10 minutes, see https://travis-ci.org/jupyter/repo2docker/jobs/448412375

image

Use `travis_wait` for Stencila tests now...

repo2docker/buildpacks/r.py
betatim: Some developer facing docs would be good. Maybe building on https://repo2docker.readthedocs.io/en/latest/architecture.html?

repo2docker/buildpacks/r.py
betatim: If there is a `binder/` subdirectory repo2docker will still spring into action. It will however only consider files in this subdirectory.

If there is a `Dockerfile` then repo2docker follows that and does nothing beyond `docker build`.

I think we should ignore the presence or absence of a `binder/` directory for deciding things for stencila.

Detect which languages are needed by looking at all the manifest files and then use the composability of repo2docker to let people install extra libraries via `requirements.txt` or `install.R` if they want to.

betatim:
Do you know what makes the stencila build take so long? If it is the libraries/dependencies used maybe we can find some that are faster to install?

repo2docker/buildpacks/base.py
```diff
@@ -518,14 +554,25 @@ def get_assemble_scripts(self):
             ))
         except FileNotFoundError:
             pass
+        if self.stencila_contexts:
```
betatim: not needed as `{'a', 'b'}.intersection(set())` works

nokome:
Hi all :wave: 

First of all, just wanted to say thanks to @nuest, @minrk and @betatim for all your work on this. @choldgraf mentioned to me a while back that some of the eLife sprint work had been integrated into `repo2docker` but I hadn't realised the extent of it until I saw this. We've been working on a similar tool recently https://github.com/stencila/dockter (which I am was going to email you all for feedback any day) - which overlaps with this work more than I expected in terms of Stencila support.

Some specific comments:

> add a check that Stencila can "run the document" - @nokome, any hint for this? Is there something like knitr:render(.) on CLI for both stencily/py and stencila/r

No, unfortunately not right now  - we had something like it about a year ago but I suspect it's broken. Maybe give me a day or two to let me think about what might be the best way to approach that.

>  "Do you know what makes the stencila build take so long? If it is the libraries/dependencies used maybe we can find some that are faster to install?"

It's a pain point for me too right now and something I've been think about how best to address. We have quite a few package dependencies https://github.com/stencila/r/blob/master/DESCRIPTION#L11 but I think some of them could be dropped pretty easily with some refactoring. That is likely to be a couple of weeks away for me - but we could perhaps chop out some of the partially implemented features more quickly (e.g. `roxygen` and `xml2` are only required for generating function manifests and we are changing approach there anyway).

Thanks again :tada: !

repo2docker/buildpacks/r.py
nuest: Should be implemented this way now.

repo2docker/buildpacks/base.py
nuest: Fixed.

repo2docker/buildpacks/base.py
```diff
@@ -271,16 +272,51 @@ def stencila_manifest_dir(self):
...
+        """Find the stencila manifest contexts if it exists"""
+        if hasattr(self, '_stencila_contexts'):
+            return self._stencila_contexts
+
+        # look at the content of the documents in the manifest to extract the required execution contexts
+        self._stencila_contexts = set()
+
+        # get paths to the article files from manifest
+        files = []
+        if (self.stencila_manifest_dir):
```
```suggestion
        if self.stencila_manifest_dir:
```

repo2docker/buildpacks/base.py
```diff
@@ -518,14 +554,24 @@ def get_assemble_scripts(self):
             ))
         except FileNotFoundError:
             pass
+        if {'py'}.intersection(self.stencila_contexts):
```
```suggestion
        if 'py' in self.stencila_contexts:
```

repo2docker/buildpacks/r.py
```diff
@@ -19,8 +19,22 @@ class RBuildPack(PythonBuildPack):
...
-    2. An optional `install.R` file that will be executed at build time,
-       and can be used for installing packages from both MRAN and GitHub.
+    2. A `DESCRIPTION` file signaling an R package
+
+    3. A Stencila document (*.jats.xml) with R code chunks (i.e. language="r")
+
+    If there is no `runtime.txt`, then the MRAN snapshot is set to latest
+    date that is guaranteed to exist across timezones.
+
+    R packages are installed if specified either
```
betatim:
```suggestion
    Additional R packages are installed if specified either
```

repo2docker/buildpacks/r.py
```diff
@@ -19,8 +19,22 @@ class RBuildPack(PythonBuildPack):
...
+
+    - in a file `install.R`, that will be executed at build time,
+      and can be used for installing packages from both MRAN and GitHub
+
+    - as dependencies in a `DESCRIPTION` file
+
+    - are needed by a specific tool, for example the package `stencila` is
+      installed and configured if a Stencila document is given.
```
betatim: I think this bit is about installing user specified packages so we should only list the ways by which a user can specify "please install X and Y" instead of listing the situations in which extra packages will be installed from a developer point of view.

The suggestion on L29 is connected to this.

tests/stencila-py/pyjp/verify
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+jupyter serverextension list 2>&1 | grep nbstencilaproxy
+jupyter nbextension list 2>&1 | grep nbstencilaproxy
+python3 -c "import stencila" 2>&1 | grep ModuleNotFoundError
```
betatim: does the `pyjp` test check for something that the `py` test doesn't already cover? I was expecting to somehow run directly from a notebook or some such but it looks like this uses the manifest file and the jats file like the other test?

nuest: The difference is in the used execution contexts that are installed. If the document only has `pyjp` cells, then the module `stencila/py` is not needed, because only Jupyter will be used to execute cells. 

In a previous version if handled the `pyjp XOR py` case in the same way, just installed the package. This is now more specific, but an extra case to handle/tests. _Should I remove this?_

nuest: The difference is in the used execution contexts that are installed. If the document only has `pyjp` cells, then the module `stencila/py` is not needed, because only Jupyter will be used to execute cells. 

In a previous version if handled the `pyjp XOR py` case in the same way, just installed the package. This is now more specific, but an extra case to handle/test.

nuest: The difference is in the used execution contexts that are installed. If the document only has `pyjp` cells, then the module `stencila/py` is not needed, because only Jupyter will be used to execute cells. 

In a previous version if handled the `pyjp XOR py` case in the same way, just installed the package. This is now more specific, but an extra case to handle/test.

repo2docker/buildpacks/base.py
```diff
@@ -271,16 +272,51 @@ def stencila_manifest_dir(self):
...
+            return self._stencila_contexts
+
+        # look at the content of the documents in the manifest to extract the required execution contexts
+        self._stencila_contexts = set()
+
+        # get paths to the article files from manifest
+        files = []
+        if (self.stencila_manifest_dir):
+            manifest = ET.parse(os.path.join(self.stencila_manifest_dir, 'manifest.xml'))
+            files = list(map(lambda x: os.path.join(self.stencila_manifest_dir, x.get('path')), manifest.findall('./documents/document')))
```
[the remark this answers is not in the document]
nuest: That's a lot of syntactic sugar that I learned there! Thanks!

Anyway, I changed this to something that I am confident I will understand in a few months time, hope you agree.

nuest:
Re. faster installation (@nokome): I played around a bit with installing pre-compiled dependencies, but that turned out to be a rabbit hole, see https://gist.github.com/nuest/8beca3b75bba97f107a314798879a2fc

nuest:
Tests pass again, I hope I could address all suggestions appropriately.

tests/stencila-py/pyjp/verify
betatim: Ah ok, I didn't know that there were two ways of executing Python code. Thanks!

betatim:
Merged! Thanks for all the patience and iterations!

Do you want to deploy this to mybinder.org? We typically do deploys on Tuesday and Thursdays so that people can plan to be around for a bit after the deploy in case of problems. To make the change to mybinder.org [checkout our SRE guide](https://mybinder-sre.readthedocs.io/en/latest/deployment/how.html#updating-repo2docker).

nuest:
@betatim Thanks to you for the good feedback and thorough checking. Yes, I very much would like to deploy this to mybinder, will take a look at the docs.

choldgraf:
wohooo! congrats @nuest and thanks for the contribution!!!

PR #461, opened 2018-11-04
[MRG] Start reusing existing docker images if content hasn't changed

### LINKED ISSUE #452: Add "caching" functionality similar to BinderHub's
In #450 we discussed the possibility of creating a container from an existing image that repo2docker built previously. Currently when you run repo2docker again it has to re-execute the steps after copying the source repo over. This can take a long time if you have a lot of dependencies or R packages.

For git repositories we could add a functionality similar to BinderHub's that checks the local registry for previously built images and reuses them. This would make it a lot faster to launch and re-launch repositories that haven't changed.

betatim:
Content providers can specify a "content ID" which is used to identify
versions of the content. The ID is used in the docker image name and if
we find an existing image for a given content ID the build step is
skipped.

(This is also physics @ctb 😀)

Closes #452

### DIFF
--- a/repo2docker/contentproviders/git.py
+++ b/repo2docker/contentproviders/git.py
@@ -53,6 +53,6 @@ class Git(ContentProvider):
     def content_id(self):
         """A unique ID to represent the version of the content.
 
-        Uses the git commit ID of the repository.
+        Uses the first seven characters of the git commit ID of the repository.
         """
-        return self._sha1
+        return self._sha1[:7]
--- a/repo2docker/contentproviders/git.py
+++ b/repo2docker/contentproviders/git.py
@@ -44,3 +44,15 @@ class Git(ContentProvider):
                                     cwd=output_dir,
                                     capture=yield_output):
                 yield line
+
+        cmd = ['git', 'rev-parse', 'HEAD']
+        sha1 = subprocess.Popen(cmd, stdout=subprocess.PIPE, cwd=output_dir)
+        self._sha1 = sha1.stdout.read().decode().strip()
+
+    @property
+    def content_id(self):
+        """A unique ID to represent the version of the content.
+
+        Uses the git commit ID of the repository.
+        """
+        return self._sha1
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -13,7 +13,6 @@ import sys
 import logging
 import os
 import pwd
-import subprocess
 import tempfile
 import time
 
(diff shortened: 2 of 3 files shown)


psychemedia:
So this make me wonder...

Could repo2docker do something like:

- rebuild an "environment" image from a repo if build related files have changed
- use pre-existing image if only "content" files (eg notebooks) have changed and mount those into a notebooks volume shared with the environment image?

betatim:
> Could repo2docker do something like:

That idea is being explored in #410. The tricky part is figuring out what is a "content" file and what is a "environment" file. For example if you use a `requirements.txt` as environment file then any file in the repo could influence what is installed as you could run `pip install -e.` in any of the subdirectories of the repo.

choldgraf:
Am I correct that this is what the PR is implementing:

* Lets people specify a kwarg "content_id" that is meant to be a unique identified for content that the *user* manually specifies
* Adds that ID to the built docker image registry somehow
* If a new build is triggered and the content ID is the same, then it just re-uses an old image?

(maybe more generally, some more context about what this PR implements, why, what the big upsides are, etc would be helpful for me to understand what's going on!)

betatim:
Right now if you run `repo2docker https://github.com/awesome-org/super-repo` twice (without the remote repo actually changing) you have to build the image twice. This PR allows the Git content provider to return a "content ID" (in this case the SHA-1 of the commit) which repo2docker uses to check if there already is an image for this. If yes it will skip building and directly launch the image.

It is like the behaviour of a BinderHub that won't rebuild a repo unless there is a new commit.

The user doesn't specify anything. It is up to a content provider to return a "content ID". The local directory content provider doesn't do this so we always rebuild the image (this is fine because if nothing has changed we profit from the caching that docker does).

So far we add a timestamp to the end of the docker image name so we have something like `r2d-https-blahblahblah-124235` as image name. If we rebuild the same repo 10 seconds later we end up with `r2d-https-blahblahblah-124245`.

If we merge this PR we will have names like `r2d-https-blahblahblah-$CONTENTID` which lets us check if an image with that name already exists and if yes skip building it.

repo2docker/contentproviders/git.py
```diff
@@ -44,3 +44,15 @@ def fetch(self, spec, output_dir, yield_output=False):
...
+        sha1 = subprocess.Popen(cmd, stdout=subprocess.PIPE, cwd=output_dir)
+        self._sha1 = sha1.stdout.read().decode().strip()
+
+    @property
+    def content_id(self):
+        """A unique ID to represent the version of the content.
+
+        Uses the git commit ID of the repository.
+        """
+        return self._sha1
```
minrk: Maybe truncate sha1 to 7 or 10 characters here? There's a limit to the length of image names, and adding a full hash could make it a little too easy to bump into that

yuvipanda:
@betatim we put this in tags in binderhub, right? If so maybe we should do that here too. But we can explore that in a different pr

betatim:
Nods, I thought about using tags but then decided not to because the timestamp (current behaviour) isn't in a tag either. I'll change to using tags and write some tests.

yuvipanda:
@betatim cool. I've merged this now, can move to tags after :)

Issue #423, opened 2018-10-03
Make a logo for repo2docker?


choldgraf:
Sometimes I'd like to have a visual way to describe repo2docker. Right now, It's hard to do this because there's no "visual" component to this project. Are folks generally +1 on having a logo of some kind for this project?

jzf2101:
@tgeorgeux do you have any thoughts on what we should do?

tgeorgeux:
I would probably go with some combination of https://git-scm.com/images/logos/downloads/Git-Icon-1788C.png, the number 2 and the whale logo. Seems like a good place to start at least :)

jzf2101:
is there an IP issue wiht that?

tgeorgeux:
This is where I would check for that. https://git-scm.com/about/trademark

betatim:
Can we use maybe a folder icon instead of a git logo so that we don't need to change the logo when we support things other than git repositories as source of the "repo"?

Some other ideas:
* a picture of a shipping container which on the left is shown as a sketch/drawing, the middle half of the container is half built and the right half is fully assembled. Kinda of showing how we go from a drawing of a repo to a fully made container
* a cartoon style production line that spits out shipping containers

Thoughts:
* don't get too tied to git
* ideally even get away from "docker" and instead focus on "container"
* has to print well to stickers
* red is a cool colour
* convey the idea that repo2docker constructs, builds or assembles stuff by following a recipe/blueprint
* the recipe is easy to follow like Ikea instructions
* repo2docker is a craftsman, proud and opinionated about what it builds
* repo2docker is simple and its logo is too

agahkarakuzu:
If I were to stick with the marine deck equipment theme, would repo2docker be the crane that puts containers on the back of the whale, or would it rather be the operator of the crane, following instructions written on a paper?

agahkarakuzu:
This illustration is from [Behance](www.behance.net). Maybe works as an inspiration trigger and can be used in a presentation (CC, non-comercial, no-derivatives) by giving credit to the source. I will try to come up with something more closer to @betatim's points. 

![image](https://user-images.githubusercontent.com/9632322/46902646-b7de1000-ce96-11e8-9e25-594c64e4d2ec.png)

betatim:
[...]

PR #463, opened 2018-11-05
[MRG] Add and use new logo

### LINKED ISSUE #455: Proposing a logo for repo2docker
repo2docker doesn't have a logo which means its documentation page looks somewhat blank and makes it tricky to represent on slides. Also, what serious project doesn't have a logo :) Here is a proposal arrived at in #423:

46944629-441e4c00-d0be-11e8-87f8-a1212f6b5ccc

To move this forward at a steady but consistent pace the idea is to agree the basic ideas of the logo (individual pages with a fork symbol on them (representing the repo) and a container) and then:

* Colour scheme (e.g. do we want to adopt any of the Jupyter colors?)
* Structure and layout

Like the code and documentation this logo doesn't have to be perfect on day one and the expectation is that we'd keep tweaking it for a while.

What do you think of this? If you'd like to change something can you try and make a specific suggestion for what to change and how. This would help us in actually making it happen. You can check #423 for how things evolved, other ideas and inspirations that lead to where we are now.

This issue will stay open for at least 5 November 2018. Unless there are strong objections by that time this would become "the logo".

here's a link to the raw logo files: https://github.com/jupyter/repo2docker/files/2548509/repo2docker_v2.zip

betatim:
Closes #455 

README preview: https://github.com/jupyter/repo2docker/blob/a9c80b0af41ae26ce9c3006424387268490c1c70/README.md

Docs preview: https://11-90722372-gh.circle-artifacts.com/0/html/index.html


betatim:
Let's merge this and then figure out the path to the PNG that is referenced in the README.md. Currently that is (I think) referencing a file on a branch in my repo. I don't know how to find out what that URL will be after merging this, hence the double step.

willingc:
FYI Cal Poly folks, @ellisonbg @tgeorgeux @zsailer @Ruv7

betatim:
In #455 we originally wrote "5 November" but there is no rush.

How about putting comments/thoughts which aren't blockers but should be addressed at some point in #455 instead of here?

choldgraf:
ahhh right - I guess I "forgot forgot the 5th of november" lol...let's leave it up a bit longer in case the cal poly folks would like to share their thoughts

choldgraf:
For anyone who's arriving at this PR for the first time:

* [Here's the original logo open design issue](https://github.com/jupyter/repo2docker/issues/423)
* [Here's the issue proposing a specific design](https://github.com/jupyter/repo2docker/issues/455)...we left it in an "open comment period" for ~10 days to get feedback, but didn't get much. I think we should have "@" notified more people

This PR is now a "final design" proposal to be merged in to the repo.

willingc:
> we left it in an "open comment period" for ~10 days to get feedback, but didn't get much. I think we should have "@" notified more people
Probably a good practice to always notify (@-mention) @ellisonbg @tgeorgeux @Ruv7 @ZSailer as a courtesy when there are UI/UX logos, visuals, branding being changed.  Let's give it 48 hours for these folks to chime in if desired. After that, I think the change can be merged with no objections.

choldgraf:
+1 to giving it several days! Sorry if this went under the radar for some folks.

choldgraf:
friendly ping on this - think that we can move things forward here?

PR #467, opened 2018-11-12
Add jupyterhub to base environment

### LINKED ISSUE #458: Re-add jupyterhub into default installs
Since https://github.com/jupyter/repo2docker/pull/111 jupyterhub is not installed by default with repo2docker. Because BinderHub doesn't require jupyterhub in images since https://github.com/jupyterhub/binderhub/pull/110. 

But now BinderHub supports authentication (https://github.com/jupyterhub/binderhub/pull/666) and it makes sense to add jupyterhub into default installations again. Because when authentication is enabled, notebook server should be aware of the hub, so hub can start/stop it with default hub home actions and users don't need a token to reach a running server (if logged out).

What do you think?

(I was thinking to install jupyterhub in appendix but it doesn't effect custom docker images)

yuvipanda:
Uses pip instead of conda, since (AFAIK) conda installs
CHP and friends too.

Fixes #458

### DIFF
--- a/repo2docker/buildpacks/conda/environment.frozen.yml
+++ b/repo2docker/buildpacks/conda/environment.frozen.yml
@@ -1,28 +1,27 @@
 # AUTO GENERATED FROM environment.py-3.6.yml, DO NOT MANUALLY MODIFY
-# Frozen on 2018-09-17 08:36:41 UTC
+# Frozen on 2018-11-12 20:48:01 UTC
 name: r2d
 channels:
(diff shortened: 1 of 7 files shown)


yuvipanda:
I don't have enough experience with environment.yml and conda to see if this is correct.

betatim:
LGTM.

This brings in a new major version of IPython (7 from 6) which as has various async features. Should we bump the image prefix on mybinder.org for this?

PR #470, opened 2018-11-14
adding symlink to roadmap in root

choldgraf:
This adds a symlink to the roadmap doc in the root of the repository, since people often look for it there (particularly more developer-types)


choldgraf:
eh I take it back, I thought the symlink would automatically direct to the target, but doesn't look like this is the case :-/

PR #472, opened 2018-11-15
Ensure Python3 before reading README with encoding arg

nokome:
If you accidentally try to install using Python 2 you get this error:

```python
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/home/nokome/nokome/repo2docker/setup.py", line 5, in <module>
        with open('README.md', encoding="utf8") as f:
    TypeError: 'encoding' is an invalid keyword argument for this function
```
Ideally, the user would be asked to use Python 3. If you remove the `encoding` argument, and try again then `pip` gives us this, more informative message:

```bash
$ python -m pip install -e .
Obtaining file:///home/nokome/nokome/repo2docker
jupyter-repo2docker requires Python '>=3.4' but the running Python is 2.7.12
```

This little PR implements a similar behaviour, whilst also reading the README using the correct encoding.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,12 @@
 from setuptools import setup, find_packages
+import sys
 import versioneer
 
-import sys
 if sys.version_info[0] < 3:
-    sys.stderr.write('jupyter-repo2docker requires Python 3 but the running Python is %s.%s.%s' % sys.version_info[:3])
-    sys.exit(1)
-
-with open('README.md', encoding="utf8") as f:
-    readme = f.read()
+    readme = None
+else:
+    with open('README.md', encoding="utf8") as f:
+        readme = f.read()
 
 setup(
     name='jupyter-repo2docker',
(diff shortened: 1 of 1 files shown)


betatim:
Seems like a good thing to do.

One pondering: should we make the reading of the README conditional on the Python version? For the case were we are on Python 2 we can set it to an empty string and then let `pip` generate the error message. The advantage would be that it shows the actual minimum required (currently 3.4 but maybe in the future we start requiring something newer and then "Python 3" isn't good enough anymore).

nokome:
> One pondering: should we make the reading of the README conditional on the Python version?

Yes, that's a better approach! I'll modify.

nokome:
Now, when trying to install with Python 2 gives the error:

```python
> python -m pip install -e .
Obtaining file:///home/nokome/nokome/repo2docker
jupyter-repo2docker requires Python '>=3.4' but the running Python is 2.7.12
```

PR #474, opened 2018-11-16
[MRG] Tweak docstring to make it more understandable

betatim:
I think this is better than the old one which was garbled as well as unclear.


choldgraf:
what you mean you don't know what a locationself is?

betatim:
I used to know but apparently forgot. It all made sense when I originally wrote it :)

Issue #310, opened 2018-05-11
Cache intermediate docker layers on travis


betatim:
Some builds tests take a long time because they have to build our repo2docker "base" again and again.

Travis can cache files between builds however they explicitly mention docker images as something [not to cache](https://docs.travis-ci.com/user/caching/#Things-not-to-cache). I think the reasoning is that pulling them from docker hub is just as fast/slow. In our case we want to cache the intermediate build layers.

What do people think about adding that and how it relates to Travis' wishes about docker image caching?

minrk:
Worth a shot

Madhu94:
Is this still relevant ?

betatim:
Would be interesting to experiment with to find out if this speeds up our CI runs.

manics:
I'll close this, we've got other issues related to caching and speeding up builds.

PR #478, opened 2018-11-29
Allow specifying images to reuse cache from

yuvipanda:
This lets us explicitly specify images that repo2docker
should try to re-use cached layers from. Docker normally only
looks for layers from images that were *built* locally - if
we want it to look in images that were *pulled* from a registry,
we need to specify it here.

This isn't a problem on binder, since we don't pull images from
elsewhere yet. However, this *is* a problem for hubploy, and
would make the life of https://github.com/yuvipanda/hubploy/pull/4
much easier.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -72,7 +72,7 @@ class Repo2Docker(Application):
         """
     )
 
-    reuse_layers_from = List(
+    cache_from = List(
         [],
         config=True,
         help="""
@@ -419,10 +419,10 @@ class Repo2Docker(Application):
         )
 
         argparser.add_argument(
-            '--reuse-layers-from',
+            '--cache-from',
             action='append',
             default=[],
-            help='Docker images to attempt to re-use cached layers from'
+            help=self.traits()['cache_from'].help
         )
 
         return argparser
@@ -564,8 +564,8 @@ class Repo2Docker(Application):
         if args.subdir:
             self.subdir = args.subdir
 
-        if args.reuse_layers_from:
-            self.reuse_layers_from = args.reuse_layers_from
+        if args.cache_from:
+            self.cache_from = args.cache_from
 
         self.environment = args.environment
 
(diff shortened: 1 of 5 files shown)


yuvipanda:
I am not sure at all how to test this.

betatim:
I like it! Two ponderings:

* should we stick to using exactly the same command-line argument name as `docker build` uses instead of changing the name slightly?
* can we offer some (automatic) behaviour to the unsuspecting user that "does the right thing" and speeds up their local builds? See last paragraph for some thoughts

re: testing, will `docker inspect` tell you if a layer was built or taken from the cache? Alternatively I think a layer that is shared between two images should have the same ID in both so if we `inspect` the cache-from image and the newly built one we should be able to find shared layers.

---

We have discussed building and pushing images that contain the layers before copying in the repo contents for each build pack to docker hub. This would speed up the 100% shared part of the builds, for the user who has never built that buildpack before locally. If we do this then we could have a `--cache-from=auto` setting that adds all those images to the list. Would we end up pulling all images or only those that are used?

There is also #310 for which this PR could be useful.

yuvipanda:
> I like it! Two ponderings:
> 
>     * should we stick to using exactly the same command-line argument name as `docker build` uses instead of changing the name slightly?

Yeah, that sounds like a good idea.

>     * can we offer some (automatic) behaviour to the unsuspecting user that "does the right thing" and speeds up their local builds? See last paragraph for some thoughts

This already happens for images already built locally. --cache-from is only needed to re-use cache from images *pulled* from an external repository.

> 
> re: testing, will `docker inspect` tell you if a layer was built or taken from the cache? Alternatively I think a layer that is shared between two images should have the same ID in both so if we `inspect` the cache-from image and the newly built one we should be able to find shared layers.

Since cache is used both for images already built locally and images pulled from a remote registry, there's no easy way to find out if this was effective or not.

> 
> We have discussed building and pushing images that contain the layers before copying in the repo contents for each build pack to docker hub. This would speed up the 100% shared part of the builds, for the user who has never built that buildpack before locally. If we do this then we could have a `--cache-from=auto` setting that adds all those images to the list. Would we end up pulling all images or only those that are used?

This seems like a larger discussion that we should have in a separate issue.

> There is also #310 for which this PR could be useful.

yuvipanda:
Maybe from https://github.com/jupyter/repo2docker/pull/482#issuecomment-444778403, another option here is to mock the docker client. I think that'd be much easier, actually! The integration test would be testing that docker itself works fine, while we only care wether we pass through the param properly

PR #481, opened 2018-12-05
Revert "[MRG] Start reusing existing docker images if content hasn't changed"

yuvipanda:
Reverts jupyter/repo2docker#461

This was causing the --image-name parameter to be ignored,
due to unconditional setting of `output_image_spec` in `fetch`.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -13,6 +13,7 @@ import sys
 import logging
 import os
 import pwd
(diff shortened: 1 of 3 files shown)


yuvipanda:
In discussion with @betatim, we are going to revert this for now. Then we can redo this, but this time with Tests and Docs, and merge it again :)

PR #482, opened 2018-12-05
Log a 'success' message when push is complete

yuvipanda:
Otherwise there's no useful info printed on completion when 
you are pushing from a local install.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -566,6 +566,7 @@ class Repo2Docker(Application):
                 self.log.info('Pushing image\n',
                               extra=dict(progress=layers, phase='pushing'))
                 last_emit_time = time.time()
+        self.log.info(f'Successfully pushed {self.output_image_spec}', extra=dict(phase='pushing'))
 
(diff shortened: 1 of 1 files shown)


betatim:
Should we go down the route of `mock`ing the docker client to test this bit?

I've used `mock` before, mostly via its `patch()` context manager but it never felt "right" (always needed several mocks mocking each others and :-/). Maybe someone with experience of using it could create a minimal example of how we'd do this?

yuvipanda:
@betatim I don't think log messages need testing - or rather, I don't think repo2docker is at a testing place where adding tests for log messages is a useful user of our time.

yuvipanda:
@betatim I'm sorry, I thought you meant mocking the *log* object to do testing.

I don't entirely understand reference to docker client in this PR. Can you expand a little bit on what you meant?

yuvipanda:
Maybe meant for https://github.com/jupyter/repo2docker/pull/478#issuecomment-444942250? I think it's a great suggestion for that PR!

PR #483, opened 2018-12-06
Make apt-get install be quiet

yuvipanda:
`apt-get install` calls `dpkg` underneath to do the
installation. Unfortunately, `-qq` does not get passed
through, and we end up with a lot of log output even with
`-qq`. Redirecting stdout to /dev/null stops this output.
Errors come out to stderr, so those will be displayed.

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -16,10 +16,10 @@ FROM buildpack-deps:bionic
 ENV DEBIAN_FRONTEND=noninteractive
 
 # Set up locales properly
-RUN apt-get update -qq && \
-    apt-get install -qq --yes --no-install-recommends locales && \
-    apt-get purge -qq && \
-    apt-get clean -qq && \
+RUN apt-get -qq update && \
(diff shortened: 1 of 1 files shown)


yuvipanda:
This isn't enough / nor make much of a difference either. Apparently the output isn't from apt-get but from dpkg that is spawned by apt-get ugh.

repo2docker/buildpacks/base.py
```diff
@@ -51,24 +51,24 @@
...
+    && apt-get -qq purge && \
+    apt-get -qq clean && \
     rm -rf /var/lib/apt/lists/*
 
 # This apt-get install is *not* quiet, so users see their packages being installed
 {% if packages -%}
-RUN apt-get update -qq && \
-    apt-get install --yes \
+RUN apt-get -qq update && \
+    apt-get -qq install --yes \
```
minrk: The comment above says this is not quiet, but the change makes it quiet.

yuvipanda: I wanted packages installed with apt.txt to not be quiet, so I've moved the comment appropriately

PR #491, opened 2018-12-10
Update the roadmap

betatim:
I've updated the roadmap to remove things which are done, add new things, etc.

There are more discussions in https://discourse.jupyter.org/t/repo2docker-roadmap-review/249 that can be added or lead to changes. As the roadmap is a living document I don't think we should wait for all discussions to come to an end and instead incrementally update it.

Preview: https://78-90722372-gh.circle-artifacts.com/0/html/contributing/roadmap.html

### DIFF
--- a/docs/source/contributing/roadmap.md
+++ b/docs/source/contributing/roadmap.md
@@ -37,7 +37,7 @@ them later)
 
 ### Reviewing and Updating the Roadmap
 
-The roadmap will get updated as time passes (next review by 1st December) based
+The roadmap will get updated as time passes (next review by 31st January 2019) based
 on discussions and ideas captured as issues.
 This means this list should not be exhaustive, it should only represent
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks for doing this, @betatim. I agree re: incremental updates.

choldgraf:
I am fine with incremental updates so long as we have semi-regular community reviews of the roadmap

betatim:
This PR was my understanding of https://discourse.jupyter.org/t/repo2docker-roadmap-review/249/16 as things we agreed on/people wanted to actively work on.

There are still unresolved things (like sub-directory naming). I decided that we didn't have to wait to conclude the discussion to update the roadmap because having that discussion is maybe something that should be on the roadmap instead of waiting to update the roadmap till we have a conclusion. So I expect that discussion to continue and to take part in it.

PR #492, opened 2018-12-10
Add codecov configuration

betatim:
This adds a codecov config that requires PRs to have a minimal amount of coverage for their diff.

We are currently at 75% coverage which is pretty low. Having an automated reminder like this should help us add tests for new PRs to see if we can increase overall coverage. At least it won't go down :)

Should check that this doesn't lead to the bot commenting on each PR as that causes a lot of noise. The goal is to have an extra check where we see the travis result.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,10 @@
+comment:
+  layout: header, changes, diff
+coverage:
+  status:
+    patch:
+      default:
+        target: '20'
+    project:
+      default:
+        target: auto


yuvipanda:
I like it! Let's see what the bot does.

Thanks for making our tests better, @betatim :)

betatim:
screen shot 2018-12-10 at 22 50 33

the second entry is what I was hoping to get. It tells you how much of the diff (changes in this PR) are covered by tests. We are aiming for at least 20% (except in this PR nothing under test changed).

Maybe the bot will also complain if your PR lowers overall coverage. Not quite sure how that could happen if your diff is covered, but yeah. Let's deal with that if/when it happens.

yuvipanda:
Awesome! This made me realize in https://github.com/jupyter/repo2docker/pull/478 I should also test the two other places `build` is called in, and I added those tests.

PR #494, opened 2018-12-11
Add unit tests for repo2docker/utils.py

yuvipanda:
The following functions have unit tests added:

- `execute_cmd`
- `chdir`
- `ByteSpecification.validate`

`maybe_cleanup` has been replaced with a `try...finally` in
its only place of use, which IMO is clearer.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -14,6 +14,7 @@ import logging
 import os
 import pwd
 import subprocess
+import shutil
 import tempfile
 import time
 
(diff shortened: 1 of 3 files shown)


yuvipanda:
I picked this out by looking through https://codecov.io/gh/jupyter/repo2docker/tree/master/repo2docker and finding files that need more coverage. Help, I've fallen into a trap set by @betatim!

betatim:
I added two commits to move the `shutil` import around. Interestingly that missing import only got noticed by one of our many tests, those that build external repos.

betatim:
screen shot 2018-12-11 at 13 04 57

Apparently something needs updating in the codecov settings as it thinks the target for the diff is 0%.

PR #496, opened 2018-12-11
Move argument parsing out of Repo2Docker class

yuvipanda:
Argument parsing should only be used when calling from
the command line, and not be deeply tied into the class
itself. This makes it easier for folks to just set
traitlets on an empty class and start it, without having to
deal with passing arguments.

This breaks how people might already be using Repo2Docker as
a library, so should not be part of 0.7


tests/test_args.py
```diff
@@ -0,0 +1,64 @@
...
+from repo2docker import __version__
+from contextlib import redirect_stdout
+import io
+
+def test_version():
+    """
+    Test passing '--version' to repo2docker
+    """
+    stdout = io.StringIO()
+    with redirect_stdout(stdout):
```
betatim: how does `redirect_stdout` compare to https://docs.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function? I've used the latter before I think.

yuvipanda: @betatim oo, that's pretty nice. I've switched to using that.

PR #497, opened 2018-12-12
Swap codecov targets around

betatim:
Third time lucky?

This swaps the targets around, we want a minimal coverage of a diff/PR of 20% and for the project as a whole we are happy as long as it doesn't decrease (auto).

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,7 +4,7 @@ coverage:
   status:
     project:
       default:
-        target: 20%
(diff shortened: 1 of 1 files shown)


yuvipanda:
@betatim I think the lesser coverage is because https://github.com/jupyter/repo2docker/pull/475 was merged, making the... code smaller? Unclear, but feels like a bug in codecov.

Feel free to merge whenever :)

PR #499, opened 2018-12-12
push tagged images on travis

minrk:
creates jupyter/repo2docker:0.7.0 when we release 0.7.0

we have to manually tag images for past releases

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ jobs:
       provider: script
       script: "./travis/travis-script.bash"
       on:
+        tags: true
         branch: master
 env:
   matrix:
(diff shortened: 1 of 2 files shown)


betatim:
@minrk could you add betatim@gmail.com as a docker hub account that can push to the repo2docker repository there?

betatim:
That worked. I tagged 0.6.0 and 0.7.0. For 0.5.0 I tried `docker pull jupyter/repo2docker:eca6f71` but that image doesn't exist :-/ Before I go looking for which existing tag is closest to this: is it worth doing this for old releases like 0.5 and before? I'd say no.

minrk:
0.6.0 seems far back enough, since this hasn't come up before

PR #500, opened 2018-12-13
Add repo2docker Dockerfile labels

### LINKED ISSUE #466: Add labels for repo2docker version and repo
it would be handy to add the labels for repo2docker version, repo URL, and ref to images when we build them:

```docker
LABEL repo2docker.version="x.y.z"
LABEL repo2docker.repo="https://..."
LABEL repo2docker.ref="abc123"
```

This records in docker image metadata info about what was built, which could be useful in the future.

jrbourbeau:
This PR adds repo2docker version, repo URL, and ref labels to the generated Dockerfile

Closes #466

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -110,7 +110,7 @@ Released 2017-04-14
 .. _@costrouc: https://github.com/costrouc
 .. _@dsludwig: https://github.com/dsludwig
 .. _@evertrol: https://github.com/evertrol
+.. _@jrbourbeau: https://github.com/jrbourbeau
 .. _@minrk: https://github.com/minrk
 .. _@nuest: https://github.com/nuest
 .. _@yuvipanda: https://github.com/yuvipanda
-.. _@jrbourbeau: https://github.com/jrbourbeau
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Release date: the-future
 
 New features
 ------------
+- Add additional metadata to docker images about how they were built `#500`_ by
+  `@jrbourbeau`_.
 
 API changes
 -----------
@@ -99,6 +101,7 @@ Released 2017-04-14
 .. _#475: https://github.com/jupyter/repo2docker/pull/475
 .. _#478: https://github.com/jupyter/repo2docker/pull/478
 .. _#482: https://github.com/jupyter/repo2docker/pull/482
+.. _#500: https://github.com/jupyter/repo2docker/pull/500
 
 .. _@agahkarakuzu: https://github.com/agahkarakuzu
 .. _@betatim: https://github.com/betatim
@@ -110,3 +113,4 @@ Released 2017-04-14
 .. _@minrk: https://github.com/minrk
 .. _@nuest: https://github.com/nuest
 .. _@yuvipanda: https://github.com/yuvipanda
+.. _@jrbourbeau: https://github.com/jrbourbeau
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -120,8 +120,8 @@ ENV {{item[0]}} {{item[1]}}
 # Container image Labels!
 # Put these at the end, since we don't want to rebuild everything
 # when these change! Did I mention I hate Dockerfile cache semantics?
-{% for k, v in labels.items() -%}
-LABEL {{k}}={{v}}
+{% for k, v in labels.items() %}
+LABEL {{k}}="{{v}}"
 {%- endfor %}
 
 # We always want containers to run as non-root
@@ -246,7 +246,7 @@ class BuildPack:
         """
         Docker labels to set on the built image.
         """
-        return {}
+        return getattr(self, 'labels', {})
 
     def get_build_script_files(self):
         """
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -171,6 +171,7 @@ class BuildPack:
     def __init__(self):
         self.log = logging.getLogger('repo2docker')
         self.appendix = ''
+        self.labels = {}
         if sys.platform.startswith('win'):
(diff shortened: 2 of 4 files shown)


jrbourbeau:
Just following up on https://github.com/jupyter/repo2docker/issues/466#issuecomment-436314202, should `repo` be considered private and not included as additional metadata?

betatim:
Thanks for this nice PR!

I think `repo` shouldn't be private. Or put differently, if we do want to consider it private we'd also have to remove it from the information in `.git` and other places where it might leak.

One thing we are currently trying to improve for repo2docker is the time taken to execute the tests. This is making me think: can we avoid having to build the image and instead only generate the `Dockerfile` commands and/or inspect the return value of the function that injects the labels? Maybe a simple unittest here together with a check in an already existing test that builds an image (maybe in `external/`?) that the labels got set correctly would work? What do you think?

CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Release date: the-future
 
 New features
 ------------
+- Add additional metadata to docker images about how they were built `#500`_ by
+  `@jrbourbeau`_.
```
betatim: Whoop! 👏 for helping maintain the change log!

CHANGES.rst
```diff
@@ -110,3 +113,4 @@ Released 2017-04-14
 .. _@minrk: https://github.com/minrk
 .. _@nuest: https://github.com/nuest
 .. _@yuvipanda: https://github.com/yuvipanda
+.. _@jrbourbeau: https://github.com/jrbourbeau
```
betatim: Tried to sort user names alphabetically to make it easy to check if someone is already in the list, can you move yourself up?

repo2docker/buildpacks/base.py
```diff
@@ -246,7 +246,7 @@ def get_labels(self):
         """
         Docker labels to set on the built image.
         """
-        return {}
+        return getattr(self, 'labels', {})
```
betatim: `getattr` makes my Python senses tingle :-/

Can we change this to `return self.labels` and set `self.labels = {}` in `__init__` instead?

tests/test_labels.py
```diff
@@ -0,0 +1,37 @@
...
+import pytest
+
+
+@pytest.mark.parametrize('ref', ['some-branch', None])
+def test_labels(ref, tmpdir):
+    app = Repo2Docker()
+    if ref is not None:
+        argv = ['--ref', ref, str(tmpdir)]
+    else:
+        argv = [str(tmpdir)]
```
betatim: For my education, what does `str(tmpdir)` do or asked differently what is `tmpdir` that we need to call `str` on it?

minrk: from [the docs](https://docs.pytest.org/en/2.8.7/tmpdir.html):

> tmpdir is a [py.path.local](
http://py.rtfd.org/en/latest/path.html) object which offers os.path methods and more.

minrk:
re: private repo, it might be appropriate to use a single `local` string when building from a local directory instead of the path, which leaks info about the build machine's directory structure. I think it should be included if it's a real repo url, though.

jrbourbeau:
> can we avoid having to build the image and instead only generate the Dockerfile commands and/or inspect the return value of the function that injects the labels?

@betatim I think this would be a nice improvement! I noticed the picked `BuildPack` for a `Repo2Docker` instance only appears in the `Repo2Docker.start()` method. I could be missing something, but I didn't see a way to retrieve the `picked_buildpack` instance to use for testing. Would it be reasonable to add a something like 

```python
self._picked_buildpack = picked_buildpack
```
to the `Repo2Docker` class? That way we would have a reference to the corresponding `BuildPack` used that could be used for testing (and potentially other purposes too).

betatim:
I'm not quite sure yet how to do it, below a first draft/thinking out loud:

We call [`BuildPack.render()`](https://github.com/jupyter/repo2docker/blob/eb272875b1daa5aa3e52b4760cf3affed01ed359/repo2docker/app.py#L749) here. This outputs the `Dockerfile` that will be used to build the image. The [`render()`](https://github.com/jupyter/repo2docker/blob/eb272875b1daa5aa3e52b4760cf3affed01ed359/repo2docker/buildpacks/base.py#L411) call itself uses jinja2 to create the file. Can we [use a mock](https://docs.pytest.org/en/latest/monkeypatch.html) to check that the jinja template was called with the right arguments (aka the labels are passed to the jinja template render call correctly? Then we don't have to call `start_container()` and can add `--no-build` to the `argv`s we set.

Overall we'd do something along the lines of `repo2docker --debug --no-buid https://example.com/here/repo | grep "LABEL correct-label-here"` but smarter and more automatic :)

jrbourbeau:
Adding a `_picked_buildpack` attribute removes the need to call `start_container()` because we can inspect the buildpack labels directly via something like `app._picked_buildpack.labels`. However, adding extra state to the `Repo2Docker` class, like a `_picked_buildpack` attribute, may be something we want to avoid, I'm not sure.

I pushed some commits that add a `_picked_buildpack` attribute and update the tests accordingly (note that images are not built in the updated tests). I can totally revert back adding `_picked_buildpack` if we want to go another direction, just thought it would be useful to have a clear example of what I was proposing.

betatim:
I'm not a huge fan of adding the attribute just for testing purposes. What do you think of the approach with a mock?

The other way to keep the attribute is if it is useful for refactoring these methods (that are far too big) which @yuvipanda is working on.

jrbourbeau:
> I'm not a huge fan of adding the attribute just for testing purposes

Agreed. After updating the test to use a `Mock`, I think it's a cleaner solution.

betatim:
Congratulations on getting your first contribution to repo2docker merged (and with such a cool PR number 🤓).

Thanks for the nice PR content and interaction!

jrbourbeau:
Thanks for the feedback and reviewing @betatim!

PR #501, opened 2018-12-13
fix push-tags condition

minrk:
it should build master and tags on this repo. #499 wasn't quite right.

`on` conditions only fire when they are *all* met, which was it's a tag *and* it's the master branch, which is never true. Use `if` instead for a proper boolean condition.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,15 +33,15 @@ jobs:
   include:
   - stage: deploy
     script: skip
-    if: type = push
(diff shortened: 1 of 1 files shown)


betatim:
Don't we tag master which would mean "tag and branch=master" is true?

(restarted failed travis build)

minrk:
tags aren't associated with a branch. The ref is either a tag or a branch, never both. While we may make tags starting from a branch, that's a local workflow implementation detail, not related to git's internal understanding.

PR #502, opened 2018-12-13
Make sure our unit tests are all run on travis

yuvipanda:
The glob used meant we only ran .py files two levels deep,
rather than one level deep. We un-nest the contentprovider
tests and simplify the glob


.travis.yml
```diff
@@ -54,7 +54,7 @@ env:
   - REPO_TYPE=nix
   - REPO_TYPE=dockerfile
   - REPO_TYPE=external/*
-  - REPO_TYPE=**/*.py
```
minrk: A simpler fix for me would be to move all the unittests into a `tests/unit` directory, then here would be `REPO_TYPE=unit`

betatim: This would allow us to organise the unit tests into sub-directories again?

yuvipanda:
Superceeded by #512

PR #507, opened 2018-12-16
Copy repo to ${REPO_DIR} rather than ${HOME}

yuvipanda:
We want to put the repo somewhere other than ${HOME}
so we can mount persistent storage in ${HOME} more easily.
Most of repo2docker assumes current directory is where
the contents are, so we should be able to get most of it
working by setting WORKDIR to ${REPO_PATH} and letting
${REPO_PATH} be configurable.

Unclear what to do for plain Dockerfiles, Legacy Dockerfiles &
nix.

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -169,6 +169,11 @@ def get_argparser():
     )
 
     argparser.add_argument(
+        '--repo-path',
+        help=Repo2Docker.repo_path.help
+    )
+
+    argparser.add_argument(
         '--appendix',
         type=str,
         #help=self.traits()['appendix'].help,
@@ -315,6 +320,9 @@ def make_r2d(argv=None):
         r2d.cleanup_checkout = False
     else:
         r2d.cleanup_checkout = args.clean
+
+    if args.repo_path:
+        r2d.repo_path = args.repo_path
 
     return r2d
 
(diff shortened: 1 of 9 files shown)


yuvipanda:
We need to basically test all our integration tests with REPO_PATH set to something other than ${HOME}. That might make our test suite much slower, unfortunately. Maybe we could get away with a subset?

betatim:
How about `REPO_DESTINATION`?

We added `~/.local/bin` to PATH explicitly to let people put things on the PATH from their repositories or from `postBuild`. I think we need to maintain that behaviour so I'd go for adding both home and the repositories `.local/bin` to the PATH.

minrk:
> Should we call it REPO_PATH or REPO_DIR?

A convention I like is PATH vars are lists of dirs (like $PATH) and DIR for single directories

minrk:
Looks excellent to me now, thanks @yuvipanda! Do we want to place any restrictions on REPO_DIR? I could imagine someone doing interesting things with REPO_DIR=/etc. But I guess in cases where you care, it's the service operator who controls REPO_DIR, right? Not users?

yuvipanda:
@minrk yep - admins control it so we should be fine. We gotta be careful if / when we let users explicitly set environment values. But I think that is unlikely since we have `start`

PR #509, opened 2018-12-16
Fix cache-busting when running tests locally

### LINKED ISSUE #508: Copying environment.frozen.yml seems to bust cache
When running tests locally, I keep seeing:

```
Step 25/38 : COPY conda/install-miniconda.bash /tmp/install-miniconda.bash
 ---> Using cache
 ---> 5d2487a23620
Step 26/38 : COPY conda/environment.frozen.yml /tmp/environment.yml
 ---> cac96dfa4930
Step 27/38 : RUN bash /tmp/install-miniconda.bash && rm /tmp/install-miniconda.bash /tmp/environment.yml
 ---> Running in df08f8e9c1bc
```

Note the lack of 'Using Cache' when copying environment.frozen.yml to /tmp/environment.yml. This means we're rebuilding the conda environment for every build from scratch, and that's slooowwwwww.

This might just be a problem on my computer, though.

yuvipanda:
When you run repo2docker locally, the uid is usually set
to the current user's uid. However, when we push the content
to Docker, we invariably set the uid of the scripts we
use as part of repo2docker at 1000. This causes a cache bust
every time.

This patch uses the correct uid, and fixes cache busting. Makes
local test runs *much* faster.

Fixes #508

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -469,8 +469,8 @@ class BuildPack:
             # https://github.com/docker/docker-py/pull/1582 is related
             tar.uname = ''
             tar.gname = ''
-            tar.uid = 1000
-            tar.gid = 1000
+            tar.uid = int(build_args.get('NB_UID', 1000))
+            tar.gid = int(build_args.get('NB_UID', 1000))
             return tar
 
         for src in sorted(self.get_build_script_files()):


betatim:
For my education: if we always copy things with the same UID (always 1000) no matter what the local user's ID is, why does it cause the caching to break? My guess is that we change the mtime (or atime?) for the file by changing the UID and this is what causes the cache to break.

yuvipanda:
@betatim it's a good question I don't have a clear answer to! :)

PR #510, opened 2018-12-16
Run tests concurrently with pytest-xdist

yuvipanda:
Should help us discover any race conditions in our
tests that prevent them from running concurrently.

Should also speed up tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ install:
 # Make a wheel and install it to test to catch possible
 # issues with releases
 - pip install --upgrade setuptools pip
-- pip install -r dev-requirements.txt
+- pip install --upgrade -r dev-requirements.txt
 - python setup.py bdist_wheel
 - pip install dist/*.whl
 script:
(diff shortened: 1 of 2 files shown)


yuvipanda:
Let's see if test are actually sped up

betatim:
I'm not a huge fan of running tests in parallel because often I end up struggling to tell which test exactly failed, why and what is its stdout. Maybe there are some tricks/hints on getting better at that which we should share somewhere.

I can't really tell you why or what it is that confuses me. Maybe the fact that tests are failing when I think they should be passing (already something I don't understand is happening) combined with things running in parallel?

Right now it seems like things don't go faster though :-/ Could that be because we don't spend an awful lot of time actually waiting but most of it using the CPU?

yuvipanda:
In the past I've found that xdist does a reasonably good deal of pointing out exactly which tests failed, but I've never run it on anything this complex.

I was hoping this would be an easy win, but unfortunately it clearly isn't :(

Travis-CI has two CPUs, so we *should* be doing ok mostly even if we're doing things with CPU. I suspect it has something to do with docker layer cache, but that's just a vague hunch.

I think we should just shelve this idea for now.

minrk:
> I think we should just shelve this idea for now.

I think so. xdlist can make failures harder to debug, and it can also make certain build tests *slower*  because it basically ensures that the cache is not re-used by running builds concurrently that may have the same first few stages.

betatim:
Let's close this for the moment and revisit at some point.

PR #511, opened 2018-12-17
[MRG] Add caching of already built repositories

betatim:
This PR is a revival of #461

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -13,7 +13,6 @@ import sys
 import logging
 import os
 import pwd
-import subprocess
 import shutil
 import tempfile
 import time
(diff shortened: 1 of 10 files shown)


betatim:
Should we try again with this @yuvipanda? :)

repo2docker/app.py
```diff
@@ -535,15 +531,78 @@ def _get_free_port(self):
...
+        # check if we already have an image for this content
+        client = docker.APIClient(version='auto', **kwargs_from_env())
+        for image in client.images():
+            if image['RepoTags'] is not None:
+                for tag in image['RepoTags']:
+                    if tag == self.output_image_spec + ":latest":
+                        return True
+        return False
+
+    def _build_image(self, checkout_path):
```
yuvipanda: I think this can be folded into the build function

betatim: I liked the separation because `build()` does a bunch of things before&after building, and `_build_image()` just builds the image. :-/ It is an attempt to make the functions a bit shorter. Happy to move it.

betatim: Private conversation with @yuvipanda just now: we put everything back in one method and he is happy with that.

Maybe look at splitting at a later point.

PR #514, opened 2018-12-17
test freezing

minrk:
notebook is bumped to 5.7.4 in #519 

just a smoke test to verify that it runs without error, for now


betatim:
I like the idea of running the freeze in travis as often as possible. How can we make it not be the blocker (mark the freeze travis job as optional?)

What is the action to take based on the output? Should we build it out to alert us (by failing) if a major version of a dependency changes (without us changing anything)?

Not sure where I'd like this to go. More automation and testing is better but I am already unhappy with how long our tests take :-/

minrk:
As it is now, it's just a smoke test checking that the freeze runs without error. I've marked it as an allowed failure for now as well.

Things we could check:

1. that required packages are installed in the final env (i.e. that notebook 5.7.4 is installed, which would catch the recent issue you found with mixing conda+pip).
2. whether there are new packages that might prompt a re-freeze (this may be useful, but definitely shouldn't be treated as a test failure)

PR #516, opened 2018-12-17
Try see if we're running test twice

yuvipanda:
I *think* we run each test twice - once to see
if notebook starts, and once to run ./verify. This 
seems fair enough, but am trying to get travis to tell
me if this is indeed the case.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -104,13 +104,6 @@ class Repo2DockerTest(pytest.Function):
 class LocalRepo(pytest.File):
     def collect(self):
         yield Repo2DockerTest(
-            'build', self,
-            args=[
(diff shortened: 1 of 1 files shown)


betatim:
The first run tries to build the image and the second tries to build+verify. This makes it easier to debug where the mistake is when things are broken.

The second build should be "free" as we should be reusing the built image.

betatim:
The first run tries to build the image and the second tries to build+verify. This makes it easier to debug where the mistake is when things are broken.

The second build should be "free" as we should be reusing the built image.

betatim:
https://travis-ci.org/jupyter/repo2docker/jobs/469294997#L733-L743 confirms that we run things twice and that the verify round is more or less free.

PR #517, opened 2018-12-18
Provide help text for commandline arguments

yuvipanda:


repo2docker/__main__.py
```diff
@@ -169,13 +169,13 @@ def get_argparser():
...
         type=str,
-        #help=self.traits()['appendix'].help,
+        help=Repo2Docker.appendix.help
     )
 
     argparser.add_argument(
         '--subdir',
         type=str,
-        #help=self.traits()['subdir'].help,
+        help=Repo2Docker.subdir.help
```
jrbourbeau: The current help message for `subdir` is 

> Subdirectory of the git repository to examine. Defaults to ‘’.

Perhaps something like "Defaults to the root of the repository" is slightly more clear. What do you think?

PR #518, opened 2018-12-18
Update RTD to install repo2docker

jrbourbeau:
I saw the current readthedocs [build is failing](https://readthedocs.org/projects/repo2docker/builds/8289685/). Looks like it's due to the changes made in #515. The issue is that `repo2docker` isn't installed in the readthedocs virtualenv where the documentation is being built, so we're getting an `ImportError` when trying to build the command line API documentation:

```
  File "/home/docs/checkouts/readthedocs.org/user_builds/repo2docker/envs/latest/lib/python3.5/site-packages/sphinxcontrib/autoprogram.py", line 120, in import_object
    mod = __import__(module_name)
ImportError: No module named 'repo2docker'
```

After reading through the readthedocs YAML configuration docs, I think we need to set `setup_py_install` to be `true` in `readthedocs.yml` (ref https://docs.readthedocs.io/en/latest/yaml-config.html#python-setup-py-install) to install `repo2docker` and fix the failing build. That's what is done in this PR.

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -2,4 +2,5 @@ name: repo2docker
 type: sphinx
 python:
   version: 3
+  setup_py_install: true
 requirements_file: docs/doc-requirements.txt


betatim:
Let's find out :)

betatim:
I think we didn't catch this because in CircleCI we explicitly run `pip install -e .` to install repo2docker. We keep getting caught out by our RTD build failing despite having the CircleCI setup. I wonder if there is anything we can do to more easily/faithfully reproduce the RTD build environment and process. At the moment we are slowly iterating towards it but someone must have already solved this problem before us :)

jrbourbeau:
Looks like we fixed the `ImportError` : )

But now there's a `SyntaxError` because of some f-strings : /
(see https://readthedocs.org/projects/repo2docker/builds/8291787/)

I opened up #520 to fix the f-string issue

PR #519, opened 2018-12-18
refreeze with notebook 5.7.4

minrk:
this is just the bump in #514 without the test


betatim:
Merging despite CodeCov being unhappy.

PR #520, opened 2018-12-18
Remove f-strings

jrbourbeau:
This PR removes f-strings from `repo2docker`

There are a few f-strings in the codebase that are causing the RTD build to fail because the docs are currently built with Python 3.5, while f-strings weren't added until 3.6 (ref https://readthedocs.org/projects/repo2docker/builds/8291787/). I was able to reproduce the f-string related `SyntaxError` when building the docs locally on `master` in a Python 3.5 environment. With the changes in this PR, the docs were successfully built locally with Python 3.5.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -426,7 +426,8 @@ class Repo2Docker(Application):
                 self.log.info('Pushing image\n',
                               extra=dict(progress=layers, phase='pushing'))
                 last_emit_time = time.time()
-        self.log.info(f'Successfully pushed {self.output_image_spec}', extra=dict(phase='pushing'))
+        self.log.info('Successfully pushed {}'.format(self.output_image_spec),
+                      extra=dict(phase='pushing'))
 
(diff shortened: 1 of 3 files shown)


betatim:
In the `README.md` we write that repo2docker supports Python 3.4+. We should update that to reflect what we actually support and then add a test to check we can run on that version.

[JupyterHub supports 3.5+](https://github.com/jupyterhub/jupyterhub#check-prerequisites), in the absence of strong opinions should we go with that?

(I don't remember if we have discussed this before or not. If yes please link us to that so we don't have to rediscuss this.)

minrk:
Our travis.yml looks like it's trying to add a test entry for Python 3.4. We should check why that's not getting included.

yuvipanda:
Thanks for the PR, @jrbourbeau! I opened #521 to talk about supported python versions.

jrbourbeau:
FYI, the RTD build is passing now

Issue #522, opened 2018-12-18
conda installing python twice?


yuvipanda:
When running some tests, I see: 

```
    python:               3.7.0-hc3d631a_0       defaults    --> 3.6.7-h5001a0f_1       conda-forge
```

In the build logs a bit. Does this mean conda is installing 3.7 first and then downgrading it to 3.6, thus building it twice?

betatim:
I think Min pointed out in a different issue that if you list `python` in your `environment.yml` and there is an update (or downgrade?) available it will take it. Whereas if you don't ever mention python in your `environment.yml` the chances of it not being changed increase.

Can you link us to which test does this? It being downgraded does seem weird. Though we might have tests that ask for a specific python version to test that that works. Shouldn't be the general approach though.

yuvipanda:
This particular one was from `tests/venv/usr-libs` which doesn't ask for a specific python version.

betatim:
What is the default version of Python in a new conda environment?

In https://github.com/jupyter/repo2docker/blob/bdd32b292a226bbc6e88e771bb3c77f0819ef4cb/repo2docker/buildpacks/conda/__init__.py#L78-L81 we set the default to 3.6

minrk:
Since it's defaults -> conda-forge, I'm guessing this is the initial Python installation. The latest miniconda itself ships with Python 3.7, so to set up the default Python 3.6 env, it's going to install Python 3.6. So while miniconda bootstraps itself with *a* Python, the initial installation is going to ~always replace it.

One thing we could do that should make setting up the base env *hugely* quicker would be to package our base environments with conda constructor or conda pack. Then instead of the miniconda bootstrap followed by an immediate install that replaces most of the packages in miniconda (mostly due to the defaults->conda-forge switch, even if versions match), we could be doing the rather quick extract-in-place bootstrap install and stop there (or proceed with a pip install for final touches).

The only tricky part of that, I think, is where to store and fetch the installers. This isn't hard for *releases*, but it's trickier if we want to publish the archives for every commit of repo2docker.

PR #523, opened 2018-12-18
Use conda only to install python

yuvipanda:
Use pip for everything else.

Testing to see if:

- This is faster
- Prevents python downgrades and what not (maybe related to
  https://github.com/jupyter/repo2docker/issues/522)


minrk:
While this may save time on the initial base install, installing base packages with pip means that any conda install by users is much more likely to get things confused and ultimately slower because a conda dependency on packages installed by pip is going to result in installing those packages again from conda (conda tells pip about conda packages, but pip doesn't tell conda about pip packages).

I also don't believe this will affect #522, which I think is the fact that we bootstrap miniconda and then install the requested Python from conda-forge afterward, which is unchanged here.

I think if we want to expedite the initial install, the right thing to do is create our own conda constructor (or more likely conda pack) for the base envs so we don't have to bootstrap miniconda and then setup the base env with a `conda env update`. That should be the quickest and most efficient, since it's a single download & extract without any communication with a package index (either conda or pypi). The main thing we have to figure out is how/where to distribute the installers.

Issue #524, opened 2018-12-18
Python 3.4 build matrix entry


betatim:
In https://github.com/jupyter/repo2docker/pull/520#issuecomment-448284175 Min mentions that we seem to have an entry for Python 3.4 in our build matrix but that it doesn't run.

This issue is about figuring out why and how to fix that.

This is also connected to #521 where we will decide the minimum Python version to support. This will influence what version this extra build matrix entry should be using (could be 3.5 for example instead of 3.4)

PR #525, opened 2018-12-18
Fix Travis Python 3.4 build matrix entry

jrbourbeau:
In this PR, I'm experimenting with the `.travis.yml` configuration file to determine why the Python 3.4 build isn't running 

Ref #524

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ sudo: required
 cache: pip
 services:
 - docker
-python:
-- 3.6
 install:
 # Make a wheel and install it to test to catch possible
 # issues with releases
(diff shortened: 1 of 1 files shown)


jrbourbeau:
Hmmm looks like Travis didn't run at all when explicitly specifying the full build matrix

minrk:
This is actually fixed by the mostly unrelated #527. Apparently using `jobs.include` supersedes `matrix.include`, and that PR removes our use of `jobs.include`.

jrbourbeau:
Awesome! Thanks for the update @minrk 

> using jobs.include supersedes matrix.include

Good to learn something new about Travis CI!

PR #527, opened 2018-12-19
Run docker builds on docker hub

### LINKED ISSUE #524: Python 3.4 build matrix entry

minrk:
Doesn't need docker credentials on travis. If it's set up right, this PR should show a CI status coming from docker cloud.

Repo changes:

- remove docker deploy step from travis
- replace with docker's `hooks/post_push` to create the tagged builds
- build docker image with python 3.7 and two-stage builds to save ~80% on size

This also somewhat accidentally closes #524 since that was caused by a conflict between jobs.include and matrix.include.


betatim:
When I click on the "Details" link for the dockercloud check I get sent to https://cloud.docker.com/redirect?resource_uri=/api/audit/v1/action/fac4aa2b-6625-43e2-8fbb-afe0663ac5fe/ which then shows me a 404. Is this because I am not logged in (as a user who has privs to look at the repo2docker build) to docker hub?

I like it!

docker-compose.test.yml
```diff
@@ -0,0 +1,3 @@
+sut:
```
betatim: What is "sut" and when do we use this file? I guess it somehow finds use in the testing of the image on docker hub?

PR #529, opened 2018-12-20
WIP: conda-pack base environments

### LINKED ISSUE #522: conda installing python twice?

minrk:
For the initial environment, installation is slow because we have to do it in stages:

1. install miniconda
2. install chosen python (never the same as miniconda)
3. install base environment (conda packages)
4. install remaining environment (pip packages)

This involves lots of talking to upstream repos and redundancies (e.g. miniconda comes with a Python that we immediately replace, as noted in #522).

This PR adds to our freeze step `conda-pack`, which doesn't just freeze the package specifications, but the package contents themselves. This lets us replace the miniconda install with our own bootstrap of the base env.

Running this on `tests/venv/default` (with docker build cache disabled) takes 2:47 on master and 1:04 on this branch. It might be a little slower if the pack came from a download instead of a local file, since it's a pretty big archive.

The things still to figure out:

- how do we distribute the packs? Is it purely a local optimization for repeated builds or do we require them strictly? If we do it as a runtime cache, including them in the repo2docker image would be easy, but it also means our tests would have a hard time benefiting from the speedup. It also means there's two code paths for initial installation, depending on whether the cache is present or not.
- Is this a good idea at all? Distributing our own binary bundles rather than specs isn't the most in-line with reproducibility. However, it is just zipping up the result of an install step of a strictly frozen env. I'd feel more sanguine about the local cache approach for this reason, but that also means it's going to benefit fewer folks (i.e. mybinder.org users, but few home users).

closes #522

### DIFF
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -97,6 +97,8 @@ class CondaBuildPack(BaseImage):
         """
         files = {
             'conda/install-miniconda.bash': '/tmp/install-miniconda.bash',
+            # FIXME: hardcoded test
+            'conda/environment.py-3.6.tar.gz': '/tmp/environment.tar.gz',
         }
         py_version = self.python_version
         self.log.info("Building conda environment for python=%s" % py_version)
--- a/repo2docker/buildpacks/conda/freeze.py
+++ b/repo2docker/buildpacks/conda/freeze.py
@@ -10,6 +10,7 @@ python freeze.py [3.5]
 """
 
 from datetime import datetime
+from distutils.version import LooseVersion as V
 import os
 import pathlib
 import shutil
@@ -23,6 +24,7 @@ from ruamel.yaml import YAML
 # since miniconda3 docker images seem to lag conda releases.
 MINICONDA_DOCKER_VERSION = '4.5.11'
 CONDA_VERSION = '4.5.11'
+CONDA_PACK_VERSION = '0.3.1'
 
 HERE = pathlib.Path(os.path.dirname(os.path.abspath(__file__)))
 
@@ -35,7 +37,7 @@ FROZEN_FILE_T = os.path.splitext(ENV_FILE_T)[0] + '.frozen.yml'
 yaml = YAML(typ='rt')
 
 
-def freeze(env_file, frozen_file):
+def freeze(env_file, frozen_file, root=True):
     """Freeze a conda environment.yml
 
     By running in docker:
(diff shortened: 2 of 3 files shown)


betatim:
>  I'd feel more sanguine about the local cache approach for this reason, but that also means it's going to benefit fewer folks (i.e. mybinder.org users, but few home users).

If you repeatedly run repo2docker (locally or on mybinder.org) these early steps get cached already by the docker build cache. At least I just tried to build a repo that has a `requirements.txt` on mybinder.org and everything up to `Step 27/39 : COPY src/ ${HOME}` comes from the build cache.

This makes me think if we want to optimise this we should be targeting the "cold start" part where I run repo2docker for the first time.

Slightly off-topic: the conda builds are slow, but the R builds are even slower. We should keep that in mind as it would be ideal if we'd only have to maintain one solution to slow builds insead of one for R and one for Python :-/

minrk:
I think it would also be easy to publish the conda-packs for *releases* of repo2docker. Doing it for every commit seems less feasible, since these are ~150-200 MB each and we'd be making one for each Python version.

I've been thinking that we should publish the pack based on the md5 of the environment.frozen.yml, so the install step is:

1. `md5sum environment.yml`
2. download `pack-repo/$md5.tar.gz`
3. proceed, depending on whether pack download was found:
  - 404: bootstrap miniconda
  - found pack: just unpack

yuvipanda:
I have some sort of really bad feeling about this, but can't quite articulate that. Shall try next week.

minrk:
@yuvipanda no worries! I'm not convinced this is a good idea either, since it significantly ups our distribution responsibilities. It does seem to really benefit impatient users, though. I'm not in any rush to get this resolved, either.

Another possible case: we could freeze the base envs for each version at release time, which would benefit all users who install stable releases and should still shrink the delta for the initial post-bootstrap install for those who run from master (mybinder.org) once we've made our first release with this.

Essentially, all this change does is replace the miniconda bootstrap with a different bootstrapped set of packages. Nothing else changes.

A further option: Since this bootstrap env unconditionally goes in /opt/conda in this docker image, we could use a totally generic tarball instead of conda-pack, since we don't actually need it to be portable in any way (conda-unpack mostly just re-runs the relocateability step to update install prefixes). This is basically what Travis does when bootstrapping its Python environments.

PR #534, opened 2018-12-21
[MRG] Add tests for port mapping conversion

betatim:
Wanted to quickly add some tests for the port mapping function and then ended up rewriting it.

I tried to follow https://docs.docker.com/config/containers/container-networking/#published-ports because I've never used a port mapping more complex than `8888:8888`.

### DIFF
--- a/repo2docker/utils.py
+++ b/repo2docker/utils.py
@@ -88,6 +88,14 @@ def validate_and_generate_port_mapping(port_mappings):
         single container_port to multiple host_ports
         (docker-py supports this but repo2docker does not)
     """
+    def check_port(port):
+        try:
+            int(port)
+        except ValueError as e:
+            raise ValueError('Port specification "{}" has '
+                             'an invalid port.'.format(mapping))
+        return port
+
     def check_port_string(p):
         parts = p.split('/')
         if len(parts) == 2:  # 134/tcp
@@ -99,11 +107,7 @@ def validate_and_generate_port_mapping(port_mappings):
             port = parts[0]
             protocol = 'tcp'
 
-        try:
-            int(port)
-        except ValueError as e:
-            raise ValueError('Port specification "{}" has '
-                             'an invalid port.'.format(mapping))
+        check_port(port)
 
         return '/'.join((port, protocol))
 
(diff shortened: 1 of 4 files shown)


choldgraf:
this seems reasonable to me, and I am always in favor of doing things w/o regex for maintainability reasons...though this also means it's a bit hard for me to know if the new way of doing things is really doing the same as the old regex stuff. That said, tests are passing so I will leave a day or so for this one to marinate, and will plan on merging if nobody protests

betatim:
I don't know why there was a regex either and I can't read regex. The docs were out of date/didn't describe what the function actually did either :-/ I think overall less regex is better.

repo2docker/utils.py
```diff
@@ -68,24 +68,17 @@ def chdir(path):
         os.chdir(old_dir)
 
 
-def validate_and_generate_port_mapping(port_mapping):
+def validate_and_generate_port_mapping(port_mappings):
     """
-    Validate the port mapping list and return a list of validated tuples.
-
-    Each entry in the passed port mapping list will be converted to a
```
yuvipanda: Is there a reason you deleted this docstring?

repo2docker/utils.py
```diff
@@ -94,55 +87,50 @@ def validate_and_generate_port_mapping(port_mapping):
...
-        Valid port mappings are:
-        - `127.0.0.1:90:900`
-        - `:999`  (match to any host port)
-        - `999:999/tcp` (bind 999 host port to 999 tcp container port)
-
-        Invalid port mapping:
-        - `127.0.0.1::999` (even though docker accepts it)
-        - other invalid ip address combinations
     """
-    reg_regex = re.compile(r"""
```
yuvipanda: less regex is good!

repo2docker/utils.py
betatim: I tried to update it as parts weren't describing what actually happened anymore.

Not sure what the best way to describe the returned dictionary is, update the docstring to link to the docker-py method that consumes this port mapping.

yuvipanda:
Thanks, @betatim

PR #536, opened 2018-12-21
Use a minimal hand-crafted Dockerfile for speed

betatim:
Instead of installing the normal stack just install the notebook server
and launch it to check for the launch URL being output correctly.

`Dockerfile` comes from Min's https://github.com/binder-examples/minimal-dockerfile/

### DIFF
--- a/tests/unit/test_connect_url.py
+++ b/tests/unit/test_connect_url.py
@@ -5,15 +5,41 @@ import requests
 import time
 from repo2docker.app import Repo2Docker
 
+
+# Minimal Dockerfile to make build as fast as possible
+DOCKER_FILE = """
+FROM python:3.7-slim
+# install the notebook package
+RUN pip install --no-cache --upgrade pip && \
+    pip install --no-cache notebook
+
+# create user with a home directory
+ARG NB_USER
+ARG NB_UID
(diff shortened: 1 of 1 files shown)


tests/unit/test_connect_url.py
```diff
@@ -5,15 +5,41 @@
...
+ARG NB_UID
+ENV USER ${NB_USER}
+ENV HOME /home/${NB_USER}
+
+RUN adduser --disabled-password \
+    --gecos "Default user" \
+    --uid ${NB_UID} \
+    ${NB_USER}
+WORKDIR ${HOME}
+USER ${NB_USER}
```
betatim: This is the only difference to [Min's original minimal Dockerfile](https://github.com/binder-examples/minimal-dockerfile/blob/master/Dockerfile). Without it the notebook complains that it is being executed as root.

choldgraf:
to be clear, this test is just checking that it can properly connect to the running server, not anything about buildpacks, right? if so this seems like a good idea to me

betatim:
Correct. It checks if the URL printed in the output is what we think it should be and that it can be used to connect to the running server.

yuvipanda:
Thanks, @betatim

PR #537, opened 2018-12-21
Remove ipyleaflet dependency from test

betatim:
Not sure it adds anything to the test except more stuff to install which makes the test slower.

### DIFF
--- a/tests/venv/start/postBuild/postBuild
+++ b/tests/venv/start/postBuild/postBuild
@@ -2,5 +2,3 @@
 
 # this value should not be visible in `verify`
 export TEST_START_VAR="var is set by postBuild"
-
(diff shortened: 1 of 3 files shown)


yuvipanda:
I think this is the only place we test that notebook extensions work. These two lines don't do the actual install - that's in requirements.txt. We should probably just keep these.

betatim:
If this is the only place we test this we should keep it. Note though that the `requirements.txt` is deleted in this PR.

yuvipanda:
Ah, fair. I think we should keep it because we aren't testing notebook extensions anywhere else.

betatim:
Closing this then. Maybe in the future we can investigate if there is a notebook extension with no dependencies that we can use in tests to speed up install times.

Issue #533, opened 2018-12-20
Run rstudio directly from repo2docker 


yuvipanda:
If you have an install.R file, we install RStudio into the image. By default, we use nbressionproxy to make this available from JupyterHub / MyBinder.org, but if you are just building images with repo2docker you don't *have* to use Jupyter at all - you can just launch directly into rstudio.

We should document this.

/cc @craig-willis

yuvipanda:
If I run `repo2docker -p 8788:8787 <path-to-repo-with-R-support> /usr/lib/rstudio-server/bin/rserver` on my local machine, I can access rstudio on localhost:8788

psychemedia:
Related to this, is there a general pattern that could be documented for running arbitrary applications under a similar model, eg @betatim's [`nbopenrefineproxy`](https://github.com/betatim/openrefineder).

betatim:
I think the general pattern is that you need to know three bits of information:
1. which port the process will be listening on (8788 in this case)
1. which command to execute to start the process
1. how to integrate these two bits of information in the `repo2docker` command-line

This means that you can use it to start openrefine directly, but you will have to know which port needs forwarding and what the openrefine command is.

craig-willis:
Has there been any discussion of putting the port and command into the generated Dockerfile instead of overriding values during `docker run`?  For the WT use case, if we build an image to run Rstudio, it seems odd to have the default command be jupyter.

psychemedia:
If the port is aliased, by a proxy path, you presumably don't need to know the port number?

Some services might allow the port on which they are run to be specified, in which case it would make sense to have a standard recipe / convention for handling that, eg a standardised way/convention for passing a port number in via a docker environment variable via `-e` arguments?

Something that struck me in context of the OpenRefine service, which was started on a specified port that was dynamically allocated, was that it would be useful to have the port number available via introspection eg within a notebook kernel, eg by writing it to a config file in a standard location.

By the by, one other thing I noticed running `repo2docker` on a mac was that the default HOME user inside the built container was the same user as the account I ran `repo2docker` with on host, whereas I'd expected it yo be `jovyan` for no very good reason. This leaks information into a Docker image for an unwary user who doesn't run `repo2docker` eg with `--user jovyan` set.

yuvipanda:
Adding aliases is easy now with https://jupyter-server-proxy.readthedocs.io/

betatim:
[...]

Issue #521, opened 2018-12-18
Determine python version support


yuvipanda:
Split off from #520. What versions of python do we support for running repo2docker? Note that this is different from what version of python repositories being built with repo2docker can request.

yuvipanda:
My personal instinct is to go with whatever the latest LTS of Ubuntu ships with - in this case, that's 3.6. But I don't have particularly strong opinions here, since we don't use the biggest improvement of 3.5 (async) here.

betatim:
I would be conservative in what we require. A bit like a while back when projects were still supporting Python 2. You'd not try to break it "on purpose" by exploiting lots of new features in Python 3 and when you did introduce something that didn't work in Python 2 there was an attempt to change the code.

The conservatism is mostly motivated by the fact that we target scientific users who are often stuck with linux systems which are ["quite old"](https://en.wikipedia.org/wiki/Scientific_Linux). As well as not under the users control. For example with f-strings: I like using them but the code will be run by many more people than those who type it -> I type out `.format(...)`.

---

One thing to look forward to when we can pin repo2docker versions #490 is that r2d will run in a container so we can use whatever Python we want (yay!) as long as we have a very small "init" script that runs on most Pythons. I have hope that we make progress on this in the near future which also makes me think we should be conservative (so as to not lose users before we reach version nirvana).

minrk:
I think part of the point of LTS versions is that they are supported for a long time even if there's a new one, so I don't think dropping support the moment a new LTS is released is quite appropriate if LTS is our reference. If we do go by Ubuntu LTS as a reference, it should probably be the *oldest* still-supported LTS, or something like "last two LTS versions" since there are always three supported, I think. 14.04 is still supported until April of next year, and it has 3.5. 16.04 is supported for another two years and also maxes out at 3.5.

On the other hand, since repo2docker is a docker tool and can itself be run in docker, I think we could also choose to be more aggressive and point documentation to running repo2docker itself in docker, since using r2d already requires everything you need to run r2d itself in docker. Then folks who are stuck with old Python (this means they don't have access to backports or conda), they can always run r2d itself via docker:

1. needs to mount docker.sock so the container can create the 'real' containers
2. needs to specify user id and name explicitly, since discovery doesn't work and `--user` at the docker level may not get the right permissions

```bash
docker run --rm -v/var/run/docker.sock:/var/run/docker.sock -it jupyter/repo2docker:0.7.0 \
[...]

PR #539, opened 2018-12-24
Bump default python version to 3.7

yuvipanda:


betatim:
#533 might be relevant and there is another issue (or PR) where we discuss what "LTS" means in terms of the default Python version.

yuvipanda:
@betatim that issue is around RStudio. Did you mean https://github.com/jupyter/repo2docker/issues/521? That mentions LTS.

I think #521 is unrelated, since it's about what version of python the *repo2docker code* should support, rather than the default for *images built with repo2docker*.

betatim:
Yes I think I meant #521. Not sure why I wrote 533. Looking now I might have meant #522 and the discussion about what the default version of Python is that will be installed in a conda environment.

I think the discussion in #521 is related despite it being about the version of Python required to run r2d. We should make some promises about which version of Python ends up in an image by default. I consider the version of the tools we install part of what we version. Similar to a library that versions its API and default arguments etc.

I don't know what our strategy should be but I do think we should decide on a strategy and write it down so that users can rely on it and plan accordingly.

yuvipanda:
Picking back up on this, I'd say the smallest thing we can say is that 'bumping default versions means we bump repo2docker version too'. This also ties back into writing a 'repo2docker spec' that will specify the default language versions (along with how each file is interpreted).

For now, I'd say bumping the version means we've to make a new release. How does that sound, @betatim and @minrk?

minrk:
I think that's sensible. Bumping the default minor version of Python would be a major version upgrade for repo2docker.

yuvipanda:
Updated, let's see if things pass

betatim:
I'm still hesitant to bump the default version of Python. Most research code lags behind by a country mile, people might not have pinned the version of Python they want and we don't yet have support to pin repo2docker versions for a repo. Mostly I am worried about breaking user's repos that used to work but now don't anymore.

Compromise: we add a note/new section/howto in the docs that explains how to pin the version of Python (`runtime.txt` or `environment.yml`) that we can send people to that will come because we broke stuff: look this is how you pin it which is a good idea anyway and will un-break you. WDY?

yuvipanda:
@betatim Sounds like a good idea to me. Where do you think that should live?

I see https://mybinder.readthedocs.io/en/latest/config_files.html#runtime-txt-specifying-runtimes now...

betatim:
I think I'd put it in the repo2docker docs (and link from the binder docs?)

There is already https://repo2docker.readthedocs.io/en/latest/howto/languages.html#specifying-a-version-of-python so maybe adding a FAQ with a title that speaks to people who are grumpy because their stuff just broke like: "How to select the version of Python used?" or "How should I freeze the version of Python used?" or some such that points to that section is enough.

---

We should try and (re)organise our various documentations? I always feel lost in them, especially what is/should be where.

yuvipanda:
Given my current situation, I'm trying to minimize typing where I can. Would be lovely if someone other than me could get the doc bit in. Thanks.

minrk:
Updated some docs on pinning runtime. I think we already had everything being asked for, I just updated the language a bit and added a reference to the default version that's pulled from the code so it'll always be correct.

- FAQ already has how to pin Python as the first two Qs, it is also mentioned in config_files and the Python section of howto/languages. Added some detail about which version is default and links to config_files
- clarification in a few places that runtime.txt is only for use with environment specifications that do not support specifying the runtime (i.e. only with requirements.txt, not environment.yml)
- get default Python version from the code so that it's harder to fall out of date

What I didn't do is add a whole new section on best practices on pinning because I think this is pretty nuanced and not a last-minute thing to do for release (the *worst* thing you can do is strictly pin only some dependencies, e.g. a requirements.txt that looks like):

```
numpy==1.15.3
```

without pinning Python as well, or with other unpinned packages, e.g. just `matplotlib`.

It's better to either:

- loosely specify all requirements, including Python, to ensure you get the most compatible latest combination of everything, or
- strictly specify all requirements and all their dependencies, including Python, to ensure you get a known-good combination frozen in time.

betatim:
Thanks!

yuvipanda:
Thank you very much, @minrk and @betatim!

PR #540, opened 2018-12-24
Ensure git submodules are updated and initialised

alimanfoo:
When obtaining content from a git repository that uses submodules, we need to ensure that the submodules are initialised and updated. It is not enough to just clone with `--recursive` option, because if a ref is also provided, the submodules will need to be updated and potentially initialised *after* checking out the provided ref. This PR adds a command `git submodule update --init --recursive` to resolve this.

I've verified this works with a github repo where the submodules were in a different state in a branch than they were in master. If you'd like some kind of unit test added, happy to add one, but would appreciate some advice on the best approach.

As I've written this, the command `git submodule update --init --recursive` will always be run, regardless of whether a ref was provided. Arguably you could run the command only if a ref was provided as it shouldn't be necessary if no ref was provided, but it seemed harmless and a bit safer to run it always. Happy to change this if you think better to run only if ref is provided.

### DIFF
--- a/repo2docker/contentproviders/git.py
+++ b/repo2docker/contentproviders/git.py
@@ -45,6 +45,12 @@ class Git(ContentProvider):
                                     capture=yield_output):
                 yield line
 
+        # ensure that git submodules are initialised and updated
+        for line in execute_cmd(['git', 'submodule', 'update', '--init', '--recursive'],
+                                cwd=output_dir,
+                                capture=yield_output):
+            yield line
+
         cmd = ['git', 'rev-parse', 'HEAD']
         sha1 = subprocess.Popen(cmd, stdout=subprocess.PIPE, cwd=output_dir)
         self._sha1 = sha1.stdout.read().decode().strip()


choldgraf:
Thanks for noticing this, and for the patch! Is there any downside to doing this? I can't think of anything maybe I am missing something?

alimanfoo:
> Thanks for noticing this, and for the patch! Is there any downside to doing this? I can't think of anything maybe I am missing something?

I don't believe so, should be completely safe, just ensures that submodules are fully synchronised with the state of the parent repo.

betatim:
Some submodules knowledge would be good, like Chris I never use them :-/ What we need (I think) is that if in repoA you add a submodule that refers to repoB you pin the SHA of repoB that you want (I am pretty sure this is what happens). We should check/confirm that updating which SHA of repoB you refer to in repoA also updates/changes the SHA of repoA (like when you edit a file).

A potential pitfall could be that referring to repoB@some-branch-name is a valid submodule but this means building the same SHA of repoA won't always lead to the same image as the some-branch-name could be pointing somewhere else. We have the same problem when people don't specify the version of a dependency so I think this is Ok. Maybe worth documenting as a "not best practice".

We will indeed need a test because otherwise codecov is unhappy :)

I would create a unit test that sets up two temporary git repositories locally: one as repoA and one as repoB which is used as a submodule from repoA. Take a look at https://github.com/jupyter/repo2docker/blob/master/tests/unit/test_connect_url.py for an example how to get a temporary directory and create files in it. This test doesn't create a git repo but I think we have some tests that do. Let me know if you push something and need some more ideas/pointers.

alimanfoo:
> One question... Is it possible to change the submodule without updating the commit hash of the parent repo? It's been a while since I've used submodules

No I don't think so. Whenever you change a submodule, to capture that you also have to make a commit in the parent repo.

The reason this comes up is that whenever you do `git checkout some-ref` in the parent repo, that does **not** automatically update the submodules to the appropriate state, you always have to do a manual ``git submodule update`` to bring the submodules in sync. It's a flaw in git design IMO, trips so many people up.

alimanfoo:
> Some submodules knowledge would be good, like Chris I never use them :-/ What we need (I think) is that if in repoA you add a submodule that refers to repoB you pin the SHA of repoB that you want (I am pretty sure this is what happens). We should check/confirm that updating which SHA of repoB you refer to in repoA also updates/changes the SHA of repoA (like when you edit a file).

I'm no git expert, but I'm pretty sure that's what happens. I.e., you cannot change anything about a submodule without also making a commit to the parent repo.
 
> We will indeed need a test because otherwise codecov is unhappy :)
> 
> I would create a unit test that sets up two temporary git repositories locally: one as repoA and one as repoB which is used as a submodule from repoA. Take a look at https://github.com/jupyter/repo2docker/blob/master/tests/unit/test_connect_url.py for an example how to get a temporary directory and create files in it. This test doesn't create a git repo but I think we have some tests that do. Let me know if you push something and need some more ideas/pointers.

OK, I'll take a look, thanks.

choldgraf:
Cool - then I'm +1 once we have a test for this! Thanks @alimanfoo :-)

betatim:
@alimanfoo do you plan on coming back to this in the next week or so? The last thing that is missing is a test and then this can be merged.

alimanfoo:
Apologies I have got swamped. I'd love to finish this but will need to find
a moment of spare time, given current workload I'm not sure when that will
be. If anyone has bandwidth to work on this please feel free to push to
this PR branch.

choldgraf:
what would the addition need to be?

1. Add a git repo w/ a submodule w/ some code that's required for a test to pass
2. Make sure that this submodule is there when repo2docker builds the image
3. Use the submodule-specific code in the test

that's it, yeah?

yuvipanda:
@choldgraf pretty much.

alimanfoo:
Would also need to make sure that the submodule is only in a branch, not in
master. To verify that repo2docker (1) clones the repo, (2) checks out the
requested branch and (3) updates the submodules. This PR adds step 3.

repo2docker/contentproviders/git.py
```diff
@@ -45,6 +45,12 @@ def fetch(self, spec, output_dir, yield_output=False):
                                     capture=yield_output):
                 yield line
 
+        # ensure that git submodules are initialised and updated
+        for line in execute_cmd(['git', 'submodule', 'update', '--init', '--recursive'],
```
minrk: Instead of adding an extra step here, does it make sense to use `git clone --recursive` as our clone command? I think that's ~always the right thing to do.

alimanfoo:
>
> Instead of adding an extra step here, does it make sense to use git clone --recursive as our clone command? I think that's ~always the right thing to do.
>
> That already happens, but it's not enough unfortunately, because if the
user specifies a branch other than master, then repo2docker does a git
checkout after it does git clone, the checkout command *does not* affect
the submodule. So if the  submodule is in a different state in the
requested branch compared to master, then it will not get checked out to
the right state, unless you also run git submodule update.

The extreme example is if the submodule is not present at all in the master
branch, but is present in the requested branch. (This is what happened to
me, and why I found this issue in the first place.)

djhoese:
This was merged in #639 but since I rebased github hasn't picked it up. So this can be closed.

betatim:
Thanks @alimanfoo for the initial PR and @djhoese for finishing it off!

PR #547, opened 2019-01-08
Fixes Travis build errors related to pytest and Sphinx 

craig-willis:
Proposed fix for https://github.com/jupyter/repo2docker/issues/548.

I'm not sure that this is the right solution here, but pinning the pytest version does work.  It may make sense to pin to a later version.  

Included a change to fix a Sphinx documentation build error apparently caused by recent changelog changes.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -3,7 +3,7 @@
 # update these accordingly
 
 [dev-packages]
-pytest="*"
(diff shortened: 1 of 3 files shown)


yuvipanda:
Thanks, @craig-willis!

Issue #490, opened 2018-12-10
Add support for pinning repo2docker version


betatim:
One thing missing for fully reproducible environments is a mechanism for pinning the version of repo2docker.

Several people have proposed ideas and it seems we have a plan: repo2docker will start, inspect a `/binder/repo2docker.version` file that specifies the version (git commit or version) that the user wants to use and then restart/start a new copy of repo2docker in that version.

The way I'd start on this is to experiment with adding a new entrypoint/CLI script that starts the current repo2docker CLI script inside a container together with the right mounts(?). Docker hub has a collection of images for lots of git revisions which we can use.

(related #170)

minrk:
Since you can run repo2docker in docker by mounting the docker socket (no mount necessary with docker-machine), this should be quite doable:

```bash
docker run --rm -it -v /var/run/docker.sock:/var/run/docker.sock -e DOCKER_HOST=unix:///var/run/docker.sock  jupyter/repo2docker:tag ...
```

So the only trick is going to be to make sure that repo2docker knows enough to start a repo2docker container with the docker socket mounted. The default case of /var/run/docker.sock ought to be straightforward. If there's any complexity, it's probably going to be in capturing and forwarding output from the sub-container.

jzf2101:
We have to publish versions on pip though- I thought that was still in process?

betatim:
> We have to publish versions on pip though- I thought that was still in process?

Releases of repo2docker are listed on PyPI: https://pypi.org/project/jupyter-repo2docker/

craig-willis:
As discussed on the call and in https://discourse.jupyter.org/t/repo2docker-roadmap-review/249/4, I'd like to work on this but will need to come up to speed on a few things.

Running a specified version of the `jupyter-repo2docker` package from within a Docker image makes sense.  What would the expected behavior be for someone running the CLI directly -- just a warning or error?  What are the current use cases for running via CLI?

betatim:
IMHO running from the CLI should do the same thing as using repo2docker as a library. This means that running `repo2docker https://github.com/org/repo` does the following:

1. fetch the repository
2. inspect the repository to find the version of repo2docker it specifies; default to the version of repo2docker that is installed
3. run `docker run --rm -it -v /var/run/docker.sock:/var/run/docker.sock -e DOCKER_HOST=unix:///var/run/docker.sock  jupyter/repo2docker:<r2dversionhere> ...` with the `...` replaced by the original CLI arguments

[...]

Issue #170, opened 2017-12-18
[Discussion] reproducibility


minrk:
Discussion issue for general topics of reproducibility and what's in and out of scope for repo2docker (and Binder).

We currently have a tension between our scientific goal of reproducibility and the maintenance goal of keeping everything up to date. We have the same issue that everyone who pursues reproducibility has, which is specifying the environment as strictly as necessary (so it's correct), but no stricter (so it stays useful). Conservative approaches are to use overly-specified environments (e.g. `pip freeze` / `conda env export`), which we should make sure to support well and document for the more reproducibility-minded users.

A user who wants to ensure a truly reproducible build must:

- use a `pip freeze` or `conda env export`-produced environment specification
- pin the Python version (for pip, already done above for conda)
- pin the distro/base image
- probably pin repo2docker itself (easy for manual use cases, not available on Binder)

Right now, the only truly reproducible builds available on Binder are custom Dockerfiles, which is something I want fewer people to use, not more. But we currently have no answer for reproducibility with any other builders, as there is no way for users to be sufficiently strict about the environment.

choldgraf:
I think this is a super important topic, especially when it comes to the publishing world. This is related to #93 though that's a more specific topic.

yuvipanda:
I really like the idea of pinning repo2docker versions, which seem like the easiest (and maybe only?) solution to this problem. If we can guarantee that a properly prepared repo will always produce the same Dockerfile (rather than image, since we can not guarantee that) for any given version of repo2docker, I think that's good enough no?

We might have to write version shims to maintain binderhub <-> repo2docker compatibility, but that seems not entirely too difficult. We could switch from passing in commandline arguments to using something more complex and versionable too if we want.

yuvipanda:
Thinking more on this, there's three things we should try to allow users to pin:

1. Versions of languages (Python, R, Julia, etc)
2. Versions of libraries for the language installed by the language specific package manager (conda, pip, whatever R uses, etc)
3. Versions of packages installed by the system package manager (apt)

We could / should use runtime.txt for (1), recommend pinning for (2), and make apt.yaml for (3). That's a good start I think, and gives us lots of low hanging fruit to work with...

betatim:
More thoughts on reproducibility: [freeze conda build numbers as well or not](https://github.com/jupyter/repo2docker/pull/193#issuecomment-357015696)

manics:
[...]

PR #550, opened 2019-01-09
`repo2docker.version` - a config file to make repo2docker run another repo2docker version

craig-willis:
WIP for #490. 

### Problem
As discussed in #490 and #170, it currently isn't possible for users to pin their repository to a specific version of `repo2docker`.  For example, the move from v0.5 to v0.6 included a change of the base image from Ubuntu artful (17.10) to Ubuntu bionic (18.04), which might cause unexpected problems for some users.  The "reproducibility-minded" user may want to have some guarantee that their repo will continue to work when repo2docker is upgraded.

## Implementation
This PR introduces two new methods in `Repo2Docker`, both called during `build()`:
*  `get_r2d_version`: After fetching the repo, reads the `repo2docker.version` file and performs a sanity check that an image with the specified version (tag) exists on Dockerhub.
* `run_r2d_version`: Runs `repo2docker` for the specified version via Docker to build the image with some special handling for local repos. 

This PR also adds:
* Unit (integration?) tests for valid and invalid versions, local and remote repositories (`test_versions.py`)
* Basic documentation describing the `repo2docker.version` file.


repo2docker/app.py
```diff
@@ -456,7 +456,7 @@ def start_container(self):
 
         docker_host = os.environ.get('DOCKER_HOST')
         if docker_host:
-            host_name = urlparse(docker_host).hostname
+            host_name = urllib.urlparse(docker_host).hostname
```
Xarthisius:
```suggestion
            host_name = urllib.parse.urlparse(docker_host).hostname
```

repo2docker/app.py
```diff
@@ -560,10 +560,103 @@ def find_image(self):
...
+        if os.path.isdir(self.repo):
+            # If the specified repo is a local path, we'll use a fixed path
+            # in the container.
+            args[-1] = '/repo2docker'
+
+            # Ensure the path is absolute 
+            if os.path.isabs(self.repo):
+                source_path = self.repo
+            else:
+                source_path = os.path.abspath(self.repo)
```
Xarthisius: I'd just always call `source_path = os.path.abspath(self.repo)` since for absolute path it's a noop.

repo2docker/app.py
```diff
@@ -560,10 +560,103 @@ def find_image(self):
...
+
+            # Mount the local path to the container
+            volumes[source_path] = {"bind": "/repo2docker", "mode": "rw"}
+
+        # Use repo2docker image with specified version tag
+        image = "jupyter/repo2docker:{}".format(version)
+
+        # Reconstruct the (potentially modified) command line 
+        cmd = "jupyter-repo2docker {}".format(" ".join(self.args))
```
Xarthisius: Did you mean `args` here instead of `self.args`?

yuvipanda:
Thanks for working on this, @craig-willis! Exciting progress.

My preliminary thoughts are:

1. repo2docker itself should verify that it is compatible with the version specified, and if not, quit with an error.
2. We should have a helper script that can read the version and figure out appropriate way to run correct version of repo2docker.
3. BinderHub should somehow find a way to respect this as well.

I don't think we should embed version checking functionality into repo2docker itself.

craig-willis:
Thanks for the very prompt feedback @Xarthisius  and @yuvipanda. 

@yuvipanda I may be misunderstanding your comment, but it sounds like you don't like the idea of running `jupyter/repo2docker` (via Docker) from within `jupyter-repo2docker` (not just the version check). Let me know if otherwise.

betatim:
> I don't think we should embed version checking functionality into repo2docker itself.

Do you mean that figuring out the version should not be part of the repo2docker package that gets installed with `pip install jupyter-repo2docker` or not part of what you get when you `import repo2docker` or the script called `repo2docker` (created via the entrypoint) or something else?

What I had in mind was that (for the moment?) add a new script called (say) `repo2docker-pinned` that fetches the repo, figures out the version, concocts a command like `docker run <dockersocketmountingstuff> jupyter/repo2docker:<tag> repo2docker ...` and then executes that. (We can also merge these two scripts into one but it seems like for discussions having different names would help as there are already many things called `repo2docker`)

The `repo2docker-pinned` script should use the `Repo2Docker` class to fetch the repo and inspect the version (via a new method). And then `subprocess.call()` the docker run command it constructed.

Good idea that r2d should check on launch that it is the version requested by the repo.

craig-willis:
Thanks, @betatim.  It sounds like I missed the mark on this PR (it's been a great experience for me to engage with the codebase, regardless).

It sounds like this is the preferred behavior?
```
$ pip install jupyter-repo2docker==0.7.0
repo2docker https://gitlab.com/org/repo-with-v0.5-specified

Error: Repo requires version 0.5.0 but current version is 0.7.0, exiting. 
          Please run repo2docker-pinned.

$ repo2docker-pinned https://gitlab.com/org/repo-with-v0.5-specified
Running repo2docker v0.5
...
Picked local content provider.
...
(works as expected)
```

The `repo2docker-pinned` CLI would handle the initial fetch and, to not fetch twice, run via local provider (modifying the path/repo passed to Repo2Docker).  It would need to accept the full set of command line arguments, but would only actually care about the repo/path.

> The repo2docker-pinned script should use the Repo2Docker class to fetch the repo and inspect the version (via a new method). And then subprocess.call() the docker run command it constructed.

Curious -- is there a benefit to using `subprocess.call()` instead of `docker.containers.run()`?

yuvipanda:
> What I had in mind was that (for the moment?) add a new script called (say) `repo2docker-pinned` that fetches the repo, figures out the version, concocts a command like `docker run <dockersocketmountingstuff> jupyter/repo2docker:<tag> repo2docker ...` and then executes that.

This sounds great to me, @betatim. Ideally, I'd like them to be two different *packages* to begin with, and repo2docker-pinned depend on the CLI interface only rather than the Python interface.

betatim:
I think this PR is pretty much what I described. The reason I spelt it out was that after asking Yuvi to be more precise I thought I should be explicit about what I was thinking as well.

The `repo2docker-pinned` script isn't required in the final thing but might help us get going because we can defer what ever logic we need to detect that we are in some kind of infinite loop.

> Curious -- is there a benefit to using `subprocess.call()` instead of `docker.containers.run()`?

I think the latter is a much better idea. I hadn't thought that far!

Not sure if we should make the optimisation of not fetching twice. Wondering what kind of bugs it might introduce because of content providers not being exactly the same. Should we start with fetching twice and see how bad it is performance wise? What do you think?

craig-willis:
Thanks, @betatim.  

I assumed creating `repo2docker-pinned` as a separate package would need to be done in an separate Github repo.   Let me know if otherwise.   If so, I could re-purpose this PR for documentation and the version check only.

A point of confirmation -- if I'm not mistaken, the `repo2docker-pinned` package will need to replicate some of the fetch, subdir, and binder_path handling if I can't use the Python package directly. 

> Should we start with fetching twice and see how bad it is performance wise? What do you think?

I'm wondering if I could use the CLI to do the fetch via `--no-build --no-run --no-clean`, possibly exposing the `git_workdir` as a CLI argument.   This would at least ensure the same fetch logic is used both places.

nuest:
@craig-willis I'd like to contribute here, if I can, because it nicely combines with #577 

@minrk @betatim Can you add your current thoughts on this? IIRC we said that cloning twice would be fine for now. The aspect of a separate script did not come up though.

craig-willis:
@nuest  Sorry to admit that I've dropped the ball on this. In the Whole Tale system, we addressed version pinning a different way and I never completed what was discussed in this PR.  If you'd like to take this work over, this PR can be closed.
  
(Did you mean to link to a different PR, #557 seems out of place?)

consideRatio:
This is a very old PR at this point, but I'm grown vary of the idea of letting repo2docker startup another version of itself for security reasons and also thinking that its a very complicated feature hard to maintain.

Since its last activity was 2019, I'll go for a close here. We can absolutely reconsider that decision!

PR #555, opened 2019-01-17
Set JULIA_DEPOT_PATH

### LINKED ISSUE #554: Julia installs packages into $HOME not /srv
We generally try to install all packages from the environment into ${APP_DIR} rather than $HOME. Julia seems  to install stuff into $HOME, causing issues when trying to start a kernel where $HOME is a persistent volume:

```
ERROR: could not open file /home/jovyan/.julia/packages/IJulia/GIANC/src/kernel.jl
Stacktrace:
 [1] include at ./boot.jl:317 [inlined]
 [2] include_relative(::Module, ::String) at ./loading.jl:1038
 [3] include(::Module, ::String) at ./sysimg.jl:29
 [4] exec_options(::Base.JLOptions) at ./client.jl:229
 [5] _start() at ./client.jl:421
[I 2019-01-17 05:51:47.383 SingleUserLabApp restarter:110] KernelRestarter: restarting kernel (4/5), new random ports
```

and over again for everything

yuvipanda:
Equivalent of JULIA_PKGDIR for Julia 1.0+

Fixes #554

### DIFF
--- a/repo2docker/buildpacks/julia/__init__.py
+++ b/repo2docker/buildpacks/julia/__init__.py
@@ -62,6 +62,8 @@ class JuliaBuildPack(CondaBuildPack):
                 will be installed
             - `JULIA_HOME`: path where all Julia Binaries will be installed
             - `JULIA_PKGDIR`: path where all Julia libraries will be installed
+            - `JULIA_DEPOT_PATH`: path where Julia libraries are installed.
+                                  Similar to JULIA_PKGDIR, used in 1.0.
             - `JULIA_VERSION`: default version of julia to be installed
             - `JUPYTER`: environment variable required by IJulia to point to
                 the `jupyter` executable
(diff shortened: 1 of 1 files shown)


yuvipanda:
Can verify that this works by locally running: 

```
 docker run -it -p 9001:8888 -v $(pwd):/home/jovyan gcr.io/ucb-datahub-2018/math124-user-image:2ceecc4 
```

For testing, I think this should verify that the JULIA_DEPOT_PATH has at least some libraries installed.

yuvipanda:
Test added!

yuvipanda:
I neither increased nor decreased test coverage in any way codecov can tell... Not sure why it dings me.

betatim:
Nice work.

There are a few issues related to Julia support that we should address all in one go/with a coherent approach.

Currently we let users choose the version of Julia they want. We need to maintain "backwards compatibility" for the oldest version we let users select (and test it actually works :-/) -> removing code yes please, if we have some tests (or a Julia expert on hand to tell us what we are about to break :) ).

StefanKarpinski:
They don't do the same thing exactly but I think that should work well enough.

PR #557, opened 2019-01-18
Support requirements.txt & runtime.txt with Julia

yuvipanda:
The comment about Julia only working with CondaBuildPack
hasn't been true since we switched to using conda for getting
our pythons. This change allows people to use requirements.txt
with Julia, rather than forcing them to use environment.yaml


yuvipanda:
@betatim When PythonBuildPack used to use a venv instead of conda, we ran into problems with Julia - https://github.com/jupyter/repo2docker/issues/23 has more information. https://github.com/JuliaPy/PyCall.jl/pull/411 was the primary underlying issue - Julia sortof pretends anyone using Python is primarily using Conda, and doesn't really support anything else well. And since PyPlot (used for plotting) depends on matplotlib, this was a no-go. So at that time, we made the Julia buildpack work only with Conda BuildPack.

Now we get conda for everything, so the restriction doesn't matter. The context for runtime.txt and requirements.txt is in PythonBuildPack, so we should get that here.

Re: tests, I think we have two PyPlot tests that do the same thing. I'm gonna remove one!

yuvipanda:
@betatim removed one test!

betatim:
Nice. With that explanation and the tests I feel like the risk of us breaking something is tiny -> merge!

Should we make a new issue to see if we can add a test of using `environment.yml` with Julia without having to build things. Just to make sure we don't break it at a later time. Tim-more-faster-tests-is-better-Head

yuvipanda:
@betatim thanks! Having an issue about being able to unit test 'does this directory tree make this kinda buildpack?' would be great.

PR #567, opened 2019-01-28
Register conda shell integration

jamesmyatt:
This allows current best practice `conda activate <name>` rather than `source activate <name>`.

See https://github.com/conda/conda/blob/master/CHANGELOG.md#440-2017-12-20

### DIFF
--- a/repo2docker/buildpacks/conda/install-miniconda.bash
+++ b/repo2docker/buildpacks/conda/install-miniconda.bash
@@ -21,7 +21,7 @@ if ! echo "${MD5SUM}  ${INSTALLER_PATH}" | md5sum  --quiet -c -; then
 fi
 
 bash ${INSTALLER_PATH} -b -p ${CONDA_DIR}
-export PATH="${CONDA_DIR}/bin:$PATH"
+. ${CONDA_DIR}/etc/profile.d/conda.sh
 
 # Allow easy direct installs from conda forge
 conda config --system --add channels conda-forge


yuvipanda:
Thanks for the PR!

This only changes the activation for this particular script - for the entire user environment, https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/conda/__init__.py#L47 is what sets the PATH. I'd like for us to make sure these two methods are consistent if possible. 

I'm not sure how to build a docker image in such a way as recommended with the CHANGELOG entry, though. I'm not sure if values exported from source are retained.

jamesmyatt:
Good question. I had assumed that any configuration, e.g. environment variables, from this script would apply to the whole environment. Where is conda added to the path in the overall context?

yuvipanda:
@jamesmyatt https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/conda/__init__.py#L47 is where it is added. Yeah, the context from this script will only apply to the end of this script.

Do you know why conda is suggesting these changes?

(and thank you for working on this!)

repo2docker/buildpacks/conda/install-miniconda.bash
```diff
@@ -21,7 +21,7 @@ if ! echo "${MD5SUM}  ${INSTALLER_PATH}" | md5sum  --quiet -c -; then
 fi
 
 bash ${INSTALLER_PATH} -b -p ${CONDA_DIR}
-export PATH="${CONDA_DIR}/bin:$PATH"
+. ${CONDA_DIR}/etc/profile.d/conda.sh
```
minrk: This is a good thing to do, but it would also probably be good to add a link to this file from `/etc/profile.d/conda.sh` so that any full shell invokes it as well.

PR #568, opened 2019-01-29
updating interfaces page and adding jupyterlab workspaces page

choldgraf:
This does the following:

* updates the interfaces page to give more useful information.
* Adds a section on Nteract and Shiny
* Adds a page about packaging workspaces for JupyterLab
* Does some minor re-structuring on the index page to make the content more discoverable (trying to riff off of @yuvipanda 's TLJH docs)

If we merge this in, then we should also update `conf.py` in the Binder docs repository to pull the JupyterLab workspaces page into the build: https://github.com/jupyterhub/binder/blob/master/doc/conf.py#L172

### DIFF
--- a/docs/source/howto/user_interface.rst
+++ b/docs/source/howto/user_interface.rst
@@ -39,17 +39,17 @@ To learn more about URLs in JupyterLab and Jupyter Notebook, visit
 `starting JupyterLab <http://jupyterlab.readthedocs.io/en/latest/getting_started/starting.html>`_.
 
 
-Nteract
+nteract
 =======
 
-`Nteract is a notebook interface <https://nteract.io/>`_ built with React.
+`nteract is a notebook interface <https://nteract.io/>`_ built with React.
 It is similar to a more feature-filled version of the traditional
 Jupyter Notebook interface.
 
-Nteract comes pre-installed in any session that has been built from
+nteract comes pre-installed in any session that has been built from
 a Python repository.
 
-You can launch Nteract from within a user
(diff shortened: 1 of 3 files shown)


choldgraf:
cc also @lheagy who I believe said she was interested in helping out w/ documentation in binderland!

docs/source/howto/user_interface.rst
```diff
@@ -25,11 +26,51 @@ To switch back to the classic notebook, add ``/tree`` to the URL like so:
...
 
+
+Nteract
+=======
+
+`Nteract is a notebook interface <https://nteract.io/>`_ built with React.
+It is similar to a more feature-filled version of the traditional
+Jupyter Notebook interface.
+
+Nteract comes pre-installed in any session that has been built from
```
lheagy: I think `nteract` is lower-case in most places in those docs - so might be better to use lower-case for consistency?

choldgraf:
cool - addressed your comments (maybe @rgbkrk can confirm that nteract should be all lower-case?)

docs/source/howto/user_interface.rst
rgbkrk: Yup, that's the case.

rgbkrk:
Yeah, nteract is with a lower case `n`.

choldgraf:
Cool, then this PR is ready to go from my end. More docs prs to come in the future

yuvipanda:
Thanks, @choldgraf!

Issue #25, opened 2017-06-02
Javascript builds


choldgraf:
We also need to support dependencies for Javascript (maybe including things that require a function call after installing, e.g. some of the jupyter extensions?). Discuss implementations etc here!

Note that our guide on repo2docker already starts out by showing how to create a builder with `npm`. Perhaps we can create a javascript build in parallel with finishing that document.

choldgraf:
just a thought: should we just capture this with `postBuild`?

manics:
If this is purely for jupyter extensions it could be part of https://github.com/jupyterhub/repo2docker/issues/868 ?

consideRatio:
Hey, starting out some triaging - this is the oldest issue open in repo2docker and seems to me like something not worth supporting without a stronger pull for it.

I'll go for a close at this point

PR #573, opened 2019-02-07
npm installation

### LINKED ISSUE #307: npm install not permitted
In https://github.com/jupyterhub/binderhub/issues/558 @aborruso reports:

> Hi,
> I have `npm install -g mapshaper` in postBuild, but I'm blocked because I'm not root.
>
> Is there another way to install a nodejs package and have it ready to use after building?
>
> Thank you

Does anyone remember why we decided that `postBuild` should not run as root?

GladysNalvarte:
- defines NPM_DIR env and creates it at APP_BASE/npm
- tells npm to use this dir for `npm install --global`

closes #307

Related to #25, but not closing since this doesn't introduce package.json support or similar.

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -356,6 +356,7 @@ class BuildPack:
         You can use environment variable substitutions in both the
         username and the execution script.
         """
+
         return []
 
     def get_assemble_scripts(self):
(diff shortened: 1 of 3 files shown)


betatim:
Thanks for working on this. Looks good to me!

I'll leave it open for others to have a chance to offer opinions, going to merge tomorrow if no one else gets to it.

minrk:
Great, thanks @GladysNalvarte!

choldgraf:
way to go @GladysNalvarte ! thanks! 🎉

PR #575, opened 2019-02-07
Check if '--debug' is set properly

### LINKED ISSUE #563: 'Repo2Docker' object has no attribute ‘debug'
As reported by @sbinet, 

When repo2docker raises an error, it is possible to get the following message on the command line:

```
'Repo2Docker' object has no attribute ‘debug'
```

This seems to come from this section: 

https://github.com/jupyter/repo2docker/blob/ff894f4404a9abcbb5163f638f952a1ba89fe8b3/repo2docker/__main__.py#L335-L344

I am not sure exactly how the configuration system works here but I notice this is the only time the `debug` argument is accessed:

https://github.com/jupyter/repo2docker/blob/ff894f4404a9abcbb5163f638f952a1ba89fe8b3/repo2docker/__main__.py#L220-L221

cc @betatim who thought this also looked like a bug

yuvipanda:
Fixes #563

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -334,12 +334,12 @@ def main():
         r2d.start()
     except docker.errors.BuildError as e:
         # This is only raised by us
-        if r2d.debug:
+        if r2d.log_level == logging.DEBUG:
             r2d.log.exception(e)
         sys.exit(1)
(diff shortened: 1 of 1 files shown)


betatim:
Merged, codecov is only unhappy because the changed lines don't have coverage (and didn't have any before).

PR #576, opened 2019-02-08
pin conda during Python-switch step

minrk:
even with auto-update-conda disabled, switching Python version (which reinstalls conda) can end up upgrading to the latest conda

Avoid this by adding the conda pinning to that install stage

This should fix the recent failures on CI, which appear to be due to a bug in the recent conda 4.6

### DIFF
--- a/repo2docker/buildpacks/conda/install-miniconda.bash
+++ b/repo2docker/buildpacks/conda/install-miniconda.bash
@@ -63,7 +63,7 @@ fi
 # which seems to result in some effective pinning of packages in the initial env,
 # which we don't intend.
 # this file must not be *removed*, however
-echo '' > ${CONDA_DIR}/envs/kernel/conda-meta/history
+echo '' > ${CONDA_DIR}/conda-meta/history
 
 # Clean things out!
 conda clean -tipsy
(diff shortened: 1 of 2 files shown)


betatim:
Is there repo we build in the CI where we can add a quick check that "conda is indeed the version we expect it to be'? It would have to be a test that has an `environment.yml` which asks for a different Python version correct?

minrk:
Good call. Added a check to the Python 3.5 test

minrk:
Will open a separate issue to track updating to conda 4.6, which is likely waiting for bugfix releases of 4.6

betatim:
We could use the quant stack repo to test this I think by editing https://github.com/jupyter/repo2docker/blob/8597520e7ad197236006b4246e4ed53516aa3c6d/tests/external/reproductions.repos.yaml#L32 to be `verify: jupyter kernelspec list && conda --version | grep 4.5.12`

betatim:
A few of the builds are failing on:
```
/tmp/install-miniconda.bash: line 66: /srv/conda/envs/kernel/conda-meta/history: No such file or directory
```
I think. :-/

betatim:
I made an edit to the path used on L66 of the install script. My guess is that the old path is the correct one and the new one (with extra `envs/kernel` in it) was a copy pasta from above. If not it seems like we should not have the same line in the if statement above and here but instead have it once.

PR #577, opened 2019-02-11
[MRG] Update ChangeLog to release 0.8.0

betatim:
Originally thought we hadn't changed much so the next release would be 0.7.1.

However while collecting things for the change log I changed my mind. There are quite a few changes, most in the "making things more stable and reliable and conformant" department. Should we call it 0.8.0?

Please make suggestions about the wording for the change log via the GitHub UI. Not quite sure what tense/form they should take.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -3,21 +3,50 @@ Changelog
 =========
 
 
-Upcoming release
-================
+Version 0.8.0
+=============
 
-Release date: TBD
+Release date: 2019-02-16
 
 New features
 ------------
 - Add additional metadata to docker images about how they were built :pr:`500` by
   :user:`jrbourbeau`.
+- Allow users to install global NPM packages: :pr:`573` by :user:`GladysNalvarte`.
+- Add documentation on switching the user interface presented by a
+  container. :pr:`568` by user:`choldgraf`.
+- Increased test coverage to ~87% by :user:`betatim` and :user:`yuvipanda`.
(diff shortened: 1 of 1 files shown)


minrk:
sounds good to me!

yuvipanda:
@minrk @betatim I'd like https://github.com/jupyter/repo2docker/pull/539 to be in this...

betatim:
The "major" version bump in #539: do you mean major as in 0.7.0 -> 1.0.0 or 0.7.0 -> 0.8.0?

If we do follow sem-ver then I think as long as it is 0.x.y we could go with 0.7.0 -> 0.8.0 as a "major" bump. It seems the rules suggest that "as long as the major is 0 anything might change at any moment".

betatim:
Looking at the change log I think we should go ahead and call this 0.8.0.

betatim:
Let's release this tomorrow before the meeting or just after as the change log is already starting to miss things.

betatim:
Merging this. If we merge more PRs that need to be in the change log please add a line as part of the PR.

docs/source/changelog.rst
```diff
@@ -3,21 +3,50 @@ Changelog
 =========
 
 
-Upcoming release
-================
+Version 0.8.0
+=============
 
-Release date: TBD
+Release date: 2019-02-16
```
betatim: Need to update the date before cutting the release :-/

yuvipanda:
Thank you very much for waiting, @betatim!

PR #579, opened 2019-02-15
Exposes CPU limit

### LINKED ISSUE #459: Expose cpu limit
We allow limiting the memory used by builds. We should allow limiting cpu as well. The simplest would be to only expose the `--cpus` arg, but docker has [several](https://docs.docker.com/config/containers/resource_constraints/#cpu) ways to tune cpu throttling.

GladysNalvarte:
Includes extra kwargs to limit CPU quota when building and running a docker image.

Closes #459

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -532,8 +532,7 @@ class Repo2Docker(Application):
 
         run_kwargs.update(self.extra_run_kwargs)
 
-        container = client.containers.run(
-            self.output_image_spec, **run_kwargs)
+        container = client.containers.run(self.output_image_spec, **run_kwargs)
 
         while container.status == 'created':
             time.sleep(0.5)
@@ -672,7 +671,7 @@ class Repo2Docker(Application):
                             self.log.info(l['error'], extra=dict(phase='failure'))
                             raise docker.errors.BuildError(l['error'], build_log='')
                         elif 'status' in l:
-                                self.log.info('Fetching base image...\r',
+                            self.log.info('Fetching base image...\r',
                                               extra=dict(phase='building'))
                         else:
                             self.log.info(json.dumps(l),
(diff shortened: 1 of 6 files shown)


repo2docker/buildpacks/base.py
```diff
@@ -463,7 +465,7 @@ def render(self):
             appendix=self.appendix,
         )
 
-    def build(self, client, image_spec, memory_limit, build_args, cache_from):
+    def build(self, client, image_spec, memory_limit, build_args, cache_from, extra_build_kwargs):
```
```suggestion
    def build(self, client, image_spec, memory_limit, build_args, cache_from, extra_build_kwargs=None):
```
What do you think of adding this as a named parameter with a default value of `None`? This means that existing calls to `build()` will not need updating. I think this is more part of the internal API of repo2docker so we don't have to do this.

betatim:
Nice work! I left a small comment on making the new argument a named parameter.

There are a few tests that fail that need fixing and then this can be merged I think. For `test_cache_from_base` we need to adjust the test but for others like `test_connect_url` I don't think we should edit the test and instead change the code of the repo2docker library itself to make the test work.

repo2docker/app.py
```diff
@@ -672,7 +671,7 @@ def build(self):
                             self.log.info(l['error'], extra=dict(phase='failure'))
                             raise docker.errors.BuildError(l['error'], build_log='')
                         elif 'status' in l:
-                                self.log.info('Fetching base image...\r',
+                            self.log.info('Fetching base image...\r',
```
betatim: Nice catch!

repo2docker/buildpacks/base.py
```diff
@@ -9,6 +9,8 @@
 import sys
 import xml.etree.ElementTree as ET
 
+from traitlets import Dict
```
betatim: Do we need this import in this file?

betatim:
Nice work! Merged this now, we can make a new PR to clean up the unused commit (if it is indeed unused).

PR #581, opened 2019-02-19
add pip freeze to travis.yml

minrk:
running pip freeze on every test is super helpful for debugging when a dependency change starts causing test failures

and stop building docs on travis; we are building them on circle now


betatim:
PR that removes our travis config from the coverage computation incoming ...

betatim:
Actually, I am not sure why this PR had negative coverage. The travis config doesn't appear in the codecov overview.

PR #583, opened 2019-02-20
install: one-liner to install most recent upstream code

haraldschilly:
This came in handy for me, to get the fixes for the python3.5 compatibility. Might be helpful for others, too.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -29,17 +29,21 @@ We recommend installing ``repo2docker`` with the ``pip`` tool::
 
     python3 -m pip install jupyter-repo2docker
 
+for the latest release. To install the most recent code from the upstream repository, run::
+
+    python3 -m pip install git+https://github.com/jupyter/repo2docker.git
+
 For information on using ``repo2docker``, see :ref:`usage`.
 
 Installing from source code
 ---------------------------
 
-Alternatively, you can install repo2docker from source,
-i.e. if you are contributing back to this project::
+Alternatively, you can install repo2docker from a local source tree,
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -29,17 +29,21 @@ We recommend installing ``repo2docker`` with the ``pip`` tool::
 
     python3 -m pip install jupyter-repo2docker
 
+for the latest release. To install the most recent code from the upstream repository, run::
+
+    python3 -m pip install git+https://github.com/jupyter/repo2docker.git
```
betatim:
```suggestion
    python3 -m pip install https://github.com/jupyter/repo2docker/archive/master.zip
```
Would the suggestion work as well? That way users wouldn't need to have git installed and don't have to transfer the history of the project, just the files that are currently in `master`. I use this form a lot instead of `git+https://` but wondering if there are edge cases?

betatim:
One question as a comment to see if we can reduce the dependencies needed for installing even further.

Another question for my education: I see `python3 -m pip` as replacement for `pip install` getting used a lot, what is the reason? Or asked differently when is it not equivalent?

haraldschilly:
> I see `python3 -m pip` as replacement for `pip install`...

There are situations (older linux distributions) where pip is tied to python 2. This makes this work without ambiguity. Another option I can think of is to change this to `pip3 install ...`, but I don't know if `pip3` is available everywhere.

docs/source/install.rst
haraldschilly: good point, yes that also works

betatim:
Thanks!

PR #594, opened 2019-02-24
WIP Add support for julia Project.toml files

### LINKED ISSUE #486: Manifest and Project .toml files in Julia Buildpacks
@jekbradbury suggested we include https://docs.julialang.org/en/v1/stdlib/Pkg/index.html#Glossary-1 in our buildpacks.  Thanks @jekbradbury !

davidanthoff:
Fixes #486.

Very WIP right now.

@NHDaly would be great if you could take a look at this. I did this without a local system to try anything out, i.e. I don't have a docker setup or anything like that. So taking this further is kind of involved for me... But maybe we can jointly somehow bring this over the finishing line?

### DIFF
--- a/repo2docker/buildpacks/julia/__init__.py
+++ b/repo2docker/buildpacks/julia/__init__.py
@@ -165,8 +165,8 @@ class JuliaBuildPack(PythonBuildPack):
                     # HACK: Can't seem to tell IJulia to install in sys-prefix
                     # FIXME: Find way to get it to install under /srv and not $HOME?
                     r"""
-                    julia -e 'using Pkg; Pkg.add("IJulia"); using IJulia; installkernel("Julia", "--project={0}");'
-                    """.format('~')
+                    julia -e "using Pkg; Pkg.add(\"IJulia\"); using IJulia; installkernel(\"Julia\", \"--project=${REPO_DIR}\");"
+                    """
                 ),
                 (
                     "${NB_USER}",
--- a/repo2docker/buildpacks/julia/__init__.py
+++ b/repo2docker/buildpacks/julia/__init__.py
@@ -120,14 +120,6 @@ class JuliaBuildPack(PythonBuildPack):
                     mkdir -p ${JULIA_PKGDIR} && \
                     chown ${NB_USER}:${NB_USER} ${JULIA_PKGDIR}
                     """
-                ),
-                (
-                    "${NB_USER}",
-                    # HACK: Can't seem to tell IJulia to install in sys-prefix
-                    # FIXME: Find way to get it to install under /srv and not $HOME?
-                    r"""
-                    julia -e 'using Pkg; Pkg.add("IJulia"); using IJulia; installkernel("Julia", "--project=.", env=Dict("JUPYTER_DATA_DIR"=>"${NB_PYTHON_PREFIX}/share/jupyter/kernels"));'
-                    """
                 )
             ]            
         else:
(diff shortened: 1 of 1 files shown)


davidanthoff:
I managed to get everything running locally on my Mac, so I can actually debug things now. I'll ping folks again once I have a working version for a review.

repo2docker/buildpacks/julia/__init__.py
```diff
@@ -132,20 +158,53 @@ def get_assemble_scripts(self):
...
-        )]
+        if self.julia_env_exists:
+            return super().get_assemble_scripts() + [
+                (
+                    "${NB_USER}",
+                    # HACK: Can't seem to tell IJulia to install in sys-prefix
+                    # FIXME: Find way to get it to install under /srv and not $HOME?
+                    r"""
+                    julia -e 'using Pkg; Pkg.add("IJulia"); using IJulia; installkernel("Julia", "--project={0}");'
+                    """.format('~')
```
davidanthoff: @betatim & @NHDaly This is a place where I'm stuck. What we need to do here is pass the absolute path to the root of the project folder inside the docker image (instead of just `~`). Is there a way that I can get that path at this point? For example, when I run some test repo on the public mybinder.org website, it seems I would have to pass `/home/jovyan` here. When I run this on my mac, it seems to use my own username instead.

davidanthoff:
@KristofferC, quick question: I want to add the ability here to decide what julia version to use, based on the content of the `Project.toml`. When I look at the [docs](https://julialang.github.io/Pkg.jl/v1/compatibility/) it seems to me that it would be good enough to look for a `julia` entry in the `[compat]` section and just decide based on that. Is that correct, or should I look at something more sophisticated? In particular, I'm slightly confused whether I first need to somehow resolve the `julia` name in some dependency section before I can look at the `[compat]` section? Or can I just ignore the entire rest of the .toml file? Thanks!

I guess if at some point `Manifest.toml` also supports pinning a julia version, then we would have to extend the logic to also look in there, but for now we can ignore that, right?

KristofferC:
> When I look at the [docs](https://julialang.github.io/Pkg.jl/v1/compatibility/) it seems to me that it would be good enough to look for a `julia` entry in the `[compat]` section and just decide based on that

Yes, that should be enough. However, the compatibility is given in "semver format" (https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format-1) so you need to translate that to actual versions. The way Pkg does this is via:

```
julia> Pkg.Types.semver_spec("1.0")
VersionSpec("1.0.0-1")
```

and from there you are unfortunately hitting internals of the package manager. We should extract all the semver stuff into its own module and document it.

> I guess if at some point `Manifest.toml` also supports pinning a julia version, then we would have to extend the logic to also look in there, but for now we can ignore that, right?

Yes, Manifest does not record julia version right now so

davidanthoff:
@KristofferC, cool thanks! We won't be able to use any julia code for handling of these semver format strings, unfortunately, since we need to process them to figure out which julia version to install, which also implies that no julia version is around at that point. So we'll have to handle that in Python.

There is a toml parser for Python, but now I'm wondering whether there might also be a python semver package that could help us with this. The format julia is using there is not julia specific, or is it?

repo2docker/buildpacks/julia/__init__.py
betatim: I think `REPO_DIR` is the variable to look at for that.

davidanthoff: Yep, that was it, thanks!

davidanthoff:
I have one question about overall design, that I think is mostly for the core maintainers of this repo here: the `JuliaBuildPack` now has a lot of `if` statements that handle the pre-julia 1.0 case and the post julia 1.0 case. Overall this makes the code kind of messy. An alternative would be to split these out into two different classes: have one `LegacyJuliaBuildPack` for julia pre 1.0 and one `JuliaBuildPack` for everything ever since.

I'm starting to think that that would make more sense: the code would be cleaner, we could leave the legacy version alone and it would just continue to work while we add features to the version that works with the non-prerelease versions of julia. I think this last point is important: if we mix these two cases into one class, then whenever someone is adding features, they need to make sure that it works with a julia version that is potentially no longer supported etc.

@betatim, any thoughts?

davidanthoff:
@betatim, oh, and one more question: right now the `get_build_scripts` function downloads and installs julia. Which version of julia gets installed depends on the content of the REQUIRE/Project.toml file in the repo. Should that code then live in `get_build_scripts`, or should it actually be in `get_assemble_scripts`?

betatim:
Short answer on just one question:

> An alternative would be to split these out into two different classes: have one `LegacyJuliaBuildPack` for julia pre 1.0 and one `JuliaBuildPack` for everything ever since.

I think I like the idea. We'd have to be a bit careful to make sure they don't both trigger and how to organise the inheritance tree. My quick fire thought is to make a `Julia1point0` and a `EarlyJulia` class that each deal with the specifics. Then use these two from within `JuliaBuildPack` which remains the only one inside the inheritance structure of build packs. More like composition than inheritance maybe.

More out loud thought: `detect()` would decide which actual Julia class to use, instantiate it and store it in `self._actual`. Other methods then become proxies/forwarders:
```python
def detect(self):
    if julia_one_detected:
        self._actual = Julia1point0()
    else:
        self._actual = EarlyJulia()

def foo(self, *args):
    return self._actual.foo(*args)

def blah(self, arg1):
    return self._actual.blah(arg1)
```

(Better names all around would be needed 😀)

davidanthoff:
I think https://github.com/jupyter/repo2docker/pull/595 at this point is clearly better, so I'm going to close this one here.

PR #595, opened 2019-02-26
[MRG+1] Split julia support into pre julia 1.0 and post julia 1.0

davidanthoff:
This is an alternative to #594, that I think is cleaner.

@betatim This doesn't follow your suggestion of composition, instead it just treats these two julia versions as entirely different stories. I think we probably eventually want to just remove support for version pre 1.0: they are not maintained, and I have a hard time imagining that anyone will use them for much longer. Removing support for this old legacy stuff would be _super_ easy with the code structure in this PR: one wouldn't even have to touch the julia 1.0 support code at all. But we can also leave it in indefinitely, given that support for the two is more or less completely isolated, it probably doesn't do any harm.

The logic of which of the julia implementations should handle a given repo is super simple, and I don't think there is a risk that they would get in each others way.

There is also a more radical suggestion: we could say that a `REQUIRE` file can only be used for julia 0.6.x, and for julia 1.0 one _must_ use a `Project.toml`. If we were designing this from scratch, I think that would actually be the proper way to handle things. `REQUIRE` is really not a thing in julia 1.0 anymore (except for some corner package cases that are not relevant to the binder story), so the `REQUIRE` support for julia 1.0 here in binder is really a binder-only thing, which doesn't seem ideal. But on the other hand, support for this has been in the shipping version of binder for a while now, so maybe we just have to keep it for backwards compat reasons?

The major thing that is still missing from this PR is the logic that selects the exact julia 1.x.x version subject to the version specifier in the `Project.toml`. Things that we need for that are a) add a dependency to https://github.com/uiri/toml so that I can read the toml file, and b) figure out whether there is a python semver package that can help us do the version resolution.

I'm stuck right now on a) already, because I don't know what I need to do to add a package dependency :) I'm not really a Python expert ;)

The support for julia 1.0 here has some other improvements: 1) I managed to install the Julia kernel into the right location right away, no manual file move needed anymore, 2) we don't need the julia code that resolves the repo dependencies, the package manager handles all of that now.

Random other question: why does the julia build pack inherit from the python build pack? Wouldn't it make more sense to inherit from something that doesn't install the python kernel? I.e. something that _does_ install the jupyter notebook & lab, but not the python kernel itself?

Also CC @NHDaly.


NHDaly:
@davidanthoff: Thanks so much for putting work into this!!

First off, I'm sorry i haven't been a part of this conversation; I'm on vacation this week and last, and haven't been looking at a computer much. This looks really great though, thanks for your work!

Second apology: I'm sorry I never posted a PR for it, but I actually started some work on this a while ago but never posted it! You can see my take on it here:
https://github.com/jupyter/repo2docker/compare/master...NHDaly:julia_projecttoml

> The major thing that is still missing from this PR is the logic that selects the exact julia 1.x.x version subject to the version specifier in the `Project.toml`. Things that we need for that are a) add a dependency to https://github.com/uiri/toml so that I can read the toml file, and b) figure out whether there is a python semver package that can help us do the version resolution.

> I'm stuck right now on a) already, because I don't know what I need to do to add a package dependency :) I'm not really a Python expert ;)

In particular I wrote a python library for (b), because I couldn't find one either. But i'm also not a super python expert and actually getting someone to check this over is one of the main reasons I never opened a PR with these changes. :)

Sorry i don't have time to look at this PR for a couple days (still on vacation), but i will! In the meantime, hopefully looking at my work will help some! I THINK it's a reasonable python implementation of Julia's semver.

repo2docker/buildpacks/__init__.py
```diff
@@ -2,6 +2,7 @@
 from .python import PythonBuildPack
 from .conda import CondaBuildPack
 from .julia import JuliaBuildPack
+from .julialegacy import JuliaLegacyBuildPack
```
betatim: I'd import both from julia and deal with splitting them up there.
```suggestion
from .julia import JuliaLegacyBuildPack
```

betatim:
I think keeping support "for longer than is reasonable" should be the strategy for repo2docker given we are in the business of reproducible science. Once we have had support for repositories specifying the version of repo2docker they want to use to build we can get more aggressive/less conservative.

I wouldn't add the `julialegacy` sub-directory and instead create two files in `julia/` one each per build pack (and a `__init__.py`). I'd also avoid using "legacy" because what happens when there is v1.2, does 1.0 now belong to legacy or not? Maybe `JuliaProjectTomlBuildPack` and `JuliaREQUIREBuildPack` instead using names with the version in it? It seems like the build packs are selected by different files and not versions.

tests/julia/julia_version-1.1/verify
```diff
@@ -1,17 +1,17 @@
...
-    # Verify that the environment variables are set correctly for julia 1.0+
-    @assert "julia" ∈ readdir(Sys.BINDIR)
-catch
-    exit(1)
-end
+#     # Verify that the environment variables are set correctly for julia 1.0+
+#     @assert "julia" ∈ readdir(Sys.BINDIR)
+# catch
+#     exit(1)
+# end
```
betatim: Is the commenting out temporary? I think we should keep these checks.

betatim:
After reading this one more time and pondering it I think we should name the build packs after the files that trigger them instead of versions because you can install Julia 1.0 and 1.1 via `REQUIRE` as well as a `Project.toml`.

I think adding `toml` as a dependency is a good idea.

semver: I'd have to do a bit of googling as well, don't have a go to library for that.

In one of the other PRs you also asked about build vs assemble scripts or some such?

betatim:
From my side we can wait for @NHDaly to come back from vacation to get some input and give you a chance to take a look.

Things we should not lose track of:
* [ ] update docs
* [ ] update change log
* [ ] add an example to https://github.com/binder-examples/ that uses the new way of specifying deps

davidanthoff:
@NHDaly I copied some of the code from your branch over here. BUT, I didn't just copy the whole julia_semver.py file in, because I don't want git to credit me with writing this code when really you did all the work :) Do you want to open a PR against this branch here that adds just that one file? If the API of it stays the same, it should hopefully work with the rest of this branch here.

There is also a test failure here that I need to figure out and fix, and then all the house keeping things that @betatim pointed out.

davidanthoff:
Ha, I figured out that I can commit something but change the author info for that commit :) So, now I copied the code from @NHDaly over into this PR, _and_ he is the author for that commit.

I removed all the asserts because they should probably become tests. We can try to sort this out once things are actually functional.

repo2docker/buildpacks/julia/julia_semver.py
```diff
@@ -0,0 +1,119 @@
...
+    # Default to "^" search if no matching mode specified (up to next major version)
+    if (first_digit.start() == 0) or (comparison_symbol == "^"):
+        if major(constraint) == 0:
+            # Also, julia treats pre-1.0 releases specially, as if the first
+            # non-zero number is actually a major number:
+            # https://docs.julialang.org/en/latest/stdlib/Pkg/#Caret-specifiers-1
+            # So we need to handle it separately by bumping the first non-zero number.
+            for i,n in enumerate(constraint):
+                if n != 0 or i == len(constraint)-1: # (using the last existing number handles situations like "^0.0" or "^0")
+                    upper = (*constraint[0:i], n+1)
```
davidanthoff: @betatim This syntax is apparently not supported on Python < 3.5... Any suggestion how we could write this instead? My Python is not exactly top notch :) And I assume we do need to support older Python versions, right?

NHDaly: I think the python2 way to do this would be "tuple concatenation" via `+`?

I just tried it, and I _think_ this should work:
```python
upper = constraint[0:i] + (n+1,)
```

NHDaly: (This is what i tried:
```python
>>> constraint = (0,6,1)
>>> i,n = (2,1)
>>> constraint[0:i] + (n+1,)
(0, 6, 2)
```
)

davidanthoff: Alright I tried that in the latest push, lets see whether it works!

NHDaly:
> Ha, I figured out that I can commit something but change the author info for that commit :) So, now I copied the code from @NHDaly over into this PR, _and_ he is the author for that commit.
> 
> I removed all the asserts because they should probably become tests. We can try to sort this out once things are actually functional.

haha neat!! :D Thanks for the credit, @davidanthoff! 😁 Sorry again that I was so out of reach!

davidanthoff:
I believe this is more or less done now and ready for a good review :)

I updated the documentation in the places that I was able to identify. I'm updating the example binder repo now.

I think what is still missing is to integrate the asserts that @NHDaly had written as tests. Maybe the easiest approach for that would be to merge this PR first, and then @NHDaly can open another PR that adds that code?

I was not sure how to update the changelog, in particular given that there didn't seem to be a section for the next version.

betatim:
> I was not sure how to update the changelog, in particular given that there didn't seem to be a section for the next version.

Oups! That is an admin fail on my side. Created #606 to fix that.

repo2docker/buildpacks/julia/julia_semver.py
betatim: We go back to Python 3.4 in terms of supporting old versions.

docs/source/faq.rst
```diff
@@ -42,12 +42,9 @@ environment.
...
 
-The following versions of Julia are supported (specified in the
-:ref:`REQUIRE <REQUIRE>` configuration file):
-
-- 1.0 (added in 0.7)
-- 0.7 (added in 0.7)
-- 0.6 (default)
+All Julia versions since Julia 0.7.0 are supported via a ``Project.toml``
+file, and this is the recommended way to install Julia environments.
+Julia 0.6.x is supported via a ``REQUIRE`` file.
```
betatim:
```suggestion
Julia 0.6.x is supported via a :ref:`REQUIRE <REQUIRE>` file.
```

I find having the cross links super useful when navigating the docs of other projects.

docs/source/faq.rst
```diff
@@ -42,12 +42,9 @@ environment.
 Julia
 ~~~~~
 
-The following versions of Julia are supported (specified in the
-:ref:`REQUIRE <REQUIRE>` configuration file):
-
-- 1.0 (added in 0.7)
-- 0.7 (added in 0.7)
-- 0.6 (default)
+All Julia versions since Julia 0.7.0 are supported via a ``Project.toml``
```
betatim:
```suggestion
All Julia versions since Julia 0.7.0 are supported via a :ref:`Project.toml <Project.toml>`
```

davidanthoff:
Cool, thanks, I now also updated the changelog in the template!

docs/source/changelog.rst
```diff
@@ -9,6 +9,7 @@ Release date: TBD
 
 New features
 ------------
+- Support for julia `Project.toml`, `JuliaProject.toml` and `Manifest.toml` files.
```
betatim:
```suggestion
- Support for julia `Project.toml`, `JuliaProject.toml` and `Manifest.toml` files in :pr:`595` by
:user:`davidanthoff`
```

So we know who to blame ... I mean credit ;)

repo2docker/buildpacks/julia/julia_project.py
```diff
@@ -0,0 +1,140 @@
...
+            project_toml = toml.load(self.binder_path('Project.toml'))
+        if 'compat' in project_toml:
+            if 'julia' in project_toml['compat']:
+                julia_version_str = project_toml['compat']['julia']
+
+                # For Project.toml files, install the latest julia version that
+                # satisfies the given semver.
+                julia_version = find_semver_match(julia_version_str, self.all_julias)
+                if julia_version is None:
+                    return _default_julia_version
```
betatim:
```suggestion
                    return default_julia_version
```
to match line 23

betatim: There is a 
```
else:
  return julia_version
```
missing here I think

betatim: Might be worth adding a test with a `Project.toml` that selects a non default version to stop this bug from coming back.

repo2docker/buildpacks/julia/julia_project.py
```diff
@@ -0,0 +1,140 @@
...
+        This creates a directory with permissions for installing julia packages
+        (from get_assemble_scripts).
+
+        """
+        return super().get_build_scripts() + [
+            (
+                "root",
+                r"""
+                mkdir -p ${JULIA_PATH} && \
+                curl -sSL "https://julialang-s3.julialang.org/bin/linux/x64/${JULIA_VERSION%[.-]*}/julia-${JULIA_VERSION}-linux-x86_64.tar.gz" | tar -xz -C ${JULIA_PATH} --strip-components 1
```
betatim: I think this should go into `get_assemble_scripts` because it depends on the contents of the repository.

The build scripts run [before the repo contents is copied over](https://github.com/jupyter/repo2docker/blob/e6cbf3e3a3e399508d5628b4ff1d30afed8c944f/repo2docker/buildpacks/base.py#L91-L96) and [assemble runs after](https://github.com/jupyter/repo2docker/blob/e6cbf3e3a3e399508d5628b4ff1d30afed8c944f/repo2docker/buildpacks/base.py#L131-L135).

We'd like to make sure that all steps before copying over the repository can be cached and shared between repos. So it should be independent of the repo contents.

(I've been pondering this for a while and am starting to think we should put things like installing Julia or RStudio up here. However I'd want to ponder that for a bit longer and get input from others. So for this PR I'd suggest we create a new issue for discussion of this. Otherwise we have to wait on a decision to merge this PR. (There is also the option that my feeling here is off and it is uncontroversial to put this step in the build part, would need some feedback. Maybe @minrk or @yuvipanda have some on how controversial this would be)

betatim: Looking at the history of `julia/__init__.py` the "install Julia version X" script has been in the build step for a good while, so maybe it is just me who thinks this is controversial. My suggestion is now "Definitely punt any discussion to a new issue".

repo2docker/buildpacks/julia/julia_semver.py
```diff
@@ -0,0 +1,119 @@
+# This file implements the julia-specific logic for handling SemVer (Semantic
+# Versioning) strings in .toml files.
+#
+# It uses the python "semver" package to do most version string comparisons, but
+# the places where julia's SemVer handling differes from the semver package have
```
betatim:
```suggestion
# the places where julia's SemVer handling differs from the semver package have
```

tests/julia/julia_version-1.1/README.rst
```diff
@@ -1,5 +1,4 @@
-Julia - REQUIRE: test version for julia v1.1
+Julia - Project.toml: test version for julia v1.1
 ---------------
 
-Starting with julia v0.7 and up, the package manager has changed, so this tests
-that the julia version can be specified and packages installed correctly.
+Test a Project.toml file for julia.
```
betatim: Related to the comment earlier ("add a test to select non default version"). Should we rename this test so it conveys the idea "this test checks that the build pack selects the default version of Julia for a Project.toml based repo". Currently that default version is 1.1 but at some point it will get bumped up (I guess).

Then with a second test that specifically selects a version that isn't the default (say 1.0?) we'd cover both options.

betatim:
I think I am happy with this PR now once we go through the above comments. Most of the mare typo fixes, one is about the version selection for which I think we should add an additional test.

I also have feelings about the semver code, mostly to make it more pythonic. I'd punt those to a new PR though where I'd add in the tests (from the asserts) and propose how I'd jiggle things around. Would that work for you?

I think adding a test to check that the build pack works properly with a non default version would also exercise some of the semver code which will make codecov happy -> we can merge.

docs/source/changelog.rst
davidanthoff: Haha, sure :)

repo2docker/buildpacks/julia/julia_project.py
davidanthoff: Alright, I changed the structure of that code part. I also added a test that should exercise it.

repo2docker/buildpacks/julia/julia_project.py
davidanthoff: Ok, so we'll leave it as is for now. I agree, I was also not sure what the right strategy is...

tests/julia/julia_version-1.1/README.rst
davidanthoff: Done!

davidanthoff:
Thanks for the great feedback! I think I addressed all of it, lets see whether travis is happy :)

repo2docker/buildpacks/julia/julia_require.py
```diff
@@ -0,0 +1,170 @@
+"""Generates a Dockerfile based on an input matrix for Julia"""
```
willingc:
```suggestion
"""Generates a Dockerfile based on an input matrix with REQUIRE for legacy Julia"""
```

repo2docker/buildpacks/julia/julia_require.py
```diff
@@ -0,0 +1,170 @@
...
+import os
+from ..python import PythonBuildPack
+
+
+class JuliaRequireBuildPack(PythonBuildPack):
+    """
+    Julia build pack which uses conda.
```
willingc:
```suggestion
    Julia build pack which uses conda and REQUIRE.
```

repo2docker/buildpacks/julia/julia_semver.py
```diff
@@ -0,0 +1,119 @@
+# This file implements the julia-specific logic for handling SemVer (Semantic
```
willingc: Let's convert to a Python file docstring `"""` instead of "#" comments.

repo2docker/buildpacks/julia/julia_semver.py
```diff
@@ -0,0 +1,119 @@
...
+# the places where julia's SemVer handling differs from the semver package have
+# been implemented directly.
+#
+# Here, we use tuples to represent a Version, and functors as "matchers". The matcher
+# functors take a version string and return true if it passes its constraints.
+
+import re
+import semver
+
+# Main algorithm:
```
willingc: Perhaps move these lines to a class docstring for `AbstractMatcher`

tests/julia/julia_version-1.0.2/README.rst
```diff
@@ -0,0 +1,4 @@
+Julia - Project.toml: test version for julia v1.0.3
+---------------
```
willingc: Extend the length of the title

tests/julia/julia_version-default/README.rst
```diff
@@ -0,0 +1,4 @@
+Julia - Project.toml: test version for julia v1.1
+---------------
```
willingc: Extend the length of the title

repo2docker/buildpacks/julia/julia_project.py
NHDaly: Sorry that might've been my bad to put it in the wrong spot in the Julia 0.7+ PR...
I also agree that it'd make sense to move it, but I also also agree that it should be done in a new issue. :)

repo2docker/buildpacks/julia/julia_require.py
```diff
@@ -0,0 +1,170 @@
...
+
+    minor_julias = {
+        '0.6': '0.6.4',
+        '0.7': '0.7.0',
+        '1.0': '1.0.3',
+        '1.1': '1.1.0',
+    }
+    major_julias = {
+        '1': '1.1.0',
+    }
```
NHDaly: Could we extend this to include all the previous julia versions, as I'd done here?: https://github.com/jupyter/repo2docker/compare/master...NHDaly:julia_projecttoml#diff-fcf7bb456679b31cc83386be059cb3abR167

It's nice to be able to share a benchmark on an older version when investigating a performance regression, for example.
But I can also see the argument for wanting to avoid the maintenance burden of supporting these old julia versions for forever...

repo2docker/buildpacks/julia/julia_require.py
```diff
@@ -0,0 +1,170 @@
...
+            # Install and pre-compile all libraries if they've opted into it.
+            # In v0.6, Pkg.resolve() installs all the packages, but in v0.7+, we
+            # have to manually Pkg.add() each of them (since the REQUIRES file
+            # format is deprecated).
+            # The precompliation is done via `using {libraryname}`.
+            r"""
+            julia /tmp/install-repo-dependencies.jl "%(require)s"
+            """ % {"require": require}
+            # TODO: For some reason, `rm`ing the file fails with permission denied.
+            # && rm /tmp/install-repo-dependencies.jl
```
NHDaly: Sorry I haven't been following all the discussion on this PR. Was anyone ever able to resolve this? There was some weird permissions thing happening where I was getting a permissions error when i tried to clean up this file.

Now that I look closely, I'm guessing it's because this command was running as `"${NB_USER}",` up above on line 138? 

Maybe if you could make this a separate step it would work? Something like:
```suggestion
           ,
           "root",  # OR whatever this is supposed to be?
           "rm /tmp/install-repo-dependencies.jl"
```
? But I actually don't know what the right user would be. Whatever it is in the `get_build_script_files` below?

willingc:
Any ideas why Travis is failing? I tried clearing the cache for the Julia build and restarting but it's still failing in the same way.

repo2docker/buildpacks/julia/julia_semver.py
davidanthoff: I believe @betatim  wanted to go over this file once it is merged and change a couple of things into a more Pythonic style. If this is a blocking issue, maybe @betatim or @NHDaly could open a PR against this branch here with such changes?

I just copied this file over from @NHDaly's branch and probably won't have the time to do much more work on this particular code.

tests/julia/julia_version-1.0.2/README.rst
davidanthoff: Done!

tests/julia/julia_version-default/README.rst
davidanthoff: Done

docs/source/config_files.rst
```diff
@@ -71,15 +71,30 @@ To install your repository like a Python package, you may include a
...
 
+.. _Project.toml:
+
+``Project.toml`` - Install a Julia environment
+==============================================
+
+A ``Project.toml`` (or ``JuliaProject.toml``) file can specify both the
+version of Julia to be used and a list of Julia packages to be installed.
+If a ``Manifest.toml`` is present, it will determine the exact versions
+of the Julia packages that are installed.
```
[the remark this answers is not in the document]
davidanthoff: Yeah, I would suggest we update this language once we have the repo with the example all sorted out. For now it doesn't seem ideal to merge new language that points to things that don't yet exist.

docs/source/howto/languages.rst
```diff
@@ -72,11 +72,8 @@ Julia
 =====
 
 To build an environment with Julia, include a configuration file called
-``REQUIRE``. Each line of this file should include a package that you wish
-to have installed with Julia. For example, the following contents of ``REQURE``
-would install the ``PyPlot`` package with your Julia environment.::
-
-  PyPlot
+``Project.toml``. The format of this file is documented at
+`the Julia Pkg.jl documentation <https://julialang.github.io/Pkg.jl/stable/>`_.
```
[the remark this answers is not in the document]
davidanthoff: Done in the next push.

repo2docker/buildpacks/julia/__init__.py
```diff
@@ -1,170 +1,2 @@
...
-        false unless an `environment.yml` is present and we do not want to
-        require the presence of a `environment.yml` to use Julia.
-
-        Instead we just check if the path to `REQUIRE` exists
-
-        """
-        # TODO(nhdaly): Add support for Project.toml here as well.
-        return os.path.exists(self.binder_path('REQUIRE'))
+from .julia_project import JuliaProjectTomlBuildPack
+from .julia_require import JuliaRequireBuildPack
```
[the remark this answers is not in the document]
davidanthoff: Done.

repo2docker/buildpacks/julia/julia_project.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        return super().get_build_env() + [
+            ('JULIA_PATH', '${APP_BASE}/julia'),
+            ('JULIA_HOME', '${JULIA_PATH}/bin'),  # julia <= 0.6
+            ('JULIA_BINDIR', '${JULIA_HOME}'),  # julia >= 0.7
+            ('JULIA_PKGDIR', '${JULIA_PATH}/pkg'),
+            ('JULIA_DEPOT_PATH', '${JULIA_PKGDIR}'), # julia >= 0.7
+            ('JULIA_VERSION', self.julia_version),
+            ('JUPYTER', '${NB_PYTHON_PREFIX}/bin/jupyter')
+        ]
```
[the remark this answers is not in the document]
davidanthoff: Done in the next push.

repo2docker/buildpacks/julia/julia_project.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def get_path(self):
+        """Adds path to Julia binaries to user's PATH.
+
+         Returns:
+             an ordered list of path strings. The path to the Julia
+             executable is added to the list.
+
+        """
+        return super().get_path() + ['${JULIA_HOME}']
```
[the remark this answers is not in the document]
davidanthoff: Done in the next push.

repo2docker/buildpacks/julia/julia_require.py
davidanthoff: I would suggest we could do that in a new PR. Right now the REQUIRE file was just moved around in this PR, with no functionality added or removed. I think adding new functionality (like supporting more julia versions) can be a new PR once this one here is merged.

I think supporting all old julia versions might not be easy, because IJulia needs to work for all of these and I'm kind of doubtful that one can just write one generic setup code for that.

I actually also think that adding support for more old julia versions is not a good idea in general. That stuff is really legacy, was never supported here, and so I don't really see any serious use for it, with a whole lot of maintenance work.

But at the end of the day, I guess I'd be fine with either solution, as long as it comes as a new PR :)

repo2docker/buildpacks/julia/julia_require.py
davidanthoff: This is unchanged from the version on `master`. I would suggest that if someone comes up with a better solution, it can come in a separate PR.

davidanthoff:
> Any ideas why Travis is failing? 

Good old bugs :) I think I found it and fixed it. Fingers crossed...

davidanthoff:
Alright everyone, thanks so much for all the feedback! I think I addressed most of them.

Tests now pass and I think the best path forward would be to merge this at this stage and then do any additional improvements in further PRs.

A couple of issues that are still open that could fall into this bucket:
- Adding the tests for the semver code
- Making comments more pythonic
- Maybe some other pythonic changes to the semver code

Maybe someone else could try to tackle these once we have things merged?

davidanthoff:
Actually, I just pushed one more commit that cleans up something :) Fingers crossed this won't change the test results :)

repo2docker/buildpacks/julia/julia_semver.py
betatim: The change here would be
```
"""This module does something with semver"""
```
instead of `# This module does something with semver`
let's do it here. We should try and stick with "don't push tidying up to tomorrow" because we never do :-/

repo2docker/buildpacks/julia/julia_require.py
betatim: 👍 to a new issue/PR when it comes to changing the behaviour of what was already there. Just from this short discussion here it sounds like something we should ponder for a bit.

My test for punting: "is this PR better than what we have now and would one of the outcomes of the discussion lead to breakage in this PR?" in this case I think the answer yes and no -> punt. (What is the logic operator for yes and no -> yes :) )

repo2docker/buildpacks/julia/julia_require.py
betatim: Agree on punting.

---

How does `/tmp/install-repo-dependencies.jl` come to exist? My guess would be that it gets copied here by `root` and that is why NB_USER can't delete it. So I'd suggest we change the UID of the user who creates this file.

tests/julia/julia_version-1.0.2/README.rst
```diff
@@ -0,0 +1,4 @@
+Julia - Project.toml: test version for julia v1.0.3
```
betatim:
```suggestion
Julia - Project.toml: test version for julia v1.0.2
```

at least that is what is in the directory name and the `Project.toml`.

betatim:
I restarted the build as some of them failed because of timeouts. Let's see how it goes now.

betatim:
The remaining failed test is because the total coverage of the project will decrease if we merge this.

https://codecov.io/gh/jupyter/repo2docker/compare/e6cbf3e3a3e399508d5628b4ff1d30afed8c944f...c3d49966403b1f2b785450f86e33e49d7989e1db/diff shows which lines of the diff are covered and which aren't. I think it is basically the semver code so I'd be ok making an exception.

tests/julia/julia_version-1.0.2/README.rst
betatim: Fixed this myself :)

willingc:
@betatim  Merge away.

@davidanthoff Thank you so much. I'm so happy Julia will be better supported 🏖

betatim:
Done!

Thanks everyone for the efficient and enjoyable PR :)

betatim:
@davidanthoff do you want to deploy these changes to mybinder.org?

We typically make deployments on Tuesdays and Thursdays. We have instructions [on making a new deployment](https://mybinder-sre.readthedocs.io/en/latest/deployment/how.html) that should let you do it. Ping me when you need a merge.

davidanthoff:
Thanks everyone for guiding me through this process and all the help :) And the vibe here is great!!

I'm wondering whether we should tackle https://github.com/jupyter/repo2docker/issues/610 and https://github.com/jupyter/repo2docker/issues/609 before we deploy... I think both have a very, very small potential for breaking things if folks rely on corner case behavior, so if we were to sort that out ASAP, we could probably just squeeze this in before anyone even has a chance to rely on any behavior that relates to the `Project.toml`. Maybe I can take a look tonight, I don't think either would be very complicated.

betatim:
Sure thing.

PR #598, opened 2019-03-02
[MRG] Add FAQ entry and Howto on exporting conda environments

betatim:
This adds a FAQ entry and a short how-to guide to explain how a user can get a strict export of their conda environment that will work in repo2docker even if they use OSX or Windows as their operating system.

I also created https://github.com/binder-examples/conda-freeze as a "base" conda env people can use. The goal is that it installs nothing extra but triggers the conda build pack.

This is in response to https://github.com/jupyterhub/binder/issues/149 and https://github.com/jupyterhub/mybinder.org-deploy/issues/906 which both had problems with this.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -55,6 +55,34 @@ R
 Only R 3.4.4 is currently supported, which is installed via ``apt`` from the
 `ubuntu bionic repository <https://packages.ubuntu.com/bionic/r-base>`_.
 
+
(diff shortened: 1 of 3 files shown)


betatim:
We might want to adjust the command used to create and export the environment after thinking about #600. Right now I am not sure which process (clicking and browser terminal vs all-in-one complex command line invocation) is easier to follow and modify for people who aren't shell + docker + conda experts.

PR #603, opened 2019-03-04
avoid touching $HOME when bootstrapping conda as root

minrk:
<del>when REPO_DIR is set to a non-home dir, $HOME may not exist.
This can prevent the image from starting and various programs from running if $HOME isn't mounted as a volume. Further, intermediate build commands that might touch files in $HOME (even if they shouldn't). `conda env update` is currently unconditionally touching `~/.conda/environments.txt`. This should be harmless, but currently fails because $HOME doesn't exist and cannot be created by the user.
</del>

$HOME already existed and had the right permissions. The actual fix was to make sure that when we are running commands as root, we are not running with $HOME pointing to a non-root user's home directory, which causes ownership issues with any commands with caching side effects (pip, conda, etc.).

cc @yuvipanda


repo2docker/buildpacks/base.py
```diff
@@ -117,6 +117,7 @@
 # Copy and chown stuff. This doubles the size of the repo, because
 # you can't actually copy as USER, only as root! Thanks, Docker!
 USER root
+RUN mkdir "${HOME}" && chown ${NB_USER}:${NB_USER} "${HOME}"
```
yuvipanda:
```suggestion
# Explicitly create $HOME, in case $REPO_DIR is explicitly set to be somewhere
RUN mkdir "${HOME}" && chown ${NB_USER}:${NB_USER} "${HOME}"
```

repo2docker/buildpacks/base.py
```diff
@@ -117,6 +117,8 @@
 # Copy and chown stuff. This doubles the size of the repo, because
 # you can't actually copy as USER, only as root! Thanks, Docker!
 USER root
+# Explicitly create $HOME, in case $REPO_DIR is explicitly set to be somewhere
```
```suggestion
# Explicitly create $HOME as some tools like conda need it to exist and won't create it themselves
```

Suggestion on the suggestion ;)

repo2docker/buildpacks/base.py
```diff
@@ -117,6 +117,8 @@
...
+RUN mkdir "${HOME}" && chown ${NB_USER}:${NB_USER} "${HOME}"
```
yuvipanda:
```suggestion
RUN mkdir -p "${HOME}" && chown ${NB_USER}:${NB_USER} "${HOME}"
```

This fails currently with

```
mkdir: cannot create directory ‘/home/travis’: File exists
```

The -p fixes that. Also, shouldn't adduser already be creating this?

yuvipanda:
@betatim incorporated your and my suggestion :)

yuvipanda:
This doesn't actually seem to fix the issue. Checking through still...

yuvipanda:
So just before we do the mkdir, $HOME and $HOME/.conda already exists...

```
total 28
drwxr-xr-x 1 jovyan jovyan 4096 Mar  4 09:54 .
drwxr-xr-x 1 root   root   4096 Mar  4 09:52 ..
-rw-r--r-- 1 jovyan jovyan  220 Mar  4 09:52 .bash_logout
-rw-r--r-- 1 jovyan jovyan 3771 Mar  4 09:52 .bashrc
drwxr-xr-x 3 root   root   4096 Mar  4 09:56 .conda
drwx------ 3 jovyan jovyan 4096 Mar  4 09:53 .config
-rw-r--r-- 1 jovyan jovyan  807 Mar  4 09:52 .profile
```

yuvipanda:
Meh, it's *hardcoded* to the user's home directory... https://github.com/conda/conda/blob/master/conda/core/envs_manager.py#L21

yuvipanda:
I think https://github.com/jupyter/repo2docker/pull/605 will solve what I want. I'm waiting for a test to run there...

minrk:
Ah, yeah, this is probably partly our fault for running install-miniconda.bash with $HOME set to the user's home, when not running as the user. That's always a good thing to avoid, and I made sure $HOME is not set during install-miniconda.bash.

repo2docker/buildpacks/base.py
```diff
@@ -120,6 +120,12 @@
 COPY src/ ${REPO_DIR}
 RUN chown -R ${NB_USER}:${NB_USER} ${REPO_DIR}
 
+# Explicitly create $HOME as some tools like conda need it to exist
+# and will not create it if it does not exist. ${REPO_DIR} defaults
+# to ${HOME}, so this problem is hidden unless ${REPO_DIR} is set to
+# something else
+RUN mkdir -p "${HOME}" && chown ${NB_USER}:${NB_USER} "${HOME}"
```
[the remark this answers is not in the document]
minrk: Yup, you're right. When we pass the right flags to adduser, this is unnecessary.

Updated the PR to the real fix, which is to avoid touching $HOME when running install-miniconda.bash as root

Issue #604, opened 2019-03-04
Make sure there's no ${HOME}/.conda dir 


yuvipanda:
Currently, when installing miniconda we end up with a ${HOME}/.conda directory, with an environments.txt file. This is apparently to help sysadmins on multi user systems discover all the environments users have created on the system (from https://github.com/conda/conda/issues/6748#issuecomment-358706178). We don't really need that inside a container...

Having ~/.conda present causes problems when REPO_DIR isn't $HOME, since we then don't explicitly recursively chown it.

PR #605, opened 2019-03-04
Make sure $HOME/.conda is owned by user

yuvipanda:
conda puts an environments.txt file under ${HOME}/.conda, and there isn't
really a way to turn it off. We explicitly make sure they're owned by the user,
since otherwise they may be owned by root when $HOME != $REPO_DIR

Ref https://github.com/jupyter/repo2docker/issues/604

### DIFF
--- a/repo2docker/buildpacks/conda/install-miniconda.bash
+++ b/repo2docker/buildpacks/conda/install-miniconda.bash
@@ -73,4 +73,10 @@ rm ${INSTALLER_PATH}
 
 chown -R $NB_USER:$NB_USER ${CONDA_DIR}
 
+# conda puts an environments.txt file under ${HOME}/.conda, and there isn't
+# really a way to turn it off. We explicitly make sure they're owned by the user,
+# since otherwise they may be owned by root when $HOME != $REPO_DIR
+# see https://github.com/jupyter/repo2docker/issues/604
+chown -R ${NB_USER}:${NB_USER} ${HOME}/.conda
+
 conda list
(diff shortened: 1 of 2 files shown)


tests/venv/repo-path/verify
```diff
@@ -7,4 +7,8 @@ assert sys.executable == '/srv/conda/bin/python'
 
 # Repo should be in /srv/repo
 assert os.path.exists('/srv/repo/verify')
-assert os.path.abspath(__file__) == '/srv/repo/verify'
\ No newline at end of file
+assert os.path.abspath(__file__) == '/srv/repo/verify'
+
+# Verify that ~/.conda is owned by current user, and not root
+assert os.stat(os.path.expanduser('~/.conda')).st_uid == 1000
```
minrk: This should probably use `os.environ['NB_UID']` since the tests aren't always run as uid 1000 (it's the calling user's uid by default, I think)

repo2docker/buildpacks/conda/install-miniconda.bash
```diff
@@ -73,4 +73,10 @@ rm ${INSTALLER_PATH}
 
 chown -R $NB_USER:$NB_USER ${CONDA_DIR}
 
+# conda puts an environments.txt file under ${HOME}/.conda, and there isn't
+# really a way to turn it off. We explicitly make sure they're owned by the user,
+# since otherwise they may be owned by root when $HOME != $REPO_DIR
+# see https://github.com/jupyter/repo2docker/issues/604
+chown -R ${NB_USER}:${NB_USER} ${HOME}/.conda
```
minrk: It might be better to just `rm -rf $HOME/.conda` since we know we don't want any of these side effects there.

yuvipanda:
Yeah, I agree - we should rm ~/.conda at the end of miniconda install

betatim:
@yuvipanda do you want to refresh this PR?

yuvipanda:
Nah, I think we can let it go.

PR #608, opened 2019-03-06
Expand environment.yml config notes

agitter:
This small config documentation update follows from https://github.com/jupyterhub/binder/issues/152#issuecomment-469668454 and https://github.com/jupyter/repo2docker/issues/411#issuecomment-462487697  Clarifying how `environment.yml` is used may help future users avoid the confusion I had.

This update is based on my understanding of
https://github.com/jupyter/repo2docker/blob/e6cbf3e3a3e399508d5628b4ff1d30afed8c944f/repo2docker/buildpacks/conda/__init__.py#L178-L186

I avoided getting too technical in the documentation about the role of frozen environments.

### DIFF
--- a/docs/source/config_files.rst
+++ b/docs/source/config_files.rst
@@ -34,6 +34,8 @@ that lets you install any kind of package,
 including Python, R, and C/C++ packages.
 ``repo2docker`` does not use ``environment.yml`` to create and activate a new conda environment.
 Rather, it updates a base conda environment with the packages listed in ``environment.yml``.
+This means that the environment will always have the same name and not follow what is
+specified in ``environment.yml``.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


docs/source/config_files.rst
```diff
@@ -32,6 +32,8 @@ Below is a list of supported configuration files (roughly in the order of build
 ``environment.yml`` is the standard configuration file used by `conda <https://conda.io>`_
 that lets you install any kind of package,
 including Python, R, and C/C++ packages.
+``repo2docker`` does not use ``environment.yml`` to create and activate a new conda environment.
+Rather, it updates a base conda environment with the packages listed in ``environment.yml``.
```
betatim:
```suggestion
Rather, it updates a base conda environment with the packages listed in ``environment.yml``.
This means that the environment will always have the same name and not follow what is
specified in ``environment.yml``.
```

What do you think of adding something like this to explain to people when this would be relevant or a  point of confusion?

agitter:
@betatim yes that is helpful.  I accepted the proposed change with another minor edit.

I suspect there are also initialization steps that aren't carried out for some environments when `conda activate` isn't called, but I'm not sure what those are or how to describe them here.

betatim:
Merged. Thanks for helping out with making the docs better and reminding me that we do weird stuff ;)

Issue #610, opened 2019-03-06
Use JULIA_LOAD_PATH for julia env handling


davidanthoff:
https://github.com/jupyter/repo2docker/pull/595 writes a jupyter kernel that automatically loads the `Project.toml` env that is in the root. This works, but it also means that any julia shell for example does not have the environment loaded by default, because only this particular Jupyter kernel automatically loads the env.

It might be a better strategy to instead add the root folder of the repo to the `JULIA_LOAD_PATH` env variable. I believe that should make everything to all julia processes available, not just Jupyter kernels.

I would suggest we first ship the current implementation in the PR (which is not incorrect), and then investigate this option here.

betatim:
If that is the right thing to do from a Julia perspective I think we should do it.

I added the labels because I think this would make a good starter issue for someone new to repo2docker.

davidanthoff:
Actually, I think we should set `JULIA_PROJECT`, see https://docs.julialang.org/en/latest/manual/environment-variables/#JULIA_PROJECT-1.

We could even set it to just `@.`, potentially... I'm playing with this now.

PR #612, opened 2019-03-06
Use JULIA_PROJECT env variable to activate julia env

davidanthoff:
Fixes https://github.com/jupyter/repo2docker/issues/610.

This should be much more robust, plus is now also means that if one starts a julia at the terminal or something like that, all the packages from the environment are also available.

@NHDaly do you think this makes sense?


repo2docker/buildpacks/julia/julia_project.py
```diff
@@ -118,7 +122,7 @@ def get_assemble_scripts(self):
             (
                 "${NB_USER}",
                 r"""
-                julia -e "using Pkg; Pkg.add(\"IJulia\"); using IJulia; installkernel(\"Julia\", \"--project=${REPO_DIR}\", env=Dict(\"JUPYTER_DATA_DIR\"=>\"${NB_PYTHON_PREFIX}/share/jupyter\"));" && \
+                JULIA_PROJECT="" julia -e "using Pkg; Pkg.add(\"IJulia\"); using IJulia; installkernel(\"Julia\", env=Dict(\"JUPYTER_DATA_DIR\"=>\"${NB_PYTHON_PREFIX}/share/jupyter\"));" && \
```
betatim: For my education: Why set it to an empty value here? Not sure what I was expecting but explicitly setting it to be empty surprised me :)

betatim:
I think my only question (as a none Julia user) is: In what ways could we be breaking users with this change? Should we add a FAQ entry to help them or even define both `JULIA_BINDIR` and `JULIA_PROJECT` for "a few releases" to allow a transition?

repo2docker/buildpacks/julia/julia_project.py
NHDaly: I assume this is to make sure we install the default version of IJulia? But actually, that is a bit interesting... might we not also want to allow the user to specify a different version of IJulia if they so choose?

NHDaly: I wonder if we should activate and instantiate the user's project first, and then only add IJulia to the default environment if they don't already have IJulia installed? And only _then_ install the kernel?

(Note that my above comments are orthogonal to this PR)

NHDaly:
Hmm, i'm not sure if this makes sense. And in fact, I actually didn't notice that you added this `installkernel(\"Julia\", \"--project=${REPO_DIR}\",` section in #595. It's cool! But sorry i didn't comment on it when reviewing.

On the one hand, I like the spirit of it: Yeah, you're using the Project.toml to specify what your environment should look like, so yeah you'd want that environment to be activated when you run notebooks! It will be nice not to have to run `Pkg.activate(".")` at the start of every notebook.

But it is a bit unexpected to me because it's not how the default IJulia kernel behaves. So if I clone this repo and run jupyter from there locally, my notebook won't work, even though the repo has the Project.toml file in it... I would need to add `using Pkg; Pkg.activate(".")`. So I kind of feel like maybe asking the user to activate the project isn't unreasonable, at least for reproducibility reasons?

I do agree that it would be nice to avoid the boilerplate `activate(".")`, but I don't know that the tradeoff is worth it. What do you think?


-----------------

That said, I _think_ I share your feeling that setting this environment variable is better than setting that flag in the Jupyter Kernel, but I'm not very sure. I don't think we'd want the shell to open into the default environment, since presumably users would be using the shell to modify things about the notebooks.

One case that may be harmed would be if they wanted to make changes to the IJulia installation itself; it's difficult to reactivate the default environment once you've opened Julia with this flag set, and people may not know _why_ that's happening, since I don't think this flag is commonly used. I could imagine people typing `]activate` into the REPL, surprised that it doesn't drop them out of the project. (That certainly surprised me in the past.)

NHDaly:
(After some investigation, i've just discovered that you can reactivate the default environment via `pkg> activate --shared v1.1`, but I don't think most people know about that.)

repo2docker/buildpacks/julia/julia_project.py
davidanthoff: Yes, @NHDaly is correct: By setting this to `""` we make sure the IJulia package gets installed into the default environment, _not_ the user specified environment. Later, these environments are automatically stacked by julia. So if a user also had an IJulia version in their `Project.toml`, it would take precedence over the one we installed in the default environment.

I don't think we should install IJulia into the user provided env: that can potentially lead to updates of over, unrelated packages, and then we really break reproducability.

davidanthoff:
> In what ways could we be breaking users with this change? 

I think none, given that this only affects the `Project.toml` story, which hasn't really shipped? Or at least not for more than 2 days ;)

> But it is a bit unexpected to me because it's not how the default IJulia kernel behaves.

Yes, that is a good point. Lets see whether we can change the IJulia default: https://github.com/JuliaLang/IJulia.jl/pull/820.

I think at this point the main question is whether we want to activate the env in the `Project.toml` automatically or not. Right now on `master`, we do. I think this PR here keeps that behavior but implements it in a more elegant way. I can't really think of any downside of this approach vs what we have on `master`, to be honest...

So I think maybe we can merge this here, but then also keep the option open to actually not activate the environment for the notebook at all, depending on how the discussion over in the IJulia issue turns out?

I guess another question is how this works for other languages in binder? I assume that if there is a requirements.txt file in the root, and then you run a notebook in that repo, the packages from that are just available, and you don't have to somehow activate things in some way?

repo2docker/buildpacks/julia/julia_project.py
NHDaly: > I don't think we should install IJulia into the user provided env: that can potentially lead to updates of over, unrelated packages, and then we really break reproducability.

Yeah, I had the same concern. I agree with you here 100%!

> So if a user also had an IJulia version in their Project.toml, it would take precedence over the one we installed in the default environment.

But what i was concerned about is that if a user specifies a different version of IJulia in their Project.toml, we're still going to install the _kernel_ from this default version, not the one they specify. Would it be crazy to peek ahead and only install IJulia into the default env if the user doesn't include it in theirs? (Either way, this seems like a feature that could be added in a later PR)

NHDaly:
> Lets see whether we can change the IJulia default: JuliaLang/IJulia.jl#820.

Yes! I love the idea of doing this by default in other kernels as well. Great idea, @davidanthoff! :) +1


> I think at this point the main question is whether we want to activate the env in the Project.toml automatically or not. Right now on master, we do. I think this PR here keeps that behavior but implements it in a more elegant way. I can't really think of any downside of this approach vs what we have on master, to be honest...

The only potential downside I could see is that this PR enables the same behavior for julia shells opened in the terminal, which is also a bit unexpected and different from the normal `julia` behavior. I don't feel strongly one or the other, but my preference would be towards only doing this in the notebooks, not the shell, via the flag, as things already are on `master`.


-----

One more question: is `REPO_DIR` the destination _after_ everything is copied? I don't remember. I just want to make sure that if the user installs more packages via the shell, or manually modifies the Project.toml file that is next to the notebooks, those changes will be reflected (after a `pkg"instantiate"`) right?

davidanthoff:
So an update on this one: IJulia now merged this behavior, so for Jupyter we should be in the clear with auto activating the environment.

So the question now is whether we also want to auto-activate the environment for any other julia process, i.e. a non-IJulia kernel, or not. I think we should, actually :) Isn't the whole point of binder that you get a pre-populated environment?

One question that might be good to sort out is how users can actually start a julia process other than an IJulia kernel? I'm just not familiar enough with the binder setup, maybe @betatim can shed some light on this? Can one e.g. SSH into a binder instance?

NHDaly:
This might be a silly question, but now that IJulia has merged this behavior, can we just rely on IJulia to do it by default, instead of implementing it ourselves?

-------------

> One question that might be good to sort out is how users can actually start a julia process other than an IJulia kernel? I'm just not familiar enough with the binder setup, maybe @betatim can shed some light on this? Can one e.g. SSH into a binder instance?

One way that I can think of:
Binder just creates a normal Jupyter notebook server, and the Jupyter notebook UI has the ability to launch a terminal (in the browser) that's running on the host. It's essentially like SSHing into the instance, but accessing it through an in-browser remote terminal UI instead.
You can get there by clicking new->terminal:
Screen Shot 2019-03-14 at 3 10 47 PM


And it looks like this:
Screen Shot 2019-03-14 at 3 10 54 PM
(For me the link was: https://hub.mybinder.org/user/nhdaly-jupyter-binder-040a6vbx/terminals/1, but that seems to be instance-specific.)

So the question is, if we open the Julia REPL through that terminal, should it activate your environment?

---------

My gut is to vote No, since that's not how the terminal behaves on your local computer. And anyone who's managed to put a Project.toml file in their repo knows how to activate a directory. But I do also see how it makes sense from a usability perspective: why would you ever NOT want your project activated? Also, since this docker instance is short lived, and about to be killed, I can't see any harm in having your project activated when you didn't expect it. You're not going to break anything!


SO I guess what i'm saying is yeah I think your change makes sense, @davidanthoff. :) I've talked myself into it, and now I vote Yes!

davidanthoff:
I think we should now first merge https://github.com/jupyter/repo2docker/pull/622, and then I can resolve any conflicts between that and this PR here, and then we can merge this PR here.

davidanthoff:
This is ready from my end. @NHDaly do you want to review this final version?

I also removed `JULIA_BINDIR` in this PR because it really is a julia 0.6 thing that serves no purpose for the Project.toml based story.

davidanthoff:
Great, thanks! @betatim ready for your review/merge!

betatim:
Thanks! LGTM -> merged.

davidanthoff:
Great, thanks!

PR #613, opened 2019-03-07
[MRG] Add tests for Julia semver matching

betatim:
This is a follow up to #595 that adds lots of the tests written by @NHDaly into our test suite.

I also applied `black` as code formatter.


betatim:
Ready for reviewing. @willingc in the original PR you had commented on using docstrings in the semver module, what do you think of this?

repo2docker/buildpacks/julia/semver.py
```diff
@@ -0,0 +1,162 @@
...
+        """Check if `v` matches the constraint"""
+        while len(v) < 3:
+            v = v + (0,)
+        v_str = ".".join(map(str, v))
+        return semver.match(v_str, self.constraint_str)
+
+    def __eq__(self, rhs):
+        return self.constraint_str == rhs.constraint_str
+
+    def __repr__(self):
```
betatim: Should this be `__str__` (create a human readable representation) instead of `__repr__` (create sth that is as close as possible to code you can paste to recreate the object).

tests/unit/test_semver.py
```diff
@@ -0,0 +1,210 @@
...
+def test_major_minor_patch():
+    V = (1, 2, 3)
+    assert (semver.major(V), semver.minor(V), semver.patch(V)) == (1, 2, 3)
+    assert semver.major((1,)) == 1
+    assert semver.minor((1,)) == 0
+    assert semver.patch((1,)) == 0
+
+
+def test_simple_matches():
+    assert repr(semver.create_semver_matcher("1.2.5")) == "[1.2.5-2)"
```
betatim: Is there a minimal change we could make to remove the need for using `repr` so much in the tests?

It somehow doesn't feel right. The best idea I have is to teach the matcher classes how to compare themselves to a version string or version tuple in `__eq__`. But I also feel like we are drifting off into the finer points of Python and it might not be worth the time (-> let's make a low priority/getting started issue out of this).

tests/unit/test_semver.py
```diff
@@ -0,0 +1,210 @@
...
+
+def test_major_minor_patch():
+    V = (1, 2, 3)
+    assert (semver.major(V), semver.minor(V), semver.patch(V)) == (1, 2, 3)
+    assert semver.major((1,)) == 1
+    assert semver.minor((1,)) == 0
+    assert semver.patch((1,)) == 0
+
+
+def test_simple_matches():
```
willingc: In a future refactor, we should use pytest's parameterize capability to feed in versions to get closer to DRY. [See example from JupyterHub tests](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/tests/test_proxy.py#L169).

repo2docker/buildpacks/julia/semver.py
```diff
@@ -0,0 +1,162 @@
...
+        return self.constraint_str
+
+
+class VersionRange:
+    """Test if a version tuple is between a lower and upper bound
```
willingc:
```suggestion
    """Represents a range of release versions.
    
    A `VersionRange` contains versions from a `lower` to `upper` bound
    which may be inclusive (default: `exclusive=False`) or exclusive (`exclusive=True`).
    
    A release version (represented by a tuple) can be checked to see if it falls within a `VersionRange`
   """
```

repo2docker/buildpacks/julia/semver.py
```diff
@@ -0,0 +1,162 @@
...
+
+    A VersionRange represents a range of versions from `lower` to `upper`
+    and lets you test if a version falls in that range. The `upper` bound
+    can either be exclusive (exclusive=True) or inclusive (exclusive=False).
+    """
+
+    def __init__(self, lower, upper, exclusive):
```
willingc: Let's set a sensible default for exclusive which I would recommend to be inclusive.

repo2docker/buildpacks/julia/semver.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # Convert special comparison strings to format accepted by `semver` library.
+    constraint_str = constraint_str.replace("≥", ">=").replace("≤", "<=")
+    constraint_str = re.sub(r"(^|\b)=\b", "==", constraint_str)
+
+    return SemverMatcher(constraint_str)
+
+
+class SemverMatcher:
+    """Match a version tuple to a given `constraint_str`
```
willingc:
```suggestion
    """Provides a utility for using `semver` package to do version matching.
    
    The `SemverMatcher` takes a `constraint_str` to represent a regex to determine if
    a version tuple matches the constraint.
```

repo2docker/buildpacks/julia/semver.py
NHDaly: Sure! I had no idea what was best for this

repo2docker/buildpacks/julia/semver.py
```diff
@@ -0,0 +1,162 @@
+"""
```
NHDaly: Just curious, why did you rename this from `julia_semver` to `semver`? I _think_ the logic in this file is fairly specific to `julia`, and it'd be a shame if someone accidentally used it for parsing semvers in an unrelated context.. What do you think?

repo2docker/buildpacks/julia/semver.py
willingc: I would recommend keeping `__repr__` as is and adding a `__str__` a human readable representation would be helpful.

repo2docker/buildpacks/julia/semver.py
NHDaly: Also, we're importing a package named `semver` from a file named `semver` which just feels weird to me. But I don't write much python these days, so I might also just not be as comfortable in python land.

repo2docker/buildpacks/julia/semver.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def patch(v):
+    return v[2] if len(v) >= 3 else 0
+
+
+def create_semver_matcher(constraint_str):
+    """Create a matcher that can be used to match version tuples
+
+    Version tuples are matched against the provided `constraint_str`.
```
willingc:
```suggestion
    Version tuples are matched against the provided regex `constraint_str`.
```

repo2docker/buildpacks/julia/semver.py
```diff
@@ -0,0 +1,162 @@
...
+
+import re
+
+import semver
+
+
+# Main algorithm:
+# Create an AbstractMatcher instance from the constraint string, and check it
+# against each version in the versions_list, returning the first success.
+def find_semver_match(constraint, versions_list):
```
willingc:
```suggestion
def find_semver_match(constraint, versions_list):
    """Finds the first match in a list of versions that match a constraint
    
    Use the `constraint` regex string to create a `SemverMatcher` which
    will be used to check for the first version tuple in the `versions_list` to
    satisfy the constraint. Returns `None` if there is no matching version.
    """
```

willingc: Remove the comments above the function declaration.

repo2docker/buildpacks/julia/semver.py
```diff
@@ -0,0 +1,162 @@
+"""
+Julia specific handling of SemVer strings
+
+It uses the python "semver" package to do most version string comparisons, but
+the places where julia's SemVer handling differs from the semver package have
+been implemented directly.
+"""
+
+# We use tuples to represent a Version, and functors as "matchers". The
```
willingc: Move this into the file docstring as a paragraph.

repo2docker/buildpacks/julia/semver.py
betatim: Mostly because I don't think anyone will use it outside of this part of the package. And if they do they'd be writing something like `from repo2docker.buildpacks.julia.semver import foo` or `from repo2docker.buildpacks.julia import semver`. If you want to use this alongside the other semver package you can always do `from repo2docker.buildpacks.julia import semver as julia_semver`.

Relying on the namespacing of python modules.

repo2docker/buildpacks/julia/semver.py
betatim: Leaving as is then.

tests/unit/test_semver.py
betatim: Good idea -> #615

betatim:
All comments resolved. Thanks for the language help and the idea of using pytest's parametrize.

willingc:
Yay!!!! Thanks @betatim @NHDaly

PR #619, opened 2019-03-10
Allow URLs for Julia package requirements

sebasguts:
This allows installing a Julia module by URL instead of name.

The main benefit is that one can also install non-registered Julia modules. The solution is not perfect, since the master branch is always used. It is however an improvement over not being able to prove URLs.

### DIFF
--- a/repo2docker/buildpacks/julia/install-repo-dependencies.jl
+++ b/repo2docker/buildpacks/julia/install-repo-dependencies.jl
@@ -21,6 +21,8 @@ else
         version = try; reqline.versions.intervals[1].lower; catch; emptyversionlower; end
         if version != emptyversionlower
           Pkg.add(PackageSpec(name=pkg, version=version))
+        elseif startswith(pkg,"http://") || startswith(pkg,"https://")
+          Pkg.add(PackageSpec(url=pkg, rev="master"))
         else
           Pkg.add(pkg)
         end


betatim:
Thanks for adding this, we should find a way to let people do this. It would be good to discuss what other options we have for achieving the same goal.

The mission of repo2docker is to make it easier to "do the right thing" in terms of making the environment reproducible than not making the environment reproducible. We do have several escape hatches that let people who know what they are doing, doing wild things, the hope is that by making it obvious you are using an escape hatch you are starting to get onto thinner ice. With that in mind:

* is it a good idea (maybe it is common practice in the Julia community) to allow people to do this from a `REQUIRE`? In Python land you an specify URLs in `requirements.txt`.
* could we use [`postBuild`](https://repo2docker.readthedocs.io/en/latest/config_files.html#postbuild-run-code-after-installing-the-environment) to let people achieve the goal of installing from a URL while making it slightly less convenient to do (signalling "you are using an escape hatch!") and letting them specify a version/SHA/tag instead of only master?
* how do the answers change now that we have support for `JuliaProject.toml` and `Project.toml`?

The one thing I don't like about this PR is that I am not a huge fan of only being able to install `master` because not explicitly specifying a version is one of the roads-to-not-reproducible-land. Given that, it feels weird to enable people to do this :-/

sebasguts:
Thank you for your feedback.

So, at the moment, both the standard `REQUIRE` nor the `Project.toml` seem to be able to actually digest URLs (see [here](https://github.com/JuliaLang/Pkg.jl/pull/1088) and [here](https://github.com/JuliaLang/Pkg.jl/issues/492) for maybe the most comprehensive write-up).

As you suggested, using `postBuild` can actually be used to install those packages, even with a specific rev/tag, so one can even get off the ice here.

About general reproducability:
repo2docker supports `REQUIRE` (I did not check for `Project.toml`) files without any version for the package dependencies defined, which means it will download whatever version is the current one in the Julia METADATA.jl. Is that already a reproducibility issue? Maybe there should be a warning if there is no version specified? I can provide a PR if you like.

With that being said, there is probably no need to keep this PR open, or even merge it. `REQUIRE` files are more or less deprecated in Julia, and this would be an extension to the standard files only for repo2docker, so probably not that necessary, as `postBuild` will do the job. Once [this PR for Julia](https://github.com/JuliaLang/Pkg.jl/pull/1088) is finished, repo2docker should probably also support it.

betatim:
Sounds good.

I think making a custom version of `REQUIRE` should be a no no. If standard `REQUIRE` files allow people to install unversioned packages we have to also support it. I just think we shouldn't add more things.

Should we make a FAQ entry that points people towards using `postBuild` to solve this problem?

davidanthoff:
I believe our existing `Project.toml` code in binder should fully support unregistered packages. All you have to do is also checkin the `Manifest.toml`, and it should all work. So if you have a custom env and you do something like `]add https://github.com/bla/bla.jl.git#v1.3.4`, it should all work, including complete reproducability.

I also agree, we shouldn't add more non-standard things to REQUIRE in binder, it really is legacy technology, and the `Project.toml` & `Manifest.toml` story is the way to go with this.

NHDaly:
+1 For full reproducibility you should just add both your `Project.toml` and `Manifest.toml` to git and (as long as you haven't `dev`'d anything) it should all just work!! :D

sebasguts:
Thank you all for your comments. Indeed, using the `Manifest.toml` in addition to the `Project.toml` works fine. That is a perfect solution, I think!

Issue #620, opened 2019-03-12
IJulia kernels installed in $HOME not /srv


rabernat:
I am using the latest repo2docker to install a Julia env using a Project.toml file ([binder config here](https://github.com/pangeo-data/pangeo-stacks/tree/master/pangeo-ocean/binder)).

I find that the IJulia kernels are installed in my home directory
```bash
$ jupyter kernelspec list
Available kernels:
  julia-1.1    /home/jovyan/.local/share/jupyter/kernels/julia-1.1
  python3      /srv/conda/share/jupyter/kernels/python3
```

It would be better if it were installed in `/srv` next to the python kernel. In particular, this causes problems in JupyterHubs which mount other volumes in $HOME; in that case the julia kernelspec disappears.

Examining the [IJulia install script](https://github.com/JuliaLang/IJulia.jl/blob/master/deps/kspec.jl), it looks like we could fix this by setting either the environment variable `JUPYTER_DATA_DIR=/srv/conda/share/jupyter` or `XGD_DATA_HOME=/srv/conda/share/jupyter`.

This issue is very similar to, but distinct from, #554.

betatim:
cc @NHDaly and @davidanthoff as resident Julia experts.

PR #621, opened 2019-03-12
[WIP] tell IJulia to install jupyter kernels in /srv

rabernat:
This PR adds an environment variable which tells IJulia to put its kernelspec in /srv
Potential fix for #620.

Todo:

- [ ] is this is the best way to find the kernel path?
- [x] figure out how to test
- [ ] update `docs/source/changelog.rst`

### DIFF
--- a/tests/julia/julia_version-default/verify
+++ b/tests/julia/julia_version-default/verify
@@ -4,6 +4,14 @@ if VERSION < v"1.1.0"
     exit(1)
 end
 
+# Verify that kernels are not installed in home directory (issue #620)
+try
+    using IJulia
+    @assert IJulia.kerneldir() == "/srv/conda/share/jupyter/kernels"
+catch
+    exit(1)
+end
+
 using Pkg
 pkg"activate ."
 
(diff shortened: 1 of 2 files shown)


tests/julia/julia_version-default/verify
```diff
@@ -4,6 +4,14 @@ if VERSION < v"1.1.0"
     exit(1)
 end
 
+# Verify that kernels are not installed in home directory (issue #620)
+try
+    using IJulia
+    @assert IJulia.kerneldir() == "/srv/conda/share/jupyter/kernels"
+catch
+    exit(1)
+end
```
rabernat: Is this test okay? Does it need to be copied to the other Julia versions?

davidanthoff:
@rabernat Sorry, I just saw this now, and had also started a PR: https://github.com/jupyter/repo2docker/pull/622.

I'm kind of surprised that your PR here works because I would have thought that [this](https://github.com/jupyter/repo2docker/blob/8dfd400c779763d0051975454050f22c1f19daf0/repo2docker/buildpacks/julia/julia_project.py#L121) line overwrites the env var that you set... But your PR passes tests, so now I'm a bit confused :)

rabernat:
Yes I agree that is weird! No idea what is going on here. I'm new to Julia packaging, so don't have much insight.

Your solution in #622 is probably better. Perhaps you could add my test to that?

davidanthoff:
Cool! I copied your tests over to #622 and also put them into all the julia tests, so that we also test the legacy `REQUIRE` case.

davidanthoff:
I think this PR can be closed, we are going with https://github.com/jupyter/repo2docker/pull/622 instead.

PR #623, opened 2019-03-13
[MRG] Julia: Two small doc updates

fredrikekre:
Two doc-changes as a follow up on #595:
- update links to Pkg.jl documentation;
- change `Compat` to `[compat]` since that is usually how we refer to said section.


betatim:
Thanks for taking the time to update the docs! Also: congrats on your first of many PRs here :)

PR #628, opened 2019-03-23
[MRG] Bump notebook package version

### LINKED ISSUE #626: Packages picked up from the `conda-forge/label/broken`
It seems that the addition of `conda-forge/label/broken` channel even as a lower-priority channel causes packages to be picked up from that channel.

An example of repository impacted is xeus-python: https://mybinder.org/v2/gh/QuantStack/xeus-python/stable?filepath=notebooks/xeus-python.ipynb

### LINKED ISSUE #627: Update the notebook in the base image
From conversation in #626 it sounds like the version of notebook that repo2docker currently uses is not working. I'm creating this issue to make sure we keep track of that issue (or we can close this if it's already been fixed, but I didn't notice a PR that seemed to suggest this)

betatim:
This updated the version of the `notebook` package and others. The `freeze.py` script wouldn't run because of some path gubbins so I've updated it as well.

Fixes #627 
Fixes #626


repo2docker/buildpacks/conda/environment.frozen.yml
```diff
@@ -1,87 +1,86 @@
-# AUTO GENERATED FROM environment.py-3.7.yml, DO NOT MANUALLY MODIFY
-# Frozen on 2019-02-11 20:26:36 UTC
+# AUTO GENERATED FROM /Users/thead/git/jupyterhub/repo2docker/repo2docker/buildpacks/conda/environment.py-3.7.yml, DO NOT MANUALLY MODIFY
```
choldgraf: why did the custom path to your local file get added here?

repo2docker/buildpacks/conda/environment.py-3.6.yml
```diff
@@ -1,11 +1,11 @@
-# AUTO GENERATED FROM environment.yml, DO NOT MANUALLY MODIFY
-# Generated on 2019-02-11 20:22:50 UTC
+# AUTO GENERATED FROM /Users/thead/git/jupyterhub/repo2docker/repo2docker/buildpacks/conda/environment.yml, DO NOT MANUALLY MODIFY
```
choldgraf: same here

repo2docker/buildpacks/conda/environment.yml
```diff
@@ -1,9 +1,9 @@
 dependencies:
   - python=3.7.*
-  - ipywidgets==7.2.1
-  - jupyterlab==0.34.9
-  - nbconvert==5.4.0
-  - notebook==5.7.4
+  - ipywidgets==7.4.2
+  - jupyterlab==0.35.4
+  - nbconvert==5.4.1
```
choldgraf: are these other packages being bumped because they need to be bumped for the notebook to work?

repo2docker/buildpacks/conda/freeze.py
```diff
@@ -67,15 +69,15 @@ def freeze(env_file, frozen_file):
         f"continuumio/miniconda3:{MINICONDA_DOCKER_VERSION}",
         "sh", "-c",
         '; '.join([
-            'set -e',
+            'set -ex',
             f"conda install -yq conda={CONDA_VERSION}",
             'conda config --add channels conda-forge',
             'conda config --system --set auto_update_conda false',
-            f"conda env create -v -f /r2d/{env_file} -n r2d",
+            f"conda env create -v -f /r2d/{env_file_in_container} -n r2d",
```
choldgraf: did something change that made these stop working? I feel like I've run the freeze script before and it worked OK...just wanna make sure that there isn't actually a different issue at play here

betatim:
As far as I can tell from testing locally there is a notebook package for Python 3.5 and 3.7 that works. I also built binder-examples/conda and binder-examples/requirements which worked and notebooks launched.

Let's see if travis is happy as well.

repo2docker/buildpacks/conda/freeze.py
betatim: I don't understand why this stopped working. But I also didn't go look at the history of the file to figure it out. The problem is that `env_file` now contains the full local path but we need the "relative to the path that got mounted" path.

repo2docker/buildpacks/conda/environment.yml
betatim: I updated them because they were quite old. Didn't check if we need to. I don't know what our policy is for bumping these dependencies so maybe they shouldn't get changed if it isnn't for security fix reasons (like the notebook).

repo2docker/buildpacks/conda/environment.frozen.yml
betatim: This is another way the "paths got broken" issue raises its ugly head. I'll fix this.

repo2docker/buildpacks/conda/freeze.py
betatim: ecb4f28 addresses this. I'll mark this as resolved so we can start a new one if we need to.

choldgraf:
sounds good - if travis is happy then I am happy (seems reasonable to bump the other dependencies too, just wasn't sure if it was a requirement)

betatim:
Travis fails because the new notebook gives an exit code of 137 when it gets killed. Investigating what the proper fix is.

betatim:
This updates the test so that it shutsdown the container but doesn't care how it exits.

choldgraf:
I am....uncertain whether that's a problem or not :-/  maybe @minrk or @yuvipanda could give their opinion on whether this is a bad idea? Otherwise I'm still +1 on the PR

betatim:
I think it is weird and needs looking at but it isn't "dangerous". It seems if you do the equivalent sequence via `docker run`, `docker exec`, `docker stop` it works as expected.

I'll merge now so we can deploy this on Sunday morning (now) as I think we should bump the build prefix which causes a lot of rebuilds and load -> would be good to start that at a quiet time of the week.

minrk:
If the test is passed by the time we kill it, then that should be fine. I think we only care about early exits.

PR #631, opened 2019-03-27
[MRG] Python 3.7 as the default in the docs

jtpio:
The latest version of repo2docker (`0.8.0`) seems to be using Python 3.7 as the default.

This change updates the docs to reflect that.


betatim:
Thank you for this. Sorry we missed updating the docs :-(

(restarted travis, it timedout fetching packages from a apt-get mirror)

PR #633, opened 2019-03-29
[MRG] Update the source links used by the legacy buildpack

betatim:
Debian Jessie is now "archived" so we need to update the links to the
sources.

This fixes the build failure we are seeing in #631.


choldgraf:

https://discourse.jupyter.org/t/debian-update-url-broken-in-repo2docker-legacy-builds/627/5

PR #635, opened 2019-04-02
[WIP] add --no-default-build flag

shibbas:
Enable users to disable the default build for when there are no special requirements found. This allows users to take other actions if no environment is specified.

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -198,6 +198,13 @@ def get_argparser():
         default=[],
     )
 
+    argparser.add_argument(
+        '--no-default-build',
+        dest='no_default_build',
+        action='store_true',
+        help='Do not run a default build if no configuration files were detected.'
+    )
+
     return argparser
 
 argparser = get_argparser()
@@ -323,6 +330,8 @@ def make_r2d(argv=None):
 
     if args.target_repo_dir:
         r2d.target_repo_dir = args.target_repo_dir
+
+    r2d.no_default_build = args.no_default_build
 
     return r2d
 
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -354,6 +354,14 @@ class Repo2Docker(Application):
         config=True
     )
 
+    no_default_build = Bool(
+        False,
+        help="""
+        Don't run a docker build if no configuration files were detected
+        """,
+        config=True
+    )
+
     def fetch(self, url, ref, checkout_path):
         """Fetch the contents of `url` and place it in `checkout_path`.
 
@@ -620,7 +628,7 @@ class Repo2Docker(Application):
                 # no need to build, so skip to the end by `return`ing here
                 # this will still execute the finally clause and let's us
                 # avoid having to indent the build code by an extra level
-                return
+                return True
 
             if self.subdir:
                 checkout_path = os.path.join(checkout_path, self.subdir)
@@ -636,6 +644,8 @@ class Repo2Docker(Application):
                         picked_buildpack = bp
                         break
                 else:
+                    if self.no_default_build: # return here if asked not to use the default buildpack.
+                        return False
                     picked_buildpack = self.default_buildpack()
 
                 picked_buildpack.appendix = self.appendix
@@ -681,12 +691,14 @@ class Repo2Docker(Application):
             # Cleanup checkout if necessary
             if self.cleanup_checkout:
                 shutil.rmtree(checkout_path, ignore_errors=True)
+        
+        return True
 
     def start(self):
-        self.build()
+        image_built = self.build()
 
-        if self.push:
+        if self.push and image_built:
             self.push_image()
 
-        if self.run:
+        if self.run and image_built:
             self.run_image()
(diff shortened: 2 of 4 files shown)


betatim:
This is a nice PR! Could you explain a bit more what you want to achieve and what the motivation is for this change?

As a "user" I am confused about what `--no-default-build` is meant to do or when it would be useful so maybe we can expand the help text or give it a different name? In particular how it is different from `--no-build`. It seems related to buildpacks?

betatim:
Also, I missed the [WIP] tag, so if things are in progress and will become clearer later I'll wait :)

shibbas:
Thanks for doing a review! Up for suggestions on the on the name. 

I wanted to make it so that when there are no "special" requirements needed for a project, the user can take other actions as needed for their scenarios, for e.g. reduce the time to container by using a pre-built image for example in binderhub scenarios.  It differs from `--no-build` because it will still build when there are configuration files detected in the repo. Just won't do a build when nothing is found. Let me know if there are any other questions.

betatim:
I made a local directory that contains some files, none of them trigger a build pack, checked out this PR and ran it:
```
$ ls /tmp/some-repo
$ repo2docker --no-default-build /tmp/some-repo
Using local repo /tmp/some-repo.
$
```
and there is no new docker image.

If I use `master` instead on the same repo it does build an image with all the basics needed to run a jupyter notebook server but not (much) more.

I like the idea of faster builds, but I am not sure I understand how this is meant to work :-/

betatim:
>  It differs from `--no-build` because it will still build

Sorry I wasn't clear enough, the similarity to `--no-build` is in the naming, which is one reason I think we should try and come up with a different name. Because as you say they aren't related.

shibbas:
Ah. Thanks, yeah, a new name makes sense. Let me think about something that's more clear. 

Yes, with this PR, no image will be generated when no build packs are picked. The idea is that the user can do something else after running repo2docker when no image is generated, like use another image that they already have access to for example (instead of building a new one even if it's basic). Or throw an error just as another basic example.

shibbas:
new name suggestions, let me know if something sounds clearer based on my explanations and i'll update the PR to reflect it: `--skip-default-image` or `--skip-default-build-pack`

betatim:
> Yes, with this PR, no image will be generated when no build packs are picked.

Ok, why use repo2docker only to fetch a repo and do nothing else with it? Especially as all the things which have been fetched will be deleted as soon as repo2docker exits. Which means you couldn't use it in another image. Could you give your concrete use case as an example maybe?

shibbas:
Basically, my scenario is:
1. Have a private registry to store my images. 
2. Given a github repo, check if there are special configuration files to run the notebooks.
3. If yes, build the image and push it to my private registry using `--push`
4. If no, don't build and push an image, just use a previously built default image.

Since repo2docker already does a great job detecting if there are any configuration files, i don't want to replicate this outside of repo2docker to determine if I should even call repo2docker. But i don't want it to build another image if there are no special requirements as well, which is what it does currently without my change.

With `--skip-default-image`, the idea is that if no configuration files are detected, no further action is taken. As a result, no image will be pushed to the registry, and i can detect that and just fall back to using my default image. Basically use the same image for all repositories that don't need a specific environment, e.g. just have a generic python image.

yuvipanda:
This is a very useful feature! From my understanding, the idea is to really check if a repo (or directory) will trigger any particular build pack we have, or if it'll just trigger the default. Is that right?

My recomendation in this case would be to have a flag that'll just output the name of the buildpack that's triggered, saying 'default' if it's the default buildpack. This will be more generally useful for the use case of 'detect what kinda stuff this repo has', rather than specifically not doing anything. What do you think?

betatim:
How about making it a second entrypoint called `repo2docker-detect` that returns the name of the build pack detected? And/or putting as much as possible inside the "library" part of repo2docker so that you could do something like `import repo2docker; print("Build pack is:", repo2docker.detect())`. 

Would that open the door to it not even downloading the source (where possible) as a future optimisation?

shibbas:
Thanks! More than just detecting, I'd like to take action after the detection; if it's not default, then build an image; otherwise don't do anything so the caller can make a call as to what to do.  

To understand the suggestion better, is the flow being recommended a two-step process?  
1. run repo2docker with something like `--detect-build-pack` or `repo2docker-detect` 
2. if the answer is anything but `default`, call repo2docker again to generate the build image?

yuvipanda:
@shibbas yep, that's the workflow I'd like. Since you have something calling repo2docker already, it should be able to do this. It'll also help other workflows that might wanna do something similar.

willingc:
Hi @shibbas :sunny:

A group of repo2docker maintainers are meeting this weekend and are triaging PRs. It looks like this WIP PR left off with a conclusion that a two-step process would be preferred.

We've opened a new issue (#783) that gives a place to discuss more specifics and get input from the community.

Would you like to keep this WIP PR open for future work? If not or if it would be easier work from a new PR, please let us know and we would be happy to close this PR.

manics:
Closing based on last comment

Issue #438, opened 2018-10-13
JupyterLab 0.35.1 is available


blink1073:
This is available for use in repo2docker, if you're ready to upgrade.

choldgraf:
+1 (if anybody has a moment to do this, [here are the instructions for bumping default versions](https://github.com/jupyter/repo2docker/blob/master/CONTRIBUTING.md#update-and-freeze-buildpack-dependencies))

jzf2101:
Don't we need to cut a release as well to put it on pip?

jzf2101:
How would we deal with freezing the bumping the conda environment when jlab requires python 3.6 and not other versions?

yuvipanda:
Closing in favor of #724, since we already have 0.35 installed.

PR #637, opened 2019-04-04
[WIP] Bump conda version 

### LINKED ISSUE #593: Changing conda versions / conda solver speed ups
Just came across this [tweet about conda solver speed issues](https://twitter.com/luizirber/status/1099046949758664705) and liked the two linked articles. It made me wonder if we should give conda 4.6 a spin (with the new solver). Which made me wonder: what is our process for upgrading conda? I understand the technical steps involved in swapping out conda versions but I am not sure I know of all (or even most) social implications (mostly concerned with breaking repo builds that currently work).

Instructions on how to help out with this https://github.com/jupyter/repo2docker/issues/593#issuecomment-476543221

jhamman:
This bumps the miniconda version to 4.5.12 (latest) and the conda version to 4.6.11 (latest). Conda 4.6 included a number of performance improvements that should help speed up install times for complex environments. It also included some improvements to conda's interoperability with pip. An overview of the conda release is available here: https://www.anaconda.com/conda-4-6-release/.

Fixes #593

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -13,7 +13,7 @@ New features
   :user:`davidanthoff`
 - Set JULIA_PROJECT globally, so that every julia instance starts with the
   julia environment activated in :pr:`612` by :user:`davidanthoff`.
-- Update Miniconda version to 4.5.12 and Conda version to 4.6.10  in :pr:`637` by
+- Update Miniconda version to 4.5.12 and Conda version to 4.6.9  in :pr:`637` by
   :user:`jhamman`
 
 API changes
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -13,6 +13,8 @@ New features
   :user:`davidanthoff`
 - Set JULIA_PROJECT globally, so that every julia instance starts with the
   julia environment activated in :pr:`612` by :user:`davidanthoff`.
+- Update Miniconda version to 4.5.12and Conda version to 4.6.11  in :pr:`637` by
+  :user:`jhamman`
 
 API changes
 -----------
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -13,7 +13,7 @@ New features
   :user:`davidanthoff`
 - Set JULIA_PROJECT globally, so that every julia instance starts with the
   julia environment activated in :pr:`612` by :user:`davidanthoff`.
-- Update Miniconda version to 4.5.12and Conda version to 4.6.11  in :pr:`637` by
+- Update Miniconda version to 4.5.12 and Conda version to 4.6.10  in :pr:`637` by
   :user:`jhamman`
 
 API changes
--- a/repo2docker/buildpacks/conda/environment.frozen.yml
+++ b/repo2docker/buildpacks/conda/environment.frozen.yml
@@ -1,5 +1,5 @@
 # AUTO GENERATED FROM environment.py-3.7.yml, DO NOT MANUALLY MODIFY
-# Frozen on 2019-03-23 18:02:15 UTC
+# Frozen on 2019-04-04 20:24:39 UTC
 name: r2d
 channels:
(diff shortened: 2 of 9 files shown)


betatim:
Thanks for helping get this done! I edited your comment to include the issue that has been tracking this so it gets closed.

We should also refreeze our environments per https://github.com/jupyter/repo2docker/issues/593#issuecomment-476553639

What is your thinking behind bumping conda but not miniconda? My expectation was that we'd keep them in sync unless there is a reason not to.

betatim:
The answer to "why not also update miniconda" is that there isn't a miniconda 4.6! See https://repo.anaconda.com/miniconda/

betatim:
I think the TL;DR is that the new conda requires a Python version newer than 3.5. So we need to think of a way to deal with that.

I think the `conda/simple-py2/verify::build FAILED` failure is due to a network issue and will resolve itself. That's why the "verify" part of that test worked even though the build part had just previously failed.

betatim:
The new version of conda doesn't work with Python 3.5 any more. At least I can't get it installed in a conda env with Python 3.5. First thought this was because conda-forge isn't building new packages for 3.5 any more but even without using conda-forge this seems to happen. There is nothing in the conda change log about this so I am suspicious/think I've made a mistake somewhere.

I think the way forward is that we need to start treating Python 3.5 separately. It already has a "deep frozen" `environment.yml` to setup the basic env. I am wondering if it is possible to adopt the same method we use for Python 2: have a "user env" and a "server env". The server env is always a modern Python 3, independent of what the user requested. Would it work to install conda there and use it in the Pythonn 3.5 env?

yuvipanda:
@betatim I agree re: separate server / user env, (or 'notebook' env and 'kernel' env) similar to python2. It'll also help with stuff like https://github.com/jupyter/repo2docker/issues/438

betatim:
After the pointer to the Jupyter lab issue I think we will want to have separate user and kernel envs for everything. Right now we do it as a special case for Python 2. Do you think that is a good goal?

I think the downsides include: harder for users to install server extensions, impossible to organise for repositories that currently install server extensions to continue working, bigger docker images.

Advantages include: we don't have to special case certain setups, harder for users to break their kernel/notebook server

Is my assumption that notebook server extensions would need installing in the kernel env correct?

Overall: we have a few tricky trade-offs ahead of us. Is there a way to upgrade conda (resolve this PR) without having to first also have a decision on all the trade-offs? Maybe we spend a bit of extra effort to do so but I think this particular issue is worth resolving "now" instead of in a few weeks, even if it costs a bit extra in engineering effort.

jhamman:
I'm going to defer to you all on how to proceed here. FWIW, a refactor of the user/kernel envs is probably beyond what I can help with right now. If it makes sense to put this on hold until that is addressed, fine by me.

yuvipanda:
@betatim yeah, if we have a 'notebook' env where notebook lives and a 'kernel env' where kernel lives, notebook extensions need to be in the notebook env.

We 'hack' this with python2 by having a 'requirements3.txt' that installs things into the notebook env. This lets people install notebook extensions.

Doing that by default for everything would definitely break lots of current repos that install server extensions. I'd say for now we special case 3.4 and 3.5, and figure out a better solution for this after.

jhamman:
Out of curiosity, do we know how ofter 3.4 and 3.5 are being used with repo2docker?

minrk:
>  do we know how ofter 3.4 and 3.5 are being used with repo2docker?

We don't, but I'm guessing it's rather few. #651 should get this unstuck because it leaves conda alone in the root env, installing the notebook in a dedicated env. This will presumably come at some cost to image size from the extra env, but may also improve installation time since it removes the step where the root env is switched to the chosen Python version.

jhamman:
Thanks all!

betatim:
What GitHub magic is this that this PR also got merged when we merged #651? This is amazing.

PR #638, opened 2019-04-04
[MRG] Remove conda download cache explicitly

### LINKED ISSUE #636: Cleaning out /srv/conda/pkgs
We are using repo2docker to build a fairly big python data science environment. (Image repo: https://github.com/pangeo-data/pangeo-stacks/tree/master/pangeo-notebook)

The compressed, built image is currently over 1GB in size. (docker hub: https://hub.docker.com/r/pangeo/pangeo-notebook/tags)

We would like the docker image to be smaller if possible. I noticed that there is currently about 2.6GB residing in `/srv/conda/pkgs`.

```
$ du -h -d0 /srv/conda/pkgs
2.6G    /srv/conda/pkgs
```

Is the expected behavior? If so, what would be the recommended way to clean this directory from my repo2docker build?

betatim:
Investigate what happens if we delete the download cache of conda packages.

If I build https://github.com/binder-examples/conda with `master` I get an image which is 2.04GB according to `docker images` and if I build the same repo with this PR I have one that is 2GB.

@rabernat what do you use to measure the image sizes? (I've never really looked into this before).

Fixes #636

### DIFF
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -181,7 +181,8 @@ class CondaBuildPack(BaseImage):
                 r"""
                 conda env update -n {0} -f "{1}" && \
                 conda clean -tipsy && \
-                conda list -n {0}
+                conda list -n {0} && \
+                rm -rf /srv/conda/pkgs
                 """.format(env_name, environment_yml)
             ))
         return super().get_assemble_scripts() + assembly_scripts


rabernat:
To measure compressed image size I am just looking on Docker Hub. Not very scientific. To measure decompressed image size (i.e. the size of the files on disk), I am using `du`.

Should I give this branch a try on the original repo which motivated the issue?

betatim:
@rabernat did you have a moment to try this out and see if it reduced the image size (or not)?

betatim:
I looked at https://microbadger.com/images/betatim/pangeo-with-patch and https://microbadger.com/images/pangeo/pangeo-notebook a bit. I think you can tell that we clean out the cache in the step that builds the user environment. It gets about 150MB smaller.

I think most of the big layers beyond the one repo2docker control are added by appendix and `postBuild` scripts in this repo. Maybe you can try out what happens if you include some clean up in those as well? My guess would be that rebuilding jupyter lab creates all sorts of files that we don't need to run lab (`node_modules`??)

I built the image with `repo2docker --subdir pangeo-notebook https://github.com/pangeo-data/pangeo-stacks/`

rabernat:
Thanks so much @betatim for your quick response and action here! 150 MB is a significant reduction in image size.

I will follow up on your suggestion to try to manually clean the post-build steps. I should have time to try this later today or tomorrow.

betatim:
Let's merge this then :)

If we find more ways to shrink images later we can make more PRs.

minrk:
I'm not 100% sure what's going to happen when packages are upgraded after this (pkgs is not a cache, it's where packages canonically live and hardlinked into envs), but it seems okay.

I'm guessing the size comes from docker layer filesystem not supporting hardlinks?

betatim:
I think on a FS that supports hardlinks it works a bit like reference counting. Not sure what happens inside docker.

From a short read on StackOverflow I got the impression that it is Ok to delete this directory but it will increase disk usage if you now install the same package in a new environment because it can't hardlink the files from `/srv/conda/pkgs`.

PR #639, opened 2019-04-07
[MRG] Ensure git submodules are updated and initialized

djhoese:
This is a continuation of #540. See that PR for details. That PR by @alimanfoo was only missing a test so this is my attempt at adding one.

Some things to note:

1. Should I merge or rebase upstream or leave this branch where it is?
2. I'm not a pytest fixture expert, but I don't think there was a cleaner way of doing what I needed to do to create the git repositories. I separated out the functionality of `repo_with_content` so I could reuse them in my new fixture.
3. After creating the fixture, I then realized there was no builtin way to get the submodule's SHA without running some commands with subprocess. It made the git unit tests uglier, but I wasn't sure what else to do.
4. As part of point 2 above, I changed the functionality of adding content to a test repository to use append so theoretically someone could call the same method multiple times to create multiple commits.

I think this tests what it needs to test, but may not be how the maintainers want this structured. Let me know what you want changed. Git tests pass locally.

### DIFF
--- a/tests/unit/contentproviders/test_git.py
+++ b/tests/unit/contentproviders/test_git.py
@@ -1,5 +1,6 @@
 import os
 import pytest
+import subprocess
 from tempfile import TemporaryDirectory
 from repo2docker.contentproviders import Git
 
@@ -16,6 +17,28 @@ def test_clone(repo_with_content):
         assert os.path.exists(os.path.join(clone_dir, 'test'))
 
         assert git_content.content_id == sha1[:7]
+
+
+def test_submodule_clone(repo_with_submodule):
+    """Test git clone containing a git submodule."""
+    upstream, sha1_upstream, sha1_submod = repo_with_submodule
+
+    with TemporaryDirectory() as clone_dir:
+        submod_dir = os.path.join(clone_dir, 'submod')  # set by fixture
+        spec = {'repo': upstream}
+        git_content = Git()
+        for _ in git_content.fetch(spec, clone_dir):
+            pass
+        assert os.path.exists(os.path.join(clone_dir, 'test'))
+        assert os.path.exists(os.path.join(submod_dir, 'test'))
+
+        # get current sha1 of submodule
+        cmd = ['git', 'rev-parse', 'HEAD']
+        sha1 = subprocess.Popen(cmd, stdout=subprocess.PIPE, cwd=submod_dir)
+        submod_sha1 = sha1.stdout.read().decode().strip()
+
+        assert git_content.content_id == sha1_upstream[:7]
+        assert submod_sha1[:7] == submod_sha1[:7]
 
 
 def test_bad_ref(repo_with_content):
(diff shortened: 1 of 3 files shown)


tests/unit/contentproviders/test_git.py
```diff
@@ -18,6 +19,28 @@ def test_clone(repo_with_content):
...
+        assert os.path.exists(os.path.join(clone_dir, 'test'))
+        assert os.path.exists(os.path.join(submod_dir, 'test'))
+
+        # get current sha1 of submodule
+        cmd = ['git', 'rev-parse', 'HEAD']
+        sha1 = subprocess.Popen(cmd, stdout=subprocess.PIPE, cwd=submod_dir)
+        submod_sha1 = sha1.stdout.read().decode().strip()
+
+        assert git_content.content_id == sha1_upstream[:7]
+        assert submod_sha1[:7] == submod_sha1[:7]
```
betatim: For my education: what are we testing here or is it a typo? Maybe it should be `sha1_submod[:7] == submod_sha1[:7]`?

The only time you want to compare `a` to `a` that I know is when you want to find out if the contents is NaN, which hopefully we don't have here :)

Again for my education: why do we need the `[:7]` bit? Both SHAs are the result of calling the same `git rev-parse...` so they should both be of the same length no?

tests/unit/contentproviders/test_git.py
```diff
@@ -18,6 +19,28 @@ def test_clone(repo_with_content):
         assert git_content.content_id == sha1[:7]
 
 
+def test_submodule_clone(repo_with_submodule):
+    """Test git clone containing a git submodule."""
+    upstream, sha1_upstream, sha1_submod = repo_with_submodule
```
betatim: Maybe we can prefix these with `expected_`? I think the similar variable names to the ones we get from the actually checked out repo are the source of the `assert` statement mixup below. Definitely took me a while to spot that `sha1_submod` and `submod_sha1` where not the same name :)

betatim:
This looks good to me. Maybe the tests could be even nicer but this is definitely better than the current state and not obviously unmaintainable/total mess so I'd merge it. You can always come back and make a second PR to clean up or refactor the tests.

> Should I merge or rebase upstream or leave this branch where it is?

I'd rebase after #640 is merged which will fix the docs failure (I think).

tests/unit/contentproviders/test_git.py
djhoese: Yes, I was just doing it for consistency to match `git_content.content_id` which was not created by me. To be clear, I am testing that the git submodule is at the commit it is supposed to be at. Since there is only one commit this is unlikely to change, but I felt like only testing that the 'test' file existed wasn't good enough.

And yeah, you're right I need to fix the variable names.

djhoese:
Ok I rebased with master. I also made it so the submodule repository has two commits, but the parent repository wants the submodule at commit 1. This makes sure that our operations don't accidentally update the commit the submodule is at.

betatim:
Merging! Thanks for picking this up and contributing the test.

PR #640, opened 2019-04-08
[MRG] Fix sphinx deprecation for recommonmark

betatim:
The way we were configuring recommonmark as a source parser has become deprecated, this gets us up the new way of doing it. Taken straight from https://www.sphinx-doc.org/en/master/usage/markdown.html

I also ran `black` on the config file in a second commit. Too much churn or not needed? If yes I can remove it again.


docs/source/conf.py
```diff
@@ -33,66 +31,64 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ['sphinx.ext.extlinks',
-              'sphinxcontrib.autoprogram',
-              ]
+extensions = ["sphinx.ext.extlinks", "sphinxcontrib.autoprogram", "recommonmark"]
```
lheagy: cool! I hadn't seen `recommonmark` mark before - looks quite handy :)

betatim:
@minrk could you add @lheagy to the people who can merge PRs in this repository please? Who else could I ask to do that if you aren't around?

minrk:
I've added both of you to the admin list. You should be able to grant others permission in the future. @lheagy go ahead and merge to test your permissions (might require accepting an invite first)

lheagy:
many thanks @minrk!

PR #643, opened 2019-04-12
[WIP] Use mamba to replace conda

yuvipanda:
https://github.com/QuantStack/mamba is a WIP, alpha quality, hopefully drop-in replacement for conda that's much faster.

This sets up CI so we can see what else needs to happen in Mamba to work
for us.


yuvipanda:
/cc @SylvainCorlay @wolfv. Everything fails now, but hopefully the CI logs are useful!

It'll be awesome to get mamba to work with repo2docker, since it'll hopefully cut down image
build times massively.

wolfv:
Hi @yuvipanda thanks for trying this. I am working on a big update that's hopefully going to solve this error soon.

yuvipanda:
@wolfv awesome! ping us here when you have a new release, and I can re-run the tests.

wolfv:
PING! just released 0.0.5 which should fix all your issues! ;)

consideRatio:
@wolfv I re-ran all the tests

wolfv:
hehe ok. well looks like mamba still got a couple of problems :D 

- we don't have a Python 3.5 build of mamba as it's been retired from conda-forge (afaik), so downgrading to 3.5 doesn't work
- you are using conda env ... which falls apparently back to "real" conda (i.e. no speed up for dependency resolution etc.). I'd need to invest some time to see how to make this command work with mamba.

consideRatio:
@wolfv ah! Wow I'm very excited about this and speedup of conda in general. When I think of conda, I instantly think of the angst associated with the gif below, and if it will ever finish ;)

conda-solving-env

consideRatio:
@wolfv the venv/conda build jobs that failed in the CI pipeline have used Python 3.6 as far as I understand, but you mentioned Python 3.5, I'm confused. Disclaimer: I'm not yet read on well on repo2docker yet, looking to make my first proper contribution soon.

wolfv:
yeah, exactly, mamba can't install environments with Python 3.5 because mamba itself doesn't exist for that version of python, thus it would need to erase itself which is currently forbidden :)

If you want to have a look at the `conda env` command and how to port it to mamba, that would be really cool and interesting. I am slightly worried that the `pip` channel which seems to be used sometimes in repo2docker is an additional challenge (but that comes after).

betatim:
> yeah, exactly, mamba can't install environments with Python 3.5 because mamba itself doesn't exist for that version of python, thus it would need to erase itself which is currently forbidden :)

We face a similar issue in #637 where we want to use a modern version of conda to create an env with Python 3.5. However that seems to fail because there is no `conda` package that we can install alongside Python 3.5. The solution proposed there is to use the hack we use for Python 2: create two separate environments: one for the notebook server and one for the kernel. The server one runs with a modern version of Python. The kernel one uses the version the user wants to run code under. Can we run mamba in a different conda env than the one it is asked to operate on? Something like `conda install -n some-other-env` where `conda` is executed in `not-some-other-env`?

consideRatio:
@betatim oh I think you maybe just explained something very relevant to https://github.com/jupyter/repo2docker/pull/649!

wolfv:
@betatim I'm pretty sure that it works when you create a new environment out of a base environment with a working Python - mamba combination.

SylvainCorlay:
>  The solution proposed there is to use the hack we use for Python 2: create two separate environments: one for the notebook server and one for the kernel.

This may actually be a cleaner way forward for everything. This would allow for users to specify frontend extensions and kernel packages separately.

minrk:
We definitely want to expand the separate kernel env to unsupported frontend Pythons, not just Python 2. The only reason I see to not *always* do it (it would be a *huge* simplification) is that letting the users only install kernel packages means that users can no longer specify versions of frontend packages (i.e. jupyterlab, lab extensions, etc.). Allowing users to *specify* packages in two environments means we need to come up with a non-standard spec for multiple environments, which I think is a pattern we should work extremely hard to avoid in repo2docker.

minrk:
mamba is a really cool experiment! How long do you think it will be before it's ready to be used in a case like repo2docker (i.e. preserving all conda features, edge cases, solver priority, etc.)? What makes it 'alpha'?

wolfv:
@minrk I think it is already at a point where it generally returns correct solutions. The support for conda's MatchSpec already landed upstream in libsolv and is used by mamba, so version numbers and build strings should mostly be correctly compared and used.

We don't have conda's "feature" support yet, and also not the strict mode of conda (I think both could be somewhat easily be done by some preprocessing of the repo data though).

Now the part that's missing for this specific use case here in repo2docker is that I haven't patched `conda_env` yet, which appears to be a seperate script for executing the `conda env ...` commands.
I was thinking about monkey patching it so that this line:

https://github.com/conda/conda/blob/7b16a9dfb0f4162908dc654eb51cd04f698e6485/conda_env/installers/base.py#L14-L18

doesn't import the conda installer, but rather imports a `mamba.conda_env...` installer that should be pretty straight forward to write.
If someone here has an idea on how to do that that'd be really cool. I briefly tried but what I did didn't really work out.

betatim:
Some data I got from running `conda` with instrumentation turned on. This is the contents of `~/.conda/instrumentation-record.csv` if you run with `CONDA_INSTRUMENTATION_ENABLED=1` in https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/conda/install-miniconda.bash which is the script we use to setup the initial conda environment(s), not the user specified environment though. Does anyone know how to read this (in particular the duplicate download entries)? Which part of it would mamba improve?
```
conda.core.solve.Solver._collect_all_metadata,11.612263
conda.core.solve.Solver._find_inconsistent_packages,0.000063
conda.resolve.Resolve.get_reduced_index,0.967496
conda.resolve.Resolve.gen_clauses,0.012656
conda.resolve.Resolve.get_reduced_index,0.000027
conda.resolve.Resolve.gen_clauses,0.012053
conda.resolve.Resolve.solve,0.041970
conda.core.solve.Solver._run_sat,1.026917
fetch_extract_prepare,0.022825
download,0.032841
download,1.632752
download,0.064413
download,0.029155
download,0.083929
download,0.034682
download,0.033897
download,0.043775
download,0.028491
download,0.040751
download,0.029912
download,0.110585
download,0.028991
download,0.054528
download,0.063211
download,0.030913
download,0.042293
download,0.207989
download,0.050900
download,0.116552
download,0.060391
download,0.056529
download,0.045303
download,0.031020
download,0.169883
download,0.030400
download,0.033375
download,0.041644
download,0.045977
download,0.027254
download,0.026921
download,0.033431
download,0.056130
download,0.136574
download,0.039796
download,0.026857
download,0.134748
download,0.029981
download,0.029963
download,0.030198
download,0.032073
download,0.030181
download,0.030978
download,0.030262
download,0.030030
download,0.041525
download,0.033822
download,0.042298
download,1.248634
download,0.080622
download,0.063832
download,0.032562
download,0.033647
download,0.409477
download,0.527558
download,0.061774
download,0.030927
download,0.040118
download,0.056374
download,0.033061
fetch_extract_execute,35.601844
unlink_link_prepare_and_verify,1.516371
unlink_link_execute,8.755935
```

yuvipanda:
I rebased this to master. It now includes https://github.com/jupyter/repo2docker/pull/651, which leaves conda in its own environment - so it can run whatever version of python it needs. Let's see what this does to the tests!

yuvipanda:
I *think* due to #674 this basically made all calls to 'conda' actuall refer to conda itself, overriding our mamba symlink :( So I *think* the tests passing is a false alarm.

yuvipanda:
I've hacked things in to have repo2docker just directly call mamba. Did you get 'conda env update' working, @wolfv? It looks like these tests are passing... Is there an easier way to tell when conda is being used vs mamba?

betatim:
I'm excited to bring mamba to the masses. I think we should do two things before we do deploy this:

* generate some numbers based on our binder-examples and other repos that show how much the total build time changes. We should measure before we cut :)
* can we rig things so that if there is an exception/error/something weird happens we rerun the command with stock conda? Basically try with `mamba` first, if exit code != 0, rerun with `conda`. This way we might make things slower for users but unlikely to break things for people (and have them come knocking on our door that stuff is broken and they have a deadline etc etc)

yuvipanda:
I don't actually know if the tests pass because mamba works or because we're just running conda... I *think* it's because mamba works, but would wanna check with @wolfv :) 

I also think we should try running it against, say, top 100 mybinder.org repo and see what happens.

Agree re: re-run with conda. Does it need to happen anywhere other than the conda build pack?

betatim:
> Agree re: re-run with conda. Does it need to happen anywhere other than the conda build pack?

Depends on if we "hide" the real `conda` or not. If a user types `conda` into a terminal once their repo is up and they get the normal conda I think we only need to make sure the conda buildpack knows how to recover. If we symlink `conda` to `mamba` we need to auto-retry.

I just ran `time repo2docker --no-run https://github.com/betatim/minimal-r2d-python` on this branch and got 2:19.67 total. When I ran "stock" repo2docker this morning I think the time was around 2:05. This repo does basically nothing but the "setup" part of repo2docker.

yuvipanda:
I think the big mamba is because I've sprinkled `mamba -V` around to see if mamba is in fact being used.

yuvipanda:
I just explicitly moved everything to 'mamba' instead of conda, so this should give us more confidence that mamba is in fact being used

yuvipanda:
Ok, there's at least one real failure I think - https://travis-ci.org/jupyter/repo2docker/jobs/529027646. Around how mamba activate behaves maybe?

consideRatio:
```
    def test_conda_activated():
>       assert os.environ.get("CONDA_PREFIX") == os.environ["NB_PYTHON_PREFIX"], dict(os.environ)
E       AssertionError: {'APP_BASE': '/srv', 'CONDA_DIR': '/srv/conda', 'CONDA_EXE': '/srv/conda/bin/conda', 'CONDA_PYTHON_EXE': '/srv/conda/bin/python', ...}
E       assert None == '/srv/conda/envs/notebook'
E         -None
E         +'/srv/conda/envs/notebook'
```

I'm guessing that the env var `CONDA_PREFIX` is longer defined as it was before.

wolfv:
hey all, love the excitement! but i've been on a vacation for two weeks, so haven't had the chance to update mamba, yet!

This means the `mamba env ...` will still forward to `conda env ...` and not use mamba (this one is actually easy to fix/reimplement).

Also, if I remember correctly, it still necessary to explicitly use `conda activate ...` as `mamba activate` would need to hook in at an even lower level (environment activation is done from bash and not python).

I'll ping you once I get `mamba env create` to work!

wolfv:
Also @betatim the parts that should be sped up is everything with solver and resolve in it. 

The duplicate download entries are things that mamba handles with conda (these are probably packages that are downloaded). But afaik conda currently downloads packages sequentially, which could be trivially parallelized.

Conda isn't actually that slow when solving for only one or two packages. It becomes very slow when solving for multiple packages (e.g. installing `xtensor xtensor-r xsimd`) or when you have some packages already installed in the workspace (since that is very similar as solving for multiple packages at the same time). That's why the solving part in your example is probably not the biggest bottleneck.

betatim:
The `conda env ...` that is measured above is installing [this environnment](https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/conda/environment.py-3.7.frozen.yml). It has a lot of packages but I think the fact that they all have explicit versions makes it "easier"?

Once we get this working nicely we can also use mamba to install the user's environment where they often don't pin all packages to explicit versions -> bigger gains.

We stand to gain about 20-25seconds on this step (assuming mamba is 10x faster). On a total runtime of 2minutes I'd say "ganz Ok" (which is Swiss German for "pretty good" ;) ). However on mybinder.org we won't see this because the step is cached so takes 0s. I think these speed up improvements should be geared towards/compared in the context of the interactive use case as that is where waiting makes a difference. This is where the user doesn't have a huge cache of previously built repos.

wolfv:
yeah, if all packages are pinned the work of the solver should be only to check that all dependencies are fulfilled and no conflicts exist, so that should be *very fast* :tm:

Let's see if we can realize any gains!

repo2docker/buildpacks/conda/install-miniconda.bash
```diff
@@ -43,9 +43,12 @@ if [[ "${CONDA_VERSION}" != "${MINICONDA_VERSION}" ]]; then
     conda install -yq conda==${CONDA_VERSION}
 fi
 
+# Install Mamba and symlink it to conda
```
minrk: I think we're no longer doing the "and symlink it to conda" step (which I agree with!)

repo2docker/buildpacks/conda/activate-conda.sh
```diff
@@ -4,8 +4,8 @@ test -f $CONDA_PROFILE && . $CONDA_PROFILE
 if [[ "${KERNEL_PYTHON_PREFIX}" != "${NB_PYTHON_PREFIX}" ]]; then
     # if the kernel is a separate env, stack them
     # so both are on PATH
-    conda activate ${KERNEL_PYTHON_PREFIX}
-    conda activate --stack ${NB_PYTHON_PREFIX}
+    mamba activate ${KERNEL_PYTHON_PREFIX}
+    mamba activate --stack ${NB_PYTHON_PREFIX}
 else
-    conda activate ${NB_PYTHON_PREFIX}
+    mamba activate ${NB_PYTHON_PREFIX}
```
minrk: Shouldn't we only be using mamba for the installation commands, not list/clean/etc.?

wolfv:
I made `mamba env create` a special case now -- with the upcoming mamba 0.0.6 (that I just released on github, coming to conda-forge in a couple of hours) it should actually use real mamba to fetch the packages and solve the environment. This is going to be a cool test! 

However, mamba already needs a rewrite :) So I don't think I would recommend at this stage to use it as the default version. There is still a bit of experimental prototype code here and there... but it becomes much clearer how a mamba 0.1.0 should look like.

In any case, if you guys could trigger another round of builds once mamba 0.0.6 hits conda-forge, that would be awesome!

minrk:
btw, I've started mamba myself locally, and it's been really great so far!

wolfv:
Mamba 0.0.6 is now available! Someone with a trigger finger?

Thanks Min! That's good to hear :)

betatim:
Restarted the build but it is now my turn to be on holiday so I won't get around to testing this locally to see if we can see a speed up.

wolfv:
@minrk is right, the conda activate script should still call conda (and not mamba). That might be the last thing for full tests passing?

Also, could we run the tests in verbose mode? I'd like to see that mamba is actually installing something :D

wolfv:
PS: Thanks @betatim and enjoy your vacation!

wolfv:
@yuvipanda do you want to quickly revert the conda-activate script? I'm still curious, I think all tests should pass then.

betatim:
I switched us back to `conda` for the activation steps. Let's see what happens.

betatim:
Good news, I think all the tests passed. The failing ones on Python 3.4 are discussed in #684 .

betatim:
With 0667db8 running `repo2docker --no-run https://github.com/betatim/minimal-r2d-python` takes about 2:05minutes and on master it takes 2:10. So timings are as in my last test.

wolfv:
the failing test is unrelated with conda/mamba right? Isn't it that ruamel.yaml requires Python > 3.4?

wolfv:
so basically there is no time improvement, right? That's probably because you explicitly set all versions down to the build string and there is no (real) solving going on -- the majority of the time is probably spend in downloading + unzipping the files from conda. Would be interesting to repeat with environments that are less tightly constrained :)

betatim:
Stopping by to let you know that we merged #699 which means we are now using [black](https://black.readthedocs.io/en/stable/) as code formatter. The motivation for doing this is to have a uniform style that will make it easier to read code.

yuvipanda:
I merged master into this, let's see what happens! I have terrible internet, so would <3 if someone else could test performance...

manics:
Done in https://github.com/jupyterhub/repo2docker/pull/962 !

PR #646, opened 2019-04-15
[MRG] Fix up the server proxy package to get websockets back

betatim:
This uses an unreleased version of jupyter-server-proxy that has a fix for the broken websocket handling.

This (I think) will fix the problems reported in https://discourse.jupyter.org/t/debugging-your-binder/33/12?u=betatim

### DIFF
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -214,10 +214,11 @@ class RBuildPack(PythonBuildPack):
                 "${NB_USER}",
                 # Install nbrsessionproxy
                 r"""
-                pip install --no-cache-dir nbrsessionproxy==0.8.0 && \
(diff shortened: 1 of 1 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/debugging-your-binder/33/13

yuvipanda:
I don't mind merging this now and making a jupyter-server-proxy release later.

betatim:
Now the question I always ask: Do you have an idea for a test that we can add to stop this from coming back? I don't really know anything about websockets other than they someone do "tricksy" things on the back of HTTP. Could we make a test based on https://github.com/binder-examples/minimal-dockerfile that starts a websocket server inside and connect to it from the outside? Would that be 20, 50 or 100 lines of code?

PR #649, opened 2019-04-20
[MRG/REVIEW] Support Pipfile / Pipfile.lock with pipenv

### LINKED ISSUE #174: Support pipfiles
Moving from https://github.com/jupyterhub/binderhub/issues/357

I think if we have a Pipfile, we should do the following:

1. Use a virtualenv
2. Ignore requirements.txt 
3. Install packages with Pipfile

consideRatio:
# To help you review this PR

I've added a new buildpack, it will allow repo2docker to detect Pipfile / Pipfile.lock and use them to install Python packages with `pipenv`. [Here](https://github.com/datastory-org/weekly-chart) is a repo that made me want this functionality.

There are 48 files changed in this PR, but most of them are relating to tests of the functionality added. Other than these tests, we have changes to...

[.gitignore](https://github.com/jupyter/repo2docker/pull/649/files#diff-a084b794bc0759e7a6b77810e01874f2) - When running tests locally I ended up with some artifacts, this change helped me avoid these.
[.travis.yml](https://github.com/jupyter/repo2docker/pull/649/files#diff-354f30a63fb0907d4ad57269548329e3) - I declared the need for another build job for our CI to run tests against this buildpack.
[docs/source/config_files.rst](https://github.com/jupyter/repo2docker/pull/649/files#diff-45f635e2feb1a1642b2b9523d9b771b4) - I documented the new build pack
[repo2docker/app.py](https://github.com/jupyter/repo2docker/pull/649/files#diff-d53de4b85f5176c7a74088518d9e0c1f) - I added an import statement, sorted the imports alphabetically, and added the buildpack in the ordered list that decides what buildpack has priority assuming it decides it can do something through the `detect` function a buildpack should have.
[repo2docker/buildpacks/__init__.py](https://github.com/jupyter/repo2docker/pull/649/files#diff-94a6626d185a09cf68dcc749f0b42c04) - Allow repo2docker.buildpacks module expose the new buildpack.
[repo2docker/buildpacks/pipfile/__init__.py](https://github.com/jupyter/repo2docker/pull/649/files#diff-0135020cc3adbe76ac8fb96443cacade) - The logic of the buildpack, **this is the key part to review I think**.

The rest of the files are for the tests within the `tests/pipfile` folder. I added quite a bit of tests I'd say!

---

# About
This PR aims to close #174. The goal is to make repo2docker consider `Pipfile` and `Pipfile.lock` dependency files using [`pipenv`](https://github.com/pypa/pipenv).

### Code strategy
The strategy adopted was to mimic sections for how a `requirements.txt` configuration file was treated but give these new `Pipfile.lock` or `Pipfile` files a higher precedence.

### Documented process
I wrote in [this discourse topic](https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839) as I was speaking to a [rubber duck](https://www.urbandictionary.com/define.php?term=Rubber%20ducking) to help myself think clearly along the way.

# What to review
### The files
- Some documentation within `docs/source/config_files.rst`.
- The actual functionality introduced all resides in `repo2docker/buildpacks/python/__init__.py`. All three functions are modified a bit, namely `detect()`, `python_version()`, and `get_assemble_scripts()`.
- Loads of added tests that can give clarity of what I wanted to accomplish in `/tests/venv/pipfile`.
- Small additions to a `.gitignore` file catching some scrap generated after running all the `pytest`.

### Concerns
- Not using [PyEnv](https://github.com/pyenv/pyenv) that is streamlined to work alongside Pipenv.
    - Not using PyEnv to manage the Python version concerns me, I choose to parse Pipfile.lock (JSON), Pipfile (TOML), and runtime.txt with Regex within [`python_version()`](https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/python/__init__.py#L11) that is overriding the CondaBuildPack's function and influences the installed Python version.
    - Not using PyEnv, we ignore patch versions of Python, just like runtime.txt does, but the Pipfile format supports `python_full_version` specifications.
- Not entering `pipenv shell`, missing out on [automatic loading of `.env` files](https://pipenv.readthedocs.io/en/latest/advanced/#automatic-loading-of-env) for example.
- Not managing to install dependencies in a `Pipfile` with Python 3.5 or lower.

    ```
    Step 40/47 : RUN ${KERNEL_PYTHON_PREFIX}/bin/pipenv lock --python >${KERNEL_PYTHON_PREFIX}/bin/python
    ---> Running in c7ae30385795
    Creating a virtualenv for this project…
    Pipfile: /home/erik/Pipfile
    Using /srv/conda/bin/python (3.5.5) to create virtualenv…
    ⠙ Creating virtual environment...Already using interpreter /srv/conda/bin/python
    Using base prefix '/srv/conda'
    New python executable in /home/erik/.local/share/virtualenvs/erik-zof0I2Qp/bin/python
    ERROR: The executable /home/erik/.local/share/virtualenvs/erik-zof0I2Qp/bin/python is not >functioning
    ERROR: It thinks sys.prefix is '/home/erik' (should be '/home/erik/.local/share/virtualenvs/erik->zof0I2Qp')
    ERROR: virtualenv is not compatible with this system or executable
    
    ✘ Failed creating virtual environment 
    [pipenv.exceptions.VirtualenvCreationException]:   File "/srv/conda/lib/python3.5/site->packages/pipenv/vendor/click/decorators.py", line 17, in new_func
    [pipenv.exceptions.VirtualenvCreationException]:       return f(get_current_context(), *args, **kwargs)
    [pipenv.exceptions.VirtualenvCreationException]:   File "/srv/conda/lib/python3.5/site->packages/pipenv/cli/command.py", line 319, in lock
    [pipenv.exceptions.VirtualenvCreationException]:       ensure_project(three=state.three, >python=state.python, pypi_mirror=state.pypi_mirror)
    [pipenv.exceptions.VirtualenvCreationException]:   File "/srv/conda/lib/python3.5/site->packages/pipenv/core.py", line 574, in ensure_project
    [pipenv.exceptions.VirtualenvCreationException]:       pypi_mirror=pypi_mirror,
    [pipenv.exceptions.VirtualenvCreationException]:   File "/srv/conda/lib/python3.5/site->packages/pipenv/core.py", line 506, in ensure_virtualenv
    [pipenv.exceptions.VirtualenvCreationException]:       python=python, site_packages=site_packages, >pypi_mirror=pypi_mirror
    [pipenv.exceptions.VirtualenvCreationException]:   File "/srv/conda/lib/python3.5/site->packages/pipenv/core.py", line 935, in do_create_virtualenv
    [pipenv.exceptions.VirtualenvCreationException]:       extra=[crayons.blue("{0}".format(c.err)),]
    [pipenv.exceptions.VirtualenvCreationException]: /home/erik/.local/share/virtualenvs/erik->zof0I2Qp/bin/python: error while loading shared libraries: libpython3.5m.so.1.0: cannot open shared >object file: No such file or directory
    
    Failed to create virtual environment.
    ```

### Opinionated
- `pipenv` will be used to do the installation as compared to using it to create a `requirements.txt` file and then do the installation with `pip`.
- The first among `environment.yml`, `Pipfile.lock`, `Pipfile`, and `requirements.txt` will be used to install packages.
- The first among `Pipfile.lock`, `Pipfile`, and `runtime.txt` will be used to decide `python_version()`.
- `setup.py` is ignored if a `Pipfile(.lock)` exists, as they can declare such installations. This can for example be seen in [this repo](https://github.com/datastory-org/weekly-chart/blob/master/Pipfile).

# Why didn't I use PyEnv?
- At the point `pipenv install` would run that in turn triggers PyEnv to setup the right Python version, a Python version would have already been installed. So using PyEnv would have required deeper changes that I wasn't ready to tackle without better understanding about repo2docker.
- Installing PyEnv wasn't something I could do with `pip` or similar, but it required adding `apt-get` packages etc. first which was a hassle.


meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

meeseeksmachine:

https://discourse.jupyter.org/t/an-unfolding-story-of-my-first-contribution-to-repo2docker/839/1

betatim:
I think we should make a new build pack for `Pipfile` support instead of modifying an existing one.

I think we should continue to use `conda` to install the version of Python needed as `pipenv` doesn't solve this and we already have `conda`.

I need to think about `python_full_version` and `python_version`. Right now I think we should extract the Python version from the `Pipfile(.lock)` and install it via `conda`. However the `python_full_version` might be incompatible with conda package version strings? To be checked.

repo2docker/buildpacks/python/__init__.py
```diff
@@ -58,26 +82,51 @@ def get_assemble_scripts(self):
...
+        else:
+            # KERNEL_PYTHON_PREFIX is the env with the kernel,
+            # whether it's distinct from the notebook or the same.
+            pip = '${KERNEL_PYTHON_PREFIX}/bin/pip'
+            requirements_file = self.binder_path('requirements.txt')
+            setup_py = 'setup.py'
+            if os.path.exists(requirements_file):
+                assemble_scripts.append((
+                    '${NB_USER}',
+                    'pip install "pip<19" && ' + \
```
betatim: We need to be careful not to reintroduce this `pip` version pin. It has been removed since this PR's branch started. The line was deleted in `master`.

betatim:
Having lots of tests is great and good work on creating lots of them already! You mentioned in the discourse post that it takes a while to run all of them. That's right. We've been trying to move more and more things to `tests/unit` to test small parts of repo2docker in isolation. Then the tests in `tests/venv` and so on can be more like integration tests.

The advantage of the unit tests is that they run very fast and don't need a docker daemon.

For example a small tests to check that version parsing does something useful is

https://github.com/jupyter/repo2docker/blob/eb551eb248e17132fd36f121ba89c0572931b94e/tests/unit/test_env_yml.py#L10-L17

I'd take a look around https://github.com/jupyter/repo2docker/tree/master/tests/unit to see how different things are tested. A lot can be done with temporary directories (for example "Does `detect()` work correctly and give precedence to the right things?"), the next level up is to use mocks to stub out other stuff that is expensive and not what you are trying to test (for example the docker client can be a mock). As we get more into mocks the complexity of the test increases (you need to understand how mocks work), you can test more things in less run time, you are really only testing things in isolation. So it is all trade offs all the way down :)

betatim:
One more thing to keep at the back of our minds: we want to eventually have separate kernel and server environments. This was also mentioned in the mamba PR and is definitely needed for updating `conda` itself.

consideRatio:
> I think we should make a new build pack for `Pipfile` support instead of modifying an existing one.

I lean towards thinking it should be one buildpack, but it becomes two I'd like to rename the current PythonBuildPack to RequirementsBuildPack or something. I figure requirements.txt and Pipfile have a lot in common, a lot more than environment.yml and Pipfile, or environment.yml and requirements.txt. Can you help me overcome my confirmation bias towards the current implementation by providing a reason?

> I think we should continue to use `conda` to install the version of Python needed as `pipenv` doesn't solve this and we already have `conda`. I agree with not using `pyenv`.

Ideally I'd like to see repo2docker function with pipenv managing Pipfile, Pipfile.lock, with PyEnv to manage the Python version without any special management. Perhaps that isnt possible, I understand it can be quite complicated to mess with the Python version at the point in time pipenv is about to install things.

Hmmm... I wonder if PyEnv can be run to parse a Pipfile(.lock) first and then later, Pipenv will install the packages but when PyEnv is run earlier, it only cares about the Python version... Speaking of which, that is what condabuildpack does right? Hmmm....

> I need to think about `python_full_version` and `python_version`. Right now I think we should extract the Python version from the `Pipfile(.lock)` and install it via `conda`. However the `python_full_version` might be incompatible with conda package version strings? To be checked.

I saw that requirements.txt only extracted major+minor version and ignored the patch version or any beta/alpha etc. So, I did the same thing for Pipfile and Pipfile.lock parsing, for python_version or python_full_version - only the major and minor version is extracted. Which is a silent failure to comply with the specification in the Pipfile / Pipfile.lock :'( I dislike this also... :hankey:



I think the requirements.txt and requirements3.txt should be named to indicate that

> One more thing to keep at the back of our minds: we want to eventually have separate kernel and server environments. This was also mentioned in the mamba PR and is definitely needed for updating `conda` itself.
> 
> I don't know yet what it means for this PR but its worth taking a look at how we handle Python 2 kernels/environments. https://github.com/binder-examples/python2_runtime#specifying-a-python-2-environment-with-runtimetxt in particular `requirements.txt` vs `requirements3.txt`. The text in the README could be improved. What it is trying to tell you is that `requirements.txt` will be installed in your kernel environment and if you need to install things in the notebook server's environment (that always uses Python 3) you should write those things in `requirements3.txt`.

Ah! This is very complicated to me and I'm struggling to grasp things still. Hmmm... Do I get it right if I think these are files that the users provide in git repositories to be built, and requirements3.txt will always be in "the jupyter server environment" (notebook environment) while requirements.txt will always be in "the kernel environment"?

---

I'll read up on the not-build-a-dockerfile tests!
I'll also make sure the pip-version used aligns like it does it the current master branch!

I'll try learn more about PyEnv, I don't like how my lack of understanding is the key reason for not using it.

I'd also love to learn if we can manage to get the `.env` files influence the environment etc like `Pipenv` does for us if we could run `pipenv shell` - this requires further learning also.

betatim:
> Can you help me overcome my confirmation bias towards the current implementation by providing a reason?

I think the code will be a lot easier to follow as it won't contain `if requirementstxt: ...do stuff ... elif pipfile: ...do something else...` and it fits with the rest of our build packs where each of them deals with one kind of scenario (a `requirements.txt` is found or a `environment.yml` is found, or ... etc). It also means we won't have to change the existing `PythonnBuildPack` with the added work of makingn sure nothing gets broken while we modify it. I don't think that the code to handle the `requirements.txt` case and the `Pipfile` case have that much in common, except for needing to have a Python interpreter provided for them (via innheriting from `CondaBuildPack`)

I won't offer an opinion on the pros and cons of renaming `PythonBuildPack` except that that should be a new PR. Changing it or not isn't part of the minimal changes needed to implement this feature so it should go in a new PR to keep the diffs manageable.

The main take away on having two environments is: the notebook server needs Python >= 3.6 so if the user wants an old version of Python we need to create a separate environment for the user's Python and packages. We will need to do this more and more in the future (is my guess).

I would extract as much version information as possible from the `Pipfile` and see how it maps to conda packaged versions of Python. We invented `runtime.txt` so it is a file that only repo2docker uses -> we shouldn't follow its lead if we can do something better.

I wouldn't use PyEnv to install the Python interpreter. The main motivation is that it will add another way of providing Python that we will need to understand, maintain, and support. I don't see any benefit from using PyEnv if we already have a well tested conda based setup. Following the spirit of minimal change per PR I'd delay that discussion to a new PR as well.

consideRatio:
@betatim thanks for helping my mind come to ease, I'll adjust!

To make the minimally viable change, I'll also not extract environment information in this PR but leave that as a future PR, sounds ok?

betatim:
> To make the minimally viable change, I'll also not extract environment information in this PR but leave that as a future PR, sounds ok?

Is this the `.env` support? If yes I think that sounds like a good plan. It is also a cool feature that helps remove the need for the special, invented here `start` file.

consideRatio:
Yepp that is what i vaguely referred to! It is automatic if we use `pipenv shell` but that seems complicated to do in a Dockerfile giving us errors, and also what would that mean really anyhow

betatim:
Figuring out how to combine `pipenv shell` with launching jupyter kernels sounds like enough work for a whole new PR :)

yuvipanda:
I've also been wondering about the long term prospects of Pipfile now that pyproject.toml is a thing and much more 'officially' sanctioned. I'm reading to understand this more, here are some links I'm reading.

- https://github.com/pypa/pipfile/issues/27
- https://github.com/pypa/pipfile/issues/3
- https://caremad.io/posts/2013/07/setup-vs-requirement/
- https://snarky.ca/clarifying-pep-518/

and lots more.

I don't understand the difference well yet, but will try and keep reading.

repo2docker/buildpacks/pipfile/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+            # using python 2 kernel,
+            # requirements3.txt allows installation in the notebook server env
+            nb_requirements_file = self.binder_path('requirements3.txt')
+            if os.path.exists(nb_requirements_file):
+                assemble_scripts.append((
+                    '${NB_USER}',
+                    # want the $NB_PYHTON_PREFIX environment variable, not for
+                    # Python's string formatting to try and replace this
+                    '${{NB_PYTHON_PREFIX}}/bin/pip install --no-cache-dir -r "{}"'.format(nb_requirements_file)
+                ))
```
consideRatio: # Q2: What about this section?
Does it make sense to have like this?

consideRatio: This was copied and not stripped out from the PythonBuildPack, but I'm not 100% it makes sense to have here for us, but I think it does.

consideRatio:
Pweh okay here comes an update! I made changes to extract the additions in this PR to a separate buildpack.

- I extracted the pipfile logic to a separate buildpack called PipfileBuildPack. I added it after CondaBuildPack (environment.yml) but before PythonBuildPack (requirements.txt).
- I have still only added full e2e tests and made the CI trigger a new job for the Pipfile buildpack and the tests I've added to it.

# Questions of my own
## Q1: Relative paths from Pipfile in a binder directory
In the added `tests/pipfile/setup-py-explicit-in-binder-dir` test. It has a Pipfile referencing a local python package (`Setup.py`) with `path=".."` rather than `path="."`. I think it makes sense but I'd like some confirmation.

See for example: https://github.com/jupyter/repo2docker/blob/5367f805411d6df5bdb721ac268244850de076a9/tests/pipfile/setup-py-explicit-in-binder-dir/.binder/Pipfile

## Q2: Python 2 stuff
See this comment: https://github.com/jupyter/repo2docker/pull/649#pullrequestreview-236425118

repo2docker/buildpacks/pipfile/__init__.py
consideRatio: @betatim delete or keep this section?

betatim:
Stopping by to let you know that we merged #699 which means we are now using [black](https://black.readthedocs.io/en/stable/) as code formatter. The motivation for doing this is to have a uniform style that will make it easier to read code.

consideRatio:
@betatim rebasing this PR caused a huge amount of formatting conflicts as I have a commit history changing within the PythonBuildPack before extracting it all to PipfileBuildPack. So I merged the new master instead and then applied black formatting.

My first impression of using black is very positive!

consideRatio:
@betatim is the CI testing every single commit leading up to the final status of the PR?

consideRatio:
# Update

## Black formatting issue resolved
I've learned that running `black` will not autoformat the `verify` files without .py file extension, while `pre-commit run --all-files` will. This caused the issue where I thought I had complied but didn't according to CI. The issue is discussed briefly [here](https://github.com/python/black/issues/491) but was closed as it would incur too much of a performance hit.

## Regarding notes in https://github.com/jupyter/repo2docker/pull/649#issuecomment-491607215
- **Q1:** I asked for confirmation, but I think its OK
- **Q2:** I'm now OK about the supporting requirements3.txt

## PR Status: MRG

consideRatio:
@yuvipanda I'm almost done reading through http://andrewsforge.com/article/python-new-package-landscape/ and that provided great insights and overview!

> PEP 517 and PEP 518—accepted in September 2017 and May 2016, respectively—changed this status quo by enabling package authors to select different build systems. Said differently, for the first time in Python, developers may opt to use a distribution build tool other than distutils or setuptools. The ubiquitous setup.py file is no longer mandatory in Python libraries.
--- NOTE: pyproject.toml is related to these PEPs

To conclude though, we need some new lines in this picture:
image

consideRatio:
I just finished adding a section to the first post of this PR that helps you overview the changes to the code, see: https://github.com/jupyter/repo2docker/pull/649#issue-272182372

I'm really looking forward to get this merged as there are some binder-unaware repositories that I want to make a PR to, where I simply update the README to include a mybinder badge ^^

repo2docker/buildpacks/pipfile/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+        Falsy empty string '' if not found.
+        """
+
+        if hasattr(self, "_python_version"):
+            return self._python_version
+
+        files_to_search_in_order = [
+            {
+                "path": self.binder_path("Pipfile.lock"),
+                "pattern": r"\s*\"python_(?:full_)?version\": \"?([0-9a-z\.]*)\"?",  # '            "python_version": "3.6"'
```
betatim: You could use `'` instead of `"` for the string itself to avoid having to escape the `"` inside the regex pattern.

Is it Ok to only match `"python_version": "3.6"` and not `"python_version": '3.6'` or other quote combos?

repo2docker/buildpacks/pipfile/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+        else:
+            # use the default Python
+            self._python_version = self.major_pythons["3"]
+            return self._python_version
+
+    def get_assemble_scripts(self):
+        """Return series of build-steps specific to this repository.
+        """
+        # If we have a runtime.txt & that's set to python-2.7,
+        # requirements.txt will be installed in the *kernel* env
```
betatim: The mention of `requirements.txt` surprised me. Won't that file be ignored if there is a `Pipfile`? My expectation was that the `Pipfile(.lock)` controls the environment the user sees (like `environment.yml` or `requirements.txt` for for other build packs) and that the "special" `requirements3.txt` allows you to install stuff in the notebook server's environment.

repo2docker/buildpacks/pipfile/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+        pipfile = self.binder_path("Pipfile")
+        pipfile_lock = self.binder_path("Pipfile.lock")
+        working_directory = self.binder_dir or "."
+        assemble_scripts.append(("${NB_USER}", "pip install pipenv"))
+        # if Pipfile.lock isn't found, Pipfile is used to create one
+        if not os.path.exists(pipfile_lock):
+            assemble_scripts.append(
+                (
+                    "${NB_USER}",
+                    "(cd {} && {} lock --python {})".format(
```
betatim: What is the working directory set to that we have to `cd`? I thought we'd be in the directory that contains the repo's content and that this is also where we wanted to be. If that isn't true maybe having a comment to explainn would help people from the future.

repo2docker/buildpacks/pipfile/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+                "(cd {} && {} install --ignore-pipfile --deploy --system --dev --python {})".format(
+                    working_directory, pipenv, python
+                ),
+            )
+        )
+
+        return assemble_scripts
+
+    def detect(self):
+        """Check if current repo should be built with the Python buildpack.
```
betatim:
```suggestion
        """Check if current repo should be built with the Pipenv buildpack.
```

repo2docker/buildpacks/pipfile/__init__.py
betatim: The `requirements3.txt` is a way to specify packages to install into the environment that your jupyter notebook server is running because if the user/kernel is Python 2 there are two environments. Having this `requirements3.txt` allows people to install server extensions which need to be in the server's env. So I'd keep that behaviour for when someone selects Python 2 in their `Pipfile`.

betatim:
With a `Pipfile` like:
```
[[source]]
url = 'https://pypi.python.org/simple'
verify_ssl = true
name = 'pypi'

[requires]
python_version = '2.7'

[packages]
requests = '*'
```

I was expecting to get a kernel that runs Python 2 and has `requests` installed. However the build crashed :-/ If instead I set the Python version to 3.6 I also get a warning that I am using Python 3.7 but asked for 3.6. Related to single quotes? I copy&pasted https://github.com/pypa/pipfile#pipfile

repo2docker/buildpacks/pipfile/__init__.py
betatim: From the experiment in https://github.com/jupyter/repo2docker/pull/649#issuecomment-504667325 I think we need to handle both kinds of quotes. Which makes me think of the old regex saying: I had a problem that I tried to solve with a regex, now I have two problems :)

repo2docker/buildpacks/pipfile/__init__.py
```diff
@@ -0,0 +1,133 @@
+"""Generates Dockerfiles based on an input matrix based on Python."""
```
betatim: Let's update this to say something about Pipenv

repo2docker/buildpacks/pipfile/__init__.py
consideRatio: lol on the quote :D I'll look into it!

consideRatio:
Thanks a lot @betatim for your review!!! As usual they are very helpful! :heart: Now I'm feeling unblocked to continue working on this as I'm eager to do!

repo2docker/buildpacks/pipfile/__init__.py
consideRatio: Resolved in a8c232e

repo2docker/buildpacks/pipfile/__init__.py
consideRatio: Meant to be resolved by: c9fd0a9

---

The working directory we `cd` into is the repo directory or the `binder` / `.binder` directory.

If there is a reference to a `setup.py` file to be installed etc as there can be in `Pipfile`'s, then they will often reference either `.` or `..`. This has been verified to function with tests, and I've reasoned that this makes the most sense, as compared to making it so that you would need to write `.` within the Pipfile's residing in a binder directory and meaning the parent directory in reality.

### An example of a `Pipfile` outside a binder dir

```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"

[packages]
there = "*"
dummy = {path=".", editable=true}
```

### An example of a `Pipfile` in a binder dir

```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"

[packages]
there = "*"
dummy = {path="..", editable=true}
```

### The code we are discussing
```python
        # install Pipfile.lock or fallback to installing Pipfile
        pipenv = "${KERNEL_PYTHON_PREFIX}/bin/pipenv"
        python = "${KERNEL_PYTHON_PREFIX}/bin/python"
        pipfile = self.binder_path("Pipfile")
        pipfile_lock = self.binder_path("Pipfile.lock")
        working_directory = self.binder_dir or "."
        assemble_scripts.append(("${NB_USER}", "pip install pipenv"))
        # if Pipfile.lock isn't found, Pipfile is used to create one
        if not os.path.exists(pipfile_lock):
            assemble_scripts.append(
                (
                    "${NB_USER}",
                    "(cd {} && {} lock --python {})".format(
                        working_directory, pipenv, python
                    ),
                )
            )
        # install Pipfile.lock
        assemble_scripts.append(
            (
                "${NB_USER}",
                "(cd {} && {} install --ignore-pipfile --deploy --system --dev --python {})".format(
                    working_directory, pipenv, python
                ),
            )
        )
```

repo2docker/buildpacks/pipfile/__init__.py
consideRatio: Resolved by 2189777

consideRatio:
I've struggled a lot (8 hours) to get go from Pipfile / Pipfile.lock to some installed python2 setup, but WOW this is hard while starting out from being inside one of two environments already.

It is hard to summarize the challenge, but pipenv leads to the creation of virtual envs, and they don't like to be created unless everything is alright, and I start to doubt I can ever get that to happen from within the realms of a buildpack assuming the other infrastructure of repo2docker is like it is. Hmmm...

What would you say about this PR if I focus on a well functional python3 only BuildPack that strips away all python2 related support @betatim?

I could for example:
- make it fail hard if we find Pipfile etc but python2 is declared (how do I propagate a failure btw? - like "this is a repo we know we will fail to manage!")
- or... make it silently do nothing when it finds Pipfiles but knows of the issues of python2 it cannot overcome so it simply ignores them, like we do currently for both py2 and py3.

consideRatio:
After 13 hours now, I'm thinking I may be more stubborn than a goat. But....

I made it! Usage of Python 2 should work now I think. :goat:  :crescent_moon: :cactus:

consideRatio:
I did a quick iteration on the documentation of Pipfiles - @betatim I think this PR may be close to complete now.

Hmmm okay I realize there may be a changelog etc I should note things in.

repo2docker/buildpacks/pipfile/__init__.py
```diff
@@ -65,51 +68,72 @@ def python_version(self):
...
+            # notebook servers Python environment
             nb_requirements_file = self.binder_path("requirements3.txt")
             if os.path.exists(nb_requirements_file):
                 assemble_scripts.append(
                     (
                         "${NB_USER}",
-                        # want the $NB_PYHTON_PREFIX environment variable, not for
-                        # Python's string formatting to try and replace this
-                        '${{NB_PYTHON_PREFIX}}/bin/pip install --no-cache-dir -r "{}"'.format(
+                        'pip install --no-cache-dir -r "{}"'.format(
```
betatim: Why not keep this as `${{NB_PYTHON_PREFIX}}/bin/pip` which is what the [Python buildpack](https://github.com/jupyter/repo2docker/blob/cd809645a9e00cf14abd110d8fb8098193379a7c/repo2docker/buildpacks/python/__init__.py#L56-L61) uses?

repo2docker/buildpacks/pipfile/__init__.py
```diff
@@ -65,51 +68,72 @@ def python_version(self):
...
+
+        # A Pipfile(.lock) can contain relative references, so we need to be
+        # mindful about where we invoke pipenv as that will dictate where .`
+        # referes to.
+        #     [packages]
+        #     my_package_example = {path=".", editable=true}
         working_directory = self.binder_dir or "."
+
+        # install pipenv to install dependencies within Pipfile.lock or Pipfile
         assemble_scripts.append(("${NB_USER}", "pip install pipenv"))
```
betatim: Here we should probably have `${KERNEL_PYTHON_PREFIX}/bin/pip` so that `pipenv` gets installed in the kernel env (where we install all the things the user actually wants)

repo2docker/buildpacks/pipfile/__init__.py
```diff
@@ -65,51 +68,72 @@ def python_version(self):
...
+        #   prioritize usage of .lock files that makes sense for
+        #   reproducibility.
+        # - The `--ignore-pipfile` requires a .lock file to be around as if
+        #   there isn't, no other option remain.
         assemble_scripts.append(
             (
                 "${NB_USER}",
-                "(cd {} && {} install --ignore-pipfile --deploy --system --dev --python {})".format(
-                    working_directory, pipenv, python
+                """(\\
```
betatim: What does a `\\` do?

tests/pipfile/py2-with-server-and-kernel-req/README.rst
```diff
@@ -0,0 +1,7 @@
+Python 2 - Pipfile and requirements3.txt
+----------------------------------------
+
+- We should get Python 2 setup for the Kernel
+   - with 'numpy' installed installed through Pipfile as listed in 'packages'
```
betatim:
```suggestion
   - with 'numpy' installed through Pipfile as listed in 'packages'
```

tests/pipfile/py2-with-server-and-kernel-req/README.rst
```diff
@@ -0,0 +1,7 @@
+Python 2 - Pipfile and requirements3.txt
+----------------------------------------
+
+- We should get Python 2 setup for the Kernel
+   - with 'numpy' installed installed through Pipfile as listed in 'packages'
+   - with 'parse' installed installed through Pipfile as listed in 'dev-packages'
```
betatim:
```suggestion
   - with 'parse' installed  through Pipfile as listed in 'dev-packages'
```

repo2docker/buildpacks/pipfile/__init__.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # A Pipfile(.lock) can contain relative references, so we need to be
+        # mindful about where we invoke pipenv as that will dictate where .`
+        # referes to.
+        #     [packages]
+        #     my_package_example = {path=".", editable=true}
+        working_directory = self.binder_dir or "."
+
+        # install pipenv to install dependencies within Pipfile.lock or Pipfile
+        assemble_scripts.append(("${NB_USER}", "pip install pipenv"))
```
betatim: should we pin the version of pipenv used? Unless there is a good reason not to, I'd vote to pin it

repo2docker/buildpacks/pipfile/__init__.py
consideRatio: Sure - let's remain explicit!

repo2docker/buildpacks/pipfile/__init__.py
consideRatio: A `\\` renders to `\` in the end within the Dockerfile, allowing content to be combined to one single line. It allowed me to split up the content that renders to the Dockerfile into multiple lines.

repo2docker/buildpacks/pipfile/__init__.py
consideRatio: Sure, so we pin it the major and minor, but leaves the patch out? like `pipenv=X.Y.*` ?

repo2docker/buildpacks/pipfile/__init__.py
consideRatio: I removed the initial one being alone.

repo2docker/buildpacks/pipfile/__init__.py
consideRatio: oh... their versions are like this: 2018.11.26

I pinned it to that.

consideRatio:
Wieeee!

image

repo2docker/buildpacks/pipfile/__init__.py
consideRatio: I also added a comment about what those were about!

betatim:
Merging this now. 👏 for the stamina to see this through!!

consideRatio:
Wieee thanks for your excellent review @betatim ! :heart:

I merged it to mybinder.org deploy it and it works on mybinder.org now!

betatim:
Noice! Do you want to make a new repository on https://github.com/binder-examples/ that demos this? Then we can send a tweet out as well.

consideRatio:
@betatim yes that would be awesome! :D

choldgraf:
yahoooo!

PR #651, opened 2019-04-25
[MRG] install notebook in its own env

minrk:
leave the conda root env alone

adds conda activation to profile.d and ensures that we start with a login shell via the ENTRYPOINT

closes #567 by adding an ENTRYPOINT that enables conda integration

cc @betatim


minrk:
I think this is going to result in a performance improvement, since it skips a whole install step (the Python switch), and creates a new env instead of updating the root env.

The result is an ENTRYPOINT that ensures we launch with a fully configured login shell (profile.d, bashrc are sourced, unlike before), including the shell integration in #567. This also means that our notebook env is fully activated, in case there are any conda activation hooks in that env.

This will also facilitate #637 because we don't care what version Python is in the conda root env

minrk:
Additionally, this moves the handling of start_script down a level, so that our ENTRYPOINT is always called. The behavior should be the same.

tests/conda/repo-path/verify
```diff
@@ -3,7 +3,7 @@ import sys
 import os
 
 # Python should still be in /srv/conda
```
betatim: -> "Python should still be in $NB_PYTHON_PREFIX"

betatim:
I think some of the failures are because we are still using conda 4.5 and not 4.6. For example the `--stack` argument seems to not exist in 4.5. I think we can update to 4.6 in this PR as well.

minrk:
@betatim good call. I merged in #637 and bumped conda up to 4.6.14 (there is now a miniconda 4.6.14) with a refreeze. This should mean significant performance improvements because we skip *two* `conda install` calls: one for the upgrade of conda (short-lived because conda and miniconda versions match for now) and another for the Python switch (permanent).

betatim:
I re-added the step to make the start script executable as I think that was the cause for some of the tests to fail. Maybe you can squash the two commits into one (I used the webinterface :-/)

betatim:
I think https://github.com/jupyter/repo2docker/blob/bc9554e18cead7336cb66daee0435a3d63d701b4/tests/julia/julia_version-default/verify#L20 is the line that fails in the Julia tests.

yuvipanda:
@minrk how do you feel about splitting up the ENTRYPOINT related stuff from here into its own PR? Will help me with #657 :D

betatim:
I think this is basically ready to be merged (I will fix the broken test now) and then we can start a new PR/rebase #657. What do you think @yuvipanda?

betatim:
I propose we merge this now. All tests pass, it gets us unblocked on conda versions, and one step closer to always splitting notebook server and user's kernel env.

minrk:
I’m happy to merge this now. Sorry I didn’t finish it earlier.

betatim:
No worries. We are IMHO doing a good job of not having to be in a hurry.

Thanks for sorting us out with a nice new codna env setup!

PR #652, opened 2019-04-27
[MRG] Fix handling of memory limit command line argument

betatim:
If a string of only numerals is passed as argument we assume it should
be converted to an integer and specifies a size in bytes.

When BinderHub constructs the `repo2docker` command-line it [passes a value in bytes without a suffix](https://travis-ci.org/jupyterhub/mybinder.org-deploy/builds/525264674#L1113-L1135). This is with a value of `"3G"` set in the `values.yaml`.

I think this will allow us to try out https://github.com/jupyterhub/mybinder.org-deploy/pull/954

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -300,7 +300,7 @@ def make_r2d(argv=None):
 
     if args.build_memory_limit:
         # if the string only contains numerals we assume it should be an int
-        # and specifies a size inn bytes
+        # and specifies a size in bytes
         if args.build_memory_limit.isnumeric():
             r2d.build_memory_limit = int(args.build_memory_limit)
         else:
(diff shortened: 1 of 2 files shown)


repo2docker/__main__.py
```diff
@@ -299,7 +299,12 @@ def make_r2d(argv=None):
         r2d.user_name = args.user_name
 
     if args.build_memory_limit:
-        r2d.build_memory_limit = args.build_memory_limit
+        # if the string only contains numerals we assume it should be an int
+        # and specifies a size inn bytes
```
yuvipanda:
```suggestion
        # and specifies a size in bytes
```

yuvipanda:
I added a unit test to make codecov happy.

PR #653, opened 2019-04-28
Support .binder directory

### LINKED ISSUE #617: support .binder instead of binder as config directory
I would like to use the path`.binder/` rather than `binder/` for my [repo](https://github.com/pangeo-data/pangeo_ocean_examples) configuration. I read the docs but couldn't figure out if it is supported to change the config path. I tried it and I think it did not work.

jhamman:
closes #617

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -189,6 +189,7 @@ class BuildPack:
         self.log = logging.getLogger('repo2docker')
         self.appendix = ''
         self.labels = {}
+        self._binder_dir = None
         if sys.platform.startswith('win'):
             self.log.warning("Windows environment detected. Note that Windows "
                              "support is experimental in repo2docker.")
@@ -414,12 +415,18 @@ class BuildPack:
         """
         return None
 
+    @property
+    def binder_dir(self):
+        if self._binder_dir is None:
+            for dirname in ['binder', '.binder']:
+                if os.path.exists(dirname):
+                    self._binder_dir = dirname
+            self._binder_dir = ''
+        return self._binder_dir
+
     def binder_path(self, path):
         """Locate a file"""
-        if os.path.exists('binder'):
-            return os.path.join('binder', path)
-        else:
-            return path
+        return os.path.join(self.binder_dir, path)
 
     def detect(self):
         return True
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -419,8 +419,9 @@ class BuildPack:
     def binder_dir(self):
         if self._binder_dir is None:
             for dirname in ['binder', '.binder']:
-                if os.path.exists(dirname):
+                if os.path.isdir(dirname):
                     self._binder_dir = dirname
+                    break
             self._binder_dir = ''
         return self._binder_dir
 
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -189,7 +189,6 @@ class BuildPack:
         self.log = logging.getLogger('repo2docker')
         self.appendix = ''
         self.labels = {}
-        self._binder_dir = None
         if sys.platform.startswith('win'):
             self.log.warning("Windows environment detected. Note that Windows "
                              "support is experimental in repo2docker.")
@@ -417,13 +416,20 @@ class BuildPack:
 
     @property
     def binder_dir(self):
-        if self._binder_dir is None:
-            for dirname in ['binder', '.binder']:
-                if os.path.isdir(dirname):
-                    self._binder_dir = dirname
-                    break
-            self._binder_dir = ''
-        return self._binder_dir
+        has_binder = os.path.isdir("binder")
+        has_dotbinder = os.path.isdir(".binder")
+
+        if has_binder and has_dotbinder:
+            raise RuntimeError(
+                "The repository contains both a 'binder' and a '.binder' "
+                "directory. However they are exclusive.")
+
+        if has_dotbinder:
+            return ".binder"
+        elif has_binder:
+            return "binder"
+        else:
+            return ""
 
     def binder_path(self, path):
         """Locate a file"""
--- a/repo2docker/buildpacks/nix/nix-shell-wrapper
+++ b/repo2docker/buildpacks/nix/nix-shell-wrapper
@@ -9,9 +9,9 @@ _term() {
 trap _term SIGTERM
 
 # find binder sub-directory (if present)
-for dir in "./binder" "./.binder"; do
+for dir in "./.binder" "./binder"; do
   if [ -e $dir ]; then
-    binder_dir = $dir
+    binder_dir=$dir
   fi
 done
 
--- a/repo2docker/buildpacks/nix/nix-shell-wrapper
+++ b/repo2docker/buildpacks/nix/nix-shell-wrapper
@@ -8,12 +8,19 @@ _term() {
 
 trap _term SIGTERM
 
+# find binder sub-directory (if present)
+for dir in "./binder" "./.binder"; do
+  if [ -e $dir ]; then
+    binder_dir = $dir
+  fi
+done
+
 # if there is a binder/ sub-directory it takes precedence
 # files outside it are ignored
-if [ -e ./binder ]; then
-  nixpath="./binder/default.nix";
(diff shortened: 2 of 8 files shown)


betatim:
Nice work. Can you add to the docs an explicit statement that you can't "stack" these directories: use either `.binder/` or `binder/` and which one takes precedence over the other if you have two. I am not sure which one should take precedence, my hunch would be to make `binder/` "win" because ... dunno.

jhamman:
I have a few tests failing here but its not clear to me why. If someone has any ideas, I'm keep to keep pushing on this.

betatim:
`tests/{venv,conda}/binder-dir/` check that the `Dockerfile` in the root directory of the repo is ignored if there is a `binder/` directory. The `venv` tests on Python 3.4 also fail because of this and the `dockerfile/` tests as well.

My guess is the nix related tests fail for a similar reason. 

As soon as there is either a `.binder/` or `binder/` directory files in the root should not be found any more. By creating a specific directory you ask repo2docker to ignore everything in the top level. This let's people who have all sorts of files that repo2docker recognises in their root and ignore them when they add repo2docker support.

Despite my earlier comment about specifying the order of precedence, I now think that for consistency we should also make `.binder/` and `binder/` exclusive. My preference would be to make it an error to have both in your repository. That way we don't have to worry whether they should be additive or not. What do you think?

To add test that check if things are working take a look at the test s in unit that setup temporary directories/repos and use just a small bit of repo2docker. Instead of the much heavier tests in `venv/` and friends that are failing now. SOme of them could/should be implemented as unit tests with better errors and faster runtimes. So for new tests e should try and create unit tests.

repo2docker/buildpacks/base.py
```diff
@@ -414,12 +415,18 @@ def get_start_script(self):
         """
         return None
 
+    @property
+    def binder_dir(self):
+        if self._binder_dir is None:
+            for dirname in ['binder', '.binder']:
+                if os.path.exists(dirname):
+                    self._binder_dir = dirname
```
consideRatio: This will currently assign `.binder` over `binder` if both exists I think. Either add a break statement or reverse the order in the array.

repo2docker/buildpacks/base.py
```diff
@@ -414,12 +415,18 @@ def get_start_script(self):
...
+            self._binder_dir = ''
```
consideRatio: This is a problematic line right? You need this to be in a conditional to not finding any dirs

consideRatio: Oh! Move it to before the for loop

consideRatio: @betatim, is there a risk that this property is called before a binder / .binder folder exists, and then isn't updated to reflect that there now is a binder / .binder folder?

@jhamman I guess the fact that I don't know at this point, makes me think that we want to avoid the preoptimization only checking for the existence of the folders once.

repo2docker/buildpacks/base.py
```diff
@@ -414,12 +415,18 @@ def get_start_script(self):
         """
         return None
 
+    @property
+    def binder_dir(self):
+        if self._binder_dir is None:
+            for dirname in ['binder', '.binder']:
+                if os.path.exists(dirname):
```
betatim: I'd switch to `os.path.isdir` to make it explicit that we need `dirname` to be a directory and not any kind of path. I think with this we get a better error message in the exception "for free". 

https://docs.python.org/3/library/os.path.html#os.path.isdir

repo2docker/buildpacks/base.py
betatim: Not sure if we could end up with a race condition. Not caching this is a good way to side step the issue and I'd be surprised if this makes any noticeable difference to the speed of running repo2docker.

betatim: I'd write something like this to make the intention of exclusive directories (very) explicit for future readers:
```python
@property
def binder_dir(self):
    has_binder = os.path.isdir("binder")
    has_dotbinder = os.path.isdir(".binder")

    if has_binder and has_dotbinder:
        raise RuntimeError("The repository contains both a 'binder' and a "
                           "'.binder' directory. However they are exclusive.")

    if has_dotbinder:
        return ".binder"
    elif has_binder:
        return "binder"
    else:
        return ""
```

repo2docker/buildpacks/base.py
jhamman: yes, I was missing an intended break statement. It looks like we're going for a more explicit rendition in the comments below. Either way, nice catch.

repo2docker/buildpacks/base.py
jhamman: Thanks for the discussion here. I've updated to match @betatim's implementation.

repo2docker/buildpacks/base.py
```diff
@@ -414,12 +414,26 @@ def get_start_script(self):
...
 
+    @property
+    def binder_dir(self):
+        has_binder = os.path.isdir("binder")
+        has_dotbinder = os.path.isdir(".binder")
+
+        if has_binder and has_dotbinder:
+            raise RuntimeError(
+                "The repository contains both a 'binder' and a '.binder' "
+                "directory. However they are exclusive.")
```
jhamman: It may make sense to raise an informative warning here instead. This implementation will render a repo unbuildable if both directories exist. The warning would just say something like: "We've found both a `.binder` and `binder` directory in your repo, we're only going to use `binder`".

betatim: I think repo2docker should stop running if there are two, not warn and continue. My feeling is that it is almost certainly an oversight on the users side when this happens. Combined with the huge amount of output repo2docker already generates my hunch is that most people will miss the warning. Can we think of a reason why a user would want to have both at the same time if they know that they are exclusive?

consideRatio: I'm unaware of how an error would propegate to MyBinder.org etc, would this be easy to get for the users of mybinder.org if it fails like this? What would happen if we try to make it a warning for users of mybinder.org or binderhubs rather?

choldgraf: As long as we can propagate the error to the user running their session on a binderhub, I'm +1 on making this error rather than warn, I think we could get into some confusing situations otherwise (partially because remembering which one takes precedence will be totally arbitrary and add cognitive burden...I think it's easier to just remember to use one or the other but not both)

betatim: You will get a failed build on a BinderHub with (I think) the traceback printed as the last thing in the logs. This can already happen today for example with https://mybinder.org/v2/git/https%3A%2F%2Fgithub.com%2Fbetatim%2Fdsgsdgdsg.git/95f2eecf43686d2a45e7c15452b99e9d6346295c (the revision doesn't exist). (We could make it a nicer error message by adding some `try ... except` handling at the highest level of the repo2docker CLI app, but that is a new PR.)

jhamman: Okay, it sounds like there in consensus here.

repo2docker/buildpacks/nix/nix-shell-wrapper
```diff
@@ -8,20 +8,27 @@ _term() {
...
+done
+
 # if there is a binder/ sub-directory it takes precedence
 # files outside it are ignored
-if [ -e ./binder ]; then
-  nixpath="./binder/default.nix";
-  if [ -f ./binder/start ]; then
-    chmod u+x ./binder/start
+if [ -z $binder_dir ]; then
+  nixpath="$binder_dir/default.nix";
```
betatim: From the failing nix tests it looks like it ends up setting `nixpath` to `/default.nix` and then erroring out. Not quite sure why it happens though :-/

betatim:
After fixing up the failing nix tests, this PR needs some simple unit style tests for the functionality added and then we should be on our way to merge city 🚂 .

jhamman:
@betatim - unit tests added. All green!

betatim:
Merged! Thanks for making this happen.

betatim:
Can you make another PR to update https://github.com/jupyter/repo2docker/blob/master/docs/source/changelog.rst with a note about this?

consideRatio:
Wohoooo @jhamman !!! :D :tada: I appreciate this feature!

choldgraf:
indeed this is awesome! thanks @jhamman :-)

PR #657, opened 2019-04-30
Make sure ENTRYPOINT is an absolute path

yuvipanda:
Unlike other parts of the generated Dockerfile,
the start script is evaluated at run time, rather
than at build time. Currently, we assume that the
current working directory is the same at runtime
as build time for the start script. This doesn't
hold true always, and particularly not in JupyterHub
environments where ${HOME} is often overlaid with
a persistent directory.

We change this to always refer to the full path,
using the ${REPO_DIR} environment variable. This
lets people building JupyterHub images to set
REPO_DIR to something like /srv/repo (like hubploy
does), and still have a working start script.

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -668,7 +668,12 @@ class BaseImage(BuildPack):
         return []
 
     def get_start_script(self):
-        start = self.binder_path('./start')
+        start = self.binder_path('start')
         if os.path.exists(start):
-            return start
+            # Return an absolute path to start
+            # This is important when built container images start with
+            # a working directory that is different from ${REPO_DIR}
+            # This isn't a problem with anything else, since start is
+            # the only path evaluated at container start time rather than build time
+            return os.path.join('${REPO_DIR}', start)
         return None


repo2docker/buildpacks/base.py
```diff
@@ -643,7 +643,12 @@ def get_post_build_scripts(self):
...
-        start = self.binder_path('./start')
+        start = self.binder_path('start')
         if os.path.exists(start):
-            return start
+            # Return an absolute path to start
+            # This is important when built container images start with
+            # a working directory that is different from ${REPO_DIR}
+            # This isn't a problem with anything else, since start is
+            # the only path evaluated at container start time rather than build time
+            return os.path.join('${REPO_DIR}', start)
```
betatim: This relies on `$REPO_DIR` being an absolute path right? Is there any benefit to calling `os.path.abspath` to make it absolute-as-it-was-at-build-time? I have no idea :-/

betatim:
LGTM. Can you add a quick test with a temporary directory (unit-test style) to make sure this remains an absolute path and respects `REPO_DIR` also in the future?

repo2docker/buildpacks/base.py
yuvipanda: @betatim yeah, we can't tell at build time if it's an absolute or relative path, so I don't think it'd be useful.

yuvipanda:
@betatim yeah, I can do that. I'm actually tempted to instead pull out just the ENTRYPOINT related stuff from https://github.com/jupyter/repo2docker/pull/651 instead.

yuvipanda:
@betatim I rebased this, but it's ready to go now (thanks to #651 being merged)

repo2docker/buildpacks/base.py
```diff
@@ -668,7 +668,12 @@ def get_post_build_scripts(self):
...
```
consideRatio: Okay so we join a placeholder path to get started, then we should not add any quotes!

image

consideRatio:
LGTM!!

repo2docker/buildpacks/base.py
betatim: What do you mean?

consideRatio: Sorry for confusion, often quotes are required in paths for bash to interpret it all as one command line argument, this as it should be, no additional quotes would make sense

Issue #166, opened 2017-12-07
Discuss `binder.yaml` files and what to support


choldgraf:
In the Binder repo we discussed a way to let users specify their distribution (e.g. artful vs. zesty). It was decided that a `binder.yaml` file would be a good way to do this. Let's use this issue to see what kind of functionality we want to support in this file.

betatim:
There is also https://github.com/jupyterhub/binder/issues/42 on `binder.yml` ideas

betatim:
https://github.com/jupyter/repo2docker/pull/148 discusses letting the user choose a base image, could also be done with a `bidner.yml`

yuvipanda:
A binder.yaml should be nothing more than something like:

```
apt.yaml:
    <contents of what would have been in apt.yaml>
requirements.txt:
    <contents of what would have been in environment.yaml>
```

We can implement it by simply exploding this into multiple files and then running normal repo2docker process. This also means we don't invent a new format.

I really wanted us to invent a new format when we started, but @willingc, @minrk and @choldgraf convinced me that we'll just be re-inventing a shitty Ansible :)

choldgraf:
a question I'm still unsure about: will there be *any* functionality in a `binder.yaml` file that can't also be accessed via a corresponding configuration file? E.g., would it be possible to specify a specific environment variable (e.g., `distribution: artful`) in a binder.yaml file, but not have a config file with which you could do the same thing. On the one hand this makes things more confusing, on the other it might reduce the number of config files in general that people need to use.

betatim:
What are the features we are missing right now that would need us to invent a way to set them? I am strongly in favour of not having a `binder.yaml`, using @yuvipanda's proposal is Ok to me.

Things I can remember that we don't support currently and users have asked for:
* setting ENV variables
* specifying/pinning the base image

choldgraf:
Those are the two I can think of as well, though I'm sure we'll have more pop up over time as well.

mukundans91:
It seems that this discussion at least the initial half coincides with a lot of what was said in #148.  But if a feature is made to allow users to determine the base image of docker(zesty or artful), then the onus for maintaining basic compatibility with those images comes into the picture as well. How to make sure that the user specified version does not break the current set of tests and basic functionality?

Another question is will there be requests to support other linux distros as well?

choldgraf:
[...]

PR #662, opened 2019-05-02
specification info page

### LINKED ISSUE #656: Give the Binder repository spec a name
We've had a [few conversations about an open standard for reproducible repositories](https://discourse.jupyter.org/t/creating-a-specification-for-reproducible-repositories/569). I think it'll take a while for that all to settle out, but in the meantime could we name the specification that Binder uses to build a repo's environment. Even if that spec changes, we could start using it in docs, presentations, etc. Something like **The [Binder] Reproducible Repository specification**.

Then rather than saying "A Binder-ready repository is a repository that can be built with repo2docker without an error" we could say "A Binder-ready repository is a repository that follows the (link to) Binder reproducible repository specification".

choldgraf:
This is a fairly lightweight information page that coins the term "Reproducible Execution Environment Specification" (REES), and provides a little bit of explanation for what it is. In the future, we could flesh out this page a bit more.

closes #656

WDYT @betatim ?

### DIFF
--- a/docs/source/specification.rst
+++ b/docs/source/specification.rst
@@ -10,18 +10,20 @@ and the resulting actions that repo2docker takes is known
 as the **Reproducible Execution Environment Specification** (or REES).
 
 The goal of the REES is to automate and encourage existing community best practices
-for reproducible computational environments. This includes installing
-community-standard specification files such as ``requirements.txt`` or ``REQUIRE`` using
-standard tools such as ``pip`` or ``conda`` or ``apt``. While repo2docker automates the
+for reproducible computational environments. This includes installing pacakges using
+community-standard specification files and their corresponding tools,
+such as ``requirements.txt`` (with ``pip``), ``REQUIRE`` (with Julia), or
+``apt.txt`` (with ``apt``). While repo2docker automates the
 creation of the environment, a human should be able to look at a REES-compliant
 repository and reproduce the environment using common, clear steps without
 repo2docker software.
 
 Currently, the definition of the REE Specification is the following:
 
-    Any collection of files taken from the :ref:`config-files`
-    list, placed either in the root of a folder or in a sub-folder called
-    either ``binder/`` or ``.binder/``.
+    Any directory containing zero or more files from the :ref:`config-files` list is a
+    valid reproducible execution environment as defined by the REES. The
+    configuration files have to all be placed either in the root of the
+    directory, in a ``binder/`` sub-directory or a ``.binder/`` sub-directory.
 
 For example, the REES recognises ``requirements.txt`` as a valid config file.
 The file format is as defined by the ``requirements.txt`` standard of the Python
--- a/docs/source/specification.rst
+++ b/docs/source/specification.rst
@@ -1,0 +1,20 @@
+.. _specification:
+
+====================================================
+The Reproducible Execution Environment Specification
+====================================================
+
+repo2docker scans a repository for particular :ref:`config_files`, such
+as ``requirements.txt`` or ``REQUIRE``. The collection of files and their contents
+that repo2docker uses is known as the **Reproducible Execution Environment Specification**.
+
+The goal of the REE Specification is to provide a structure that is clearly-defined and that
+can be extended to accomodate more components of a reproducible workflow.
+
+Currently, the definition of the REE Specification is the following:
+
+> Any collection of files taken from the :ref:`config_files`
+> list, placed either in the root of a folder or in a sub-folder called either ``binder/`` or ``.binder/``.
+
+In the future, the repo2docker team plans to formalize this specification into a pattern
+that can also be followed in other ways, such as by creating a JSON or YAML file.
--- a/docs/source/specification.rst
+++ b/docs/source/specification.rst
@@ -4,17 +4,27 @@
 The Reproducible Execution Environment Specification
 ====================================================
 
-repo2docker scans a repository for particular :ref:`config_files`, such
-as ``requirements.txt`` or ``REQUIRE``. The collection of files and their contents
-that repo2docker uses is known as the **Reproducible Execution Environment Specification**.
+repo2docker scans a repository for particular :ref:`config-files`, such
+as ``requirements.txt`` or ``REQUIRE``. The collection of files, their contents,
+and the resulting actions that repo2docker takes is known
+as the **Reproducible Execution Environment Specification** (or REES).
 
-The goal of the REE Specification is to provide a structure that is clearly-defined and that
-can be extended to accomodate more components of a reproducible workflow.
+The goal of the REES is to automate and encourage existing community best practices
+for reproducible computational environments. This includes installing
+community-standard specification files such as ``requirements.txt`` or ``REQUIRE`` using
+standard tools such as ``pip`` or ``conda`` or ``apt``. While repo2docker automates the
+creation of the environment, a human should be able to look at a REES-compliant
+repository and reproduce the environment using common, clear steps without
+repo2docker software.
 
 Currently, the definition of the REE Specification is the following:
 
-> Any collection of files taken from the :ref:`config_files`
-> list, placed either in the root of a folder or in a sub-folder called either ``binder/`` or ``.binder/``.
+    Any collection of files taken from the :ref:`config-files`
+    list, placed either in the root of a folder or in a sub-folder called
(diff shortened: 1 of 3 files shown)


docs/source/specification.rst
```diff
@@ -0,0 +1,20 @@
...
+The goal of the REE Specification is to provide a structure that is clearly-defined and that
+can be extended to accomodate more components of a reproducible workflow.
+
+Currently, the definition of the REE Specification is the following:
+
+> Any collection of files taken from the :ref:`config_files`
+> list, placed either in the root of a folder or in a sub-folder called either ``binder/`` or ``.binder/``.
+
+In the future, the repo2docker team plans to formalize this specification into a pattern
+that can also be followed in other ways, such as by creating a JSON or YAML file.
```
betatim: I'm not sure I understand what you mean with this.

docs/source/specification.rst
```diff
@@ -0,0 +1,20 @@
+.. _specification:
+
+====================================================
+The Reproducible Execution Environment Specification
+====================================================
+
+repo2docker scans a repository for particular :ref:`config_files`, such
+as ``requirements.txt`` or ``REQUIRE``. The collection of files and their contents
```
betatim:
```suggestion
as ``requirements.txt`` or ``REQUIRE``. The collection of files, their contents, and actions to take based on them
```
I think we should include something that says that the spec not only specifies which files are recognised, what their format/contents is but also specifies what you should do with each file.

In my mind this is what will form part of this spec for the `requirements.txt`:

> REE Spec recognises a `requirements.txt` as a valid config file. The file format is as defined by the `requirements.txt` standard of the Python community. A REE spec compliant tool will install a Python interpreter (of unspecified version) and perform the equivalent action of `pip install -r requirements.txt` so that the user can afterwards run `python` and use the packages installed.

The wording is not very good but in terms of topics covered this is what I imagine the spec will be.

docs/source/specification.rst
```diff
@@ -0,0 +1,20 @@
...
+that repo2docker uses is known as the **Reproducible Execution Environment Specification**.
+
+The goal of the REE Specification is to provide a structure that is clearly-defined and that
+can be extended to accomodate more components of a reproducible workflow.
```
minrk: I would add some text along the lines of "automating existing community best practices", e.g. why we support `requirements.txt` and not `repo2docker-packages.fmt`.

minrk: How about something like this:

> The goal of REES is to automate and encourage existing community best practices for reproducible computational environments. This includes installing community-standard specification files such as `requirements.txt` or `REQUIRE` using standard tools such as `pip` or `conda` or `apt`. While repo2docker automates the creation of the environment, a human should be able to look at a REES-following repository and reproduce the environment using common, clear steps without repo2docker software.

docs/source/specification.rst
minrk: I think this is referring to #166

I think we can probably omit future plans from the spec file until they are more concrete.

docs/source/specification.rst
betatim: I like the spirit and the wording of this.

choldgraf:
Thanks all for the great comments + edits, what do folks think about the latest push?

betatim:
Ready to merge or do you want to tweak/work on it some more?

choldgraf:
I'm fine merging as-is with the idea that we'll keep iterating on this moving forward. Understand if folks want to hold off though!

docs/source/specification.rst
```diff
@@ -0,0 +1,30 @@
...
+standard tools such as ``pip`` or ``conda`` or ``apt``. While repo2docker automates the
+creation of the environment, a human should be able to look at a REES-compliant
+repository and reproduce the environment using common, clear steps without
+repo2docker software.
+
+Currently, the definition of the REE Specification is the following:
+
+    Any collection of files taken from the :ref:`config-files`
+    list, placed either in the root of a folder or in a sub-folder called
+    either ``binder/`` or ``.binder/``.
```
betatim: Could we turn L20-24 into something like "Any directory containing zero or more files from the config-files list is a valid *reproducible execution environment* as defined by the REES. The configuration files have to all be placed either in the root of the directory, in a `binder/` sub-directory or a `.binder/` sub-directory."?

In my mind a specification is a document that describes a thing which it defines. What I was expecting is something like "The HTTP spec defines that a HTTP request consists of a header and possibly a body. The header is separated from the body by two `\n`. A valid header consists of zero or more lines ... etc" https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html Something like this is what I think of when I think of REES.

docs/source/specification.rst
```diff
@@ -0,0 +1,30 @@
...
+
+repo2docker scans a repository for particular :ref:`config-files`, such
+as ``requirements.txt`` or ``REQUIRE``. The collection of files, their contents,
+and the resulting actions that repo2docker takes is known
+as the **Reproducible Execution Environment Specification** (or REES).
+
+The goal of the REES is to automate and encourage existing community best practices
+for reproducible computational environments. This includes installing
+community-standard specification files such as ``requirements.txt`` or ``REQUIRE`` using
+standard tools such as ``pip`` or ``conda`` or ``apt``. While repo2docker automates the
```
betatim: I'd mention the files and corresponding tools instead of having `requirements.txt` which matches with `pip` but then `REQUIRE` which doesn't match with `conda` and `apt` which is all by itself.

docs/source/specification.rst
```diff
@@ -0,0 +1,30 @@
...
+The Reproducible Execution Environment Specification
+====================================================
+
+repo2docker scans a repository for particular :ref:`config-files`, such
+as ``requirements.txt`` or ``REQUIRE``. The collection of files, their contents,
+and the resulting actions that repo2docker takes is known
+as the **Reproducible Execution Environment Specification** (or REES).
+
+The goal of the REES is to automate and encourage existing community best practices
+for reproducible computational environments. This includes installing
```
betatim: I don't think you install a `requirements.txt`. You install the packages listed in the file and the required dependencies of these packages.

docs/source/usage.rst
```diff
@@ -10,7 +10,8 @@ Using ``repo2docker``
    order to run ``repo2docker``. For more information on installing
    ``repo2docker``, see :ref:`install`.
 
-``repo2docker`` is called with a URL/path to a git repository. It then
+``repo2docker`` can build a reproducible computational environment for any repository that
+follows :ref:`specification`. repo2docker is called with a URL/path to a git repository. It then
```
betatim: Not just Git repositories, local directories work just as well.

betatim:
I left several more comments which feel a bit language lawyer-y but then a specification is like a legal text so ... :-/

docs/source/specification.rst
choldgraf: I think eventually we should have that kind of more specific language in there, I agree

docs/source/specification.rst
choldgraf: see latest push, how's that?

betatim: Didn't seem to change these lines?

choldgraf: oops, I forgot to save the file before pushing the latest commit :-)

betatim:
For a new PR/something to ponder when we extend this document: we focus on the fact that a human should be able to look at a directory that is REES compliant and be able to figure out what to do. THis is a key feature. How about when we describe a config file and what a REES compliant tool should do with it we also have a "For a human" section that says what a human would have to do to (roughly) achieve the same goal (I think often to achieve exactly the same goal we'd have to link to the (many) lines of code with corner cases etc in repo2docker). `install-python; pip install -r requirement.txt` is probably what a human should do when they find a `requirements.txt`. Even though what repo2docker exactly does is to add flags like `--no-cache-dir` and use conda, etc.

Having the "as a human" section would make it clear we are serious about that angle and instil confidence in the reader that this is actually possible (and not some kind of fantasy).

choldgraf:
@betatim that makes sense to me - I agree we should highlight the human aspect of the specification, along with the desire to co-opt other workflows rather than define new ones. Wanna throw that into a new issue assuming that this spec will be merged?

betatim:
Pressing merge despite the CI failures. I think the docker cloud part failed because they were doing maintenance during most of Saturday.

PR #663, opened 2019-05-03
tweak language to make clear `exec "$@"` works

fomightez:
originally in description of the `start` config file it read:
 "The last line must be ``exec "$@"`` equivalent." 

Seemed open to interpretation as I needed to deduce what the equivalent of that would be here. 
I think adding 'or' before 'equivalent' shows that will work, but I assume other options exist? Or why not say  "The last line must be ``exec "$@"``"?


betatim:
I like the suggestion. The "or equivalent" is correct because `start` doesn't have to be a bash script, it could just as well be a Python script in which case you need to do "the equivalent of `exec $@` in Python.

Issue #632, opened 2019-03-28
Command to run already built image


saulshanabrook:
Often, I would like to play with what volumes are mounted, or what command is executed, without having to rebuild the docker image.

This takes a long time when working on a JupyterLab extension that depends on JupyterLab master and so has to rebuild all of JupyterLab every image build.

Currently, I am looking for the image name manually with `docker image ls` and running it manually. This isn't ideal, because then it isn't run in the same way as `repo2docker` is running it. You pass some extra options on start, like a port and the custom display URL. I would like to keep all these nicetities and not have to hunt in the image list for the most recent one.

I thought this is what the `--no-build` argument would do, but I guess it not only doesn't build but doesn't start. For example, [`docker-compose up`](https://docs.docker.com/compose/reference/up/) has a `--no-build` argument that won't build the images, but will start up the services.

betatim:
This would be a great feature to have. I think it would make a good first issue for people who are already familiar with docker, the concept of images and containers.

A way to get started is to look at the mechanism used to generate the docker image names used by repo2docker, check if there is already one (with a similar name) and then launch that instead when the user specifies a particular command-line flag.

If someone wants to work on this please open a PR. It doesn't have to be perfect and it is even encouraged that you open a PR long before you think it is ready as a way to get feedback and help when needed.

tmtaybah:
Hi, can I work on this issue?

betatim:
Yes please! A good way to "claim" an issue is to start work on it and soon after open a Pull Request with `[WIP] ` at the start of the title. That way we know ho is working on what. The WIP tag means "this is still work in progress, not ready to be reviewed".

betatim:
@saulshanabrook does the new docker build caching (fast builds if you don't edit `environment.yml`) and `--editable` solve your use case?

PR #665, opened 2019-05-03
Add `--reuse-image` to not rebuild an image but instead re-populate content after r2d starts the container

tmtaybah:
WIP for #632

Command to pull in remote a repository and run an existing image instead of rebuilding because of the updated repo.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -313,13 +313,12 @@ class Repo2Docker(Application):
         config=True,
     )
 
-
     reuse_image = Bool(
         False,
         help="""
         Copies new contents to existing image.
         """,
-        config=True
+        config=True,
     )
 
     # FIXME: Refactor classes to separate build & run steps
@@ -554,7 +553,6 @@ class Repo2Docker(Application):
         # store ports on self so they can be retrieved in tests
         self.ports = ports
 
-
         container_volumes = {}
         if self.volumes:
             api_client = docker.APIClient(
@@ -575,7 +573,7 @@ class Repo2Docker(Application):
             detach=True,
             command=run_cmd,
             volumes=container_volumes,
-            environment=self.environment
+            environment=self.environment,
         )
 
         run_kwargs.update(self.extra_run_kwargs)
@@ -600,8 +598,7 @@ class Repo2Docker(Application):
                 self.log.error("Failed to copy repo contents into containe")
                 return container
 
-
-        while container.status == 'created':
+        while container.status == "created":
             time.sleep(0.5)
             container.reload()
 
@@ -644,13 +641,13 @@ class Repo2Docker(Application):
         if self.dry_run:
             return False
         # check if we already have an image for this content
-        client = docker.APIClient(version='auto', **kwargs_from_env())
+        client = docker.APIClient(version="auto", **kwargs_from_env())
         # looking for partial match without commit tag
         repo_image_name = self.output_image_spec[:-7]
 
         for image in client.images():
-            if image['RepoTags'] is not None:
-                for tag in image['RepoTags']:
+            if image["RepoTags"] is not None:
+                for tag in image["RepoTags"]:
 
                     if allow_revision and repo_image_name in tag:
                         self.output_image_spec = tag
@@ -686,7 +683,7 @@ class Repo2Docker(Application):
         else:
             if self.git_workdir is None:
 
-                checkout_path = tempfile.mkdtemp(prefix='repo2docker')
+                checkout_path = tempfile.mkdtemp(prefix="repo2docker")
                 self.git_workdir = checkout_path
             else:
                 checkout_path = self.git_workdir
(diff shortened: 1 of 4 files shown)


betatim:
My understanding of #632 was that the repository itself hadn't changed in content but what had changed is what volumes might be mounted or what command you want to execute in the container.

This means I was thinking the process we would be trying to implement is something like performing the following manual steps:
```
$ repo2docker some/directory/to/a/git/repo
$ docker images
$ docker run <image-name-found-from-previous-command> some/command.sh
$ docker run <image-name-found-from-previous-command> some/other-command.sh
$ docker run <image-name-found-from-previous-command> -v/tmp:/host-tmp some/command.sh
```
and so on. So you aren't changing the contents of the repository. Is this also what you had in mind? I'm asking because from the text in your comment I think not. Want to make sure we are on the same page :)

tmtaybah:
I think I might be missing something. It seems like a combination of the command line args '--volumes' and '[cmd](https://github.com/jupyter/repo2docker/blob/df89849e21a5c9d97d3b5a06dd049e8d18c7ec49/repo2docker/__main__.py#L96)' coupled with [image caching](https://github.com/jupyter/repo2docker/blob/df89849e21a5c9d97d3b5a06dd049e8d18c7ec49/repo2docker/app.py#L621) (which avoids a rebuild) can execute the steps in your example.

betatim:
Good point. Ok so we are trying to do something else :)

The idea then is to allow the user to specify a repository and if there is already an image for that repository we use it. As well as populating it with the latest content of the repository?

What do you think of making this a separate command to avoid having lots of command-line flags and having to check which of them are exclusive/don't work together (for example what should `--image-name` do in this case)? Some off the cuff ideas: `repo2docker-from-cache`, `repo2docker-cached`, `repo2docker-reuse-image`, `repo2docker-with-new-content`. Neither of them really gets me excited but I still think moving towards having separate commands for separate jobs is the future :-/

betatim:
I think we should modify the `build()` function in `app.py` to do something along the lines of:
1. search for an existing (potentially old) image for this "repo"
1. if there is one copy the contents of `checkout_dir` into it
1. return without performing further build steps

By copying the contents into the image we keep things closer to how building the image works the first time around. This means we can keep the clean up logic as it is in `master`. Instead of having to add new clean up logic which runs after the container stops which would be quite tricky to do.

I think by not mounting the directory we also side step some/all of the reason for prepending `/private` to the checkout path.

betatim:
Another idea for names: `repo2docker-refresh-content`, `repo2docker-with-fresh-content` or maybe as a flag `--refresh-content`? I think we should try and find a name where (ideally) the user knows straight away what it does. For me `run_changes` is too ambiguous (does it run changes that somehow have been queued up or ...?)

betatim:
https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.put_archive is probably what is needed if we want to copy things into the container. It doesn't work with an image so we'd have to start a container from the found image, use `put_archive`, commit the image(?) and then use it. It would mean we get a new image for each new revision that is run without the cost of building the whole image.

tmtaybah:
I changed the flag name to **--reuse-image** but we can change it to something better if it could potentially be confused with **--no-build**. I also merged **find_image()** and **find_repo_image()** into one function with a _allow_revision_ flag, like you suggested. And used the copy function (put_archive) instead of mounting a volume. 

Before copying the new contents of the image into the container we delete the old contents to ensure that no deleted files from the old image remain. Is there a better way to remove the contents of the container rather than deleting _home/user_ ? Maybe we can change workdir to _home/user/repo_name_ ?

tmtaybah:
This PR is ready for review, please take a look at your earliest convenience!

repo2docker/app.py
```diff
@@ -549,6 +557,24 @@ def start_container(self):
...
+            repo_archive = archive_repo(repo_path)
+
+            # delete old container contents
+            cmd = "sh -c 'rm -rf {}/*'".format(image_workdir)
+            container.exec_run(cmd, stderr=True, stdout=True)
+
+            copied = container.put_archive(image_workdir, repo_archive)
+            shutil.rmtree(self.git_workdir, ignore_errors=True)
+            if not copied:
+                self.log.error("Failed to copy repo contents into containe")
```
nuest: Just a little typo here, should be "container".

tmtaybah: Fixed! Thanks :)

consideRatio:
Hi! I'm reviewing PRs in the project to help with this project maintenance.

---

I perceive this feature to be too fragile to work reliably short term, and too complicated to maintain support for long term. Due to that I'm opined towards not accepting this PR implementation.

I'm thinking that:
- Files are copied after the container is started, making the container startup command possibly break. I've seen this in Kubernetes with a postStart lifecycle hook - where something is executed after the container starts, but people (myself for example) initially has the expectation that it will be instant before other things happen.
- The copying of files seems quite complicated, making it hard to maintain.

I'll make a judgement call and close this PR for now, returning to development of a feature like this is absolutely an option still though if a case is made for it an capacity is allocated towards it!

I'm sorry we couldn't follow up in time with reviews on this work @tmtaybah!

Issue #395, opened 2018-09-10
Error message when running repo2docker as root


DrosoNeuro:
I tried running repo2docker as root and got a rather cryptic error message. 

```Removing intermediate container 22753025f42a
 ---> 48ed20783f45
Step 13/32 : RUN adduser --disabled-password     --gecos "Default user"     --uid ${NB_UID}     ${NB_USER}
 ---> Running in 7ece88c37beb
adduser: The user `root' already exists.
Removing intermediate container 7ece88c37beb
The command '/bin/sh -c adduser --disabled-password     --gecos "Default user"     --uid ${NB_UID}     ${NB_USER}' returned a non-zero code: 1(rcs18) root@DrosoNeuro:/home/drosoneuro
```
Is there any way to tell repo2docker to create a user under a different name?

betatim:
You can pass `--user MYUSERNAMEHERE` to ask repo2docker to use a different user when creating the container.

Having a better error message that recommends not running repo2docker as root would be a good first contribution.

21BITGUY:
Can I give it a try?

betatim:
Yes please!

PR #676, opened 2019-05-07
[MRG] Do not try to build the image with root as the primary user.

### LINKED ISSUE #267: adduser: The user `root' already exists.
In a fresh installation of `docker-ce` on `Ubuntu-16.04`, if I try to locally build a binder repository, then I get the following logs (and finally the error):
```
root@myserver:/mount/SDE/crowdAI/conda# repo2docker . --no-run
Using CondaBuildPack builder
Step 1/28 : FROM buildpack-deps:artful
 ---> 766d0aa6a29e
Step 2/28 : RUN apt-get update &&     apt-get install --yes --no-install-recommends locales &&     apt-get purge &&     apt-get clean &&     rm -rf /var/lib/apt/lists/*
 ---> Using cache
 ---> b173f69c6e6a
Step 3/28 : RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen &&     locale-gen
 ---> Using cache
 ---> 3a5031e18af9
Step 4/28 : ENV LC_ALL en_US.UTF-8
 ---> Using cache
 ---> a221ace95304
Step 5/28 : ENV LANG en_US.UTF-8
 ---> Using cache
 ---> e0315886a2f4
Step 6/28 : ENV LANGUAGE en_US.UTF-8
 ---> Using cache
 ---> 0fc8bee8cc99
Step 7/28 : ENV SHELL /bin/bash
 ---> Using cache
 ---> 5fd1340b2573
Step 8/28 : ARG NB_USER
 ---> Using cache
 ---> 2b0535a59b11
Step 9/28 : ARG NB_UID
 ---> Using cache
 ---> 655d60cd28de
Step 10/28 : ENV USER ${NB_USER}
 ---> Using cache
 ---> 1bc2aa837513
Step 11/28 : ENV HOME /home/${NB_USER}
 ---> Using cache
 ---> ab2784542f9e
Step 12/28 : RUN adduser --disabled-password     --gecos "Default user"     --uid ${NB_UID}     ${NB_USER}
 ---> Running in bc1f2314bb1d
adduser: The user `root' already exists.
Removing intermediate container bc1f2314bb1d
The command '/bin/sh -c adduser --disabled-password     --gecos "Default user"     --uid ${NB_UID}     ${NB_USER}' returned a non-zero code: 1
``` 

This can be deterministically replicated by : 
```
pip install jupyter-repo2docker
git clone https://github.com/binder-examples/conda
cd conda
repo2docker . --no-run
```

Xarthisius:
Fixes #267 and #395. Solution suggested in https://github.com/jupyter/repo2docker/issues/267#issuecomment-478691267

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -16,6 +16,8 @@ API changes
 
 Bug fixes
 ---------
+- Prevent building the image as root if --user-id and --user-name are not specified
+  in :pr:`676` by :user:`Xarthisius`.
 
 
 Version 0.9.0
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -8,6 +8,7 @@ Usage:
     python -m repo2docker https://github.com/you/your-repo
 """
 import argparse
+import errno
 import json
 import sys
 import logging
@@ -650,6 +651,19 @@ class Repo2Docker(Application):
                                extra=dict(phase='building'))
 
                 if not self.dry_run:
+                    if os.geteuid() == 0:
+                        self.log.error(
+                            'Root as the primary user in the image is not permitted.\n'
+                        )
+                        self.log.info(
+                            "The uid and the username of the user invoking repo2docker "
+                            "is used to create a mirror account in the image by default. "
+                            "To override that behavior pass --user-id <numeric_id> and "
+                            " --user-name <string> to repo2docker.\n"
+                            "Please see repo2docker --help for more details.\n"
+                        )
+                        sys.exit(errno.EPERM)
+
                     build_args = {
                         'NB_USER': self.user_name,
                         'NB_UID': str(self.user_id),
--- a/tests/unit/test_app.py
+++ b/tests/unit/test_app.py
@@ -1,3 +1,5 @@
+import errno
+import pytest
 from tempfile import TemporaryDirectory
 from unittest.mock import patch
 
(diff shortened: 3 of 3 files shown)


Xarthisius:
Did I draw a short straw and ubuntu removed `artful` within past few days/hours?

betatim:
I think it is more likely this was a temporary connection thing as some of the tests passed that use the same docker file 🤞.

betatim:
Though there is no good reason to use artful here, so we should switch them to bionic which we use for the rest of repo2docker.

Xarthisius:
> I think it is more likely this was a temporary connection thing as some of the tests passed that use the same docker file .

I can reproduce that locally: `docker run --rm -ti ubuntu:artful apt-get update -qqy`. I think it's gone, gone. I will bump to bionic.

betatim:
lol. It looks like the first test "passes" because we expect it to fail and it does, just not because of overallocation of RAM but because there is a problem with building the image. This means your suggestion of "artful support stopped" is probably right. I'll make a PR to switch the tests to bionic.

Xarthisius:
> lol. It looks like the first test "passes" because we expect it to fail and it does, just not because of overallocation of RAM but because there is a problem with building the image. This means your suggestion of "artful support stopped" is probably right. I'll make a PR to switch the tests to bionic.

OK, will wait for you PR then.

betatim:
https://github.com/jupyter/repo2docker/pull/677 for updating the base image.

betatim:
#677 is merged. If we rebase this PR we should be good to go.

This PR is a great example of why it is so hard to estimate how long it will take to do something. "Change up a few lines to give a better error message, child's play!" ... except then we go down the rabbit hole of Ubuntu releases and how the Docker API works :)

Xarthisius:
> #677 is merged. If we rebase this PR we should be good to go.

Already did that :)

betatim:
Ah yes, I should make use of GitHub snitching on the `git push -f` crowd. It used to be that well kept secret between me and github when I did that ... and now they tell everyone! :)

PR #677, opened 2019-05-07
[MRG] Update base image used for memory limit checks

betatim:
Locally this still fails :-/

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -523,13 +523,17 @@ class BuildPack:
         tar.close()
         tarf.seek(0)
 
-        limits = {
-            # Always disable memory swap for building, since mostly
-            # nothing good can come of that.
-            'memswap': -1
-        }
+        # If you work on this bit of code check the corresponding code in
+        # buildpacks/docker.py where it is duplicated
+        limits = {}
         if memory_limit:
-            limits['memory'] = memory_limit
+            # We'd like to always disable swap but all we can do is set the
+            # total amount. This means we only limit it when the caller set
+            # a memory limit
+            limits = {
+                'memory': memory_limit,
+                'memswap': memory_limit + 1
+            }
 
         build_kwargs = dict(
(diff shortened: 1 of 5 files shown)


betatim:
It seems like (locally) the limit doesn't actually get applied/used :-/ I'll need a bit of time to investigate what is going on.

Xarthisius:
> Locally this still fails :-/

@betatim I vaguely remember that I had problem imposing mem limits for docker using python API. It accepted the size notation but it wasn't setting the limit properly. I'd try to use number of bytes explicitly.

betatim:
Mystery solved. The [docker-py documentation](https://docker-py.readthedocs.io/en/stable/api.html#module-docker.api.build) says that a value of `-1` will disable swap, which we were using. However https://docs.docker.com/engine/reference/commandline/build/ specifies that `-1` means "unlimited swap". In the [docker-py code](https://github.com/docker/docker-py/blob/23c5c001c6345c431fe31d8dfb62a38cc59720bc/docker/api/build.py#L123-L175) I can't find anywhere where they do processing of the values we passed in.

Updated our code to reflect this new knowledge. Though I am confused why all tests have passed on travis (for 5adc4b2) when they fail locally. Speculation: there is no swap for the whole machine on travis?

yuvipanda:
I dug around a bit more, and found https://github.com/docker/docker.github.io/issues/4335. Is there a reason we're setting memory swap to be one byte more, instead of exactly the same?

betatim:
When I had it set to 1 (a guess for "if we can't disable then let's use 1byte") I got an error the said "swap has to be bigger than memory limit", I didn't check if they actually meant "equal or larger". Checking now.

betatim:
Looks like setting it to the same amount works! That is a nicer solution :D Thanks for asking.

yuvipanda:
@betatim awesome! I changed the comment around it a little, otherwise it LGTM.

yuvipanda:
Merged with @betatim's blessing. <3

PR #679, opened 2019-05-09
Do not try to build the image with root as the primary user

Xarthisius:
This a redo of https://github.com/jupyter/repo2docker/pull/676. It errors r2d when root users try to run it without `--user-id` specified. The difference between this PR and #676 is that check is done against `self.user_id`. PTAL @betatim

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -8,6 +8,7 @@ Usage:
     python -m repo2docker https://github.com/you/your-repo
 """
 import argparse
+import errno
 import json
 import sys
 import logging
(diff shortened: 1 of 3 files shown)


betatim:
Let's try this again :)

repo2docker/app.py
```diff
@@ -650,6 +651,19 @@ def build(self):
...
+                            'Root as the primary user in the image is not permitted.\n'
+                        )
+                        self.log.info(
+                            "The uid and the username of the user invoking repo2docker "
+                            "is used to create a mirror account in the image by default. "
+                            "To override that behavior pass --user-id <numeric_id> and "
+                            " --user-name <string> to repo2docker.\n"
+                            "Please see repo2docker --help for more details.\n"
+                        )
+                        sys.exit(errno.EPERM)
```
yuvipanda: Since r2d can be used as a library, this code should raise an exception that's caught elsewhere and the message printed. Otherwise, the calling code just exits and can not actually handle anything here.

See https://github.com/jupyter/repo2docker/issues/696.

Previously, #496

PR #680, opened 2019-05-12
Diff generated by applying `black` to the code base

choldgraf:
This is a diff to show off what happens in the repo2docker repository when you run the default black command:

```
cd path/to/repo2docker
black .
```

Mostly this is just for comparison's sake to help discussion!


minrk:
We picked up black as an experiment [in jupyterhub](https://github.com/jupyterhub/jupyterhub/pull/2418) via [pre-commit](https://pre-commit.com), along with a few other auto-formatting hooks (import order, etc.) I think it's been a mostly positive experience. Having formatting/lint-fixing be fully automatic with a single command is pretty handy. As a pre-commit hook, I've been frustrated a few times as it balks when it has to fix things and I have to commit twice (it's still auto-fixed, but the hook exits with nonzero status if it makes changes so the commit succeeds on second attempt after a message about the fix), rather than what I'd really prefer, which is quietly fixing everything it can fix without my having to do anything.

I don't love all the changes black makes, but it is very nice to eliminate most style from the discussion.

One thing I'm not really a fan of is black's string quote defaults, so the jupyterhub black config skips quote-normalization, which is the vast majority of changes here.

choldgraf:
any particular reason that you don't like those string quote defaults? Is it because you want to have double-quotes in the strings themselves?

I think it'll also be the case that no single code style will make us all happy, so the bigger question is whether there are "dealbreakers" that we can't live without (and that break from black).

betatim:
I would skip the string quotes normalisation not so much because of preference over which type is "better" but because it causes a lot of code churn and it isn't that big a deal IMHO.

Because of the annoying feature of not re-formatting on the fly that Min mentioned I end up running `black .` frequently even without committing or have it configured to run on save in my editor.

One thing I'd do if I was the lone developer on a project is configure black to stick to 79 characters as line length instead of its default for longer lines. Happy to go with the majority vote on that though.

I think we should take over all of the JupyterHub setup in terms of pre-commit hooks and CI setup.

choldgraf:
(fwiw, it seems like formatting automatically when a python file is saved [is possible with vscode](https://code.visualstudio.com/updates/v1_6#_format-on-save), so I imagine it'd be possible with other editors as well)

I think my general preference is to just use defaults unless we have a really strong reason and general agreement on a particular deviation we'd want. However I also don't want voting to become a blocker on this happening, so it seems reasonable to start with the jupyterhub defaults that are already there.

minrk:
> Is it because you want to have double-quotes in the strings themselves?

Black's smart about this - it allows single-quotes on strings with double-quotes inside to avoid escapes:

```python
" 'single quote inside' "
' "double quote inside" '
' \'single escaped quote inside\' '
" \"double escaped quote inside\" "
```

becomes

```python
" 'single quote inside' "
' "double quote inside" '
" 'single escaped quote inside' "
' "double escaped quote inside" '
```

The biggest reason for disabling quote normalization in the black docs is what @betatim mentioned, that it reduces churn when adopting black partway through the life of a project (in my experience, this one option is responsible for most of the changes black makes). It's also the change that black makes where I most often look at it and wish the change wasn't made. That part is not important, just a personal feeling.

> One thing I'd do if I was the lone developer on a project is configure black to stick to 79 characters as line length instead of its default for longer lines. Happy to go with the majority vote on that though.

...and one of my favorite things about black defaults is that it *doesn't* use strict 79 characters, which I think doesn't fit well with modern displays or typical indentation in Python. I see why black resists options, because it only has two and we are already disagreeing on both 😉.

I'm okay with black defaults, have a slight preference for disabling quote normalization, and stronger preference for not shortening the line length. If we disagree, I'm happy to defer to black defaults, which are likely to become widespread as black is adopted. I don't want to argue myself for picking up jupyterhub's own settings since those are basically my preferences applied when I did the black setup over there, but I'd certainly be happy with the consistency.

choldgraf:
In that case, my preference is that we just go with the defaults and decide how painful it feels after a few weeks/months of use. I'm happy to re-discuss enabling/disabling particular options in the future! To be honest it is hard for me to assess the primary negative people have mentioned (code churn) because I haven't been a part of any projects that have used Black yet (other than jupyterhub/jupyterhub)

betatim:
Given that luckily(!!) I am not the lone developer on this project: let's go with blacks defaults on line length :)

The only downside to code churn during the transition: you need to click one more time to see who actually last edited a line. I rarely look at the history though so happy to stick to black defaults on this as well.

yuvipanda:
LET'S DO IT!

betatim:
Closing this now that #699 got merged.

PR #681, opened 2019-05-20
[MRG] Allow absolute paths in build_script_files

### LINKED ISSUE #673: Add a hook for injecting files to tarfile with a Docker build context
### Proposed change
I'd like to propose adding a hook that would allow to add files directly to the tarfile that's used as a Docker build context. See: https://github.com/jupyter/repo2docker/blob/d674eceeba77ec3fd16b7123bf248b0aede2bf7e/repo2docker/buildpacks/base.py#L493-L521

Right know, there's an implicit assumption that all files injected to the context are a part of r2d itself:

https://github.com/jupyter/repo2docker/blob/d674eceeba77ec3fd16b7123bf248b0aede2bf7e/repo2docker/buildpacks/base.py#L516

### Rationale
This feature is useful(/necessary) for developing custom buildpacks as repo2docker plugins (following @yuvipanda's suggestion: https://github.com/jupyter/repo2docker/issues/487#issuecomment-479794426) in case they need to inject their own script files.

### Alternative options
Alternative, would be to somehow modify base template and `get_build_script_files` to take into account that files can reside outside of repo2docker repository. However, this sounds a lot more invasive.

Not much, because I'm a lazy bum:
```diff
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -488,6 +488,10 @@ class BuildPack:
             appendix=self.appendix,
         )
 
+    def add_files_to_docker_context(self, tar, filter=None):
+        """Inject files directly to the tar file used as the Docker build context."""
+        pass
+
     def build(self, client, image_spec, memory_limit, build_args, cache_from, extra_build_kwargs):
         tarf = io.BytesIO()
         tar = tarfile.open(fileobj=tarf, mode='w')
@@ -514,12 +518,15 @@ class BuildPack:
         for src in sorted(self.get_build_script_files()):
             src_parts = src.split('/')
             src_path = os.path.join(os.path.dirname(__file__), *src_parts)
-            tar.add(src_path, src, filter=_filter_tar)
+            if os.path.isfile(src_path):
+                tar.add(src_path, src, filter=_filter_tar)
 
         tar.add(ENTRYPOINT_FILE, "repo2docker-entrypoint", filter=_filter_tar)
 
         tar.add('.', 'src/', filter=_filter_tar)
 
+        self.add_files_to_docker_context(tar, filter=_filter_tar)
+
         tar.close()
         tarf.seek(0)
```

### Who can do this work?
I volunteer as a tribute!

Xarthisius:
Fixes #673

This PR allows to provide an absolute path for source files in `get_build_script_files`. As a result, plugins can inject files outside of r2d's repository. In case, source file path is relative, the old behavior is preserved, i.e. path is assumed to be relative to the directory where `repo2docker/buildpacks/base.py` is.

@yuvipanda PTAL!

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -5,8 +5,10 @@ import io
 import os
 import re
 import logging
-import docker
+import string
 import sys
+import hashlib
+import escapism
 import xml.etree.ElementTree as ET
 
 from traitlets import Dict
@@ -509,6 +511,17 @@ class BuildPack:
                 "RUN {}".format(textwrap.dedent(script.strip("\n")))
             )
 
+        # Based on a physical location of a build script on the host,
+        # create a mapping between:
+        #   1. Location of a build script in a Docker build context
+        #      ('assemble_files/<escaped-file-path-truncated>-<6-chars-of-its-hash>')
+        #   2. Location of the aforemention script in the Docker image
+        # Base template basically does: COPY <1.> <2.>
+        build_script_files = {
+            self.generate_build_context_filename(k)[0]: v
+            for k, v in self.get_build_script_files().items()
+        }
+
         return t.render(
             packages=sorted(self.get_packages()),
             path=self.get_path(),
@@ -518,11 +531,40 @@ class BuildPack:
             build_script_directives=build_script_directives,
             pre_assemble_script_directives=pre_assemble_script_directives,
             assemble_script_directives=assemble_script_directives,
-            build_script_files=self.get_build_script_files(),
+            build_script_files=build_script_files,
             base_packages=sorted(self.get_base_packages()),
             post_build_scripts=self.get_post_build_scripts(),
             start_script=self.get_start_script(),
             appendix=self.appendix,
+        )
+
+    @staticmethod
+    def generate_build_context_filename(src_path, hash_length=6):
+        """
+        Generate a filename for a file injected into the Docker build context.
+
+        In case the src_path is relative, it's assumed it's relative to directory of
+        this __file__. Returns the resulting filename and an absolute path to the source
+        file on host.
+        """
+        if not os.path.isabs(src_path):
+            src_parts = src_path.split("/")
+            src_path = os.path.join(os.path.dirname(__file__), *src_parts)
+
+        src_path_hash = hashlib.sha256(src_path.encode("utf-8")).hexdigest()
+        safe_chars = set(string.ascii_letters + string.digits)
+
+        def escape(s):
+            return escapism.escape(s, safe=safe_chars, escape_char="-")
+
+        src_path_slug = escape(src_path)
+        filename = "build_script_files/{name}-{hash}"
+        return (
+            filename.format(
+                name=src_path_slug[: 255 - hash_length - 20],
+                hash=src_path_hash[:hash_length],
+            ).lower(),
+            src_path,
         )
 
     def build(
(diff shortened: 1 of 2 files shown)


repo2docker/buildpacks/base.py
```diff
@@ -481,13 +482,38 @@ def render(self):
...
+
+        In case the src_path is relative, it's assumed it's relative to directory of
+        this __file__. Returns the resulting filename and an absolute path to the source
+        file on host.
+        """
+        if not os.path.isabs(src_path):
+            src_parts = src_path.split('/')
+            src_path = os.path.join(os.path.dirname(__file__), *src_parts)
+        BLOCKSIZE = 65536
+        hash_md5 = hashlib.md5()
```
yuvipanda: I think we can get away with hashing the absolute path instead of having to hash the contents of the file. What do you think?

yuvipanda: Also let's use sha256, it should be fast enough - especially if we're just hashing the filename.

repo2docker/buildpacks/base.py
```diff
@@ -481,13 +482,38 @@ def render(self):
...
+        hash_md5.update(src_path.encode())
+        with open(src_path, 'rb') as fh:
+            buf = fh.read(BLOCKSIZE)
+            while len(buf) > 0:
+                hash_md5.update(buf)
+                buf = fh.read(BLOCKSIZE)
+        return hash_md5.hexdigest(), src_path
```
yuvipanda: We should probably include at least some part of the filename in the resulting filepath we return, to make debugging easier. How about something like: `assemble_files/<escaped-file-path-truncated>-<6-chars-of-hash>` or something like that?

See https://github.com/jupyterhub/binderhub/blob/6b2908d7aaf4a7ec62beed0019de54db06494214/binderhub/builder.py#L127 where we do something like that, keeping in mind the 255 char linux filepath limit.

repo2docker/buildpacks/base.py
```diff
@@ -481,13 +482,38 @@ def render(self):
             labels=self.get_labels(),
             build_script_directives=build_script_directives,
             assemble_script_directives=assemble_script_directives,
-            build_script_files=self.get_build_script_files(),
+            build_script_files={
```
yuvipanda: Should probably move this to a temporary variable for clarity, and add a comment about what it's doing.

repo2docker/buildpacks/base.py
Xarthisius: I assumed that build scripts are small enough so that it won't matter, but I'm happy to oblige.

Xarthisius: Changed to `sha256` and computing slug using filepath only.

repo2docker/buildpacks/base.py
Xarthisius: Fixed in 5447508

Xarthisius:
@yuvipanda I believe I addressed all your inline comments and added tests.

repo2docker/buildpacks/base.py
```diff
@@ -481,13 +494,39 @@ def render(self):
...
+            src_path = os.path.join(os.path.dirname(__file__), *src_parts)
+
+        src_path_hash = hashlib.sha256(src_path.encode('utf-8')).hexdigest()
+        safe_chars = set(string.ascii_letters + string.digits)
+
+        def escape(s):
+            return escapism.escape(s, safe=safe_chars, escape_char='-')
+
+        src_path_slug = escape(src_path)
+        filename = 'assemble_files/{name}-{hash}'
```
yuvipanda:
```suggestion
        filename = 'build_script_files/{name}-{hash}'
```

yuvipanda:
@Xarthisius one minor change I proposed (making up for an original mistake I made in my review!) otherwise LGTM.

The failing test is due to #684 I believe.

yuvipanda:
@Xarthisius I've now merged #684. Can you rebase to see if all the tests pass?

Thank you

repo2docker/buildpacks/base.py
Xarthisius: Updated

repo2docker/buildpacks/base.py
Xarthisius: Updated

PR #683, opened 2019-05-21
[MRG] Use getpass instead of pwd to fetch username

betatim:
`getpass` is available on windows as well as unix, whereas `pwd` is not
available on windows.

`getpass.getuser()` also looks are environment variables which means users can trivially change the value. I don't think this is a problem for us as we don't use it for access control and give users a way to set this to arbitrary values via CLI flag already.

Using this instead of `pwd` means we are a bit closer to working on Windows which is nice even though we don't officially support the windows platform. Unless there is a strong reason to I'd vote for taking the option that does not break Windows.

Not quite sure how to test this :-/

### DIFF
--- a/repo2docker/__init__.py
+++ b/repo2docker/__init__.py
@@ -1,3 +1,5 @@
 from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
+
+from .app import Repo2Docker
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -7,13 +7,12 @@ Usage:
 
     python -m repo2docker https://github.com/you/your-repo
 """
-import argparse
 import errno
 import json
 import sys
 import logging
 import os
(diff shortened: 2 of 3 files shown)


yuvipanda:
WFM! I restarted the failing test. I would say this could be unit tested by setting an enviy variable that getpass reads

betatim:
The test failed because Python 3.4 has been EOL and ruamel.yaml has dropped support for it (no more wheels) in their latest release: https://bitbucket.org/ruamel/yaml/src/572b7bdf491d897f309442b00a23297a84c1c611/README.rst?at=default

Good idea with the test, added that.

betatim:
The travis failure for Python 3.4 is waiting on #684.

betatim:
All tests pass. Ready to be merged.

yuvipanda:
Thanks, @betatim

PR #684, opened 2019-05-21
[MRG] Drop support for Python 3.4

betatim:
This PR drops support for running repo2docker on a system that has Python 3.4 only. This does not effect users ability to build repositories that need Python 3.4

Why? One of our dependencies has stopped shipping wheels for Python 3.4 because https://www.python.org/dev/peps/pep-0429/ states that the last release of Python 3.4 was in March 2019.

There are two questions here:
* should we also drop Python 3.4 support?
* why does our travis setup not install ruamel.yaml "from source" but fail when there are no wheels? https://travis-ci.org/jupyter/repo2docker/jobs/535148903#L643


yuvipanda:
Most of our other projects require 3.5+ so am happy with this. I think we should give others an opportunity to object though. Not sure what our process for something like this should be

yuvipanda:
Based on @minrk's comment in https://github.com/jupyter/repo2docker/issues/521#issuecomment-448630192, I'm happy to drop 3.4 support.

betatim:
I tried to also switch us to testing on Python 3.7 by default (instead of 3.6) but apparently there is no Python 3.7 for Ubuntu 14.04, which is the base version we have in travis. Maybe that is a project for another day to move us forwards in time a bit there.

yuvipanda:
Opened https://github.com/jupyter/repo2docker/issues/685 to talk about switching base image in Travis to 16.04

yuvipanda:
As reference, Ubuntu 14.04 was EOL in April 2019, and shipped with Python 3.4. Python 3.5 was released in September 2015, almost 4 years ago.

I am going to merge this now, but we can always revert if needed.

Issue #686, opened 2019-05-21
Add support for packagcompiler for julia support


jlperla:
cc: @davidanthoff @arnavs  This is a specific implementation of the discussion in #601 

We have now discussed how best to annotate project files for packagecompiler, which is a prerequisite for getting immediate support in repo2docker

**Background**
Right now, repo2docker doesn't support the package compiler but:
- @arnavs has a proof of concept in https://github.com/arnavs/compiled-binder-example which relies on the fancy https://github.com/arnavs/compiled-binder-example/blob/master/Dockerfile  to do the package compilation.  It is not possible to do this with a `postBuild` 
- In practice, it is not possible or desireable to package compile every package, so you need a list of settings and whitelisted packages to compile into the system image.
- The `Project.toml` and `Manifest.toml` file need to be the only sources of setup for the repository

**Extension to the Project File**
After discussion with maintainers of Julia and the package manager (cc: @KristofferC @aviks) the proposal was to standardize on a section such as the following inside of the Project.toml files: 
```
[deps]
Markdown = "XXXXXXXXXXX"
Random = "XXXXXXX"

[packagecompiler]
compilerflags = "-O3"
packages = ["Markdown", "Random"]
```
Note that this is not repo2docker specific, and could be used by other backend systems for packagecompile'ing based on a Project and Manifest.  It would be hand-modified in the Project.toml.

@SimonDanisch any thoughts on this including the naming or additional options to be able to pass to the `compile_incremental`?

**Implementation of PackageCompiler Support**

Now, to implement support repo2docker, we can produce a PR which does the following:
- Around https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/julia/julia_project.py#L30 where the `Project` is loaded, we would check `if 'packagecompiler' in project_toml:` and if so, load up all of the `packages` inside of it into a list for later.  Similarly, if `compilerflags` was given, we would store it.
- In https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/julia/julia_project.py#L126   we would check to see if any packagecompiler packages were given.  If so, then we do a variation of the logic in https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/julia/julia_project.py#L130-L131   where we packagecompile the passed in list, and pass in the `compilerflags` as an option.
- Adds additional documentation comment to https://mybinder.readthedocs.io/en/latest/config_files.html#project-toml-install-a-julia-environment about package compiling

[...]

PR #688, opened 2019-05-23
Julia PackageCompiler [WIP] 

arnavs:
This handles `repo2docker`'s side of the changes in #686.

See https://github.com/arnavs/binder-bakery for an example repo to test against. 

The main changes to code are: a `self.baked_goods()` method to digest the `[packagecompiler]` bit of TOML, and then feeding that into the Julia script (if the TOML section is present.)

cc: @jlperla, @betatim 

Edit: Will change the names to be less whimsical.


arnavs:
@betatim Ready for you.

arnavs:
Sorry, made a few fixes. Turns out the errors weren't simply local. 

@betatim, do you know how I could run the output of `get_assemble_scripts(self)` with more permissions? It's successfully building a new system image, but returning an `EACCESS` error when we try to unlink the old one.

Note: PackageCompiler is still fragile. This is the "state of the art" branch, but it's still rickety.

arnavs:
Just wanted to check in about this (@betatim and others.) The changes we've made to the `get_assemble_script` are correct (i.e., it returns the right commands.) The problem is that whatever's running them doesn't have the appropriate unlinking permissions.

betatim:
I've not had time to read and think about this PR yet, sorry! Fantastic weather for outdoors activities and general busy-ness. I will try and look at this towards the end of this week.

I did want to stop by to let you know that we merged #699 which means we are now using [black](https://black.readthedocs.io/en/stable/) as code formatter. The motivation for doing this is to have a uniform style that will make it easier to read code.

arnavs:
Sounds good, I'll rebase and run the linter. And likewise, the weather here is almost unbearably good. 

Whenever you get around to it, the tl;dr is that we can spit out the right commands in the Julia assemble script. They're just not being run with the appropriate permissions (at the stage where we unlink the old Julia system image, and link the new one we've built.)

arnavs:
Hey @betatim, do you think you'll be able to take a look this week?

arnavs:
I'll close this PR per @davidanthoff's comments in the issue. We can re-open it if we decide to go this route.

betatim:
Thanks a lot for all the work on this @arnavs! Sorry I didn't get to this earlier/at all. Maybe in this case we lucked out though given @davidanthoff's thoughts.

PR #689, opened 2019-05-23
Switch Travis CI to Ubuntu Xenial 16.04

### LINKED ISSUE #685: Switch base travis CI to use Ubuntu 16.04
### Proposed change

Ubuntu 16.04 is available on Travis, and we should probably switch. I think 14.04 is EOL, and this will let us test on Python 3.7.

Theoretically, we should be able to flip the appropriate config value in .travis.yml and that should be it.

jrbourbeau:
This PR updates Travis CI to use Ubuntu 16.04. Fixes #685

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: python
 sudo: required
 cache: pip


minrk:
Thanks!

PR #690, opened 2019-05-24
set CONDA_DEFAULT_ENV

minrk:
so that `conda install` commands affect the kernel environment instead of the base env

This should preserve `conda install` behavior for Dockerfiles that inherit from r2d images.
This behavior is already managed at runtime by the ENTRYPOINT which properly activates the env.

cc @tam203 who reported the issue on gitter

I'm still testing this locally, and trying to think of a reasonable way to include an automated test for this.

### DIFF
--- a/repo2docker/buildpacks/conda/activate-conda.sh
+++ b/repo2docker/buildpacks/conda/activate-conda.sh
@@ -3,9 +3,16 @@ CONDA_PROFILE="${CONDA_DIR}/etc/profile.d/conda.sh"
 test -f $CONDA_PROFILE && . $CONDA_PROFILE
 if [[ "${KERNEL_PYTHON_PREFIX}" != "${NB_PYTHON_PREFIX}" ]]; then
     # if the kernel is a separate env, stack them
-    # so both are on PATH
+    # so both are on PATH, notebook first
     conda activate ${KERNEL_PYTHON_PREFIX}
     conda activate --stack ${NB_PYTHON_PREFIX}
+
+    # even though it's second on $PATH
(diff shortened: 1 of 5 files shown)


minrk:
Added a test that covers `postBuild`, which is *currently* equivalent to the reported `RUN conda install ...`, but not strictly the same since we could in the future add activation-wrappers around running postBuild, but not downstream RUN.

repo2docker/buildpacks/conda/activate-conda.sh
```diff
@@ -3,9 +3,16 @@ CONDA_PROFILE="${CONDA_DIR}/etc/profile.d/conda.sh"
...
-    # so both are on PATH
+    # so both are on PATH, notebook first
     conda activate ${KERNEL_PYTHON_PREFIX}
     conda activate --stack ${NB_PYTHON_PREFIX}
+
+    # even though it's second on $PATH
+    # make sure CONDA_DEFAULT_ENV is the *kernel* env
+    # so that `!conda install PKG` installs in the kernel env
+    # where user packages are installed, not the notebook env
+    # which only contains UI when the two are different
```
betatim: I had to read this a few times for it to make sense but I can't suggest a different wording that is easier to parse :-/

You always want to install into the "kernel env" and never into the "notebook env", but sometimes the two are the same in which case you are still installing into the kernel env, it just happens to also be the notebook env. You just gotta have your wits about you with all these envs :)

betatim:
I think it is fine that we don't have a perfect test as it gets a bit involved to create one that does actually test the full thing. Worth creating an issue for this? Dunno.

LGTM, waiting for travis

PR #692, opened 2019-05-26
Adding contentprovider documentation

### LINKED ISSUE #444: Document new content provider infra
In https://github.com/jupyter/repo2docker/pull/242 @betatim added some new infrastructure for defining new content providers. However, I don't believe that any documentation was added as a part of this. We should:

* Document how to add a new content provider
* Document what content providers are currently supported
* Make it clear that people can use different content providers in their repos

choldgraf:
This adds mention of the ContentProvider class in the "architecture" page (that currently explains what BuildPacks are. I imagine that this information might be incomplete or maybe even incorrect, so it would be great if @betatim and @pdurbin could give their own edits or comments.

This might close #444 if we're happy enough with the information.

### DIFF
--- a/docs/source/architecture.md
+++ b/docs/source/architecture.md
@@ -119,7 +119,7 @@ a path to a repository. This might be a local path or a URL. Upon being called,
 commands:
 
 * Run the `detect()` method on the repository path given to `repo2docker`. This
-  should return `True` if the path matches what the ContentProvider is looking
+  should return any value other than `None` if the path matches what the ContentProvider is looking
   for.
 
   > For example, the [`Local` ContentProvider](https://github.com/jupyter/repo2docker/blob/80b979f8580ddef184d2ba7d354e7a833cfa38a4/repo2docker/contentproviders/base.py#L64)
--- a/docs/source/architecture.md
+++ b/docs/source/architecture.md
@@ -125,7 +125,8 @@ commands:
   > For example, the [`Local` ContentProvider](https://github.com/jupyter/repo2docker/blob/80b979f8580ddef184d2ba7d354e7a833cfa38a4/repo2docker/contentproviders/base.py#L64)
   > checks whether the argument is a valid local path. If so, then `detect(`
   > returns true.
-* If `detect(` returns `True`, run `fetch(` with the input path. This should
+* If `detect()` returns something other than `None`, run `fetch()` with the
+  returned value as its argument. This should
   result in the contents of the repository being placed locally to a folder.
 
 For more information on ContentProviders, take a look at
--- a/docs/source/architecture.md
+++ b/docs/source/architecture.md
@@ -124,7 +124,8 @@ commands:
 
   > For example, the [`Local` ContentProvider](https://github.com/jupyter/repo2docker/blob/80b979f8580ddef184d2ba7d354e7a833cfa38a4/repo2docker/contentproviders/base.py#L64)
   > checks whether the argument is a valid local path. If so, then `detect(`
-  > returns true.
+  > returns a dictionary: `{'path': source}` which defines the path to the repository.
+  > This path is used by `fetch()` to check that it matches the output directory.
 * If `detect()` returns something other than `None`, run `fetch()` with the
   returned value as its argument. This should
   result in the contents of the repository being placed locally to a folder.
--- a/docs/source/architecture.md
+++ b/docs/source/architecture.md
@@ -118,7 +118,7 @@ a path to a repository. This might be a local path or a URL. Upon being called,
 `repo2docker` will loop through all ContentProviders and perform the following
 commands:
 
-* Run the `detect(` method on the repository path given to `repo2docker`. This
+* Run the `detect()` method on the repository path given to `repo2docker`. This
   should return `True` if the path matches what the ContentProvider is looking
   for.
 
--- a/docs/source/architecture.md
+++ b/docs/source/architecture.md
@@ -3,7 +3,7 @@
 This is a living document talking about the architecture of repo2docker
 from various perspectives.
 
-## Buildpack
+## Buildpacks
 
 The **buildpack** concept comes from [Heroku](https://devcenter.heroku.com/articles/buildpacks)
 and Ruby on Rails' [Convention over Configuration](http://rubyonrails.org/doctrine/#convention-over-configuration)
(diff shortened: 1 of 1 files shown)


pdurbin:
@choldgraf looks good! Thanks!

choldgraf:
@pdurbin do you think this would have made your life easier when you were extending repo2docker?

pdurbin:
@choldgraf sorry, let me clear up some confusion. I certainly haven't extended repo2docker but I'd love for someone out there to make it work with Dataverse. That's what https://github.com/IQSS/dataverse/issues/4714 is about. If you check out the thread at https://groups.google.com/d/msg/dataverse-community/VG6gTMEd_Ps/Xy7jDhVoBwAJ you can see how much excitement there was when the idea of integrating Dataverse with Binder was brought up but I believe the effort has stalled. There are some notes from a UC Berkely class called "Spring 2018 CloudWG: Dataverse + Binderhub" at https://docs.google.com/document/d/1ZR9AhhAAqCRmXjFkqAm7UAJlxEUe4ooyOgWOILwOIXw/edit?usp=sharing but it's been quite a while since I've heard from any of those students. I do think the docs you added help so thank you!

docs/source/architecture.md
```diff
@@ -95,16 +95,41 @@ so less restrictions apply to this than to `get_build_scripts`.
...
+
+ContentProviders provide a way for `repo2docker` to know how to find and
+retrieve a repository. They follow a similar pattern as the BuildPacks
+described above. When `repo2docker` is called, its main argument will be
+a path to a repository. This might be a local path or a URL. Upon being called,
+`repo2docker` will loop through all ContentProviders and perform the following
+commands:
+
+* Run the `detect(` method on the repository path given to `repo2docker`. This
+  should return `True` if the path matches what the ContentProvider is looking
```
betatim: it should return something that `is not None`. So for example you could return an empty dictionary which is `False` (try `bool({})`) but `{} is not None == True`. The return value is then used as the argument to `fetch`. This allows detect to modify/change/clean up/add to the argument passed in from the CLI.

docs/source/architecture.md
```diff
@@ -95,16 +95,41 @@ so less restrictions apply to this than to `get_build_scripts`.
...
+## ContentProviders
+
+ContentProviders provide a way for `repo2docker` to know how to find and
+retrieve a repository. They follow a similar pattern as the BuildPacks
+described above. When `repo2docker` is called, its main argument will be
+a path to a repository. This might be a local path or a URL. Upon being called,
+`repo2docker` will loop through all ContentProviders and perform the following
+commands:
+
+* Run the `detect(` method on the repository path given to `repo2docker`. This
```
betatim:
```suggestion
* Run the `detect()` method on the repository path given to `repo2docker`. This
```
I've not seen the `detect(` notation before and find it a bit weird, it feels like writing a sentence with no punctuation at the end 

Somehow unfinished :)

betatim:
Left some comments, otherwise LGTM.

docs/source/architecture.md
```diff
@@ -95,16 +95,41 @@ so less restrictions apply to this than to `get_build_scripts`.
...
+commands:
+
+* Run the `detect()` method on the repository path given to `repo2docker`. This
+  should return `True` if the path matches what the ContentProvider is looking
+  for.
+
+  > For example, the [`Local` ContentProvider](https://github.com/jupyter/repo2docker/blob/80b979f8580ddef184d2ba7d354e7a833cfa38a4/repo2docker/contentproviders/base.py#L64)
+  > checks whether the argument is a valid local path. If so, then `detect(`
+  > returns true.
+* If `detect(` returns `True`, run `fetch(` with the input path. This should
```
choldgraf:
```suggestion
* If `detect()` returns `True`, run `fetch(` with the input path. This should
```

docs/source/architecture.md
```diff
@@ -95,16 +95,41 @@ so less restrictions apply to this than to `get_build_scripts`.
...
+
+ContentProviders provide a way for `repo2docker` to know how to find and
+retrieve a repository. They follow a similar pattern as the BuildPacks
+described above. When `repo2docker` is called, its main argument will be
+a path to a repository. This might be a local path or a URL. Upon being called,
+`repo2docker` will loop through all ContentProviders and perform the following
+commands:
+
+* Run the `detect()` method on the repository path given to `repo2docker`. This
+  should return `True` if the path matches what the ContentProvider is looking
```
choldgraf:
```suggestion
  should return any value other than `None` if the path matches what the ContentProvider is looking
```

docs/source/architecture.md
```diff
@@ -95,16 +95,41 @@ so less restrictions apply to this than to `get_build_scripts`.
...
+  for.
+
+  > For example, the [`Local` ContentProvider](https://github.com/jupyter/repo2docker/blob/80b979f8580ddef184d2ba7d354e7a833cfa38a4/repo2docker/contentproviders/base.py#L64)
+  > checks whether the argument is a valid local path. If so, then `detect(`
+  > returns true.
```
choldgraf:
```suggestion
  > returns True.
```

is this correct?

choldgraf:
thanks for the comments @betatim - I made a couple quick modifications w/ some suggested changes. How does that look to you?

docs/source/architecture.md
betatim: No. All content providers should (and do) return `None` if they don't match and something "that isn't `None`" when they do match. What precisely it is they return is left to them. The only requirement is that the `fetch()` method of the provider knows what to do with it.

docs/source/architecture.md
betatim: Apparently I can't commit these :( but I would if I could :)

Screen Shot 2019-05-28 at 10 17 57

docs/source/architecture.md
choldgraf: I was asking in this case specifically about the `Local` contentprovider rather than in general. I'll dig into it a bit more to figure it out

choldgraf: see updated edit above, LMK if that's OK

betatim: > was asking in this case specifically

Ah ok. I missed that. 

Yes the suggestion is correct.

betatim: Do you have any idea why I can't commit the suggestion? Normally it works but it seems for this PR it consistently doesn't (and there are other PRs where it doesn't). So confusing.

choldgraf:
@betatim I'm not sure why it wasn't allowing commits, that's annoying :-/ maybe github changed something in their permissions? Eitherway, I've committed it!

betatim:
This is failing because of the problem that #695 fixes. Once #695 is merged (the PR is ready and green) we can merge this.

Issue #234, opened 2018-02-12
Running repo2docker on non-git repositories?


choldgraf:
**A PR gets us part of the way in https://github.com/jupyter/repo2docker/pull/236 but it could use help from someone to push it over the finish line**

In this zenodo issue (https://github.com/zenodo/zenodo/issues/1416) somebody brought up the challenge that zenodo repositories are essentially zip files, they're not git repositories.

It looks like this would *not* be handled by a build-pack, since the fetching happens prior to config file detection (https://github.com/minrk/repo2docker/blob/3ab4503f7c161abc25a809bfe207234aa6471e21/repo2docker/app.py#L152). To that extent, a couple thoughts:

1. We should have documentation explaining the possibilities one way or another (e.g. a section in the docs called "does it *have* to be a git repository?") where we say either "yes" or "right now yes, but we'd love to support more workflows in the future".
2. Once we agree on 1, I can open a PR to add a docs change, and either close or convert this issue into something actionable.

yuvipanda:
#102 and #133  are related, and have useful pointers I think.

There's two things here - one is support for this in repo2docker, and support for it in binder. Binder already has a concept of Providers, which repo2docker should also have.

Agree with documenting in the short term!

yuvipanda:
#236 starts work on this, would be nice if someone can continue that :)

rlconley:
I'm taking a look at continuing #236.

betatim:
@rlconley sorry I didn't reference this issue as well, take a look at https://github.com/jupyter/repo2docker/pull/242 if you are working on this.

nuest:
@betatim @rlconley Has there been any work on a Zenodo provider since this issue was last updated?

betatim:
I don't know of any work on this.

The foundations are there (content providers). We'd need a new content provider that can take a Zenodo DOI and figure out the URL of the related ZIP file, download and unzip it.

nuest:
That does not seem too hard to do :-)

I'll have to figure out first if I can run our own UI (which uses multiple microservices..) within a Binder, but I am optimistic. If so, then I'll take this on (and potentially also #102). No clear ETA, hopefully May.

betatim:
This issue and #102 should probably be treated together.

betatim:
Closing this as we now have Zenodo. We can create a new one for specific other sources when we think of them.

PR #693, opened 2019-05-27
[MRG] Zenodo content provider

betatim:
The goal of this PR is to make `repo2docker 10.5281/zenodo.3229823` work.

related: #234, #102

### DIFF
--- a/repo2docker/contentproviders/zenodo.py
+++ b/repo2docker/contentproviders/zenodo.py
@@ -15,18 +15,34 @@ class Zenodo(ContentProvider):
     """Provide contents of a Zenodo deposit."""
 
     def detect(self, doi, ref=None, extra_args=None):
+        doi = doi.lower()
         # 10.5281 is the Zenodo DOI prefix
-        if doi.startswith('10.5281'):
+        if doi.startswith("10.5281/"):
             resp = urlopen("https://doi.org/{}".format(doi))
             self.record_id = resp.url.rsplit("/", maxsplit=1)[1]
-            return {'record': self.record_id}
+            return {"record": self.record_id}
+
+        elif doi.startswith("https://doi.org/10.5281/") or doi.startswith(
+            "http://doi.org/10.5281/"
+        ):
+            resp = urlopen(doi)
+            self.record_id = resp.url.rsplit("/", maxsplit=1)[1]
+            return {"record": self.record_id}
+
+        elif doi.startswith("https://zenodo.org/record/") or doi.startswith(
+            "http://zenodo.org/record/"
+        ):
+            self.record_id = doi.rsplit("/", maxsplit=1)[1]
+            return {"record": self.record_id}
 
     def fetch(self, spec, output_dir, yield_output=False):
-        record_id = spec['record']
+        record_id = spec["record"]
 
         yield "Fetching Zenodo record {}.\n".format(record_id)
-        req = Request("https://zenodo.org/api/records/{}".format(record_id),
-                      headers={"accept": "application/json"})
+        req = Request(
+            "https://zenodo.org/api/records/{}".format(record_id),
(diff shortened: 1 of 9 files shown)


repo2docker/contentproviders/zenodo.py
```diff
@@ -0,0 +1,99 @@
...
+        yield "Fetching Zenodo record {}.\n".format(record_id)
+        req = Request(
+            "https://zenodo.org/api/records/{}".format(record_id),
+            headers={"accept": "application/json"},
+        )
+        resp = urlopen(req)
+
+        record = json.loads(resp.read().decode("utf-8"))
+
+        def _fetch(file_ref, unzip=False):
```
betatim: This function feels quite messy, but it needs to handle weird cases like "there is only one ZIP file in this zenodo record so lets unpack it, and if that generates a new sub directory move stuff from it to the top, then delete the sub dir and the zip file" and the variations on this.

betatim:
Before we merge this: needs documentation!

repo2docker/contentproviders/zenodo.py
```diff
@@ -0,0 +1,99 @@
...
+from zipfile import ZipFile, is_zipfile
+
+from .base import ContentProvider
+from ..utils import copytree
+
+
+class Zenodo(ContentProvider):
+    """Provide contents of a Zenodo deposit."""
+
+    def detect(self, doi, ref=None, extra_args=None):
```
yuvipanda: I like this!

How can we support zenodo installations that aren't at zenodo.org?

repo2docker/contentproviders/zenodo.py
```diff
@@ -0,0 +1,99 @@
...
+        # 10.5281 is the Zenodo DOI prefix
+        if doi.startswith("10.5281/"):
+            resp = urlopen("https://doi.org/{}".format(doi))
+            self.record_id = resp.url.rsplit("/", maxsplit=1)[1]
+            return {"record": self.record_id}
+
+        elif doi.startswith("https://doi.org/10.5281/") or doi.startswith(
+            "http://doi.org/10.5281/"
+        ):
+            resp = urlopen(doi)
```
yuvipanda: Do we already use requests? If not maybe we should for this. I also think we should set a user-agent to be good citizens.

yuvipanda: We're already depending on requests because docker-py does.

repo2docker/contentproviders/zenodo.py
betatim: Good question. We'd have to maintain a list of DOI prefixes that we know map to a Zenodo instance and its hostname.

When we start doing this I'd also swap what `detect` returns to include the hostname instead of constructing the URL in `fetch`. I added a comment with this thinking in it.

I'd punt actually implementing it unless we know of an instance (I don't know any :( ).

repo2docker/contentproviders/zenodo.py
betatim: Ahhh! I had a look but didn't see `requests` in the direct dependencies and (re)learnt how `urlopen` works. Because I ended up using `urlopen` the tests now rely on being able to open a `file:///path/some/where/temporary/fake.zip` (for example [here](https://github.com/jupyter/repo2docker/pull/693/files#diff-6be528671d2ae05d5d68659f729a276bR64)) which we can't do with `requests`. How strong are your feelings on switching?

User-agent is a good idea, done.

yuvipanda:
@betatim this looks good to me. What do you think about adding an integration test in the externals list? I'd like that...

betatim:
Added an external test and merged.

Code coverage went down a bit (hence the red tick) but most of the uncovered lines are in the `copytree` function taken from the stdlib.

PR #699, opened 2019-05-31
[MRG] Add auto-formatting setup

### LINKED ISSUE #616: black based autoformatting
I'd like to propose we follow JupyterHub's lead and adopt black as commit hook and use it to format our code.

https://github.com/jupyterhub/jupyterhub/issues/2416

I'm not sure we need to go as far as JupyterHub and sort imports etc, just using black as a pre-commit hook seems like a good compromise.

It will cause a bit of disruption/code churn so I think we should schedule it for some time in the future so we can merge or close as many outstanding PRs as possible to minimise the number of "nothing but formatting" commits.

What do people think?

betatim:
Trying to make one complete PR that gets us from no autoformatting to "all the autoformatting". We have attempted this before but the longer a PR with these changes sits the harder it gets to rebase etc. hence trying to make one PR that hopefully ticks all the boxes so it can be merged quickly.

The way things are setup in this PR travis will enforce black's formatting, you can make commits without installing pre-commit, we install `black` as a dev dependency so that people can run it manually, the contrib docs explain how to setup the pre-commit hooks. If we start using more pre-commit hooks we need to make it mandatory to install them but I'd propose we delay that. Getting black is more important to me than implementing some of the other hooks.

This PR:
* [x] sets upauto-formatting enforcement on travis,
* [x] installs the necessary tools, and
* [x] adds instructionns to the contributing docs
* [x] run auto formatting

This PR supersedes #680 and closes #616.


choldgraf:
I think we had enough convergence in #616 that we should give this a shot, especially since @betatim put in all the legwork to get the auto-formatting working. I'm happy to keep iterating on this in the coming weeks as we get a feel for how this works. For now, I'll plan to merge this later today unless somebody speaks up

choldgraf:
Alrighty then, let the great blackening begin!

PR #703, opened 2019-06-13
add `git-lfs` to dockerfile used for the repo2docker image

minrk:
ensures clones are in-tact when there are LFS-stored files.

Test repo added at binderhub-ci-repos/lfs

I don't think the test really covers this, since I think the tests don't use the repo2docker in the image, they use the local install.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,11 @@ RUN mkdir /tmp/wheelhouse \
 FROM python:${PYTHON_VERSION}-slim
 
 # we do need git, though
-RUN apt-get update \
- && apt-get -y install --no-install-recommends git \
+# git-lfs is in backports
+ARG DEB_RELEASE=stretch
+RUN echo "deb http://deb.debian.org/debian ${DEB_RELEASE}-backports main" > /etc/apt/sources.list.d/backports.list \
+ && apt-get update \
+ && apt-get -t ${DEB_RELEASE}-backports -y install --no-install-recommends git git-lfs \
  && rm -rf /var/lib/apt/lists/*
 
 # install repo2docker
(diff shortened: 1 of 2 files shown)


betatim:
> I don't think the test really covers this, since I think the tests don't use the repo2docker in the image, they use the local install.

This means `git-lfs` happens to already be installed on the machine we run the tests on right?

Should we add a comment in the install docs/FAQ that if you want git LFS support you need to install it? I don't think there is anyway for us to specify that `git-lfs` is a requirement for running repo2docker so docs is all we have really.

betatim:
I tweaked the title a bit to (hopefully) make it easier to remember that this isn't about the `Dockerfile` used as a template when repo2docker builds an image but the `Dockerfile` used to make the repo2docker container.

Another idea: Is there an easy way for the Git content provider to notice that there are git-lfs files in the repo and print a warning so we can tell people to install it?

minrk:
> This means git-lfs happens to already be installed on the machine we run the tests on right?

Yes. `git-lfs` is pretty common, so I

> Another idea: Is there an easy way for the Git content provider to notice that there are git-lfs files in the repo and print a warning so we can tell people to install it?

Easy-ish. LFS is implemented as git filters, so it can be found in `.gitattributes` like [this one](https://github.com/binderhub-ci-repos/lfs/blob/9abf54a17c3d31823b037b80e9ab0903c0e40694/.gitattributes#L1):

```ini
*.dat filter=lfs diff=lfs merge=lfs -text
```

The first version is to check if lfs is in git-attributes at all, assuming if it's configured that it's used, the second version is to parse that file and search for matching files. That's probably harder than we want to do.

betatim:
Let's add the warning as a new PR or make an issue for it. Merging.

PR #704, opened 2019-06-13
[MRG] Generalize Zenodo content provider to support other Invenio repositories

tmorrell:
Updates the Zenodo content provider to:

- Use a more general DOI -> URL conversion
- Support different hosts and different metadata organizations

My interest is supporting CaltechDATA, which is based on Zenodo but has a different metadata scheme.  I'll also work on getting Binder buttons into CaltechDATA, which would be cool.

I'm still working on figuring out how to generalize the object path stuff.  The idutils functions could also be copied into utils; don't know what your preference is on dependencies.  Any and all suggestions are appreciated!

### DIFF
--- a/repo2docker/contentproviders/zenodo.py
+++ b/repo2docker/contentproviders/zenodo.py
@@ -1,15 +1,16 @@
 import os
 import json
 import shutil
+import copy
 
 from os import makedirs
 from os import path
 from urllib.request import build_opener, urlopen, Request
 from zipfile import ZipFile, is_zipfile
-from idutils import normalize_doi, is_doi
 
 from .base import ContentProvider
 from ..utils import copytree
+from ..utils import normalize_doi, is_doi
 from .. import __version__
 
 
@@ -40,6 +41,24 @@ class Zenodo(ContentProvider):
         else:
             return doi
 
+    def _getfromdict(self, datadict, dotpath):
+        # Use a dotpath (string separated by periods)
+        # to access vaules in a dictionary
+        # data.files.0 returns value at dataDict[data][files][0]
+        split = dotpath.split(".")
+        # We check if we have any digits and convert these to
+        # ints for list access
+        mapList = []
+        for s in split:
+            if s.isdigit():
+                mapList.append(int(s))
+            else:
+                mapList.append(s)
+        values = copy.deepcopy(datadict)
+        for k in mapList:
+            values = values[k]
+        return values
+
     def detect(self, doi, ref=None, extra_args=None):
         """Trigger this provider for things that resolve to a Zenodo/Invenio record"""
         # We need the hostname (url where records are), api url (for metadata),
@@ -51,7 +70,7 @@ class Zenodo(ContentProvider):
                 "hostname": ["https://zenodo.org/record/", "http://zenodo.org/record/"],
                 "api": "https://zenodo.org/api/records/",
                 "filepath": "files",
-                "filename": "files.key",
+                "filename": "filename",
                 "download": "links.download",
                 "type": "metadata.upload_type",
             },
@@ -61,8 +80,9 @@ class Zenodo(ContentProvider):
                     "http://data.caltech.edu/records/",
                 ],
                 "api": "https://data.caltech.edu/api/record/",
-                "filepath": "files",
-                "filename": "electronic_location_and_access.electronic_name.0",
+                "filepath": "metadata.electronic_location_and_access",
+                "filename": "electronic_name.0",
+                "download": "uniform_resource_identifier",
                 "type": "metadata.resourceType.resourceTypeGeneral",
             },
         ]
@@ -91,8 +111,8 @@ class Zenodo(ContentProvider):
         def _fetch(file_ref, unzip=False):
             # the assumption is that `unzip=True` means that this is the only
             # file related to the zenodo record
-            with self._urlopen(file_ref["links"]["download"]) as src:
-                fname = file_ref["filename"]
+            with self._urlopen(self._getfromdict(file_ref, host["download"])) as src:
+                fname = self._getfromdict(file_ref, host["filename"])
                 if path.dirname(fname):
                     sub_dir = path.join(output_dir, path.dirname(fname))
                     if not path.exists(sub_dir):
@@ -126,9 +146,13 @@ class Zenodo(ContentProvider):
                         copytree(path.join(output_dir, d), output_dir)
                         shutil.rmtree(path.join(output_dir, d))
 
-        is_software = record["metadata"]["upload_type"] == "software"
-        only_one_file = len(record["files"]) == 1
-        for file_ref in record["files"]:
+        is_software = self._getfromdict(record, host["type"]).lower() == "software"
+        files = self._getfromdict(record, host["filepath"])
+
+        #
+
+        only_one_file = len(files) == 1
+        for file_ref in files:
             for line in _fetch(file_ref, unzip=is_software and only_one_file):
                 yield line
 
--- a/repo2docker/contentproviders/zenodo.py
+++ b/repo2docker/contentproviders/zenodo.py
@@ -33,20 +33,10 @@ class Zenodo(ContentProvider):
         # Transform a DOI to a URL
         # If not a doi, assume we have a URL and return
         if is_doi(doi):
-            # Zenodo instances will most likely use DataCite DOIs
-            # Could also resolve the DOI to get the url
-            # Get url from DataCite
             doi = normalize_doi(doi)
 
-            req = Request(
-                "https://api.datacite.org/dois/{}".format(doi),
-                headers={"accept": "application/json"},
-            )
-            resp = self._urlopen(req)
-
-            record = json.loads(resp.read().decode("utf-8"))
-            url = record["data"]["attributes"]["url"]
-            return url
+            resp = self._urlopen("https://doi.org/{}".format(doi))
+            return resp.url
         else:
             return doi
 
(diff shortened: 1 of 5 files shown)


betatim:
This is cool!

I think I'd import the `is_doi` and `normalize_doi` functions (with a reference to the original and copyright notice) as they are both one-liners so not worth adding a dependency that has dependencies IMHO.

One thing I couldn't figure out from a quick reading of the code: we should organise for this content provider to only trigger for some DOIs (right now ones that point to Zenodo.org or CaltechDATA, in the future other Invenio instances (is Invenio really the software behind Zenodo? I always thought of it as a web framework kinda thing as the agenda and document servers at CERN were also "Invenio")) so that other content providers can deal with DOIs that point to other types of repositories. Same logic if someone passes a URL to either a Invenio record or dx.doi.org with a DOI that points to one of them.

repo2docker/contentproviders/zenodo.py
```diff
@@ -28,39 +29,59 @@ def _urlopen(self, req, headers=None):
...
-    def detect(self, doi, ref=None, extra_args=None):
-        """Trigger this provider for things that resolve to a Zenodo record"""
-        # To support Zenodo instances not hosted at zenodo.org we need to
-        # start maintaining a list of known DOI prefixes and their hostname.
-        # We should also change to returning a complete `record_url` that
-        # fetch() can use instead of constructing a URL there
-        doi = doi.lower()
-        # 10.5281 is the Zenodo DOI prefix
-        if doi.startswith("10.5281/"):
+    def _process_doi(self, doi):
```
betatim: could we call it `doi2url` or `resolve_doi` or something? I had to read it a few times before I understood what it does and doesn't do (not helped in this case by the diff view :) ) ... I am a fan of names that tell you what the thing does.

tmorrell: Yup!

tmorrell:
I don't think we need an explicit DOI check.  I resolve all DOIs to a url, but then see if the url is on the list of known hosts.  So if it's a DOI for a different repository, the url won't match and you won't get anything back.

There are two versions of Invenio, and there's going to be a lot of new development around Zenodo and Zenodo flavors in the next year.  It's why I didn't rename the content provider...Zenodo has better name recognition at this point.

betatim:
The thinking behind making it work with any kind of record was that even an empty Git repository or one containing no files recognised by repo2docker is a valid input repository. Which made me think that we could use that logic to allow datasets as a way to make it easy for people to explore a dataset. A notebook and a Python interpreter is already a good starting point for exploring data. If people use it to create very large images I'd solve this at the deployment level (by configuring BinderHub to reject DOIs that don't point to software records. What do you think of that thinking?

repo2docker/contentproviders/zenodo.py
```diff
@@ -28,39 +30,78 @@ def _urlopen(self, req, headers=None):
...
+        split = dotpath.split(".")
+        # We check if we have any digits and convert these to
+        # ints for list access
+        mapList = []
+        for s in split:
+            if s.isdigit():
+                mapList.append(int(s))
+            else:
+                mapList.append(s)
+        values = copy.deepcopy(datadict)
```
betatim: Why do we need to copy the dict? `import copy` is on my list of code smells as 95% of the time it turns out it isn't needed, hence the question.

Not a huge fan of `mapList` as a name, we should try and stick with `map_list` style and find a name that tells us what is in the variable.

What do you think of the following as an alternative which we could even place in `utils.py` as it is bound to be useful in other places at some point:
```python
def deep_get(dikt, path):
    """Get a value located in `path` from a nested dictionary.

    Use a string separated by periods as the path to access
    values in a nested dictionary:

    deep_get(data, "data.files.0") == data["data"]["files"][0]
    """
    value = dikt
    for component in path.split("."):
        if component.isdigit():
            value = value[int(component)]
        else:
            value = value[component]
    return value
```

repo2docker/utils.py
```diff
@@ -391,3 +391,24 @@ def copytree(
...
+
+
+# Code segments below from idutils (https://github.com/inveniosoftware/idutils)
+# Copyright (C) 2015-2018 CERN.
+# Copyright (C) 2018 Alan Rubin.
+# Licensed under BSD-3-Clause license
+doi_regexp = re.compile(
+    "(doi:\s*|(?:https?://)?(?:dx\.)?doi\.org/)?(10\.\d+(.\d+)*/.+)$", flags=re.I
+)
+"""See http://en.wikipedia.org/wiki/Digital_object_identifier."""
```
betatim: can we turn this into a comment preceding the regex. This isn't a doc string really/it isn't customary to have doc strings for variables.

repo2docker/utils.py
```diff
@@ -391,3 +391,24 @@ def copytree(
...
+
+
+def is_doi(val):
+    """Test if argument is a DOI."""
+    return doi_regexp.match(val)
+
+
+def normalize_doi(val):
+    """Normalize a DOI."""
```
betatim: In the doc string can we explain a bit what kind of normalisation happens? As is the reader who needs more than the function name is left none the wiser after reading this :) We should either make it contain more info (or scrap it) IMHO

tests/unit/contentproviders/test_zenodo.py
```diff
@@ -24,13 +24,47 @@ def test_detect():
...
+            == "3232985"
+        )
+        assert (
+            Zenodo().detect("https://zenodo.org/record/3232985")["record"] == "3232985"
+        )
+
+        # only two of the three calls above have to resolve a DOI
+        assert fake_urlopen.call_count == 2
+
+    with patch.object(Zenodo, "_urlopen") as fake_urlopen:
```
betatim: I'd make this a new test method named something like `test_caltech_instance` or some such. This will help people from the future when part of the test breaks. I even wonder if we could combine the two tests into one where the input and expected output is passed in as parameters (https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions)

repo2docker/contentproviders/zenodo.py
```diff
@@ -105,9 +146,13 @@ def _fetch(file_ref, unzip=False):
                         copytree(path.join(output_dir, d), output_dir)
                         shutil.rmtree(path.join(output_dir, d))
 
-        is_software = record["metadata"]["upload_type"] == "software"
-        only_one_file = len(record["files"]) == 1
-        for file_ref in record["files"]:
+        is_software = self._getfromdict(record, host["type"]).lower() == "software"
+        files = self._getfromdict(record, host["filepath"])
+
+        #
```
betatim: Was there a comment here that got lost? If yes we should add it back, otherwise remove one of the blank lines and this line.

repo2docker/contentproviders/zenodo.py
```diff
@@ -28,39 +30,78 @@ def _urlopen(self, req, headers=None):
...
+                "filename": "electronic_name.0",
+                "download": "uniform_resource_identifier",
+                "type": "metadata.resourceType.resourceTypeGeneral",
+            },
+        ]
+
+        url = self._doi2url(doi)
+
+        for host in hosts:
+            if any([url.startswith(s) for s in host["hostname"]]):
```
betatim: Nice!

repo2docker/contentproviders/zenodo.py
```diff
@@ -28,39 +30,78 @@ def _urlopen(self, req, headers=None):
...
+        for k in mapList:
+            values = values[k]
+        return values
+
+    def detect(self, doi, ref=None, extra_args=None):
+        """Trigger this provider for things that resolve to a Zenodo/Invenio record"""
+        # We need the hostname (url where records are), api url (for metadata),
+        # filepath (path to files in metadata), filename (path to filename in
+        # metadata), type (path to type in metadata)
+
```
betatim: I'd remove this to make it clear that the comment belongs to the `hosts` variable that follows.

betatim:
Nice work making the code more general! I left some comments on cleaning up stray white space, potentially parametrizing a test (we should definitely split it) and made a suggestion for a "deep get" that I find easier to follow (and maybe eliminate a use of `deepcopy()`).

betatim:
To [please the formatting robot](https://travis-ci.org/jupyter/repo2docker/jobs/547787240) `pip install black` and `black .` in the top level directory.

tmorrell:
Thanks for the great suggestions @betatim!  I think I've implemented everything, and I added a couple of new tests as well.  

For functional testing, I've got a notebook in CaltechDATA that works: `repo2docker 10.22002/d1.1250`

And for fun, here is how a binder badge will look in the repository: https://doi.org/10.22002/d1.1250

betatim:
Merged!

Nice work and with all those tests how could I not merge it :)

tests/unit/test_utils.py
```diff
@@ -82,3 +82,31 @@ def test_invalid_port_mapping(port_spec):
...
+def test_deep_get():
+    data = {"data": {"files": [1, 2, 3]}}
+    assert utils.deep_get(data, "data.files.0") == 1
+    assert utils.deep_get(data, "data.files.1") == 2
+    assert utils.deep_get(data, "data.files") == [1, 2, 3]
+    assert utils.deep_get(data, "data") == {"files": [1, 2, 3]}
+
+
+def test_is_doi():
+    assert utils.is_doi("10.1234/jshd123") != None
```
betatim: For even more Pythonicness people usually do `something is None` and `something is not None` instead of `==` and `!=`. To be honest I'd have to dig a bit to remember/explain why or in which edge cases `is` is better :D

betatim:
And it is live: https://mybinder.org/v2/zenodo/10.22002/d1.1250

PR #705, opened 2019-06-17
base docker image on alpine

minrk:
for a much smaller image (93 vs 231 MB) and faster builds.

The issue that prevented us from using alpine (#160) has since been resolved.


minrk:
FWIW, Apline 3.9 is still using Python 3.6, not 3.7.

PR #706, opened 2019-06-18
[WIP] contributing: black install / docs and pipenv over venv

consideRatio:
# PR Summary

It is a pure documentation update, regarding suggested way to work with the repo2docker source code. The changes can be [overviewed best in this view](https://github.com/jupyter/repo2docker/pull/706/files#diff-5f56cf96128c62808603c6f5d2c5b564) I think.

__Summary of changes__
Apparently `black .` will only autoformat files with .py extension, and ignore all our [`verify`](https://github.com/jupyter/repo2docker/blob/master/tests/conda/repo-path/verify) files for example. But, if `black` is passed the files explicitly, it will format those after finding `#!/usr/bin/env python` within them. If the Python package `pre-commit` is used with our in-repo [`.pre-commit-config.yaml`](https://github.com/jupyter/repo2docker/blob/master/.pre-commit-config.yaml) file, then we end up passing the files explicitly to `black` like that and it will succeed to format all the files as compared to only those with `.py` extension. This is what I try to explain and why this PR now suggest the use of `pre-commit run` or installing a hook to get it done automatically with `pre-commit install`. I ended up here because it caused an issue for me with our CI setup after having run `black .` and became really confused as the CI system rejected my PR. The difference was that CI used `pre-commit run --all files` instead.

---

While refreshing #649 I ended up learning various things.

1. ~I think usage of `pipenv` should be favored over `venv` in the contribution docs - so I moved instructions to use `pipenv` above `venv`.~ - I reverted this commit with a force push.
2. Black has an issue about not formatting Python files without `.py` extension, like our `verify` files in the tests folder. If a pre-commit git hook is used to invoke black, this is no longer an issue.
3. `pre-commit` was installed from `dev-requirements.txt` but `black` wasn't, but still explicitly written out to be installed. I moved `black` to `dev-requirements.txt`. Hmmm but I think `pre-commit` will install the `black` specified by `.pre-commit-config.yaml` separately anyhow, so in reality we should probably remove all dependency of `black` and always invoke it through `pre-commit`.
    **UPDATE:** I made it so in the latest commit.

# Review suggestions
- Form an opinion about 3 and ask me to act on it.
   **UPDATE:** I made a fourth commit directly... I removed black as a dependency and shifted recommendation to invoking it through `pre-commit run` as it will catch python files like `verify` as well.

### DIFF
--- a/docs/source/contributing/contributing.md
+++ b/docs/source/contributing/contributing.md
@@ -130,7 +130,6 @@ source bin/activate
 pip3 install -e .
 pip3 install -r dev-requirements.txt
 pip3 install -r docs/doc-requirements.txt
-pip3 install black
 ```
 
 This should install all the libraries required for testing & running repo2docker!
--- a/docs/source/contributing/contributing.md
+++ b/docs/source/contributing/contributing.md
@@ -152,13 +152,13 @@ We use [`black`](https://black.readthedocs.io/en/stable/) as code formatter to
 get a consistent layout for all the code in this project. This makes reading the
 code easier.
 
-While you can format your code with `black .` in the top-level directory of this
-repository, we suggest the use of a pre-commit git hook as it avoids a [known
-limitation](https://github.com/python/black/issues/491) for Python files without
-`.py` extension.
+While you can install `black` and format code with `black .` from the top-level
+directory of this repository, we recommend the use of a pre-commit git hook as
+it avoids a [known limitation](https://github.com/python/black/issues/491) for
+Python files without `.py` extension.
 
-To install a pre-commit git hook for black into your `.git/hooks` folder, we
-rely on the [pre-commit](https://github.com/pre-commit/pre-commit) python
+To install a pre-commit git hook to run `black` into your `.git/hooks` folder,
+we rely on the [pre-commit](https://github.com/pre-commit/pre-commit) python
 package to help us together with our `.pre-commit-config.yaml` file.
 
 ```bash
@@ -166,9 +166,19 @@ pre-commit install
 ```
 
 With the pre-commit git hook installed, `black`'s autoformatting will be
-executed before any commit completes. If additional changes are made by `black`
-your commit will fail. Then you need to stage the autoformat changes with `git
-add .` and try again.
+executed before any commit completes. If the pre-commit git hook causes
+additional changes by `black` your commit will fail. You then need to stage the
+autoformat changes with `git add .` and try again.
+
+You can run the installed pre-commit git hook manually as well:
+
+```bash
+# for the stages files to be committed
+pre-commit run
+
+# for all files
+pre-commit run --all-files
+```
 
 Our continuous integration will check that code is formatted properly and report
 failure if this is not the case.
--- a/docs/source/contributing/contributing.md
+++ b/docs/source/contributing/contributing.md
@@ -120,7 +120,10 @@ After cloning the repository, you should set up an
 isolated environment to install libraries required for running / developing
 repo2docker.
 
-There are many ways to do this but here we present you with two approaches: `virtual environment` or `pipenv`.
+There are many ways to do this but here we present you with two approaches:
+`virtual environment` or `pipenv`. In both approaches, you will end up
+installing all dependencies required for development within a virtual
+environment that is created and activated.
 
 - Using `virtual environment`
 
@@ -132,8 +135,6 @@ pip3 install -r dev-requirements.txt
 pip3 install -r docs/doc-requirements.txt
 ```
 
-This should install all the libraries required for testing & running repo2docker!
-
 - Using `pipenv`
 
 Note that you will need to install pipenv first using `pip3 install pipenv`.
@@ -141,9 +142,10 @@ Then from the root directory of this project you can use the following commands:
 
 ```bash
 pipenv install --dev
+pipenv shell
 ```
 
-This should install both the dev and docs requirements at once!
+This should install all requirements at once within an virtual environment that and enter the virtual environment!
 
 
 ### Code formatting
--- a/docs/source/contributing/contributing.md
+++ b/docs/source/contributing/contributing.md
@@ -149,39 +149,37 @@ This should install both the dev and docs requirements at once!
 ### Code formatting
 
 We use [`black`](https://black.readthedocs.io/en/stable/) as code formatter to
-get a consistent layout for all the code in this project. This makes reading the
-code easier.
+get a consistent layout for all the code in this project. Our Continuous
+Integration (CI) will check that code is formatted accordingly and report
+failure if isn't.
 
-While you can install `black` and format code with `black .` from the top-level
-directory of this repository, we recommend the use of a pre-commit git hook as
-it avoids a [known limitation](https://github.com/python/black/issues/491) for
-Python files without `.py` extension.
-
-To install a pre-commit git hook to run `black` into your `.git/hooks` folder,
-we rely on the [pre-commit](https://github.com/pre-commit/pre-commit) python
-package to help us together with our `.pre-commit-config.yaml` file.
+To perform autoformatting while also avoiding a [known
+limitation](https://github.com/python/black/issues/491) of using `black`
+directly, we use the already installed Python package called
+[`pre-commit`](https://github.com/pre-commit/pre-commit). `pre-commit` will
+utilize a config file in this repo (`.pre-commit-config.yaml`) to understand it
+should use `black`.
 
 ```bash
+# run black on all files
+pre-commit run --all-files
+
+# run black only on the staged files to be committed
+pre-commit run
+```
+
+`pre-commit` the Python package, can also help us autoformat our code before
+each commit by installing a *git hook*:
+
+```bash
+# installs a git hook for this repository (a script within .git/hooks)
 pre-commit install
 ```
 
 With the pre-commit git hook installed, `black`'s autoformatting will be
 executed before any commit completes. If the pre-commit git hook causes
-additional changes by `black` your commit will fail. You then need to stage the
-autoformat changes with `git add .` and try again.
-
-You can run the installed pre-commit git hook manually as well:
-
-```bash
-# for the stages files to be committed
-pre-commit run
-
-# for all files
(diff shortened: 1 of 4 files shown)


betatim:
When we introduced black the feeling was that `pre-commit` was too confusing and/or failed for some. This is why we have the "entry level" setup that asks people to `pip install black` instead of using `pre-commit`. As you point out `pre-commit` installs black itself so this is somewhat redundant. At the time @choldgraf had problems while following the `pre-commit` instructions. We should check if that is still the case and what people think about what the lowest barrier to entry setup is.

In general we have tried to only have one way to setup things in the contributing guide. The goal when writing the instructions is to have a very high chance of them "just working" for someone who isn't a regular Python user and being easy to translate into other ways of setting things up for more expert users. For example we never mention conda but assume that a conda user knows how to translate the instructions.

I am a little unconformable with `pipenv` because to me it feels like hiding steps. As in: I am an very experienced Python user and from the one `pipenv` command I can't instantly translate what I should do to get up and running with my preferred setup method. So while I don't actually use our venv instructions verbatim (and maybe even think thats not the best way to setup things) I do know how to translate them because they are so explicit.

Another problem I think we should ponder is how popular `pipenv` is amongst users. I am not sure how we'd get data on "venv users", "conda users", "pipenv users", etc. The only information I do have is that pipenv got hyped by its author and that he took several steps that in retrospect turned out to be premature (for example declaring it "the official way to install stuff") and aimed at misleading people.

In conclusion:

* I think we should keep the (super) explicit setup steps as the way we explain how to setup stuff and assume that people with opinions/other preferences are experienced enough to translate them.
* How do we balance explaining different methods of getting setup and keeping the guide short and clear? Should we list `venv` and `pipenv` and `conda` steps? Should we list none? My preference is to not cover any of the alternatives after we have the "basic and explicit" one (lower maintenance and clearer docs)

consideRatio:
Thank you for reviewing this Tim!

My experience is very clear, but I'm okay to adjust the PR for concrete change suggestions or even scrap most changes and open up a new PR with the single change of adding a note about [the issue I experienced](https://github.com/python/black/issues/491) with running the command: `black .`

## pipenv / venv
pipenv / venv - discussion about what that is listed first among two options that were already there, and me moving pipenv as a first entry. 

I've struggled with having multiple requirements files in general, it took me a long time before I grasped what went on within z2jh. This repo has that, and that is a key benefit of using pipenv + Pipfile. One more reason for using `pipenv` and `Pipfile` is that it allows us to `Pipfile.lock` and get something very reproducible, something I consider relevant for this repo.

In my opinion, and with less background in python about the history of tools etc, I'm less confused about a short instructions of pipenv, it does things I understand as well.

## Black / pre-commit
Regarding pre-commit focused instructions vs black focused instructions.

I'm one of the users that felt pre-commit was overkill initially, so I opted to not do that. I ended up needing to learn all about it in order to debug and conclude the difference between `black .` and `pre-commit run --all-files`. I then felt obliged to help others avoid this issue by documenting that and figured it would be easier to strongly encourage the use of pre-commit right away.

---

> I think we should keep the (super) explicit setup steps as the way we explain how to setup stuff and assume that people with opinions/other preferences are experienced enough to translate them.

I didn't grasp if/how that translates to action suggestions for this PR.

> How do we balance explaining different methods of getting setup and keeping the guide short and clear? Should we list venv and pipenv and conda steps? Should we list none? My preference is to not cover any of the alternatives after we have the "basic and explicit" one (lower maintenance and clearer docs)

I'd vote for pipenv + venv as it is no change to have that, or only pipenv for lowered maintenance and clearer docs.

betatim:
> I didn't grasp if/how that translates to action suggestions for this PR.

The action would be: don't remove/hide/make less prominent the venv based steps

> I'm one of the users that felt pre-commit was overkill initially

Good, well, bad you went in circles but good that we gained some experience :) I am all for the pre-commit based setup as that is what I actually use.

betatim:
For me the thing we should aim to achieve with the instructions is to present them in a language that as many readers as possible understand. Either because they can "blindly" copy and paste them and succeed or because they can read them and translate into their favourite language. For me this shared language is `pip install` based commands (even though I don't actually use it when I setup this repo for work). The way I'd pick the shared language is by number of users and "built-in'ness". For me `pip` ticks both those boxes: pip is used by more humans then any of the other options (this is an informed hunch*) and it is available in most Python installs (via system package managers, conda, pyenv, home brew, etc).

I use conda every day, all day to create environments and install software. However I think compared to the total number of Python users "no one else" uses conda which makes it a bad option for the shared language of install instructions. In order to achieve the "just copy&paste it" goal we'd have to also provide instructions for installing conda itself. I've chosen to be a conda user and learnt how to translate `pip install -r some-file.txt` into `conda install -n my-fave-env -c conda-forge <packages>`. I think that the effort to learn how to translate from the mainstream option to my niche option is on my plate.

I am not sure we need a lock file. IMHO that is relevant if you want to reliably produce deployments, however repo2docker is more like a library which means it should not pin its versions super tightly. We also create deployments of repo2docker but those are via docker images and we have one image for ~every commit on master so you can recover exactly the state things were in for each deployment.

---

* The hunch is based on watching Python users and what the PyPA starts with when you google for "install python package" https://packaging.python.org/tutorials/installing-packages/

consideRatio:
@betatim I removed the commit where I reordered the venv / pipenv instructions

consideRatio:
I consider pip to be easy, but not venv. Since we in the contributing docs suggest the use of pip+venv rather than only pip, i think pipenv is easier and more understandable to do the job.

betatim:
Nods, I don't really like venv either, compared to conda environment they feel like hacks. However at the last iteration of editing these docs there was support to have venv mentioned because (and here I agree) it makes sense to have some form of one-environment-per-project. Maybe we can change the start to have a sentence for humans without commands saying something like "It makes sense for you to create a environment (via venv, conda, pipenv, or other) for this project and setup the dependencies: ...now come the pip commands...". This way we don't tell people how to make an env, but we do tell them there are several ways and they need to pick one.

consideRatio:
This PR, only does this within that section currently:
image

Is it okay that we shift this discussion to another PR?

---

Somewhat hooked by the discussion though:

When I try to make what you suggest concrete I come up with the following idea:
- tell the users that venv/conda can be used to create an environment without saying how, and then pip install commands
- point out it is also possible to install and create an environment using the Pipfile / Pipfile.lock files in a single command after installing pipenv.

When I consider what I'd suggest myself, I conclude that I'd embrace pipenv and the Pipfile + Pipfile.lock we currently have and remove the [xxx-]requirements.txt / environment.yml files that are redundant. Perhaps adding a note that we can run commands to generate requirements.txt files from the Pipfile if we want.

consideRatio:
I did some rewording of the Code formatting section.

choldgraf:
mmmm, I got the same error I got before:

```bash
(dev) choldgraf@barbecue ~/github/forks/python/jupyter/jupyterhub/repo2docker (master) $ pre-commit run --all-files
[INFO] Installing environment for https://github.com/ambv/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: Command: ('/home/choldgraf/anaconda/envs/dev/bin/python', '-mvirtualenv', '/home/choldgraf/.cache/pre-commit/repo46rqbbtl/py_env-python3', '-p', 'python3')
Return code: 100
Expected return code: 0
Output:
    Using base prefix '/home/choldgraf/anaconda/envs/dev'
    New python executable in /home/choldgraf/.cache/pre-commit/repo46rqbbtl/py_env-python3/bin/python3
    Also creating executable in /home/choldgraf/.cache/pre-commit/repo46rqbbtl/py_env-python3/bin/python
    ERROR: The executable /home/choldgraf/.cache/pre-commit/repo46rqbbtl/py_env-python3/bin/python3 is not functioning
    ERROR: It thinks sys.prefix is '/' (should be '/home/choldgraf/.cache/pre-commit/repo46rqbbtl/py_env-python3')
    ERROR: virtualenv is not compatible with this system or executable
    Running virtualenv with interpreter /home/choldgraf/anaconda/envs/dev/bin/python3

Errors:
    /home/choldgraf/.cache/pre-commit/repo46rqbbtl/py_env-python3/bin/python3: error while loading shared libraries: libpython3.6m.so.1.0: cannot open shared object file: No such file or directory
```

consideRatio:
@choldgraf can you try leaving all envs, conda etc, and try using pipenv?

pip install pipenv
pipenv install --dev
pipenv shell
pre-commit run

consideRatio:
I've tried and tested the following on my Windows computer successfully. I corrected some details.

- The documentation for pipenv also needed a note about `pipenv shell` to activate the virtual environment.
- `pre-commit` needed to be added as a `--dev` package within our `Pipfile`, and `Pipfile.lock` needed to be updated to reflect this.

docs/source/contributing/contributing.md
```diff
@@ -130,37 +133,56 @@ source bin/activate
 pip3 install -e .
 pip3 install -r dev-requirements.txt
```
choldgraf: I believe we need to add `black` to this list, yes?

choldgraf:
Looking into this a bit more, it looks like in general pre-commit doesn't work well with conda unless it's installed *by* conda:

https://github.com/pre-commit/pre-commit/issues/701

So, I uninstalled pre-commit and then ran `conda install -c conda-forge pre-commit`

and then it worked!

So, we should update these instructions to tell people what to do if they're using `conda`. I am -1 *requiring* that people use `pipenv` instead of other package managers (though fine in general with package managers being in the docs)

willingc:
I would be hesitant to discontinue the recommendation `python3 -m venv named_env`. pipenv relies on the virtualenv package which was designed to meet the needs of creating a virtual environment in Python 2 and is a separate package. venv is available as part of CPython.

While I like the concept of a Pipfile.lock, I would prefer to have the option to use venv over pipenv.

consideRatio:
Hi all, I'll close this PR as I've lost momentum on it - anyone is of course welcome to continue from this point if wanted! Thanks everyone for working with me on this!

Issue #707, opened 2019-06-19
The need for speed


betatim:
This issue is about collecting ideas that could make the images produced by repo2docker smaller, faster to push/pull or faster to build.

I envision this thread to be a meta thread with lots of ideas that then either turn into PRs or get marked as "already done" or "not possible". This way we can use it as an entry point to finding other related issues or PRs.

Why make images smaller and faster to build? From our own experience and the first results from the [binder user survey](https://docs.google.com/forms/d/e/1FAIpQLSd3fiLCMuQsc48_ga2q_FJFqgFcVUkie7RBex4DtzzOyyNWHg/viewform) it is clear that faster builds and faster launches is something people really care about.

---

## Smaller images

A few ideas via https://jcrist.github.io/conda-docker-tips.html

* don't use MKL: we already use conda-forge as default -> openblas is our default, no more gains possible
* run `conda clean -afy`: already implemented
* `--freeze-installed`: not currently used, unsure if it would help, worth trying
* remove additional unnecessary files: we should do this
* use a smaller base image: not applicable as the ubuntu base we use should be present on all nodes in a BinderHub cluster -> this is "free" as it doesn't need pulling or pushing. Double check this is actually true

---

## Reordering build steps for faster builds

Right now we have to rebuild the whole image (from the point onwards where we copy in the contents of the repo) even if the user only changes a typo in the README. The reasoning behind this is that a `requirements.txt` could contain something like `-e .` which leads to the `setup.py` in the repo being executed. This in turn means the setup process could be executing "anything and depends on everything in the repo". There is no way of knowing that the one character change in the README won't change the build result.

However I think this is a fringe case and the common case is that people only install packages from PyPI and don't depend on the rest of the repository. How can we make it so this common case is faster and still get the rare case right?

 The obvious way to speed up builds and rebuilds is to copy only the `requirements.txt` into the repo, run the install step and then copy over the rest of the repository. This way a change in the README won't break the docker layer cache, which means rebuilds are fast.

One thing we could try is to copy the `requirements.txt` early, run `pip install -r requirements.txt` wrapped in a "if this fails just continue" block, then copy the full repo, rerun `pip install -r requirements.txt` which will either be a no-op (if the early run succeeded) or will clean up the breakage from the first run.

We invented `install.R` so we could declare it a mistake to rely on anything in the repository. This means we can copy it over early. This would save a large amount of user pain because R builds are some of the slowest builds we have. (see #716)

[...]

PR #709, opened 2019-06-20
[MRG] Remove more files during the image build to slimdown the image

betatim:
This is a test related to #707: removing static libraries and such.

### DIFF
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -191,6 +191,9 @@ class CondaBuildPack(BaseImage):
                     r"""
                 conda env update -p {0} -f "{1}" && \
                 conda clean --all -f -y && \
+                find ${{CONDA_DIR}} -follow -type f -name '*.a' -delete && \
+                find ${{CONDA_DIR}} -follow -type f -name '*.pyc' -delete && \
+                find ${{CONDA_DIR}} -follow -type f -name '*.js.map' -delete && \
                 conda list -p {0}
                 """.format(
                         env_prefix, environment_yml
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -191,9 +191,9 @@ class CondaBuildPack(BaseImage):
                     r"""
                 conda env update -p {0} -f "{1}" && \
                 conda clean --all -f -y && \
-                find ${{CONDA_DIR}} -follow -type f -name '*.a' -delete && \
-                find ${{CONDA_DIR}} -follow -type f -name '*.pyc' -delete && \
-                find ${{CONDA_DIR}} -follow -type f -name '*.js.map' -delete && \
+                find -L ${{CONDA_DIR}} -type f -name '*.a' -delete && \
+                find -L ${{CONDA_DIR}} -type f -name '*.pyc' -delete && \
+                find -L ${{CONDA_DIR}} -type f -name '*.js.map' -delete && \
                 conda list -p {0}
                 """.format(
                         env_prefix, environment_yml
--- a/repo2docker/buildpacks/conda/install-miniconda.bash
+++ b/repo2docker/buildpacks/conda/install-miniconda.bash
@@ -71,9 +71,9 @@ fi
 conda clean --all -f -y
 
 # Remove static libaries, compiled Python byte code, JS map files
-find ${CONDA_DIR} -follow -type f -name '*.a' -delete
-find ${CONDA_DIR} -follow -type f -name '*.pyc' -delete
-find ${CONDA_DIR} -follow -type f -name '*.js.map' -delete
+find -L ${CONDA_DIR} -type f -name '*.a' -delete
+find -L ${CONDA_DIR} -type f -name '*.pyc' -delete
+find -L ${CONDA_DIR} -type f -name '*.js.map' -delete
 
 # Remove the big installer so we don't increase docker image size too much
 rm ${INSTALLER_PATH}
--- a/repo2docker/buildpacks/conda/install-miniconda.bash
+++ b/repo2docker/buildpacks/conda/install-miniconda.bash
@@ -70,6 +70,11 @@ fi
 # Clean things out!
 conda clean --all -f -y
 
+# Remove static libaries, compiled Python byte code, JS map files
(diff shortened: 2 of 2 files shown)


betatim:
Just built https://github.com/betatim/requirements with `master` and this branch:
```
master_r2d   latest                 ea3cfcabc70d        30 seconds ago      1.72GB
slim_r2d     latest                 bc5f9a5997f4        12 minutes ago      1.45GB
```

Which is surprising. 300MB gone like that? Makes me suspect we deleted something that was needed.

betatim:
I think we should merge this as it already shrinks the image size and has no downsides as far as I can tell.

betatim:
Switched to using `-L`. Nice find!

I think the summaries you found are spot on. The ideas for which files to delete come via [the blog post](https://jcrist.github.io/conda-docker-tips.html) linked in #707. I think it is safe to delete them and/or some shouldn't be there in the first place (like the `.js.map` files). I pinged the nteract team about getting a smaller "production" build of nteract_on_jupyter which is one of the largest contributors in terms of `.js.map` files. They will look into it.

consideRatio:
Wow excellent resource thanks for linking!

What about the mentioned environment variable `PYTHONDONTWRITEBYTECODE=true` associated with removing the .pyc files? Do we have such variable set? ([Reference for PYTHONDONTWRITEBYTECODE](https://docs.python.org/3/using/cmdline.html#id1))

betatim:
I think that we should allow Python to write `.pyc` files when the user is using the pod as that doesn't change the image's size. Before adding it as default to the image, with the intention of preventing installers from creating `.pyc` files I think we'd have to do more digging to understand if there are ways for install commands to ignore this variable. At least I have a vague memory of having seen log output when installing packages that explicitly refers to "byte compiling packages" and I wonder if that would be prevented by this env variable (which in turn would allow us to remove the `find` command to delete them.

TL;DR: We'd have to do more digging so let's punt it.

consideRatio:
I'm okay with all options and i'm currently clueless if it would be good to set the env var or not as it may or may not slow down the initial run a lot or not for example.

LGTM!

consideRatio:
@betatim id like someones help on reviewing/merging Pipfile buildpack, but im afriad everybody is out of time to spend and it has gotten stale...

I put in quite an effort to make it quicker and easier to overview recently, with full understanding of limited time and that it is hard to split up focus on many things by adding another thing to do on the list, id be very appreciative if you could find the time to look it through.

betatim:
Merging as you approved it.

SylvainCorlay:
This breaks some binders that we are working on, because of the deletion of the static libraries (`.a`).

Indeed, we perform some compilation in the postbuild scripts, and these static libraries may be needed at this time. Beyond that, even if we actually do not use the static libraries, if the cmake targets file was generated assuming that they would be present, we can't make use of these cmake targets file anymore once the static library has been deleted.

A solution would be to perform that deletion *after* the postbuild script has run.

cc @vasavanthiru.

betatim:
Can you explain a bit more (or link to an example) of when you create the `.a` files and when you use them?

VasavanThiru:
Thanks, @betatim  If you want a clear example where the binder seems to delete the static libraries, you can follow this link for more information about our issue.

[Xeus-Calc Binder ](https://gke.mybinder.org/v2/gh/QuantStack/xeus-calc/master)

SylvainCorlay:
So indeed, it is generally reasonable to remove the static libraries (.a) since there are typically not used at runtime unlike share objects (.so).

However, there are several usecases of binder that this breaks:

1. cling world, where the C++ interpreter may make use of static libraries *at runtime*. But let's put the cling-crazyness aside for now, and this is not the issue here.
2. binders in which the postBuild scripts performs some compilation, which is a nice way to produce a binder for in-development stage of a project that would not be on conda-forge yet.

Expanding on (2):

This bit us for a project by two @QuantStack interns (@vasavanthiru and @thibaultlacharme), that is a xeus-based "calculator" kernel for Jupyter called [xeus-calc](https://github.com/QuantStack/xeus-calc). By default, the xeus package includes both a shared object and a static library for xeus, and the `xeusTargets.cmake` manifest declares both.

Unfortunately, even if they want to use only the shared object, they include xeusTargets in their cmake and cmake crashes complaining that the files refered-to by the targets are missing.

A solution to that issue would be to only perform the deletion of static libraries *after* the postBuild stage has completed.



When installing the static library is produced, we also declare it in the `xeusTargets.cmake` manifest that is installed with the library, just like the shared object. Any package including

minrk:
I think we should definitely not remove static libs. Removing static libraries is not a safe thing to do, since there are plenty of packages that may do run-time compilation (cython, fenics, cling, etc.) which this can break. If a package includes static libs, they should definitely be left alone.

I'm also a bit skeptical that removal of .pyc files is also recommended, since this adds a runtime cost to rebuild these files on load. But this is optimizing image size vs runtime cost, which is less clear what the best thing to do is in any given situation. In particular, this trades once-per-node startup time (image pulling .pyc files) with a cost paid every startup (recompiling used .pyc files, which is certainly fewer).

In general, I think we should try to avoid making a repo2docker install different from a normal installation, at least except where they are *known* cache/optimization files (.pyc qualifies, .a does not) since it deviates from the "automate best practices" goal.

Would it be okay to revert this PR and try again, perhaps?

betatim:
I think reverting this PR is a good idea.

I'd probably restart with just deleting the JS map files.

For `pyc` I'd be Ok with the runtime overhead though if we are revisiting it I'd investigate how much space this actually saves.

minrk:
Thanks! It would be cool, too, to breakdown the filetypes by the size impact as well, to see which ones are helping the most.

For .a files, it's possible that the 'right' thing to do is to push conda packages to split static libs into a separate package (this is the direction many are going), so that they will just start showing up less in dependencies when they aren't really needed, rather than trying to optimize here.

SylvainCorlay:
> For .a files, it's possible that the 'right' thing to do is to push conda packages to split static libs into a separate package (this is the direction many are going), so that they will just start showing up less in dependencies when they aren't really needed, rather than trying to optimize here.

+1. If conda-forge was to adopt more 'debian style' naming conventions, we could have split packages for  `doc`, `dev` and the like...

minrk:
conda-forge is in the process of splitting at least some static libs: https://github.com/conda-forge/hdf5-feedstock/pull/112

PR #710, opened 2019-06-20
Add support for julia 1.0.4 and 1.1.1

davidanthoff:


davidanthoff:
I don't understand the build failure, and I'm not even sure it is related to the Julia stuff at all?

tmorrell:
It's a new formatting check.  If you install black using `pip install black` and type `black .` in the top level directory it should pass.

davidanthoff:
> All hail our new formatting god

It did insert a comma that was not there originally, that I then removed manually :) The comma was placed after the last element in the array, and hadn't been there in the original code.

betatim:
Nods, for elements spread out over several lines there is a comma after every element. This isn't the case when all elements fit on a single line.

It seems a bit draconian at first but the fact that there are no choices left for us humans to make or config options to tweak means we get consistently formatted code and no bike shedding about what the "correct" formatting is.

davidanthoff:
Ah, so, I shouldn't have made that manual edit! Alright, next try :)

betatim:
Correct. https://github.com/python/black has some more explanation on the choices made and why.

PR #714, opened 2019-06-24
Use the `strict` config option for conda

ocefpaf:
Conda will be adopting `strict` as the default option soon but until then it is recommend to enable it when using third party channel like conda-forge. For a longer explanation see https://youtu.be/9Ft20NizQA4?t=930.

The gist of it is:
- env/install solutions will be faster;
- the stack of packages are consistent "within the ecosystem" and errors are easier to debug;
- [avoid the mixing channel problem](https://conda-forge.org/docs/user/tipsandtricks.html#using-multiple-channels).

Ping @jhamman

### DIFF
--- a/repo2docker/buildpacks/conda/install-miniconda.bash
+++ b/repo2docker/buildpacks/conda/install-miniconda.bash
@@ -30,6 +30,7 @@ export PATH="${CONDA_DIR}/bin:$PATH"
 
 # Allow easy direct installs from conda forge
 conda config --system --add channels conda-forge
+conda config --set channel_priority strict
 
 # Do not attempt to auto update conda or dependencies
 conda config --system --set auto_update_conda false


ocefpaf:
I see that the `broken` label is used in some places :scream_cat: 

I'll take a closer look tomorrow and try to fix the failures and adapt it to a more reliable set of packages.

scottyhq:
On a related note, @yuvipanda, are there reasons for installing `jupyterhub` and `nteract_on_jupyter` via pip instead of conda-forge or conda defaults? https://github.com/jupyter/repo2docker/blob/9c52e5abea3127750e2748eba334df4787257354/repo2docker/buildpacks/conda/environment.yml#L7-L9

I think this is how we're ending up with pip packages on pangeo-stacks base-notebook even after removing pip packages (https://github.com/pangeo-data/pangeo-stacks/blob/master/base-notebook/binder/environment.yml)

binder link:
https://binder.pangeo.io/v2/gh/scottyhq/pangeo-binder-test/pangeo-base?urlpath=lab

output of `conda list | grep pypi`
```
alembic                   1.0.9                    pypi_0    pypi
chardet                   3.0.4                    pypi_0    pypi
idna                      2.8                      pypi_0    pypi
jupyterhub                0.9.4                    pypi_0    pypi
mako                      1.0.9                    pypi_0    pypi
nteract-on-jupyter        2.0.12                   pypi_0    pypi
python-editor             1.0.4                    pypi_0    pypi
python-oauth2             1.1.0                    pypi_0    pypi
requests                  2.21.0                   pypi_0    pypi
sqlalchemy                1.3.3                    pypi_0    pypi
urllib3                   1.24.2                   pypi_0    pypi
```

betatim:
I don't think there is a reason to use JupyterHub from pip. Maybe at the time there wasn't a conda-forge package yet.

I don't think there is a conda package for `nteract_on_jupyter`.

The CI failures are related to not being able to install nbconvert which we pin as part of the base environment. I'd apply the same `strict` setting to the base env freezing and refreeze to see if that fixes it (https://repo2docker.readthedocs.io/en/latest/contributing/tasks.html#update-and-freeze-buildpack-dependencies).

yuvipanda:
Maybe the conda-forge JupyterHub brings chp with it?

My personal preference is to use conda only to get python and use pip for everything else in the base image :) 🔥

ocefpaf:
> I don't think there is a conda package for `nteract_on_jupyter`.

Working on that now...

> I'd apply the same strict setting to the base env freezing and refreeze to see if that fixes it 

Sorry, I'm new here and a bit I'm confused. What do you mean by base env? If I understood `install-miniconda.bash` applying the `strict` config here should be the "first one" in the installation chain, no?

> Maybe the conda-forge JupyterHub brings chp with it? I've no idea...

What is `chp`?

> My personal preference is to use conda only to get python and use pip for everything else in the base image :)  conda is an extremely confusing and opaque world for me. I do recognize that this isn't true for most others in this community...

That is a fine strategy for the base image. Still, looks like user downstream would like conda support for non-python or hard to install python packages and the `strict` option in the canonical config helps a lot.

TL;DR as long as the base image wants to "ship" conda, even if it is not entirely build with it, I recommend the `strict` option to be enabled.

I'll work on the envs with the broken label later today. Conda is not opaque to me :smile:

betatim:
> Sorry, I'm new here and a bit I'm confused. What do you mean by base env?

I should have been clearer :) We create a conda environment with certain things installed, no matter what the repo that is being built does. For example we pre install Jupyter. The packages and versions that get used in this "base env" come from a `environment.yml` which has very specific pins for every package. We generate that file from a looser spec. The process of generating that file is the "freezing" I linked to.

My guess as to why nbconvert won't install is that in the frozen `environment.yml` we refer to a version (and build) that is incompatible with the other selected packages and versions if `strict` is enabled. Probably because `strict` wasn't enabled when we froze the environment in the first place.

-> Add `strict` to the options in the `freeze.py`, refreeze and see if things work now.

ocefpaf:
> I should have been clearer :) We create a conda environment with certain things installed, no matter what the repo that is being built does. For example we pre install Jupyter. The packages and versions that get used in this "base env" come from a `environment.yml` which has very specific pins for every package. We generate that file from a looser spec. The process of generating that file is the "freezing" I linked to.

Thanks. I'll look into that.

> I don't think there is anything to do regarding the broken channel.

As someone who packages these packages, maintain conda-forge, and move packages to broken I do think we should remove those use. It is a flaky solution for conda's solver, there are packages with bad metadata there, and, of course, broken packages too! I'm navigating the requirements now and, if there are any versions that only exists on `broken` we should rebuild them and put them on `main`.

> Before creating a conda-forge package for nteract_on_jupyter I'd chat with the maintainers about who is going to keep it up to date. If there is no one who wants to keep it recent I don't think we should switch to using it.

Conda-forge is a community of packagers and the maintainer submits it (I'm assuimg by maintainer you meant the authors and software maintainers, not the conda package per se). If the package authors also want to help, that is great! But most of the time package authors are not involved with downstream packaging (be that conda, yum RPMs, apt-get deb, chocolatey, etc).

> Is there an explanation of what it does, the problems it solves and what the downsides are of turning it on that isn't a video?

Sure:

- https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html#strict-channel-priority
- https://www.anaconda.com/understanding-and-improving-condas-performance
- https://conda-forge.org/docs/user/introduction.html?#how-can-i-install-packages-from-conda-forge
- https://conda-forge.org/docs/user/tipsandtricks.html#how-to-fix-it

scottyhq:
adding a link to this related pull request on bumping to conda 4.7: https://github.com/jupyter/repo2docker/pull/719

minrk:
@ocefpaf as far as I can tell, the broken channel must be included to support reproducible environments. See #731 for a discussion of why we need it and how we might be able to get away from it when it's not needed. The vast majority of packages in broken aren't really broken, but instead have insufficiently strict pinning of dependencies.

A critical use case for repo2docker is running `conda env export` to freeze the environment and then installing it later. Conda-forge frequently labelling packages as broken regularly makes working, frozen environments not installable. Ideally, we would only be allowing the broken channel when packages are strictly specified, but I'm not sure of a good way to do that. I agree that we don't want to pull `numpy` from broken, but if an env specifies `numpy-1.16.0-py37_blas_openblashb06ca3d_0` which is marked as broken, that install should not fail.

betatim:
It seems for now that we won't be able to enable `strict` out of the box. At least it will take a bit more thinking.

The second topic being discussed here is use of the broken channel. I think this is settled by Min's comment above and other discussions on this repo.

I'd move to close this PR for now and maybe create an issue to summarise the discussion about `strict` and what the next step towards enabling it would be. WDYT?

minrk:
Closing for now, since strict priority cannot work for r2d until issues like https://github.com/conda/conda/pull/8961 are resolved in conda.

meeseeksmachine:

https://discourse.jupyter.org/t/pass-channel-priority-strict-to-conda-with-repo2docker/10551/2

Issue #711, opened 2019-06-20
Configurable docker push timeout


betatim:
via https://discourse.jupyter.org/t/i-get-read-timed-out-errors-on-when-pushing-image-ovh/1394/7?u=betatim

Is there a way we can control/increase the timeout for the push operation at the end of the build? We are getting timeouts for this step on the OVH deployment and a potential way to help here is to increase (or make configurable) the push timeout.

The first step here is to find out how to control the timeout in the docker python client library we use.

mael-le-gal:
I succeed in reproducing the error locally.

When we create a docker `docker.APIClient` there is a `timeout` parameter that we could give to the constructor.

Documentation say :

```
timeout (int): Default timeout for API calls, in seconds.
```

I let you now if changing this parameter fix the issue.

mael-le-gal:
Indeed increasing that timeout fixed the issue.

The line of code is that [one](https://github.com/jupyter/repo2docker/blob/70819410f1d4cc2a31d82411646e141630e305be/repo2docker/app.py#L451)

It needs to be changed to something like that :  
```python
client = docker.APIClient(version="auto", timeout=CHOSEN_TIMEOUT, **kwargs_from_env())
```

Maybe this would be great to make it configurable. In this initial command line (like `--push-timeout xxx`) ?

mael-le-gal:
Or by changing the `kwargs_from_env()` function to take the parameter from environment variables.

What do you think ?

betatim:
Taking it from the environment sounds like a good idea to me. Mostly because we have a huge number of CLI flags already :-/ and my guess is this would be mostly used by deployments like BinderHub and not humans.

I think `kwargs_from_env()` is part of the docker-py library so maybe it already looks at an environment variable for this?

mael-le-gal:
> Taking it from the environment sounds like a good idea to me. Mostly because we have a huge number of CLI flags already :-/ and my guess is this would be mostly used by deployments like BinderHub and not humans.

agree

> I think kwargs_from_env() is part of the docker-py library so maybe it already looks at an environment variable for this?

Indeed, I didn't saw that it was from `docker-py` but sadly the `timeout` parameter is not set in the function.

I'll then write my own function to get this parameter from env variables.

betatim:
Something like this sounds like a good idea or maybe we can contribute to docker-py to get `kwargs_from_env()` changed?
```
def get_timeout():
  return value_from_env

[...]

PR #717, opened 2019-06-25
Reading Docker timeout value from environment variable

mael-le-gal:
Fixing the following [issue](https://github.com/jupyter/repo2docker/issues/711) by adding `DOCKER_REGISTRY_TIMEOUT` environment variable

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -14,6 +14,8 @@ New features
   to Python 3.5 in :pr:`684` by :user:`betatim`.
 - Support for `Pipfile` and `Pipfile.lock` implemented in :pr:`649` by
   :user:`consideratio`.
+- Add an environment variable `REPO2DOCKER_DOCKER_TIMEOUT` to configure the timeout (in seconds)
+  on the docker push timeout in :pr:`717` by :user:`mael-le-gal`
 
 API changes
 -----------
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -19,6 +19,8 @@ import time
 
 import docker
 from urllib.parse import urlparse
+
+from docker.constants import DEFAULT_TIMEOUT_SECONDS
 from docker.utils import kwargs_from_env
 from docker.errors import DockerException
 import escapism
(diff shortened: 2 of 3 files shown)


consideRatio:
Hi @mael-le-gal I mistakenly cancelled your CI build and restarted it again! Sorry!

I noted that one job of all the jobs had failed relating to not having run the autoformatter `black`.

consideRatio:
My understanding is that this PR makes it so that repo2docker configures docker to explicitly use its own default value, unless a environment variable is set, called `DOCKER_REGISTRY_TIMEOUT`.

Considerations:
- I wonder, would docker adjust itself if we simply provided an env var called `DEFAULT_TIMEOUT_SECONDS` ?
- I wonder, if we add a new env that has meaning to repo2docker, should we perhaps prefix it with `REPO2DOCKER` or `R2D`? I've come to appreciate such things a lot as it can be quite a hurdle to figure out why and who considers what env vars if you have something nesting something that potentially even nests something else etc.

consideRatio:
I've learned that no, Docker does not care about an environment variable named `DOCKER_TIMEOUT_SECONDS`.

__Hardcoded constant...__
https://github.com/docker/docker-py/blob/20b5491363a9124dbf53b00fb6562f8f257a6228/docker/constants.py#L1-L6

__... used as fallback if no timeout argument was provided.__
https://github.com/docker/docker-py/blob/20b5491363a9124dbf53b00fb6562f8f257a6228/docker/client.py#L82

mael-le-gal:
> I noted that one job of all the jobs had failed relating to not having run the autoformatter black.

Indeed, it should be fixed now

> I wonder, would docker adjust itself if we simply provided an env var called DEFAULT_TIMEOUT_SECONDS ?

No there is only a hardcoded default timeout but it is never override by environment variable

> Regarding my second consideration, I'm leaning towards naming the variable REPO2DOCKER_DOCKER_TIMEOUT or something making it obviously ours.

Fixed, I changed to `REPO2DOCKER_DOCKER_TIMEOUT`

consideRatio:
Should we set the timeout explicitly, by using the docker default value imported from the docker package at? It will entwine us with the docker package slightly. If we need that, because we need to know the timeout someplace for some reason, then I think we should. If we don't need to know the implicitly set timeout, then I think it could be beneficial to avoid explicitly passing a `timeout` argument when we don't detect a `REPO2DOCKER_DOCKER_TIMEOUT` variable.

betatim:
Two general comments:

* we use https://github.com/ipython/traitlets for the configuration of repo2docker, lets use it also for this timeout. I need to dig a bit to find an example/docs of how traitlets and env variables work together. We should use that machinery as much as we can
* not sure how I feel about fetching the default value and using it. I'd have used `None` as a value for the "default" value in r2d which signals use what ever the default in docker-py is. However [docker-py sets the default](https://github.com/docker/docker-py/blob/4d8e5eb2ada3dfbffd5721271c1c7c6bec476dd4/docker/api/client.py#L102-L105) to a particular value (not `None`) so I think in this case getting the value and passing it in is the thing to do.

betatim:
I'd suggest that we close this PR. It hasn't been a problem in a long time and we can open a new PR using traitlets to implement this if we need to.

(We are trying to reduce the number of outstanding/stale PRs)

Issue #487, opened 2018-12-07
Make it possible to configure the base image


choldgraf:
I feel like we've discussed this a few times but I can't find a specific issue, so:

What if we exposed the ability for CLI users of repo2docker to specify a base image to use instead of the ubuntu base image. I could see this being useful for:

1. People who want to use a Rocker- or Neurodocker-style image w/ a really nasty stack, and then build on top of it
2. People who want to build a more usecase-specific BinderHub deployment (e.g. for pangeo, everybody starts w/ an image that has many basic earth analytics packages and then they can specify on top of this.

@craig-willis if you have thoughts on the above that'd be helpful!

I think the trick here is that we'd need to lay down some clear rules for what would need to be in the image. It'd have to run the same Ubuntu version, and would need to have jupyterhub / server stuff ready to go. Perhaps it could be treated as an "advanced use case, you should know what you're doing" kinda thing.

choldgraf:
interesting idea - one thought: do we currently have a way of people to define a buildpack without merging it directly into repo2docker? I wonder if we could let people define new buildpacks and point to them as a part of a repo2docker build. Neurodocker, for example, is basically doing the same thing as repo2docker (it's a command line tool where you say "I want this packages installed" and it builds a Dockerfile with the relevant lines in there)

betatim:
We could explore how to use [entrypoints](https://entrypoints.readthedocs.io/en/latest/#) to allow for plugins/extensions. I think that would be interesting both for build packs and content providers!

craig-willis:
I came across a couple of discussions about supporting different distros in https://github.com/jupyter/repo2docker/issues/166 and  pinning the distro/base image in https://github.com/jupyter/repo2docker/issues/170.

I think I need to clarify what we're talking about when we say "base image", because there are at least two cases. 
1. Base image is a child of Ubuntu but with some domain/use-case specific packages
2. Root image is another distro (e.g., Debian, CentOS) 

One possible addition/expansion is supporting other versions (i.e., a use case that must have `trusty`).

> It'd have to run the same Ubuntu version, and would need to have jupyterhub / server stuff ready to go.

Does repo2docker require the JupyterHub/server packages, or is this a requirement for BinderHub specifically?  Must this be a hard requirement for every Dockerfile produced by `repo2docker`?  Or rather, must every Dockerfile produced by repo2docker be potentially runnable on BinderHub (I think "yes").

> If you really want to use arbitrary bases you can already do this by using a Dockerfile

[...]

PR #718, opened 2019-06-25
[MRG] Install R packages before copying repo contents

betatim:
This is the R part of #716. Needs rebasing after #716 gets merged.

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -90,7 +90,7 @@ ENV PATH {{ ':'.join(path) }}:${PATH}
 
 {% if build_script_files -%}
 # If scripts required during build are present, copy them
-{% for src, dst in build_script_files.items() %}
+{% for src, dst in build_script_files|dictsort %}
 COPY {{ src }} {{ dst }}
 {% endfor -%}
 {% endif -%}
@@ -127,7 +127,7 @@ ENV {{item[0]}} {{item[1]}}
 # example installing APT packages.
 {% if preassemble_script_files -%}
 # If scripts required during build are present, copy them
-{% for src, dst in preassemble_script_files.items() %}
+{% for src, dst in preassemble_script_files|dictsort %}
 COPY src/{{ src }} ${REPO_DIR}/{{ dst }}
 {% endfor -%}
 {% endif -%}
@@ -151,7 +151,7 @@ RUN chown -R ${NB_USER}:${NB_USER} ${REPO_DIR}
 # Container image Labels!
 # Put these at the end, since we don't want to rebuild everything
 # when these change! Did I mention I hate Dockerfile cache semantics?
-{% for k, v in labels.items() %}
+{% for k, v in labels|dictsort %}
 LABEL {{k}}="{{v}}"
 {%- endfor %}
 
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -125,7 +125,14 @@ ENV {{item[0]}} {{item[1]}}
 # of the repository but don't access any files in the repository. By executing
 # them before copying the repository itself we can cache these steps. For
 # example installing APT packages.
-{% for sd in pre_assemble_script_directives -%}
+{% if preassemble_script_files -%}
+# If scripts required during build are present, copy them
+{% for src, dst in preassemble_script_files.items() %}
+COPY src/{{ src }} ${REPO_DIR}/{{ dst }}
+{% endfor -%}
+{% endif -%}
+
+{% for sd in preassemble_script_directives -%}
 {{ sd }}
 {% endfor %}
 
@@ -379,6 +386,19 @@ class BuildPack:
 
         return []
 
+    def get_preassemble_script_files(self):
+        """
+        Dict of files to be copied to the container image for use in preassembly.
+
+        This is copied before the `build_scripts`, `preassemble_scripts` and
+        `assemble_scripts` are run, so can be executed from either of them.
+
+        It's a dictionary where the key is the source file path in the
+        repository and the value is the destination file path inside the
+        repository in the container.
+        """
+        return {}
+
     def get_preassemble_scripts(self):
         """
         Ordered list of shell snippets to build an image for this repository.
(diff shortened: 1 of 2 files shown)


betatim:
A good example and alternative to moving `install.R` early is linked in https://github.com/jupyter/repo2docker/issues/487#issuecomment-479825150. It is a buildpack that uses a different base image.

One thing I don't understand is how installing a specific R package (selected by MRAN date) would work out faster than with the current approach if we start from (say) `rocker/binder:3.6.0` as base image. Chances are tidyverse from 2019-04-11 isn't the one already installed so we have to install it again.

The other thing that isn't clear to me: if I run `install.package('random-package')` in the `rocker/binder:3.6.0` image before and after `random-package` make a release, how is the version that gets installed selected?

---

I see three options:

1. move `install.R` forwards (propose here)

2. adopt the buildpack-with-custom-base from Yuvi's example and make a `RockerBuildPack`, recommend people use that over the R buildpack

3. make installing popular R packages from a specific MRAN date part of the "basics" you get when you select the R buildpack in repo2docker. Right now we already select the Shiny version for users.

yuvipanda:
Of the 4, I like 4 as an experiment :)

I think for (2), my intent wasn't to make it be part of repo2docker, but to upstream it to https://github.com/rocker-org/binder.

betatim:
And because there is never anything new under the sun: some prior art #410 that we should look at and ponder.

betatim:
Option 4 is now implemented. I decided to use a empty file to store the state of "pre-assembled worked" as sometimes re-running an `install.packages()` did take just as long as the first time :-/

Test it with something like https://github.com/betatim/r-caching (which tries to read the `README.md` in `install.R`).

yuvipanda:
Thanks for implementing that, @betatim. This is definitely a lot cleaner in terms of useful error messages and what not.

I'm very conflicted on this particular PR, on two issues:

1. install.R performance is absolutely, almost unacceptably terrible right now, and something like this will definitely make it a lot better
2. This might still have weird side effects. I don't actually know what R does if a package install fails halfway through (because it dependended on local filesystem?) and then you re-do it. This kinda consistency issues aren't something we have to deal with right now, since we just evict cache instead.

So it's a tradeoff between performance and a cleaner conceptual mental model. Based on the work you had done earlier looking at repos that have an install.R, and conversations in #716, I feel this is an ok tradeoff if we make sure to document it heavily - including the fact that this is an explicit tradeoff we made for this case. 

We should also add a section in https://repo2docker.readthedocs.io/en/latest/design.html about performance tradoffs and how to evaluate them. (doesn't need to block this PR though)

Again, thank you for diligently pushing this through :D <3

yuvipanda:
I restarted the travis job, which I think was a flake.

yuvipanda:
I'm happy to merge this after:

- [ ] Add doc blurb to install.R docs
- [ ] Ping @cboettig and @karthik to see what they think
- [ ] Open an issue to figure out how we can test 'cache re-use' when building images

betatim:
The "external" job in travis will keep failing till https://github.com/jupyter/repo2docker/pull/732 or something like it is merged.

It'll be good to hear where @cboettig and @karthik think this approach could fail, how likely it is that someone will hit that problem and how serious the outcome would be.

minrk:
This is looking so familiar that I thought we already had this, and now I realize that I never finished #410 . Maybe it's worth pulling that in here, since it's the conda and Python implementation of this same thing?

repo2docker/buildpacks/base.py
```diff
@@ -125,7 +125,14 @@
 # of the repository but don't access any files in the repository. By executing
 # them before copying the repository itself we can cache these steps. For
 # example installing APT packages.
-{% for sd in pre_assemble_script_directives -%}
+{% if preassemble_script_files -%}
+# If scripts required during build are present, copy them
+{% for src, dst in preassemble_script_files.items() %}
```
minrk: use `sorted(preassemble_script_files.items())` to ensure that files are added in a stable order. Dictionary randomization could potentially result in cache busting here for the same collection of files.

betatim:
#410 is indeed the parent of all these ideas! I didn't pull in the pip and conda changes here because I thought we should focus on the R part. In particular the fact that we can't tell if `install.R` will succeed or not before executing it. This seems like the trickiest part here.

Once we merge this we can reboot #410 and it should be a "no brainer" to merge (at least I can't think why you'd merge this but oppose #410 :) ).

repo2docker/buildpacks/base.py
betatim: Good catch! I added two more `sorted()` statements to the template for other loops over `items()`.

minrk:
My only concern with running the install twice is that I'm not 100% sure that a failed execution followed by a successful one is idempotent with a single successful execution. E.g. if it unconditionally does something that will fail on the second run, but this bit is before the step when install.R fails.

I could write a totally fake install.R that does this, but I'm not sure it's a realistic concern. It's just that there would be no real recourse for folks where install really does require the whole repo to skip the first failing install. I hate adding more files, but is there a way to opt into or out of the first try? E.g. a magic comment at the top of install.R?

betatim:
From a totally unscientific study of maybe 30 or 40 `install.R` files I could find via GitHub search I think the chances are low we will run into the "`install.R` isn't idempotent" or the weaker "running `install.R` without repo and then with repo context will not do the right thing" (this is what we need).

So I would chance it and when someone reports a problem (or we spot one on https://grafana.mybinder.org/d/fZWsQmnmz/pod-activity?refresh=1m&panelId=8&fullscreen&orgId=1&from=now-12h&to=now&var-cluster=prometheus) we should introduce the "magic comment to opt out".

It felt like most `install.R`s (that weren't clearly "not made for Binder") call `install.packages()` and the "install from GitHub" version of it. Some define a list of package names and then iterate over it. I didn't see one that read the package list in from a file (in Python `setup.py` often reads the `requirements.txt`). I don't know if there is an equivalent of `pip install -e.`, if there is I didn't see it used.

Tl;DR: I'd wait to see if we get build failures on mybinder.org where people would need to opt out, then implement a magic comment.

betatim:
ps. @minrk do you want to reboot #410 or should/could someone else make a new PR?

minrk:
Makes sense. I'll give resurrecting #410 a go. Thanks!

minrk:
#743 is the re-issue of #410

meeseeksmachine:

https://discourse.jupyter.org/t/tip-embed-custom-github-content-in-a-binder-link-with-nbgitpuller/922/50

PR #722, opened 2019-06-27
[WIP] Bump xeus-cling tag used in our tests

betatim:
Investigating if the xeus-clinng problem in #719 is related to the conda update.


betatim:
It has to be something to do with the conda update as building xeus-cling worked  with the old version.

PR #723, opened 2019-06-28
[MRG] Handle root user case more gracefully. Fixes #696

### LINKED ISSUE #696: Setting r2d.user_id = 0 doesn't cause exception
### Bug description

When using repo2docker as a library, setting `r2d.user_id = 0` prints a message and exits.

#### Expected behaviour

It should raise an exception I can catch in my code.

#### Actual behaviour

It exits with a confusing (to users of my program) message about commandline parameters.

Ref #126

Xarthisius:
Trying to atone for bugs introduced by #679. With this PR exiting happens only when r2d is invoked from the commandline. In case user tries to use r2d as library, `ValueError` is raised instead.

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -303,6 +303,16 @@ def make_r2d(argv=None):
         r2d.user_id = args.user_id
     if args.user_name:
         r2d.user_name = args.user_name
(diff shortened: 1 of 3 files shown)


Xarthisius:
Third time is the charm I hope!

PR #727, opened 2019-07-02
include full docker progress events in push progress events

minrk:
for backward-compatibility, add a `layers` key with the full data,
preserving the `progress` field, which has only partial data, depending on state of each layer.

The most useful field, I think, is `progress`, which shows the docker cli-rendered progress bar.

Strangely, when I sampled the log data we get from ovh, current and total in progressDetail (which we already publish under 'progress') were on totally different scales,
so rendering our own progress bars with `$current/$total` doesn't seem trustworthy. Not sure what's up with that.

Sample log data:

```json
  "layers": {
    "25b373f5f7f1": {
      "status": "Waiting",
      "progressDetail": {},
      "id": "25b373f5f7f1"
    },
    "e9f2b5eca21e": {
      "status": "Pushing",
      "progressDetail": {
        "current": 747589632,
        "total": 758965327
      },
      "progress": "[=================================================> ]  747.6MB/759MB",
      "id": "e9f2b5eca21e"
    },
    "7753d7e0913b": {
      "status": "Pushed",
      "progressDetail": {},
      "id": "7753d7e0913b"
    },
    "ed03b06eb165": {
      "status": "Pushed",
      "progressDetail": {},
      "id": "ed03b06eb165"
    },
    "01cb88e6c1af": {
      "status": "Pushed",
      "progressDetail": {},
      "id": "01cb88e6c1af"
    },
    "adda4de99b3d": {
      "status": "Mounted from library/buildpack-deps",
      "progressDetail": {},
      "id": "adda4de99b3d"
    },
    "3a034154b7b6": {
      "status": "Mounted from library/buildpack-deps",
      "progressDetail": {},
      "id": "3a034154b7b6"
    }
```

Related to https://github.com/jupyterhub/binderhub/issues/852 . Passing through the `progress` fields might be the easiest implementation.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -457,6 +457,8 @@ class Repo2Docker(Application):
         layers = {}
         last_emit_time = time.time()
         for chunk in client.push(self.output_image_spec, stream=True):
+            # each chunk can be one or more lines of json events
+            # split lines here in case multiple are delivered at once
             for line in chunk.splitlines():
                 line = line.decode("utf-8", errors="replace")
                 try:
@@ -478,11 +480,10 @@ class Repo2Docker(Application):
                 layers[progress["id"]] = progress
                 if time.time() - last_emit_time > 1.5:
                     self.log.info(
-                        "Pushing image\n", extra=dict(
-                            progress=progress_layers,
-                            layers=layers,
-                            phase="pushing",
-                        )
+                        "Pushing image\n",
+                        extra=dict(
+                            progress=progress_layers, layers=layers, phase="pushing"
+                        ),
                     )
                     last_emit_time = time.time()
         self.log.info(
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -453,24 +453,38 @@ class Repo2Docker(Application):
         client = docker.APIClient(version="auto", **kwargs_from_env())
         # Build a progress setup for each layer, and only emit per-layer
         # info every 1.5s
+        progress_layers = {}
         layers = {}
         last_emit_time = time.time()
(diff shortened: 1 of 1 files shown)


repo2docker/app.py
```diff
@@ -453,24 +453,39 @@ def push_image(self):
...
-                layers[progress["id"]] = progress["progressDetail"]
-            else:
-                layers[progress["id"]] = progress["status"]
-            if time.time() - last_emit_time > 1.5:
-                self.log.info(
-                    "Pushing image\n", extra=dict(progress=layers, phase="pushing")
-                )
-                last_emit_time = time.time()
+        for chunk in client.push(self.output_image_spec, stream=True):
+            for line in chunk.splitlines():
```
minrk: You might be wondering why this extra iteration was added. When I tested this locally, each chunk could be *multiple* JSON lines, split by `\r\n`. So I added splitting lines to the iteration. I'm not sure why this hasn't come up on mybinder.org. I'm using `docker 3.7.1`

betatim: Should we add this comment as a comment to the code? I'd vote yes

betatim:
This looks nice.

To display this in BinderHub we will need to add a bit of logic to the JS that displays this to include the right magic/invisible characters to rewind the cursor position before writing the progress info again right?

minrk:
> we will need to add a bit of logic to the JS that displays this to include the right magic/invisible characters

Yeah, if we want the progress bars to clear nicely, we'll need to do that. We can also start with a more crude version that prints a static view over and over again (updates are limited to 1.5s, so this wouldn't b a flood) and refine it to update in-place later.

PR #730, opened 2019-07-03
[MRG] Remove print statement and unused import

betatim:
Remove a stray print statement and an unused import in the zenodo content provider.


minrk:
ignoring test failure, which is due to external factors

Issue #698, opened 2019-05-30
In a Python 2 repository `python` gives you Python 3


betatim:
### Bug description
In a repo with Python 2 selected via a `runtime.txt` the command `python` launches Python 3.

#### Expected behaviour

`python` should launch version 2.

### How to reproduce
* [Launch binder](https://mybinder.org/v2/gh/mbcxqcw2/EEModel/v1.03)
* open a console
* run `python`

I think it is because of the order in which environments appear on the search path:

```
$ echo $PATH
/home/jovyan/.local/bin:/home/jovyan/.local/bin:/srv/conda/bin:/srv/conda/envs/kernel/bin:/srv/npm/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

Not sure if this is a bug or merely "surprising behaviour" that needs documenting. Maybe we could write down somewhere why `/srv/conda/bin` does come before `/srv/conda/envs/kernel/bin` to help people understand why it is this way.

betatim:
related/original issue https://github.com/jupyterhub/binderhub/issues/871

nuest:
Activating the `kernel` environment switches to the correct kernel as suggested in #871

![image](https://user-images.githubusercontent.com/1325054/60945796-d6b9e780-a2ec-11e9-8d16-e57bd921e6c4.png)

So :+1: on activating that in the terminal by default, see my attempt below.

--------

**Digging into the issue**

AFAICS the conda environment should be activated via file `conda-activate.sh`: https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/conda/activate-conda.sh#L17

But looking at the debug output of repo2docker locally, that file does not seem to be called. Should it be picked up by `conda` automatically?

Could the problem be that `conda activate` does not work?

```
jovyan@jupyter-nuest-2dpluc-5fmozambique-2difrz8msn:~$ conda activate /srv/conda/envs/notebook

CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.
To initialize your shell, run

    $ conda init <SHELL_NAME>

Currently supported shells are:
  - bash
  - fish
  - tcsh
  - xonsh
  - zsh
  - powershell

See 'conda init --help' for more information and options.

IMPORTANT: You may need to close and restart your shell after running 'conda init'.
```

When I run `conda init bash` and start a new terminal, I can activate the different envs:

```
[...]

PR #737, opened 2019-07-10
always activate default conda environment

nuest:
Suggestion for fixing #698 - feedback by conda expert needed :-)

### DIFF
--- a/docs/source/howto/export_environment.rst
+++ b/docs/source/howto/export_environment.rst
@@ -41,7 +41,7 @@ you need and then perform a strict export of the environment.
 #. open a terminal by clicking "New -> Terminal" next to the "Upload" button on the
    right hand side of the webpage
 #. install the packages your project requires with ``conda install <yourpackages>``
-#. use ``conda env export -n root`` to print the environment
+#. use ``conda env export`` to print the environment
 #. copy and paste the environment you just printed into a ``environment.yml`` in
    your projects repository
 #. close your browser tabs and exit the repo2docker session by pressing Ctrl-C.
--- a/repo2docker/buildpacks/conda/activate-conda.sh
+++ b/repo2docker/buildpacks/conda/activate-conda.sh
@@ -1,3 +1,4 @@
+#!/bin/bash -x
 # enable conda and activate the notebook environment
 CONDA_PROFILE="${CONDA_DIR}/etc/profile.d/conda.sh"
 test -f $CONDA_PROFILE && . $CONDA_PROFILE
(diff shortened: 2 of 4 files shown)


nuest:
Note the tests also changed. Is is a problem that Python 3 is not listed as a kernel anymore? :thinking:

meeseeksmachine:

https://discourse.jupyter.org/t/creating-environment-yml-for-python-2-via-binder/1609/7

minrk:
> Is is a problem that Python 3 is not listed as a kernel anymore? 🤔

Possibly. We'll need to be careful with this one if the kernel env is unconditionally first on PATH. Make sure this case is tested: the `notebook` package is installed in both envs and the notebook env is the one that's used. I'm pretty sure this won't be the case because we launch `jupyter notebook ...` relying on PATH to resolve correctly. If we switched the launch command to $NB_PYTHON_PREFIX/bin/jupyter-notebook, it would probably be correct, but wouldn't be as portable to e.g. Dockerfile builds that can use different paths.

The alternative is to try to fix more directly, and make sure that env activation is different for subprocesses of the notebook vs the notebook process. This can be achieved by performing some custom startup in the notebook process and/or entrypoint that's not reflected by terminal subprocesses

nuest:
@minrk Thanks for your comment, though I have to admit I am bit out of my depth here, if the fix is not as simple as it seemed to me.

_Which approach do you think is worth trying out: switching the launch command, or activate envs?_

With the changes in this PR, I get the following:

```
daniel@687f78fd20b2:~$ conda env list
# conda environments:
#
base                     /srv/conda
notebook              *  /srv/conda/envs/notebook

daniel@687f78fd20b2:~$ /srv/conda/bin/python --version
Python 3.7.3
daniel@687f78fd20b2:~$ /srv/conda/envs/notebook/bin/python --version
Python 2.7.15
daniel@687f78fd20b2:~$ echo $KERNEL_PYTHON_PREFIX
/srv/conda/envs/notebook
daniel@687f78fd20b2:~$ echo $NB_PYTHON_PREFIX
/srv/conda/envs/notebook
daniel@687f78fd20b2:~$ echo $PATH
/srv/conda/envs/notebook/bin:/home/daniel/pcraster/bin:/srv/conda/envs/notebook/bin:/srv/conda/condabin:/home/daniel/.local/bin:/home/daniel/.local/bin:/srv/conda/envs/notebook/bin:/srv/conda/bin:/srv/npm/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
daniel@687f78fd20b2:~$ ls /srv/conda/envs/
notebook
```

- The base environment does _not_ have the `notebook` package.
- `$KERNEL_PYTHON_PREFIX` is the same as `$NB_PYTHON_PREFIX`, namely `/srv/conda/envs/notebook`

Re. "the notebook package is installed in both envs and the notebook env is the one that's used.": There are no "both" envs. Is that an issue, i.e. should the Notebook always run Python 3, and the kernel be configurable by the user?

Should it be possible to have multiple Python versions and environments?

--------

**Related question:** For clarity, could the `CMD ["jupyter", "notebook", "--ip", "0.0.0.0"]` in the template in `base.py` be removed, because the `if not self.run_cmd` - code `app.py` will either override the command or use the custom one provided by the user? Or is the one in `base.py` the one for running the image on JupyterHub, and the one created in `app.py` the one for local use? :confused:

consideRatio:
Currently, you can get python2 by writing `python2`, or open a python2 kernel.

image

Since this PR didn't resolve in the last three years, and python2 is even more outdated now, I think this behavior suffice and its more pressing to ensure sustainable maintenance of repo2docker as a project by not making it more complicated.

I'll go for a close for now, but feel free to open it again!

Issue #736, opened 2019-07-09
Binderhub unable to access private repo with git-credential-env 


eexwhyzee:
### Bug description

When running binderhub with repo2docker, binderhub is unable to clone from a private repo.

#### Expected behaviour

Binderhub should be able to clone from a private repo and start building the image.

#### Actual behaviour

<img width="729" alt="Screen Shot 2019-07-09 at 10 02 41 AM" src="https://user-images.githubusercontent.com/13672394/60908729-6c903c80-a231-11e9-99f1-6269b7b0c6b0.png">

### How to reproduce

1. Deploy binderhub using 0.2.0-c8ab0a3 with repo2docker a251ad26 w/ a github access token
2. Try to create a new binder using a link to a private github repo
3. Build will fail with error above

### Your personal set up

 - OS: macOS 10.13.6
 - Docker version: Docker version 18.09.2, build 6247962
 - repo2docker version: image tag a251ad26
 - binderhub version: 0.2.0-c8ab0a3

Github section of my `secrets.yaml`:
```
config:
  GitHubRepoProvider:
    access_token: ****
```
Access token has the proper permissions:
<img width="730" alt="Screen Shot 2019-07-08 at 3 23 57 PM" src="https://user-images.githubusercontent.com/13672394/60909157-66e72680-a232-11e9-804c-0fcbfe782329.png">

### Workaround
I was able to narrow down the cause of the issue to a git install change in the `Dockerfile`: https://github.com/jupyter/repo2docker/commit/ee93bcf1a6155a71c8978b78b8bf03282dbdf088#diff-3254677a7917c6c01f55212f86c57fbf

I reverted this change to the `Dockerfile` (used `python:3.7` instead of `alpine`) and built the image locally to use for my binderhub deployment and was able to successfully clone and build from a private github repo.

betatim:
Thanks for finding this bug and the excellent bug report!

Do you know what exactly the new `apt-get install` command/setup does in the commit you linked to? My guess would be that we now install an older version of git or an otherwise broken version? Unfortunately I don't know enough about the Ubuntu/Debian package management system to have an idea off the top of my head. If you have time and/or knowledge to investigate or suggest a fix that would be great and if not that is (of course!) also fine.

eexwhyzee:
[...]

PR #738, opened 2019-07-10
[MRG] Adding bash to Dockerfile to fix git-credential-env

eexwhyzee:
Re: #736 

Since `alpine` images  don't include bash by default, this breaks the usage private repos with `git-credential-env` which requires [bash](https://github.com/jupyter/repo2docker/blob/master/docker/git-credential-env#L1).....small PR to just add bash to the `Dockerfile` (image size increased ~5MB)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,8 @@ RUN mkdir /tmp/wheelhouse \
 
 FROM alpine:${ALPINE_VERSION}
 
-# install python, git
-RUN apk add --no-cache git git-lfs python3
+# install python, git, bash
+RUN apk add --no-cache git git-lfs python3 bash
 
 # install repo2docker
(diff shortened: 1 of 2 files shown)


manics:
Is bash necessary, or would changing https://github.com/jupyter/repo2docker/blob/master/docker/git-credential-env#L1 to `/bin/sh` also work?

eexwhyzee:
@manics that was my initial thought at first too, but because of https://github.com/jupyter/repo2docker/pull/389 I wanted to keep git-credential-env consistent by using bash

betatim:
LGTM! Thanks again for the excellent bug report, tracking it down and fixing it. The open-source trifecta 🏆!

eexwhyzee:
@betatim no problem! glad i could contribute

PR #739, opened 2019-07-10
[MRG] Dataverse content provider

Xarthisius:
This PR add support for Dataverse datasets as content provider for repo2docker, e.g.:

```
repo2docker doi:10.7910/DVN/6ZXAGT
repo2docker https://dataverse.harvard.edu/file.xhtml?persistentId=doi:10.7910/DVN/6ZXAGT/3YRRYJ
```

It's very much a WIP, but I'd like to solicit feedback at this point if that's a good way to go. Related to https://github.com/IQSS/dataverse/issues/4714

Since it's very similar to Zenodo provider, I abstracted parts of it into base `DOIHandler` class. Additional refactoring is probably still possible, as I duplicated a lot of code.

TODO:
- [x] add tests
- [x] refactor `fetch` to remove code duplicated with Zenodo content provider.


betatim:
Do we want to allow `repo2docker doi:10.7910/DVN/6ZXAGT`? It looks like a URL but isn't. I don't think we can make it `doi://` can we/should we? Do we even need a prefix or can we keep it prefix-less? The only time we end up with a problem is when there is a directory with the same name as the DOI and the user didn't mean to use it. I'd be Ok with saying that is a rare edge case.

Is it "standard" to do something like `doi:...`? I have no idea :-/

Xarthisius:
> Do we want to allow `repo2docker doi:10.7910/DVN/6ZXAGT`? It looks like a URL but isn't. I don't think we can make it `doi://` can we/should we? Do we even need a prefix or can we keep it prefix-less? The only time we end up with a problem is when there is a directory with the same name as the DOI and the user didn't mean to use it. I'd be Ok with saying that is a rare edge case.
> 
> Is it "standard" to do something like `doi:...`? I have no idea :-/

I'm certainly not an expert in this field, but there are other types of persistent digital objects identifiers (DOI is kinda a loaded term...), one example is handle.net (e.g. hdl:1902.1/12379). They have similar structure as you can see some_number_prefix/some_other_number, but they resolve the same way. Check out:

```
curl -IL https://doi.org/hdl:1902.1/12379
```

With Zenodo you just got lucky ;)

betatim:
I am now thinking that I misunderstood what you are proposing :) I thought you were suggesting that we should make `repo2docker doi:1902.1/12379` the required way of specifying a DOI and remove `repo2docker 1902.1/12379` as a legal option. Now I think you are not suggesting that and instead pointing out that DOIs come in many flavours and that we can treat them all at the same time. If that isn't right then please help me ;)

```
curl -IL https://doi.org/hdl:1902.1/12379
curl -IL https://doi.org/doi:1902.1/12379
curl -IL https://doi.org/1902.1/12379
curl -IL https://doi.org/doi:10.5281/zenodo.3208700
curl -IL https://doi.org/10.5281/zenodo.3208700
```
all seem to be legal and "do the right thing". Which is pretty cool actually :) Now I'm thinking that we should allow both forms and encourage people to use one with "prefix" when there could be ambiguity (is it true that no prefix means "use the default" and the default is "doi:"?).

i will try and take a look at the code tomorrow.

pdurbin:
> DOIs come in many flavours

Yes. The craziest DOI I'm aware of is [https://doi.org/10.1002/(sici)1099-1409(199908/10)3:6/7<672::aid-jpp192>3.0.co;2-8](https://doi.org/10.1002/(sici)1099-1409(199908/10)3:6/7<672::aid-jpp192>3.0.co;2-8)

That said, in practice, I don't think you'll see DOIs in the wild like this very often, thankfully. 😄 

@betatim mostly I dropped by to say thanks for taking a look at this pull request!! 🎉 🎉 🎉  (And thank you to @Xarthisius for making it, of course!! 🎉 🎉 🎉 ) I have some thoughts on URLs to pass to mybinder.org (what's possible today from "external tools" in Dataverse) but I decided to throw it into a comment in our "Binderverse" issue for now: https://github.com/IQSS/dataverse/issues/4714#issuecomment-510594346 . I'm wondering where to have the conversation about Binder URLs. I assume the BinderHub repo but wherever is fine with me. Please let me know. 😄 

@Xarthisius is aware that Dataverse supports an alternative to DOI called Handle. Collectively, people seem to refer to DOIs, Handles, and other identifiers as Persistent IDs (PIDs). There's even a conference called https://pidapalooza.org some colleague of mine attended earlier in the year and a new forum at https://www.pidforum.org where we can ask questions about PIDs.

http://hdl.handle.net/11529/10016 is an example of what a Handle might look like. When you click it, it should resolve to a landing page on an installation of Dataverse that looks like the screenshot below.

Screen Shot 2019-07-11 at 6 16 09 PM

I don't have the numbers on the 46 installations of Dataverse but I'd say most of them use DOIs rather than Handles.

I can't emphasize enough how thrilled I am about this pull request. A few weeks ago a gave a demo of launching a Jupyter Notebook from a dataset in Dataverse using Whole Tale and while I did touch on Binder in my talk, I'd love to do a similar demo with Binder in the future. You'll find a Binder screenshot and a transcript in the write up of my talk: https://scholar.harvard.edu/pdurbin/blog/2019/jupyter-notebooks-and-crazy-ideas-for-dataverse

betatim:
After a quick browse on the train to EuroPython: LGTM 🎉!

A few questions for my education, maybe we can turn some into comments in the code so people from the future find them.

* the `dataverse.json` is needed to detect if the link we get from resolving a DOI points to a dataverse instance right? We should add instructions on how to get and update this file. I assume it'll need updating every few months? Is there a way to make a request to a "potential dataverse" host to find out if it is indeed one? I checked the `Server` response header but thtat doesn't contain anything. Maybe there is a well known endpoint like `/versions` or `/status` or some such that we could query instead of having a list that we will need to update?
* when fetching files we build a list of file IDs we want to fetch and then receive a zip file containing all of them? I left a comment in `fetch` where maybe we can add a comment. I read the code a few times to convince myself that it did do that (I think), so a comment would help

For the code formatting pedant in me: can we stick to `snake_case` as much as possible please?

I look forward to tests and refactoring. (Will the `is_doi()` functions keep working with `doi:1234.2/3425`?)

pdurbin:
> * I checked the `Server` response header but thtat doesn't contain anything. Maybe there is a well known endpoint like `/versions` or `/status` or some such that we could query instead of having a list that we will need to update?

Yes! Once you know the hostname of the potential Dataverse installation (due to the DOI resolving to it), you can check if the host indeed is running Dataverse by going to /api/info/version like this:

https://demo.dataverse.org/api/info/version

The expected response is something like this:

```
{
  "status": "OK",
  "data": {
    "version": "4.15.1",
    "build": "1377-701b56b"
  }
}
```

Obviously, this **could** be some other software but I suspect it's enough of a fingerprint to be pretty sure it's Dataverse. 😄 

> * when fetching files we build a list of file IDs we want to fetch and then receive a zip file containing all of them?

One of the things to look out for here is that Dataverse installations can limit the size of the zip file that is constructed on the fly based this (potentially long) list of file IDs. The default is 100 MB as mentioned at http://guides.dataverse.org/en/4.15.1/installation/config.html#zipdownloadlimit . If you can't get all the files at once, you could switch to downloading them individually.

nuest:
Needs an update if #788 is merged, overlapping refactoring of an abstract DOI-provider class.

Xarthisius:
> Needs an update if #788 is merged, overlapping refactoring of an abstract DOI-provider class.

Thanks for the heads up @nuest ! I'll rebase when it happens. I'd really love to finish this off, but I'm constantly swamped with other work :(

pdurbin:
@Xarthisius thanks for resolving those merge conflicts!! 🎉 🎉 🎉  One thing I noticed over at pull request #788 is that @nuest had "WIP" in the title of his pull request initially but @betatim encourage him to switch it to "MRG" at https://github.com/jupyter/repo2docker/pull/788#issuecomment-530384418 saying, "We also change the title from WIP to MRG when it is ready for reviewing (and then maybe merge)."

So should we switch the title to MRG? Are you ready for review? If not, is there any help you need?

It looks like the entire process is documented at https://repo2docker.readthedocs.io/en/latest/contributing/contributing.html#guidelines-to-getting-a-pull-request-merged and here's a screenshot with the MRG bit highlighted in blue:

Screen Shot 2019-09-13 at 8 20 21 AM

Thanks!!!

Xarthisius:
Thank you @pdurbin, I'm aware of r2d contributing process. This PR is still a WIP cause it's literally a work in progress. If you could kindly read this PR's description, you'll notice a TODO list that I'd like to get done before switching status to MRG.

However, if you want to push it across the finish line faster, by all means, just take the code and continue working on it.

pdurbin:
@Xarthisius perfect! Thanks for explaining! I was completely unaware of the repo2docker contribution process before this week but it makes sense that because you build on top of this project you know all about it. Probably the best way for me to help is with advocating for the "add config to test Dataverse support in repo2docker" issue over at https://github.com/IQSS/dataverse-jenkins/issues/13

Additionally, I could put a call out on the dataverse-community mailing list to get you some help on the code itself. Just say the word and I'll put out the call!

Thanks!!!

Xarthisius:
I believe it's now in a "reviewable" state. I significantly reduced the complexity of `detect()`. Right now it doesn't handle all possible cases, but I focused on two that are IMHO gonna be most common: doi pointing to a Dataverse resource and url created by Dataverse's External Tools.

repo2docker/contentproviders/dataverse.py
```diff
@@ -0,0 +1,129 @@
...
+            search_url = urlunparse(
+                parsed_url._replace(path="/api/search", query=search_query)
+            )
+            resp = self.urlopen(search_url).read()
+            data = json.loads(resp.decode("utf-8"))["data"]
+            if data["count_in_response"] != 1:
+                self.log.debug("Dataverse search query failed!")
+                self.log.debug("  - doi = " + doi)
+                self.log.debug("  - url = " + url)
+                self.log.debug("  - resp = " + json.dumps(data))
```
nuest: Not that I should convey the impression I know what "pythonic" is, but maybe this log statements looks better this way:

```Python
self.log.debug("Dataverse search query failed!\n - {}\n - {}\n - {}\n".format(doi, url, json.dumps(data))
```

repo2docker/contentproviders/dataverse.py
```diff
@@ -0,0 +1,129 @@
...
+            return self.detect(new_doi)
+        elif parsed_url.path.startswith("/api/access/datafile"):
+            # Raw url pointing to a datafile is a typical output from an External Tool integration
+            entity_id = os.path.basename(parsed_url.path)
+            search_query = "q=entityId:" + entity_id + "&type=file"
+            # Knowing the file identifier query search api to get parent dataset
+            search_url = urlunparse(
+                parsed_url._replace(path="/api/search", query=search_query)
+            )
+            resp = self.urlopen(search_url).read()
```
nuest: Suggest to `log.debug` the URL before you're calling it, could be important error-finding information for users.

repo2docker/contentproviders/dataverse.py
```diff
@@ -0,0 +1,129 @@
...
+            data = json.loads(resp.decode("utf-8"))["data"]
+            if data["count_in_response"] != 1:
+                self.log.debug("Dataverse search query failed!")
+                self.log.debug("  - doi = " + doi)
+                self.log.debug("  - url = " + url)
+                self.log.debug("  - resp = " + json.dumps(data))
+                return
+
+            self.record_id = deep_get(data, "items.0.dataset_persistent_id")
```
nuest: Could there be a case where more than 1 item will be returned? Should you warn about that here?

repo2docker/contentproviders/dataverse.py
```diff
@@ -0,0 +1,129 @@
...
+        elif is_doi(doi):
+            self.record_id = "doi:" + normalize_doi(doi)
```
nuest: In the other DOI-supporting content providers we only store the repository specific identifier here, not the full DOI. Can you briefly confirm that this does not work here, because multiple Dataverse installations can have different DOI schemes?

repo2docker/contentproviders/dataverse.py
```diff
@@ -0,0 +1,129 @@
...
+            headers={"accept": "application/json"},
+        )
+        resp = self.urlopen(req)
+        record = json.loads(resp.read().decode("utf-8"))["data"]
+
+        # In order to fetch entire dataset we build a list of file IDs we want to fetch
+        # and then receive a zip file containing all of them.
+        # TODO: Dataverse has a limit for the zipfile size (see
+        # https://github.com/jupyter/repo2docker/pull/739#issuecomment-510834729)
+        # If size of the dataset is grater than 100MB individual files should be downloaded.
```
nuest: Suggest to open an issue for that in repo2docker and remove the TODO here. 

OTOH, why not only implement an individual file download?

repo2docker/contentproviders/dataverse.py
```diff
@@ -0,0 +1,129 @@
...
+        # if there is only one new subdirectory move its contents
+        # to the top level directory
+        if len(new_subdirs) == 1:
+            d = new_subdirs[0]
+            copytree(os.path.join(output_dir, d), output_dir)
+            shutil.rmtree(os.path.join(output_dir, d))
+
+    @property
+    def content_id(self):
+        """The Dataverse persistent identifier (could use internal dataset_id too)."""
```
nuest: If you stick to the `doi:` prefix, suggest to mention that here. Also, remove the "could use internal dataset_id", if that is not used eventually - see comment above.

setup.py
```diff
@@ -48,7 +78,7 @@
     ],
     packages=find_packages(),
     include_package_data=True,
-    cmdclass=versioneer.get_cmdclass(),
+    cmdclass=__cmdclass,
```
nuest: Interesting mechanism! Does this man the command is executed every time `repo2docker` is installed?
Could you also do it every time a new release is created?

@betatim @minrk Curious what you think about this.

tests/unit/contentproviders/test_dataverse.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+        assert Dataverse().detect("https://doi.org/10.21105/joss.01277") is None
+
+
+@contextmanager
+def dv_archive(prefix="a_directory"):
+    with NamedTemporaryFile(suffix=".zip") as zfile:
+        with ZipFile(zfile.name, mode="w") as zip:
+            zip.writestr("{}/some-file.txt".format(prefix), "some content")
+            zip.writestr("{}/some-other-file.txt".format(prefix), "some more content")
```
nuest: Looking at the test record, it seems like Dataverse also support directories, right?

image

Or is that only a metadata entry of the file source?

In case Dataverse does support nested directories, it would be good to have a test for that, too.

setup.py
Xarthisius: No, that command is only executed whenever you explicitly call `python setup.py generate_dataverse_file`. If you want it to be executed during `install` I'd need to override default `InstallCommand` and add something like `self.run_command("generate_dataverse_file")` to `InstallCommand.run()`.

BTW, I don't think it needs to be regenerated every release (definitely not every install). I added it so that it doesn't get lost.

repo2docker/contentproviders/dataverse.py
Xarthisius: I might be wrong here, but I got the impression that a scenario with a lot of small files is way more popular than just a few big files case in Dataverse repositories. Going with that hunch, I'd rather see one API call than a bazillion calls.

tests/unit/contentproviders/test_dataverse.py
Xarthisius: Sure, I could add that. However, I don't really make any assumptions about zip content in the Dataverse provider, and as such subdirectories don't affect it. There's only one edge case (single directory with content) which is already covered here. All the other cases boil down to unpack the zip and return root directory where it was unpacked.

repo2docker/contentproviders/dataverse.py
pdurbin: @nuest Dataverse supports both DOIs and Handles and collectively we refer to these as "persistent identifiers" or PIDs.

- DOI example: https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/TJCLKP 
- Handle example: https://data.cimmyt.org/dataset.xhtml?persistentId=hdl:11529/10016

Does this help?

repo2docker/contentproviders/dataverse.py
pdurbin: Let me try to explain the pros and cons of downloading files individually from a Dataverse dataset vs. downloading them as a zip.

## Pros of downloading files individually

- Easier on Dataverse servers that store files on S3 because a "redirect" feature can be enabled such that files are downloading directly from S3 rather than passing through Dataverse servers.
- Easier on Dataverse servers because Dataverse servers do not have to create a zip file on the fly

## Cons of downloading files individually

- In order to know the file hierarchy, you have to parse metadata about the files and look for `directoryLabel` to get path/to/file information. (There can be other potentially interesting metadata about files in there, by the way, such as descriptions.)
- Lots of requests. Bazillion calls.

## Pros of downloading zip files

- Single request.
- Easier to deal with a single file. Just unpack it. The file hierarchy is in the zip file.

## Cons of downloading zip files

- May crash Dataverse servers. Dataverse servers have to construct the zip on the fly. Also when not using the S3 redirect option mentioned above, files pass through Glassfish and Apache.

I hope this helps! I might be forgetting a point or two.

pdurbin:
@Xarthisius @nuest I left a couple comments just now. I hope they're helpful. Please let me know if you have any questions! Thanks!

repo2docker/contentproviders/dataverse.py
pdurbin: Let me try to explain the pros and cons of downloading files individually from a Dataverse dataset vs. downloading them as a zip.

## Pros of downloading files individually

- Easier on Dataverse servers that store files on S3 because a "redirect" feature can be enabled such that files are downloading directly from S3 rather than passing through Dataverse servers.
- Easier on Dataverse servers because Dataverse servers do not have to create a zip file on the fly

## Cons of downloading files individually

- In order to know the file hierarchy, you have to parse metadata about the files and look for `directoryLabel` to get path/to/file information. (There can be other potentially interesting metadata about files in there, by the way, such as descriptions.)
- Lots of requests. Bazillion calls.

## Pros of downloading zip files

- Single request.
- Easier to deal with a single file. Just unpack it. The file hierarchy is in the zip file.

## Cons of downloading zip files

- May crash Dataverse servers. Dataverse servers have to construct the zip on the fly. Also when not using the S3 redirect option mentioned above, files pass through Glassfish and Apache.

I hope this helps! I might be forgetting a point or two.

setup.py
nuest: Thanks for the explanation!

repo2docker/contentproviders/dataverse.py
nuest: Thanks for the elaborate description! That's very helpful.

Re. the `directoryLabel`: that is what is done now in the Zenodo and Figshare providers although I don't think directories are even possible in those repositories, see how `host["download"]` and `host["filename"]` are used in https://github.com/jupyter/repo2docker/blob/master/repo2docker/contentproviders/doi.py#L53

@betatim and @minrk should comment on what is preferable for BinderHub operators.

From a user perspective, I tend towards downloading files individually, because I want things not to break (e.g. if the ZIP would larger then 1 GB) and don't care about load on the servers. The comment says

> "Dataverse has a limit for the zipfile size"

so I think there is no way around implementing individual download (maybe as a second step if the zipfile download fails) unless you'd be fine with not supporting large repos.

tests/unit/contentproviders/test_dataverse.py
nuest: AFAICS this is related to the zip vs. individual download discussion above - can close here.

repo2docker/contentproviders/dataverse.py
pdurbin: I mean, if it's easier to implement zip download first just to get something working, that's fine with me. It should work fine with datasets with small files. 😄 As you see, I'm just trying to explain the pitfalls of creating zip files on the fly from the Dataverse server perspective. (And it's a bad experience for users if the zip file isn't complete, if it only has 5 of the 17 files the user wants because they exceeded the zip size limit.) So yes, downloading individual files, even if it's a bazillion requests, is probably safer.

repo2docker/contentproviders/dataverse.py
Xarthisius: modified the code as suggested

repo2docker/contentproviders/dataverse.py
Xarthisius: I reimplemented fetch to use individual files download

tests/unit/contentproviders/test_dataverse.py
Xarthisius: add code as suggested

repo2docker/contentproviders/dataverse.py
```diff
@@ -0,0 +1,129 @@
...
+        host = next(
+            (
+                host
+                for host in self.hosts
+                if urlparse(host["url"]).netloc == parsed_url.netloc
+            ),
+            None,
+        )
+        if host is None:
+            return
```
betatim: To check I parsed the above lines properly (the use of `next()` to handle the "didn't work" case made me read it a few times), is the following code equivalent (I think yes):
```python
for host_ in self.hosts:
    if urlparse(host_["url"]).netloc == parsed_url.netloc:
        host = host_
else:
    return
```

betatim:
LGTM! Not quite sure why codecov is (that0 unhappy. Merging this now. Thanks for the work and then coming back to finish it off :)

Two more todos for a quick PR:
* add an entry to the changelog
* mention Figshare and Dataverse support in the first paragraph of https://repo2docker.readthedocs.io/

repo2docker/contentproviders/dataverse.py
nuest: Ah OK, I see. Thanks!

nuest:
Working on the docs update PR...

PR #743, opened 2019-07-16
[MRG] preassembly for conda/python

minrk:
re-issue of #410 based on #716

- conda preassembles *unless* there are local pip requirements
- requirements.txt preassembles *unless* there are local refs (e.g. `-e .`)
- <del>pipenv always preassembles (may need local ref check like requirements.txt?)</del>

### DIFF
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -145,7 +145,7 @@ class CondaBuildPack(BaseImage):
             # check if the env file is empty, if so instantiate an empty dictionary.
             if env is None:
                 env = {}
-            # check if the env file provided a dick-like thing not a list or other data structure.
+            # check if the env file provided a dict-like thing not a list or other data structure.
             if not isinstance(env, Mapping):
                 raise TypeError(
                     "environment.yml should contain a dictionary. Got %r" % type(env)
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -6,6 +6,7 @@ from collections import Mapping
 from ruamel.yaml import YAML
 
 from ..base import BaseImage
+from ...utils import is_local_pip_requirement
 
 # pattern for parsing conda dependency line
 PYTHON_REGEX = re.compile(r"python\s*=+\s*([\d\.]*)")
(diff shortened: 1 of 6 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/tip-embed-custom-github-content-in-a-binder-link-with-nbgitpuller/922/50

betatim:
I think a `environment.yml` with `- -r requirements.txt` in the pip section is legal. At least I have a vague memory that when I investigated last I found this out. Something to double check.

repo2docker/buildpacks/conda/__init__.py
```diff
@@ -178,7 +178,19 @@ def py2(self):
...
+    def get_preassemble_script_files(self):
+        """preassembly only requires environment.yml
+
+        enables caching assembly result even when
+        repo contents change
+        """
+        assemble_files = super().get_preassemble_script_files()
+        environment_yml = self.binder_path("environment.yml")
+        if os.path.exists(environment_yml):
+            assemble_files[environment_yml] = environment_yml
```
betatim: Should we attempt to sort the contents of a `environment.yml` so that two repos asking for the same packages in a different order get a cache hit? Or is this a level two optimisation for a new PR?

minrk: ooh, that's a very interesting question. I think that's a job for a second PR. I'm also not sure how often this is likely to happen, since it has to be a truly identical environment specification (not just equivalent) unless we import conda and start resolving dependencies, which would slow down builds quite a bit, I think, since it would have to happen prior to generating the Dockerfile.

Before spending much time on that, I think it'd be worth trying to survey environment.ymls in the wild and seeing how many:

1. are not identical byte-wise, but
2. are identical, dependency-wise

minrk:
I bumped the local pip check up one so that it applies to conda as well (adding unittests, too)

minrk:
I decided to save pipenv for the preassembly for another PR, since that should be done by parsing the Pipfile toml, which is added in #748.

repo2docker/buildpacks/conda/__init__.py
betatim: Created an issue to tackle this https://github.com/jupyter/repo2docker/issues/749

repo2docker/buildpacks/conda/__init__.py
```diff
@@ -127,6 +128,50 @@ def get_build_script_files(self):
...
+        if not os.path.exists(environment_yml):
+            self._environment_yaml = {}
+            return self._environment_yaml
+
+        with open(environment_yml) as f:
+            env = YAML().load(f)
+            # check if the env file is empty, if so instantiate an empty dictionary.
+            if env is None:
+                env = {}
+            # check if the env file provided a dick-like thing not a list or other data structure.
```
betatim:
```suggestion
            # check if the env file provided a dict-like thing not a list or other data structure.
```
the schoolboy in me chuckled

PR #748, opened 2019-07-18
parse pipfiles to determine Python version

minrk:
these are structured files, no need to use regular expressions to parse them


betatim:
Thanks!

PR #752, opened 2019-07-23
[MRG] Call parent preassemble scripts methods

### LINKED ISSUE #751: runtime.txt for installing R and requirements.txt clashing
Previously, I was installing R/RSTudio using `runtime.txt` and designating some Python packages to install in same build with `requirements.txt`. ([This repository](https://github.com/fomightez/b-roll-binder) allowed me to launch sessions up until today when I rebuilt it to add a `start` script.) Now this combination of `runtime.txt` and `requirements.txt` doesn't work.  
I made a simpler repository to demonstrate as the one I was working with when I found this issue is rather complex.  

TEST EXAMPLE:   
repo: https://github.com/fomightez/testruntimeANDreqs
binder launch link: https://mybinder.org/v2/gh/fomightez/testruntimeANDreqs/master (currently fails to build)

ERROR SEEN:  
```bash
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
Removing intermediate container 927ccc3e1883
The command '/bin/sh -c ${KERNEL_PYTHON_PREFIX}/bin/pip install --no-cache-dir -r "requirements.txt"' returned a non-zero code: 1
```

NOTE:  
Hiding `runtime.txt` by appending `.` in front makes `requirements.txt` be seen again and build works. However, there is no R then.

betatim:
Without this we don't collect the files that other build packs need to
copy them over to the container in the pre-assemble stage.

Needs a test to stop it from coming back.

closes #751

### DIFF
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -292,7 +292,7 @@ class RBuildPack(PythonBuildPack):
         return super().get_build_scripts() + scripts
 
     def get_preassemble_script_files(self):
-        files = {}
+        files = super().get_preassemble_script_files()
         installR_path = self.binder_path("install.R")
         if os.path.exists(installR_path):
             files[installR_path] = installR_path
--- a/tests/unit/test_preassemble.py
+++ b/tests/unit/test_preassemble.py
@@ -1,0 +1,27 @@
+import os
+
+import pytest
+
+from repo2docker import buildpacks
+
+
+@pytest.mark.parametrize("binder_dir", ["", ".binder", "binder"])
+def test_combine_preassemble_steps(tmpdir, binder_dir):
+    tmpdir.chdir()
+    if binder_dir:
+        os.mkdir(binder_dir)
+
+    # create two empty files for the build pack to use for pre-assembly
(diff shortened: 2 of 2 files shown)


choldgraf:
I see you've added a test now?

choldgraf:
I see you've added a test now?

betatim:
Yes I have. Forgot to update the top comment

choldgraf:
This seems good to me - the only thing that seems a bit strange is the chain of sub-classing we've got going on. The RBuildPack inherits from the PythonBuildPack, which inherits from the CondaBuildPack, and so on. I don't see any reason that this PR should be the one to change that pattern, but just something I noticed that made it harder to figure out what was happening in that `super()` call.

betatim:
`CondaBuildPack` is the actual "base buildpack" because we use it to install conda which we use to run the notebook server. This is why everything (or pretty much everything) inherits from the conda buildpack. Does that make sense?

(and it sometimes surprises me as well that the PythonBuildPack inherits from the conda one, but I think it makes sense. Maybe different names would help?)

Issue #753, opened 2019-07-25
bot for refreezing


minrk:
### Proposed change

Every once in a while, we want to rebuild the base environments. This is done by running the `refreeze.py` script.

### Alternative options

keep manually refreezing

### Who would use this feature?

Maintainers of repo2docker and folks contributing updates to the base environments who don't want to run refreeze themselves.

One way is to do periodic refreeze, to keep things up to date, as is done with henchbot. Another is to allow pull requests to update the environment spec or other files and then post a comment like `@r2d-bot please refreeze`.

### How much effort will adding it take?

Not sure! Hopefully not too much. Requires writing a bot that talks to the GitHub API. It would need to:

1. respond to a trigger somehow (e.g. a `@refreeze-bot please refreeze` comment to update a PR or a new `please refreeze` Issue to trigger a new PR)
2. make a commit in an existing pull request and/or open a new pull request

### Who can do this work?

- Interest in / knowledge of the GitHub API
- Interest in / knowledge of writing bots and deploying them somewhere (on heroku, etc.)

References:

- [henchbot](https://blog.jupyter.org/automating-mybinder-org-dependency-upgrades-in-10-steps-bb5e38542059) which does pull requests for updates to repo2docker and binderhub for mybinder.org
- [conda-forge-admin bot](https://conda-forge.readthedocs.io/en/latest/webservice.html) which does things like add commits to pull requests to run common commands like `@conda-forge-admin please rerender`

henchc:
I'd be happy to take this up, I've been wanting to get more into repo2docker but hadn't found a good entry point yet.

I don't see a "refreeze.py" just "freeze py" in the conda and legancy buildpacks. Am I missing something?

Also, does this depend on builds existing outside of the state of the GitHub repo? One of the reasons henchbot just goes every hour is because the images maybe hadn't made it to Dockerhub yet, so it checks there specifically.

betatim:
I think it is a typo. `freeze.py` is what we use to (re)freeze the environments.

I think it uses a miniconda docker image so we shouldn't have the problem where we needto wait for an image to build.

minrk:
> Am I missing something?

@betatim is right. I updated the description: We only want to run freeze.py in the conda buildpack. This will regenerate the eight `environment.pyX.Y[.frozen].yml` files.

> does this depend on builds existing outside of the state of the GitHub repo?

[...]

PR #754, opened 2019-07-26
[DO NOT MERGE] refreeze PR check

henchc:
This is confirming what should be included in a PR generated by the proposed bot in #753 .
I ran the `freeze.py` script in both conda and legacy buildpacks. A few questions:

1. Do we refreeze the legacy too? Or just conda?
2. Is this a decoupled update or tied to a change in this repo? It sounds like @minrk implied that it *is* decoupled (as in the state of the repo may not change but a refreeze is still desired) but *also* that changes to the base environments in the repo would require a refreeze. So maybe we want *both* periodic stand-alone PRs *and* a responsive bot to refreeze on top of specific PRs?
3. Is there specific text we want in the PR body that would be helpful?

### DIFF
--- a/repo2docker/buildpacks/conda/environment.frozen.yml
+++ b/repo2docker/buildpacks/conda/environment.frozen.yml
@@ -1,5 +1,5 @@
 # AUTO GENERATED FROM environment.py-3.7.yml, DO NOT MANUALLY MODIFY
-# Frozen on 2019-07-16 13:18:20 UTC
+# Frozen on 2019-07-25 23:43:10 UTC
 name: r2d
 channels:
   - conda-forge
@@ -43,7 +43,7 @@ dependencies:
   - parso=0.5.1=py_0
   - pexpect=4.7.0=py37_0
   - pickleshare=0.7.5=py37_1000
-  - pip=19.1.1=py37_0
+  - pip=19.2.1=py37_0
   - prometheus_client=0.7.1=py_0
   - prompt_toolkit=2.0.9=py_0
   - ptyprocess=0.6.0=py_1001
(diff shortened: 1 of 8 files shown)


henchc:
I'm guessing "no" to the legacy freeze?

minrk:
> Do we refreeze the legacy too? Or just conda?

Just conda. We don't want to update legacy under normal circumstances, since that's freezing a point in time well in the past. We may do it by hand from time to time for security reasons.

>  So maybe we want both periodic stand-alone PRs and a responsive bot to refreeze on top of specific PRs?

I think both will be useful, but they can be taken as separate tasks, perhaps. I'd trigger the responsive one on PR comments a la [conda-forge](https://github.com/conda-forge/pyzmq-feedstock/pull/25) rather than PR contents.

> Is there specific text we want in the PR body that would be helpful?

A summary of package version updates would be great. E.g. in this PR, a message like:

> Python 3.7:

> - updated pip from `19.1.1 to 19.2.1` (version changed, no need to report build string, but fine if to include if it's easier)
> - update pyzmq build from `18.0.2=py37hc4ba49a_1` to `18.0.2=py37h1768529_2` (only build changed)
> ...

> Python 3.6:

> ...

And for changes like the ones in legacy:

- added package `ca-certificates 2019.5.15`

and removals, if they happen.

To parse this, you'll need to load the contents of the files into a data structure to compare, parsing packages like:

```python
# for conda deps 
name, version, build = line.split('=')
# for pip deps (the last item in the conda deps is a dict that looks like `{'pip': ['pkg==x.y.z',...]}`)
name, version = line.split('==')
```

PR #757, opened 2019-08-02
[MRG] Rewrite tests in test_semver.py #615

### LINKED ISSUE #615: Rewrite semver tests to use pytest's parametrize functionality
via https://github.com/jupyter/repo2docker/pull/613#discussion_r263836315:

> In a future refactor, we should use pytest's parameterize capability to feed in versions to get closer to DRY. [See example from JupyterHub tests](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/tests/test_proxy.py#L169).

This would make a good issue for getting started on contributing to repo2docker and/or someone who wants to learn more about pytest.

Geektrovert:
Rewrite tests using Pytest's [parametrizing](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) functionality.

### DIFF
--- a/tests/unit/test_semver.py
+++ b/tests/unit/test_semver.py
@@ -1,9 +1,10 @@
+import pytest
 from repo2docker.buildpacks.julia import semver
 
 
-def test_str_to_version():
-    assert semver.str_to_version("1.5.2") == (1, 5, 2)
-    assert semver.str_to_version("1") == (1,)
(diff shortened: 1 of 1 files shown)


betatim:
The coverage test fails because the coverage apparently went down. This sometimes happens when you delete a lot of code. I'll take a look if I can see something suspicious or if this is is "noise".

betatim:
Looks good to me. I think the coverage computation was ab it confused. Merged :)

Geektrovert:
It was a good learning experience! Thanks :)

PR #768, opened 2019-08-26
Add support for Julia 1.2.0

davidanthoff:
Julia 1.2.0 is released :)

I believe the julia_version-default test should now automatically test on 1.2.0 with this PR.


betatim:
Thanks :)

Issue #764, opened 2019-08-16
Handle requirements.txt files that use alternative encodings


betatim:
This issue originates from https://github.com/jupyterhub/mybinder.org-deploy/issues/1125

### Bug description

A `requirements.txt` using an encoding that isn't utf-8 causes repo2docker to raise an error:
```
$ repo2docker https://github.com/YoniChechik/computer_vision_course
Picked Git content provider.
Cloning into '/var/folders/rr/69pl0sh155l1ncw782764xq00000gn/T/repo2dockerkjatc_o3'...
remote: Enumerating objects: 55, done.
remote: Counting objects: 100% (55/55), done.
remote: Compressing objects: 100% (55/55), done.
remote: Total 55 (delta 2), reused 40 (delta 0), pack-reused 0
Unpacking objects: 100% (55/55), done.
Traceback (most recent call last):
  File "/Users/thead/miniconda/envs/bhub/bin/repo2docker", line 11, in <module>
    load_entry_point('jupyter-repo2docker', 'console_scripts', 'repo2docker')()
  File "/Users/thead/git/jupyterhub/repo2docker/repo2docker/__main__.py", line 354, in main
    r2d.start()
  File "/Users/thead/git/jupyterhub/repo2docker/repo2docker/app.py", line 714, in start
    self.build()
  File "/Users/thead/git/jupyterhub/repo2docker/repo2docker/app.py", line 668, in build
    self.log.debug(picked_buildpack.render(), extra=dict(phase="building"))
  File "/Users/thead/git/jupyterhub/repo2docker/repo2docker/buildpacks/base.py", line 519, in render
    for user, script in self.get_assemble_scripts():
  File "/Users/thead/git/jupyterhub/repo2docker/repo2docker/buildpacks/python/__init__.py", line 124, in get_assemble_scripts
    if not self._should_preassemble_pip:
  File "/Users/thead/git/jupyterhub/repo2docker/repo2docker/buildpacks/python/__init__.py", line 90, in _should_preassemble_pip
    for line in f:
  File "/Users/thead/miniconda/envs/bhub/lib/python3.7/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

#### Expected behaviour

If the `requirements.txt` uses an encoding supported by `pip` the repository should build and launch as normal.

### How to reproduce

[...]

PR #771, opened 2019-08-28
Handle different file encodings

GeorgianaElena:
This should fix https://github.com/jupyter/repo2docker/issues/764

Decided to use *chardet* to guess the encoding because I liked the idea of reusing the function for ``environment.yml`` also (not sure if this was the right call).

* [x] implement function to detect encoding
* [x] add tests

### DIFF
--- a/repo2docker/utils.py
+++ b/repo2docker/utils.py
@@ -3,6 +3,7 @@ from functools import partial
 import os
 import re
 import subprocess
+import chardet
 
 from shutil import copystat, copy2
 
@@ -68,6 +69,22 @@ def chdir(path):
         yield
     finally:
         os.chdir(old_dir)
+
+@contextmanager
+def open_utf8convert_read(path):
+    with open(path, "rb") as f:
+        file_to_encode = f.read()
+
+    encoding_detection_result = chardet.detect(file_to_encode)
+    if not "utf-8" in encoding_detection_result:
+        with open(path, "wb") as f:
+            f.write(file_to_encode.decode(encoding_detection_result["encoding"]).encode("utf-8"))
+
+    file = open(path)
+    try:
+        yield file
+    finally:
+        file.close()
 
 
 def validate_and_generate_port_mapping(port_mappings):
--- a/repo2docker/utils.py
+++ b/repo2docker/utils.py
@@ -81,7 +81,6 @@ def open_guess_encoding(path):
     with open(path, "rb") as f:
         for line in f.readlines():
             detector.feed(line)
-            print(str(i) + str(detector.done))
             if detector.done:
                 break
     detector.close()
--- a/repo2docker/utils.py
+++ b/repo2docker/utils.py
@@ -70,17 +70,23 @@ def chdir(path):
     finally:
         os.chdir(old_dir)
 
+
 @contextmanager
-def open_utf8convert_read(path):
+def open_guess_encoding(path):
+    """
+    Open a file in text mode, specifying its encoding,
+    that we guess using chardet.
+    """
+    detector = chardet.universaldetector.UniversalDetector()
     with open(path, "rb") as f:
-        file_to_encode = f.read()
+        for line in f.readlines():
+            detector.feed(line)
+            print(str(i) + str(detector.done))
+            if detector.done:
+                break
+    detector.close()
 
-    encoding_detection_result = chardet.detect(file_to_encode)
-    if not "utf-8" in encoding_detection_result:
-        with open(path, "wb") as f:
-            f.write(file_to_encode.decode(encoding_detection_result["encoding"]).encode("utf-8"))
-
-    file = open(path)
+    file = open(path, encoding=detector.result["encoding"])
     try:
         yield file
     finally:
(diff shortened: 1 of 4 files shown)


betatim:
I like the idea of reusing it from `environment.yaml`. No idea where to check what legal encodings are for that file though. Maybe they don't allow anything except utf-8 in which case we should probably also not allow that. It should be written down somewhere (deep) in the conda docs or code.

repo2docker/utils.py
```diff
@@ -69,6 +70,22 @@ def chdir(path):
     finally:
         os.chdir(old_dir)
 
+@contextmanager
+def open_utf8convert_read(path):
```
betatim: I'm not a huuuge fan of the name as it is long and still I feel like I have to go read its code to know what it does.

Some ideas, neither of which I really like either: `open_with_chardet`, `open_guess_encoding`, `guess_encoding`, `open_auto_encoding` :-/

repo2docker/utils.py
```diff
@@ -69,6 +70,22 @@ def chdir(path):
...
+    with open(path, "rb") as f:
+        file_to_encode = f.read()
+
+    encoding_detection_result = chardet.detect(file_to_encode)
+    if not "utf-8" in encoding_detection_result:
+        with open(path, "wb") as f:
+            f.write(file_to_encode.decode(encoding_detection_result["encoding"]).encode("utf-8"))
```
betatim: I don't think we should overwrite the original file. If I was a user who carefully chose the encoding (for what ever reason) I'd be annoyed at repo2docker for overwriting it with some other encoding.

In (new-ish?) versions of Python we can pass the encoding of the file to `open`: https://docs.python.org/3/library/functions.html#open maybe we can ask chardet for the name of the encoding and pass it to `open()`. Or if this feature is too new (we support Python 3.5 I think) then maybe `chardet` has a `open()` that supports passing it an encoding?

repo2docker/utils.py
```diff
@@ -69,6 +70,22 @@ def chdir(path):
     finally:
         os.chdir(old_dir)
 
+@contextmanager
+def open_utf8convert_read(path):
+    with open(path, "rb") as f:
+        file_to_encode = f.read()
+
+    encoding_detection_result = chardet.detect(file_to_encode)
```
betatim: Not sure it makes much difference here in terms of performance so this is more for my education/old memory: is there a way to read only a small part of the file and guess the encoding instead of having to read the whole file?

repo2docker/utils.py
GeorgianaElena: Sorry, but I'm not that good at naming things. I changed the name 3 times before coming up with ``open_utf8convert_read`` (the other ones were waaay worse :disappointed_relieved:).

I really like ``open_guess_encoding``, thanks :heart:

repo2docker/utils.py
GeorgianaElena: According to the docs we can open a file specifying its encoding in Python 3.5 also, yay :confetti_ball:

repo2docker/utils.py
GeorgianaElena: Seems like a good idea and apparently chardet's docs have an example about [detecting encoding incrementally](https://chardet.readthedocs.io/en/latest/usage.html#advanced-usage) :confetti_ball:. 
I *think* that even though it might not impact the performance that much in *requirements.txt*'s case (I'm guessing the file is not very large in general), it might make a difference if used somewhere else, on other files :thinking:

GeorgianaElena:
Thanks for the feedback @betatim :sun_with_face:
I ended up not using the function for ``environment.yml`` because it seems that only utf-8 is accepted (according to [this line](https://github.com/conda/conda/blob/master/conda_env/specs/yaml_file.py#L18) and [this line](https://github.com/conda/conda/blob/aaad747a882ad97628beea092a51579bec79805e/conda_env/env.py#L149)).

repo2docker/utils.py
betatim: No worries! Finding good names (not too long, tell you what the thing is, etc, etc) is one of the hardest things. Also totally worth doing I think because future us will thank past us for giving the thing a good name :)

repo2docker/utils.py
betatim: Just checked we require Python 3.5+ so good to go!

repo2docker/utils.py
betatim: Took a quick look at the example in the docs that you linked to and my first impression was "woah what is going on here, no context managers? do we need code to handle errors?". Maybe we keep the "read it all at once" because it is nice and clear. Maybe we leave a comment pointing to the incremental one for later? Maybe there is a nice way to write the incremental example with modern Python? What do you think?

betatim: Too slow in commenting/didn't see the "outdated" flag. Your code doesn't look as scary as the example so let's keep it :)

GeorgianaElena:
@betatim, related to the tests, do you think adding some unit tests in ``test_utils.py`` covering more or less the cases [here](https://github.com/pypa/pip/blob/3c9770d9b22b6e3e4de927ef8e1aa9c5155b74d6/tests/unit/test_utils.py#L531) is enough? :eyes:

betatim:
Those look like some good tests to copy!

I was thinking "take a string with 'weird' characters, encode it to bytes using a legal but different from utf-8 encoding, write bytes to temporary file, read them back in with our `open_guess_encoding`, check the string is the same as at the beginning".

PR #772, opened 2019-09-03
[MRG] Add support for installing different versions of R

### LINKED ISSUE #661: Different versions of R or staying just a point release behind the current version
### Proposed change
Allowing different versions of R (and possibly RStudio) to be run (R 3.6 vs. 3.5 vs. 3.4 etc).

### Alternative options
Assuming that installing the most recent version of R would be less intensive that installing a pre-specified version, just keeping up to date with the major release versions would be a fantastic alternative - but maybe a point or two behind.  R 3.6 has just been released - so maybe the most stable release is the previous version (3.5.3) 

Reproducibility has become increasingly important for researchers in Psychology and many groups and labs have switched to R for open and reproducible research (incl. in teaching).  Given that reproducing the computational environment is arguably the gold standard of reproducibility, it would be hugely beneficial to be able to launch a particular version of R using repo2docker - so the runtime.txt file wouldn't just determine what version of packages are pulled from MRAN, but it could also contain a specification of which version of R to run.  

I can imagine a future where research journal articles each contain a link in them so that reviewers and readers can launch a Binder to see the entire analysis script and data exactly as it was on the date the authors carried out the analysis.  That would be a huge benefit to the community and a massive boon for reproducibility.

I'm not sure I can estimate this.

### Who can do this work?
I'm guessing someone who knows how to install different versions of R via Ubuntu’s package manager - and how repo2docker can read info in the runtime.txt file to determine which version of R to then install.  I'm happy to help where I can - although I'm only a psychologist!

betatim:
Closes #661 

This uses the instructions from https://cran.r-project.org/bin/linux/ubuntu/README.html to allow users to select a different version of R. We use the `runtime.txt` to let users specify which version they want via `r-3.6-YYYY-MM-DD` (selects R 3.6, 3.6.1 to be precise) or `r-YYYY-MM-DD` (selects the default, currently 3.4).

We can also add support for R 3.5 but I wanted to open the PR before being done. Ideally `r-3.4-YYYY-MM-DD` should be equivalent to `r-YYYY-MM-DD` as well.

This is pretty exciting as it adds a frequently requested feature without a lot of new code.

What do people think of using `runtime.txt` like this? I like it and prefer it over trying to guess a R version from the date or introducing an additional file. Maybe in a second PR we can check if a `DESCRIPTION` file can specify a R version as well.

It would be good to get some feedback from R packaging experts on https://cran.r-project.org/bin/linux/ubuntu/README.html.

* [x] R 3.5 support
* [x] make 3.6 the default
* [x] tests
* [ ] docs

### DIFF
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -2,6 +2,8 @@ import re
 import os
 import datetime
 
+from distutils.version import LooseVersion as V
+
 from .python import PythonBuildPack
 
 
@@ -170,7 +172,7 @@ class RBuildPack(PythonBuildPack):
         ]
         # For R 3.4 we use the default Ubuntu package, for other versions we
         # install from a different PPA
-        if self.r_version.startswith("3.4"):
+        if V(self.r_version) < V("3.5"):
             packages.append("r-base")
 
         return super().get_packages().union(packages)
@@ -216,7 +218,7 @@ class RBuildPack(PythonBuildPack):
         scripts = []
         # For R 3.4 we want to use the default Ubuntu package but otherwise
         # we use the packages from a PPA
-        if not self.r_version.startswith("3.4"):
+        if V(self.r_version) >= V("3.5"):
             scripts += [
                 (
                     "root",
@@ -342,7 +344,7 @@ class RBuildPack(PythonBuildPack):
 
         if "r" in self.stencila_contexts:
             # new versions of R require a different way of installing bioconductor
-            if self.r_version.startswith("3.4"):
+            if V(self.r_version) <= V("3.5"):
                 scripts += [
                     (
                         "${NB_USER}",
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -214,7 +214,8 @@ class RBuildPack(PythonBuildPack):
         )
 
         scripts = []
-        # For R 3.4 we want to use the default Ubuntu package
+        # For R 3.4 we want to use the default Ubuntu package but otherwise
+        # we use the packages from a PPA
         if not self.r_version.startswith("3.4"):
             scripts += [
                 (
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -59,15 +59,30 @@ class RBuildPack(PythonBuildPack):
     def r_version(self):
         """Detect the R version for a given `runtime.txt`
 
-        Will return '' (false-y string) as the default or other 'x.y'
-        version if found.
+        Will return the version specified by the user or the current default
+        version.
         """
+        version_map = {
+            "3.5": "3.5.3-1bionic",
+            "3.5.0": "3.5.0-1bionic",
+            "3.5.1": "3.5.1-2bionic",
+            "3.5.2": "3.5.2-1bionic",
+            "3.5.3": "3.5.3-1bionic",
+            "3.6": "3.6.1-3bionic",
+            "3.6.0": "3.6.0-2bionic",
+            "3.6.1": "3.6.1-3bionic",
+        }
+        # the default if nothing is specified
+        r_version = "3.6"
+
         if not hasattr(self, "_r_version"):
             parts = self.runtime.split("-")
             if len(parts) == 5:
-                self._r_version = parts[1]
-            else:
-                self._r_version = ""
+                r_version = parts[1]
+
+            # see if we need to translate the specified version into a
+            # "full" version string
+            self._r_version = version_map.get(r_version, r_version)
 
         return self._r_version
 
@@ -79,7 +94,7 @@ class RBuildPack(PythonBuildPack):
         Returns '' if no date is specified
         """
         if not hasattr(self, "_checkpoint_date"):
-            match = re.match(r"r-(\d.\d-)?(\d\d\d\d)-(\d\d)-(\d\d)", self.runtime)
+            match = re.match(r"r-(\d.\d(.\d)?-)?(\d\d\d\d)-(\d\d)-(\d\d)", self.runtime)
             if not match:
                 self._checkpoint_date = False
             else:
@@ -153,7 +168,9 @@ class RBuildPack(PythonBuildPack):
             "sudo",
             "lsb-release",
         ]
-        if not self.r_version:
+        # For R 3.4 we use the default Ubuntu package, for other versions we
+        # install from a different PPA
+        if self.r_version.startswith("3.4"):
             packages.append("r-base")
 
         return super().get_packages().union(packages)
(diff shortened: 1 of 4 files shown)


trallard:
Some thoughts here:
- It would make more sense to have the latest R version as the "empty one". So that the most recent version of R would be installed by default. This will make it consistent with the Rocker Docker images behaviour. 
- I quite like the idea of having the R version explicitly declared in the `runtime.txt` as `r-3.6-YYYY-MM-DD` 
- I vote in favour of directly supporting 3.5 - I believe a good chunk of R users will either be running on 3.5 or 3.6 (probably more than 3.4) as this is quite old now

betatim:
> It would make more sense to have the latest R version as the "empty one"

I was wondering about that. Right now I went with "nothing should change when we merge this", currently you can't select the R version so after merging this PR if you don't specify anything we stick (unfortunately) with what used to be the only version. This is the "never break users" mantra. I don't actually know how many existing repos would break if we suddenly switch them from 3.4 to 3.6. Is there a way to guess-timate that (without having to go and build a whole bunch of repos)?

For Python we didn't follow this mantra and at some point switched to Python 3.7 as the default. There were a few (but less than expected) people who had problems with their repos not building any more who showed up on twitter/gitter/the forum. What is unknown is how many repos kept building but now produced different results (probably less than those that broke).

I need to check what we do for Julia (no spec -> we update your version or no spec -> we stick to some old version).

trallard:
This would be particularly hard to guesstimate, but I do not expect a huge number of repos to break if we upgrade to 3.6 right away.
That said I imagine there will be a few (maybe more than a few) repos currently broken as some R packages tend to drop support for old R base versions. Of course there are workarounds to install packages for an older R version.

Something that could be done would be upgrading to 3.6 (as in Python) and perhaps add a FAQ/detailed documentation on the repo2docker and binder pages so that this does not result in loads of issues around the line "my repo worked a week ago and now it's broken"?

NHDaly:
When I first made the change in Julia, I kept everyone on the old version.

But the Julia transition from 0.6 (the default before) to 1.0 was more
significant -- 0.6 is no longer supported, and almost no one uses it, so we
then changed the default to 1.0.

Now, I don't think we've been updating the version to run the latest Julia,
which probably makes sense. 1.0 is the "long-term supported" version, so I
think continuing to update the default to the latest patch release of 1.0
is reasonable (1.0.4, 1.0.5, etc).

I'm not sure how much of that maps to the situation for R?

For the full details on Julia's version numbers, they just released this
thorough blog post that discusses the above in more detail:
https://julialang.org/blog/2019/08/release-process

Hopefully that's helpful! :) Sorry if it was too much information.

NHDaly:
> Now, I don't think we've been updating the version to run the latest Julia,
which probably makes sense. 1.0 is the "long-term supported" version, so I
think continuing to update the default to the latest patch release of 1.0
is reasonable (1.0.4, 1.0.5, etc).

Oh, just kidding, apparently we just use the latest released version of julia as the default:
https://github.com/jupyter/repo2docker/blob/46cd5670e821f210930562a42e2769488bd23149/repo2docker/buildpacks/julia/julia_project.py#L30
🤷‍♂i think that's reasonable too. :) Supposedly, the upgrades are always supposed to be backwards compatible within `1.*`, even if downgrades aren't. So this is probably fine as well.

Anyway, hope that's helpful. :)

repo2docker/buildpacks/r.py
```diff
@@ -180,7 +195,34 @@ def get_build_scripts(self):
             self.checkpoint_date.isoformat()
         )
 
-        scripts = [
+        scripts = []
+        if self.r_version == "3.6":
```
choldgraf: Maybe I am being silly, but what happens if other R versions are given? I couldn't find a spot where a different R installation command is triggered if the specified version isn't 3.6

choldgraf: oops, I just read through your first comment again, I see there's no 3.5 support yet but that's on the horizon, so feel free to ignore this :-)

choldgraf:
This seems pretty reasonable to me - it feels like a good piggy-backing of the runtime buildpack. Just to have 👀 on it, one other pattern we could use is something like:

```
r-VERSION
mran-YYYY-MM-DD
```

which would be more explicit (although two separate lines) and would disentangle the two pieces of information. I can see "pros" and "cons" with both approaches, but curious what other folks think about this

repo2docker/buildpacks/r.py
betatim: Nothing yet because this PR doesn't even work properly if you don't specify 3.6 :)

betatim:
This is ready for review now @minrk or @nuest

repo2docker/buildpacks/r.py
```diff
@@ -55,6 +55,37 @@ def runtime(self):
...
+        r_version = "3.6"
+
+        if not hasattr(self, "_r_version"):
+            parts = self.runtime.split("-")
+            if len(parts) == 5:
+                r_version = parts[1]
+
+            # see if we need to translate the specified version into a
+            # "full" version string
+            self._r_version = version_map.get(r_version, r_version)
```
nuest: I think I would have an easier time reading this as `version_map.get(r_version, default = r_version)`.

What about the following test, which fails with the current code:

```python

def test_fallback_version_completion(tmpdir):
    tmpdir.chdir()

    with open("runtime.txt", "w") as f:
        f.write(f"r-3.8-2019-01-01")
    r = buildpacks.RBuildPack()
    assert r.r_version == "3.6.1-3bionic"

    with open("runtime.txt", "w") as f:
        f.write(f"r-2.15-2019-01-01")
    r = buildpacks.RBuildPack()
    assert r.r_version == "3.6.1-3bionic"
```

Do we want to capture the case where someone specifies an unsupported version and go to the default? We probably should fail then, right? It was just confusing to me to because `version.map(..)` will not return a full version string if there is no runtime version specified.

repo2docker/buildpacks/r.py
```diff
@@ -55,6 +55,37 @@ def runtime(self):
...
         return self._runtime
 
+    @property
+    def r_version(self):
+        """Detect the R version for a given `runtime.txt`
+
+        Will return the version specified by the user or the current default
+        version.
+        """
+        version_map = {
```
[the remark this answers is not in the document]
betatim: Which version would that install? The latest 3.x that exists? Maybe we keep it for a later PR when there is R 4.x?

nuest:
Exactly, the latest R version in the 3 series. It would move along similar to a x.y version.

You are right, it probably makes more sense to do it when there are alternatives (like support for 2.x.z or >3. Maybe R stays with 3 for a long time!).

repo2docker/buildpacks/r.py
betatim: I made it so that if the user specifies a version that isn't supported an exception is raised. I think that is what should happen in this case as my guess would be that the user made a typo when there is a version specified but it isn't a valid one.

nuest:
LGTM

betatim:
I updated the documentation to mention R version support. Ready to merge if the bots are happy.

betatim:
Now with more `LooseVersion` @minrk

betatim:
Merging this now. Min's last (in person) comment on this was positive except for the part about not using `LooseVersion` which is now fixed.

PR #786, opened 2019-09-08
[MRG] Restore the hooks directory when building the docker image

betatim:
This is an attempt to fix [the fact that the repo is still marked as dirty](https://github.com/jupyterhub/mybinder.org-deploy/pull/1151#issuecomment-529171875) because now the files are missing.

* [ ] check the docker cloud logs https://cloud.docker.com/u/jupyter/repository/docker/jupyter/repo2docker/timeline


betatim:
I've cancelled most of the travis builds as this PR doesn't need them and this way we can save some resources.

Issue #744, opened 2019-07-16
Use Azure pipelines for tests


minrk:
repo2docker tests take a long time! There's not a whole lot we can do about this other than try to increase parallelism and caching (these are usually mutually exclusive) and use more unittests instead of full builds. We're already well passed the limit of parallelism available on Travis.

One way to increase parallelism is to switch to a CI provider other than Travis that offers more at the free tier. It's been popular recently to use Microsoft's Azure Pipelines for CI for projects that have really high demand (conda-forge, jupyterlab) and I think they offer the most for free (presumably since they are a. Microsoft and b. late to the game). Another, perhaps more tedious option, would be to spread tests over a few CI services if one service doesn't offer the parallelism we need, but that might make the maintenance burden more than it's worth.

cc @SylvainCorlay who brought up r2d tests clogging the Jupyter account travis queue.

betatim:
At EuroPython I chatted with @ogrisel about CI and he said he was happy camper with Azure Pipelines and found that (because they were late to the party) the queues are short and things run fast.

I think we should try out moving to a different provider and only if that doesn't work out consider splitting because of the additional maintenance you mentioned. Going to a different provider would also fix the pain we cause others (as long as they don't also switch ;) )

trallard:
I recently helped Carol et al. to migrate to Azure pipelines for a couple of repos to in the nteract org. 
Thus I'd bee happy to implement this in a fork of mine of repo2docker so that you can compare?

betatim:
> I'd bee happy to implement this in a fork of mine of repo2docker so that you can compare?

Could we add it to the main repo and then run things in parallel? That way we'd have the two right next to each other in each PR which makes it easy to build up a feeling for which one is faster, more stable, nicer output, $stuff. Do you need an admin to enable/disable things in the settings?

trallard:
Yes- I was suggesting implementing on a fork of mine as I would need elevated privileges to enable the Azure DevOps Github app

betatim:
@minrk @Carreau or @willingc could one of you add the https://github.com/marketplace/azure-pipelines app to this repository please? (I think that is the right app to add)

trallard:
I can work on this send the PR and on merge, this can be added/approved by @willingc (that's what we did for scrapbook - of course if Carol is happy with this)

willingc:
Will do in a bit..

betatim:
Thanks to @willingc and friends we now have a basic pipeline on Azure 🎉!

[...]

PR #787, opened 2019-09-08
Add baseline infra for Azure Pipelines

willingc:
Partially addresses #744 

- Enable Azure Pipelines
- Add minimal tests
- Azure Pipelines needed requests so adding to dev requirements

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,20 @@
+# https://docs.microsoft.com/azure/devops/pipelines/ecosystems/python
+pool:
+  vmImage: "ubuntu-16.04"
+
+steps:
+- task: UsePythonVersion@0
+  inputs:
+    versionSpec: '3.6'
+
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,21 @@
...
+
+- script: pip install -r dev-requirements.txt
+  displayName: 'Install dev requirements'
+
+- script: pip install -e .
+  displayName: 'Install r2d'
+
+# venv conda 
+- script: pytest --durations 10 --cov repo2docker -vs tests/unit
+  displayName: 'Run pytest'
```
betatim: @willingc if I wanted to add more tests like we have for travis

Screenshot 2019-09-08 at 11 37 31

then the thing to do is add new versions of these last two lines here?

willingc: Yes. This was just to get the bare tooling in before I leave. Next steps would be to matrix the tests and add the operating systems.

PR #766, opened 2019-08-23
[WIP] add hydroshare as a content provider

sblack-usu:
https://www.hydroshare.org is a data repository for Hydrologists to share water data.

I will add tests soon

PR #788, opened 2019-09-08
[MRG] Add Figshare content provider

nuest:
Also needs UI update: https://github.com/jupyterhub/binderhub/issues/938

### DIFF
--- a/repo2docker/contentproviders/figshare.py
+++ b/repo2docker/contentproviders/figshare.py
@@ -20,6 +20,7 @@ class Figshare(DoiProvider):
 
     Examples:
       - https://doi.org/10.6084/m9.figshare.9782777
+      - https://doi.org/10.6084/m9.figshare.9782777.v2
       - https://figshare.com/articles/binder-examples_requirements/9784088 (only one zipfile, no DOI)
     """
 
@@ -38,7 +39,7 @@ class Figshare(DoiProvider):
             }
         ]
 
-    url_regex = re.compile(r"(.*)/articles/([^/]+)/(\d+)(/\d)?")
+    url_regex = re.compile(r"(.*)/articles/([^/]+)/(\d+)(/)?(\d)?")
 
     def detect(self, doi, ref=None, extra_args=None):
         """Trigger this provider for things that resolve to a Figshare article"""
@@ -53,7 +54,14 @@ class Figshare(DoiProvider):
                 match = self.url_regex.match(url)
                 if match:
                     self.article_id = match.groups()[2]
-                    return {"article": self.article_id, "host": host}
+                    self.article_version = match.groups()[4]
+                    if not self.article_version:
+                        self.article_version = "1"
+                    return {
+                        "article": self.article_id,
+                        "host": host,
+                        "version": self.article_version,
+                    }
                 else:
                     return None
 
@@ -62,9 +70,13 @@ class Figshare(DoiProvider):
         article_id = spec["article"]
         host = spec["host"]
 
-        yield "Fetching Figshare article {}.\n".format(article_id)
+        yield "Fetching Figshare article {} in version {}.\n".format(
+            self.article_id, self.article_version
+        )
         req = Request(
-            "{}{}".format(host["api"], article_id),
+            "{}{}/versions/{}".format(
+                host["api"], self.article_id, self.article_version
+            ),
             headers={"accept": "application/json"},
         )
         resp = self.urlopen(req)
@@ -84,3 +96,8 @@ class Figshare(DoiProvider):
     def content_id(self):
         """The Figshare article ID"""
         return self.article_id
+
+    @property
+    def content_version(self):
+        """The Figshare article ID"""
+        return self.article_version
--- a/repo2docker/contentproviders/figshare.py
+++ b/repo2docker/contentproviders/figshare.py
@@ -94,9 +94,4 @@ class Figshare(DoiProvider):
     @property
     def content_id(self):
         """The Figshare article ID"""
-        return self.article_id
-
-    @property
-    def content_version(self):
-        """The Figshare article ID"""
-        return self.article_version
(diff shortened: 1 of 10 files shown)


repo2docker/contentproviders/zenodo.py
```diff
@@ -4,78 +4,46 @@
...
-            except HTTPError:
-                return doi
-            return resp.url
-        else:
-            # Just return what is actulally just a URL
-            return doi
+    # We need the hostname (url where records are), api url (for metadata),
+    # filepath (path to files in metadata), filename (path to filename in
+    # metadata), download (path to file download URL), and type (path to item type in metadata)
+    hosts = [
```
betatim: Could we move this into a `__init__` method and make it define `self.hosts` instead of here?

Mostly because class attributes and such are traps people eventually fall into later in a Python project when modifying it for once instance seems to modify it for all instances "as if by magic". Hence I'd prefer setting it explicitly as a instance variable.

repo2docker/contentproviders/figshare.py
```diff
@@ -0,0 +1,85 @@
...
+    """Provide contents of a Figshare article.
+    
+    See https://docs.figshare.com/#public_article for API docs.
+
+    Examples:
+      - https://doi.org/10.6084/m9.figshare.9782777
+      - https://figshare.com/articles/binder-examples_requirements/9784088 (only one zipfile, no DOI)
+    """
+
+    hosts = [
```
betatim: Same comment below about class variables, instance variables etc

repo2docker/contentproviders/figshare.py
```diff
@@ -0,0 +1,85 @@
...
+        only_one_file = len(files) == 1
+        for file_ref in files:
+            unzip = file_ref["name"].endswith(".zip") and only_one_file
+            for line in self.fetch_file(file_ref, host, output_dir, unzip):
+                yield line
+
+    @property
+    def content_id(self):
+        """The Figshare article ID"""
+        return self.article_id
```
betatim: Just to double check: Can you have versioned DOIs on figshare and if yes does a different version of a DOI have a different article ID? If not we need to include the version here somehow.

nuest:
Thanks for taking a look so quickly! I will investigate the versioning.

repo2docker/contentproviders/figshare.py
nuest: The article ID stays the same:

- https://figshare.com/articles/Binder-ready_openSenseMap_Analysis/9782777/1
- https://figshare.com/articles/Binder-ready_openSenseMap_Analysis/9782777/2 (hope you don't mind the co-authorship, needed to trigger new version)

The DOI changed from `..v1` to https://doi.org/10.6084/m9.figshare.9782777.v2

nuest: BUT the problem with that is of course that I can _not_ ignore the version, because a different DOI should also lead to a different version being downloaded!

betatim: Ok. We can return something like `"{}.{}".format(self.article_id, self.article_version)` as content_id to solve the problem (I think).

nuest: True. I have a working implementation with a separate version field though. Just want to add one more test but Something is off... Will look with fresh eyes tomorrow.

betatim: The value returned from `content_id` has to be a string as we use it in the name of the docker image that r2d creates (in case that helps with the thing that is off).

nuest:
Merging this will affect two other open PRs (#739, #766). I'd like to help the authors with updating their code or can make the updates myself.

repo2docker/contentproviders/figshare.py
```diff
@@ -84,3 +96,8 @@ def fetch(self, spec, output_dir, yield_output=False):
     def content_id(self):
         """The Figshare article ID"""
         return self.article_id
+
+    @property
+    def content_version(self):
```
betatim: The `content_id` is part of the contract between the rest of repo2docker and content providers as the (opaque to the user) value that can be used to identify if the content is the same or not. So we should include the version there instead of creating a new attribute. Put a different way: there can be a new attribute in this content provider if it wants to organise itself that way however for talking to the rest of r2d the `content_id` is what counts.

nuest:
@betatim Shall I spend some time the code coverage? I can surely improve the ones submitted now (`figshare.py` is far away from the targeted 80%)

pdurbin:
> Merging this will affect two other open PRs (#739, #766). I'd like to help the authors with updating their code or can make the updates myself.

Thank you, @nuest!! Any help getting those other pull requests merged is much appreciated!! :tada:

repo2docker/contentproviders/figshare.py
nuest: I see - so the `content_id` is more outward facing, and internally I can keep the `article_version`, right?

nuest: Fixed now

nuest:
@betatim I think this is ready now, can you take another look, please?

betatim:
Sure thing! We also change the title from WIP to MRG when it is ready for reviewing (and then maybe merge). MRG is a bit of a misnomer really it is more like "review please and then merge if you are happy" but yeah.

repo2docker/contentproviders/figshare.py
```diff
@@ -0,0 +1,97 @@
...
+        only_one_file = len(files) == 1
+        for file_ref in files:
+            unzip = file_ref["name"].endswith(".zip") and only_one_file
+            for line in self.fetch_file(file_ref, host, output_dir, unzip):
+                yield line
+
+    @property
+    def content_id(self):
+        """The Figshare article ID"""
+        return "{}.v{}".format(self.article_id, self.article_version)
```
betatim: 🎉

betatim:
Merged! Could you make a PR that adds a line to the change log for this?

PR #789, opened 2019-09-08
use getpass.getuser instead of os.getlogin

minrk:
os.getlogin apparently doesn't work everywhere (can raise OSError, inappropriate ioctl for device)

and is failing on azure pipelines tests


betatim:
We need to add a step to the azure pipeline that sets up Git credentials as the tests that are now failing are around creating temporary commits.

PR #790, opened 2019-09-08
[MRG] Add more test scenarios to Azure pipeline

betatim:
Add more of our test scenarios to the Azure setup.


willingc:
Nicely done @betatim. Do you want to remove the WIP and merge?

betatim:
Merged! A thing to look up is if we can specify a priority for the jobs so that for example the "unit" and "lint" jobs start amongst the first. They should complete quickly and provide feedback to users. Something to find out from the docs :D

PR #793, opened 2019-09-12
[MRG] Make sure the exit status gets passed along in Azure pipelines

betatim:
See if @manics idea from https://github.com/jupyter/repo2docker/issues/744#issuecomment-530867092 is it. Don't merge for now as it contains a test that always fails.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -62,7 +62,7 @@ steps:
 - script: pip install -e .
   displayName: 'Install repo2docker'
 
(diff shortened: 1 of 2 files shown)


betatim:
Let's see if this produces "the right failures" and then I'll remove the always fail test.

betatim:
Looks like it all works as expected now. I remove the lint step from Azure. We should add it back before we finally remove travis.

We need to figure out what a nice way to change the test command when the "job type" is "lint".

betatim:
cc @minrk

minrk:
Had to restart some Azure tests due to network errors. Not sure what's up there, but it all succeeded eventually.

PR #794, opened 2019-09-13
[MRG] Update nteract-on-jupyter to 2.1.3

betatim:
Updated the version and refroze.


betatim:
I think this should be uncontroversial so merging.

PR #796, opened 2019-09-18
[MRG] Add docs for Figshare and Dataverse; update Dataverse installations file formatting

nuest:
This provides the docs requested in https://github.com/jupyter/repo2docker/pull/739#issuecomment-532534037 and also formats and sorts the entries in the JSON file with Dataverse installations. The latter has the advantage that we'll be able to track changes in the installation list without relying on the structure on the server, only changing the lines that did change, in a human-friendly way (at the cost of some whitespace).

@pdurbin @Xarthisius


pdurbin:
@nuest these docs look great. I'm sort of wondering how the dataverse.json file gets update over time but I'm glad to see it's up to date with all 48 Dataverse installations represented!

On a related note, can I ask you to please take a look at https://github.com/jupyterhub/binderhub/issues/900 ?

PR #800, opened 2019-09-25
[MRG] add Hydroshare as a content provider

sblack-usu:
https://www.hydroshare.org is a data repository for Hydrologists to share water data.


sblack-usu:
@nuest Tests around the Hydroshare content provider have been written.  I also updated it to use the DoiProvider.

repo2docker/contentproviders/hydroshare.py
```diff
@@ -0,0 +1,79 @@
...
+        files = os.listdir(contents_dir)
+        for f in files:
+            shutil.move(os.path.join(contents_dir, f), output_dir)
+        yield "Finished, cleaning up.\n"
+        shutil.rmtree("temp")
+
+    @property
+    def content_id(self):
+        """The HydroShare resource ID"""
+        return self.resource_id
```
sblack-usu: Hydroshare resources can change up until they are published.  Should this include the version of the resource as well?

repo2docker/contentproviders/hydroshare.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # bag downloads are prepared on demand and may need some time
+        conn = self.urlopen(bag_url)
+        total_wait_time = 0
+        while conn.getcode() == 200 and conn.info().get_content_type() != "application/zip":
+            wait_time = 10
+            total_wait_time += wait_time
+            if total_wait_time > timeout:
+                yield "Bag taking too long to prepare, exiting now, try again later."
+                return
```
sblack-usu: Should I raise an exception to signal a failure with the fetch?

jhamman:
@sblack-usu - do you have a public hydroshare repository that we can try this out with?

sblack-usu:
@jhamman any resource should work for pulling the contents and placing the files in jupyter.

sblack-usu:
travis-ci appears to have failed when running the lint.  Anyway I can access the reformatted file to get it in a commit?

sblack-usu:
binderhub changes:
https://github.com/jupyterhub/binderhub/pull/967

betatim:
To fix the formatting checkout https://repo2docker.readthedocs.io/en/latest/contributing/contributing.html#code-formatting

repo2docker/contentproviders/hydroshare.py
```diff
@@ -0,0 +1,92 @@
...
+from .doi import DoiProvider
+from .base import ContentProviderException
+from ..utils import normalize_doi, is_doi
+
+
+class Hydroshare(DoiProvider):
+    """Provide contents of a Hydroshare resource."""
+
+    def detect(self, doi, ref=None, extra_args=None):
+        """Trigger this provider for things that resolve to a Zenodo/Invenio record"""
```
betatim: Needs to be updated to mention hydroshare

betatim:
I took a look at the code to start reviewing it but there are quite a few things that made me think "ah this is still a rough draft/work in progress". When you are ready set the title to start with `[MRG]` instead of `[WIP]` and ping me :)

jhamman:
@scottyhq - it looks like you have [two](https://travis-ci.org/jupyter/repo2docker/jobs/589639536#L776-L825) CI failures. Any idea what's up here?

You also have a linter failure. I think you can correct this by running `black` on the source files you've added.

repo2docker/contentproviders/hydroshare.py
sblack-usu: updated

sblack-usu:
@jhamman - thanks for pointing out black.  I've got the files reformatted.  As for the tests that fail, I'm unsure why they're failing here.  When I run them locally, all 5 new tests pass.  The part that fails is the version piece of the content_id does not match the expected integer which is an epoch string representation of the last modified date on the hydroshare resource.  It shouldn't evaluate differently on my machine... maybe it has something to do with timezones, now that I'm considering it.  I'll look into it.

sblack-usu:
@betatim - Got the tests sorted out and I believe it's ready for a thorough review.  Thanks for your patience.

repo2docker/contentproviders/hydroshare.py
```diff
@@ -0,0 +1,110 @@
...
+        return urlretrieve(bag_url)
+
+    def fetch(self, spec, output_dir, yield_output=False, timeout=120):
+        """Fetch and unpack a Hydroshare resource"""
+        resource_id = spec["resource"]
+        host = spec["host"]
+
+        yield "Fetching HydroShare Resource {}.\n".format(resource_id)
+
+        bag_url = "{}{}".format(host["django_irods"], resource_id)
```
nuest: From a user perspective, I would like to see the full URL also printed so I know that the download has started, maybe `yield "Downloading {}.}n".format(bag_url)`.

The DOI from the tests actually returns a 2.7 GB large repo... I know it is not actually downloaded during the tests... but maybe we can (also) have a smaller one there for tests?

nuest:
LGTM, I ran the contentprovider tests locally and 

```
repo2docker 10.4211/hs.e5ff94b3e9a442faa32618425cad36f6
```

gave me a JupyterHub with the expected files.

repo2docker/contentproviders/hydroshare.py
sblack-usu: @nuest - Thanks for taking a look.  Could you clarify what you'd like to test with the smaller resource?  Do you just want me to add a comment pointing to a smaller resource to test locally with?  I'm not sure what you're looking for.

nuest: Yes, something smaller for manual testing. 

_BUT maybe also for a test_: I'm not sure what the state of the other DOI-based providers is, but it could be worth having at least one "integration test" that actually tries to create a Binder from a DOI. @betatim What's your take? Should we open a new issue on that?

In any case, that should not hinder this PR. I might be the only person you looks into the source code for a example DOI anyway.

sblack-usu: Here's a much smaller DOI.  https://doi.org/10.4211/hs.142c59757ed54de1816777828c9716e7

sblack-usu:
@nuest @betatim - Anything else you need from me to get this merged?

nuest:
Not from my side. @betatim is probably just busy.

@sblack-usu Is there any deadline for you (a project meeting soon) or so?

repo2docker/contentproviders/hydroshare.py
betatim: For "integration" tests we usually make a copy of https://github.com/binder-examples/requirements and put that into what ever "storage" we want to test. Generally we want to test as much as possible with unit tests though or tests that fake stuff because the repo2docker tests as a whole take long enough to execute that glaciers melt while we do that.

So I'd say we don't need to check that r2d will successfully build a container from a known good content (like binder-examples/requirements). What we could test "integration test" style is actually downloading a small file from hydroshare to catch when our fake download tests are outdated.

repo2docker/contentproviders/hydroshare.py
```diff
@@ -0,0 +1,110 @@
...
+                item for item in json_response["dates"] if item["type"] == "modified"
+            )["start_date"]
+            # Hydroshare timestamp always returns the same timezone, so strip it
+            date = date.split(".")[0]
+            parsed_date = datetime.strptime(date, "%Y-%m-%dT%H:%M:%S")
+            epoch = parsed_date.replace(tzinfo=timezone(timedelta(0))).timestamp()
+            # truncate the timestamp
+            return str(int(epoch))
+
+        url = self.doi2url(doi)
```
betatim: The indentation of this and the following code block is wild. Should it be indented by one more level? If it is here on purpose then we need to move it as class level code between two methods is not something we do ;)

betatim: Ah, it turns out this is actually the right level of indentation but I got distracted by `def fetch_version`'s indentation. is there a reason to have that function inline like that? We could just make it a method. But this is nitpicking

tests/unit/contentproviders/test_hydroshare.py
```diff
@@ -0,0 +1,212 @@
...
+        def read():
+            return '{"dates": [{"type": "modified", "start_date": "2019-09-25T16:09:17.006152Z"}]}'
+
+        fake_urlopen.return_value.read = read
+        # valid Hydroshare DOIs trigger this content provider
+        assert Hydroshare().detect(test_input[0]) == expected
+        assert Hydroshare().detect(test_input[1]) == expected
+        assert Hydroshare().detect(test_input[2]) == expected
+        # only two of the three calls above have to resolve a DOI
+        assert fake_urlopen.call_count == 5
```
betatim: Why do we expect five calls here? From the comment above I was guessing that this should be two.

tests/unit/contentproviders/test_hydroshare.py
```diff
@@ -0,0 +1,212 @@
...
+            },
+            "resource": "b8f6eae9d89241cf8b5904033460af61",
+            "version": "1569427757",
+        },
+    )
+]
+
+
+@pytest.mark.parametrize("test_input,expected", test_hosts)
+def test_detect_hydroshare(test_input, expected):
```
betatim: Could you change this to

```suggestion
def test_detect_hydroshare(test_input):
```

and define `expected` inside the test? I think right now this test only gets called once with these arguments (based on your use of `test_input[0]` and `test_input[1]` etc). However a parametrised test in my mind would run once for each input (once with `test_input[0]` as argument and the expected value for that input, then for `test_input[1]` with expected value for that input, etc).

I think part of my confusion with "what should `fake_urlopen.call_count` be?" comes from the fact that I thought we run this test three times, once for each input. But in reality we run it once.

tests/unit/contentproviders/test_hydroshare.py
```diff
@@ -0,0 +1,212 @@
...
+                        "http://www.hydroshare.org/resource/",
+                    ],
+                    "django_irods": "https://www.hydroshare.org/django_irods/download/bags/",
+                },
+                "resource": "123456789",
+            }
+            with TemporaryDirectory() as d:
+                output = []
+                try:
+                    for l in hydro.fetch(spec, d, timeout=0):
```
betatim: Instead of `try-except`ing yourself to find out if an exception was raised can you use `pytest.raises` as context manager here?

```python
import pytest


def myfunc():
    raise ValueError("Exception 123 raised")


def test_match():
    with pytest.raises(ValueError, match=r".* 123 .*"):
        myfunc()
```

tests/unit/contentproviders/test_hydroshare.py
sblack-usu: It's 5 because each of the DOI calls call urlopen twice (once to fetch the version and another to resolve the DOI to the Hydroshare url).  The HydroShare url only makes the call to fetch version.  I can see how this is unclear, I'll update comments to clarify.

betatim:
I don't think the build failure is related to the changes here. Looks like it is related to the legacy build pack.

sblack-usu:
@betatim - Is there anything I can do to help get this build failure sorted out?  I believe I addressed all of your code review comments.

betatim:
It would be great (if you have time) to [look at the build failure](https://travis-ci.org/jupyter/repo2docker/jobs/622090584?utm_medium=notification&utm_source=github_status) and try to understand why the build fails. Fixes would be even more awesome (in a new PR).

It seems like something recently changed in how conda is installed(??). I don't think our legacy buildpack instructions have changed in a long, long time because its is a legacy thing :)

sblack-usu:
I'm afraid it would be a large time sink for me as I'm not familiar with the build pack.

choldgraf:
I just opened up https://github.com/jupyter/repo2docker/issues/828 in case we'd like to punt the Travis buildpack question for a different issue, as it doesn't seem to be related to this PR

betatim:
One of the guidelines for repo2docker development is to not merge if travis is red. As this feature isn't critical/time sensitive I would prefer we wait and work on fixing the legacy buildpack before merging this.

Otherwise we set a precedent and have a failing `master` which generally leads to an increase in exceptions for failing builds when merging PRs. Starting us down a spiral from which is gets ever more tricky to recover. Especially because no one wants to invest the (large) amount of time it will take to fix the breakage. It is issues like this that make building software hard :) (yay)

manics:
https://github.com/jupyter/repo2docker/pull/829 should've fixed the failing test. Can someone close and re-open this PR to rerun all tests?

choldgraf:
just did, let's see what happens :-)

choldgraf:
it passed! are we ready to :shipit: now that all the tests are happy? I believe that this should also be paired with a merge in BinderHub, yeah?

sblack-usu:
yeah, looks like there are some merge conflicts in the binderhub PR.  I'll work on resolving those today and let you decide whether you want to merge this before I get the conflicts straightened out or not.

sblack-usu:
Conflicts resolved and I tested a deploy of binderhub with these changes on google cloud.  Ready to be merged along with the binderhub changes.

choldgraf:
fantastic! I will plan to merge this over the weekend unless anybody has objections. It sounds like we've got all comments addressed and all the tests are happy. @betatim you had some comments earlier on, do you think we are good to 🚢 ?

choldgraf:
ok let's see how this goes

PR #803, opened 2019-10-02
[MRG] Upadte RStudio 1.1 - > 1.2, minor update to RShiny

pablobernabeu:
Based on [this Discourse discussion with Tim](https://discourse.jupyter.org/t/widget-failing-in-shiny-app), regarding an error I have on a widget in a Shiny app, I made two small edits in r.py:

1. Updated the version of RStudio Server to the latest, stable release available on the official RStudio website (URL commented in code). This involved two code lines, namely, the URL and the MD5.

2. Updated the version of Shiny Server, too, as it could also be relevant to the present error. This involved the same types of edits as above.

Thank you!

### DIFF
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -201,13 +201,12 @@ class RBuildPack(PythonBuildPack):
         We set the snapshot date used to install R libraries from based on the
         contents of runtime.txt.
         """
-        # Latest version as of October 2, 2019
-        rstudio_url = "https://download2.rstudio.org/rstudio-server-1.2.5001-amd64.deb"
+        # Via https://rstudio.com/products/rstudio/download-server/debian-ubuntu/
+        rstudio_url = "https://download2.rstudio.org/server/trusty/amd64/rstudio-server-1.2.5001-amd64.deb"
         # This is MD5, because that is what RStudio download page provides!
         rstudio_checksum = "d77039693dd09c3ca48038aa74cd921a"
 
         # Via https://www.rstudio.com/products/shiny/download-server/
-        # Latest version as of October 2, 2019
         shiny_url = "https://download3.rstudio.org/ubuntu-14.04/x86_64/shiny-server-1.5.12.933-amd64.deb"
         shiny_checksum = "9aeef6613e7f58f21c97a4600921340e"
 
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -202,7 +202,7 @@ class RBuildPack(PythonBuildPack):
         contents of runtime.txt.
         """
         # Via https://rstudio.com/products/rstudio/download-server/debian-ubuntu/
-        rstudio_url = "https://download2.rstudio.org/server/trusty/amd64/rstudio-server-1.2.5001-amd64.deb"
+        rstudio_url = "https://download2.rstudio.org/server/bionic/amd64/rstudio-server-1.2.5001-amd64.deb"
         # This is MD5, because that is what RStudio download page provides!
         rstudio_checksum = "d77039693dd09c3ca48038aa74cd921a"
 
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -248,13 +248,14 @@ class RBuildPack(PythonBuildPack):
                     "root",
                     r"""
                     apt-get update && \
-                    apt-get install --yes r-base={} && \
-                    apt-get install --yes r-base-dev={} && \
+                    apt-get install --yes r-base={R_version} \
+                         r-base-dev={R_version} \
+                         libclang-dev && \
                     apt-get -qq purge && \
                     apt-get -qq clean && \
                     rm -rf /var/lib/apt/lists/*
                     """.format(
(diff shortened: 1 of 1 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/widget-failing-in-shiny-app/2226/6

repo2docker/buildpacks/r.py
```diff
@@ -201,13 +201,14 @@ def get_build_scripts(self):
         We set the snapshot date used to install R libraries from based on the
         contents of runtime.txt.
         """
-        rstudio_url = "https://download2.rstudio.org/rstudio-server-1.1.419-amd64.deb"
+        # Via https://rstudio.com/products/rstudio/download-server/debian-ubuntu/
+        rstudio_url = "https://download2.rstudio.org/server/trusty/amd64/rstudio-server-1.2.5001-amd64.deb"
```
betatim:
```suggestion
        rstudio_url = "https://download2.rstudio.org/server/bionic/amd64/rstudio-server-1.2.5001-amd64.deb"
```

The repo2docker base image is based on Ubuntu "Bionic" so I think we should fetch this instead

repo2docker/buildpacks/r.py
```diff
@@ -201,13 +201,14 @@ def get_build_scripts(self):
...
         # This is MD5, because that is what RStudio download page provides!
-        rstudio_checksum = "24cd11f0405d8372b4168fc9956e0386"
+        rstudio_checksum = "d77039693dd09c3ca48038aa74cd921a"
```
betatim:
```suggestion
        rstudio_checksum = "d33881b9ab786c09556c410e7dc477de"
```

betatim:
Looks good. I left two suggestions that you should be able to apply (if you agree with them) by clicking on them in the web UI.

Otherwise this looks good to me.

pablobernabeu:
> Looks good. I left two suggestions that you should be able to apply (if you agree with them) by clicking on them in the web UI.
> 
> Otherwise this looks good to me. Thanks a lot for helping out!

Thank you!

pablobernabeu:
The check 'continuous-integration/travis-ci/pr' failed. 

Just as a guess, I noticed that the Bionic RStudio Server download used Ubuntu 18, whereas the Shiny Server download used Ubuntu 14.

betatim:
I think the problem is with the install command. We have one that is different from what is suggested on the webpage. Not sure if it is different on purpose or only because the instructions used to be different.

pablobernabeu:
I see. For instance, a `gdebi` package isn't used in the r.py script. The [RStudio info](https://rstudio.com/products/shiny/download-server/ubuntu/) says:

> Once you’ve installed R and the Shiny package, execute the following commands in a terminal window to install gdebi (which is used to install Shiny Server and all of its dependencies) and Shiny Server.

I wish I could try this out, but I don't know how to.

I will try updating RStudio in a first pull and, if it works, Shiny Server later.

repo2docker/buildpacks/r.py
```diff
@@ -243,6 +248,7 @@ def get_build_scripts(self):
                     r"""
                     apt-get update && \
                     apt-get install --yes r-base={} && \
+                    apt-get install --yes r-base-dev={} && \
```
betatim: We also need `libclang-dev` to add it I'd replace these lines with:
```
                    apt-get install --yes r-base={R_version} \
                         r-base-dev={R_version} \
                         libclang-dev && \
```
and then edit the
```
                    """.format(
                        self.r_version
                    ),
```
a few lines later to read:
```
                    """.format(
                        R_version=self.r_version
                    ),
```

repo2docker/buildpacks/r.py
```diff
@@ -178,6 +180,7 @@ def get_packages(self):
         # install from a different PPA
         if V(self.r_version) < V("3.5"):
             packages.append("r-base")
+            packages.append("r-base-dev")
```
betatim:
```suggestion
            packages.append("r-base-dev")
            packages.append("libclang-dev")
```

seems like we also need `libclang-dev` now

repo2docker/buildpacks/r.py
pablobernabeu: thank you!

betatim:
How have you been making your changes? The test that is failing is one that checks the code is formatted according to the style guide we follow. If you have a local checkout and Python installed the easiest way to make the check happy is to `pip install black` and then run `black .` in the directory where you checked out this repo. Then commit any changes it suggests.

If you don't have a local copy (or want to go through the process of setting one up) I'd suggest that I'll fix the formatting, add the commit and then merge this PR.

Which one would you prefer?

pablobernabeu:
I've done the changes right here on Github, in my fork. I don't know any Python, unfortunately, so I would definitely prefer the second approach.

betatim:
Good to hear your issue got fixed and thanks for helping to update RStudio!

The formatting turned out to be something about white-space, so just a minor thing.

Going to merge once the robots have done their checking.

betatim:
Congratulations on your first merged PR to repo2docker!

pablobernabeu:
Thank you very much! :blush:

Issue #682, opened 2019-05-20
Provide 'podman' as a build backend


yuvipanda:
### Proposed change

Currently, only docker is available to build images. We should:

1. Abstract the interface that builds a Dockerfile
2. Implement docker (via docker-py) as the default implementation
3. Allow [podman](https://podman.io/) to be the second implementation, so we can make sure our design isn't docker specific.

podman supports building Dockerfiles directly, so this should be a fairly minimal change.

### Alternative options

We do (1) and (2) in core, but leave (3) to be a plugin.

### Who would use this feature?

Anyone who doesn't want to use the docker daemon. There are good technical and social reasons for this.

### How much effort will adding it take?

Abstracting the current docker calls behind an interface shouldn't be too much work. Podman is going to be interesting. It seems pretty production ready, but treats running on ubuntu as an extreme second class citizen - I haven't been able to get it to run yet. So that might be more challenging - making the case for keeping the podman build bits in an extension.

### Who can do this work?

You'd need some familiarity with the docker python client and Dockerfile.

betatim:
Are there other backends we should also take a look at (not implement though) to inform how we design the abstraction layer?

👍  on attempting to work towards `repo2docker` being able to use things that aren't docker!

yuvipanda:
To begin with I think we should only support things that understand Dockerfiles. Off the top of my head, that is BuildKit, Podman and possibly buildah

yuvipanda:
See also #560

PR #806, opened 2019-10-23
WIP: Podman backend

manics:
I had a go at implementing a Podman backend (Docker has mysteriously broken itself on my Fedora install)

Main issues:
- There may be some bugs with hard-links (affects conda, needs to be fixed in the upstream podman project, but coming up with a minimal example for a bug report is tricky)
- Travis VM is too old to run podman in user-mode (requires the fuse-overlayfs driver). The options are to use VFS (works but extremely inefficient) or run as root. On a modern OS such as Fedora 30 running Podman as a normal user with `fuse-overlayfs` should work.
- The [Podman Python client library](https://github.com/containers/python-podman) uses the [varlink interface](https://podman.io/blogs/2019/01/16/podman-varlink.html) which adds unnecessary complication since one of the aims of Podman is to run with minimal privileges and no daemon so I've written a wrapper library that executes `podman`
- `cache_from` isn't implemented by podman

Todo:
- [ ] Implement `push_image`
- [ ] Implement volumes for `podman run`
- [ ] Get rid of print/debug statements
- [ ] Failing test: `REPO_TYPE=stencila-py ENGINE=podman`
- [ ] Check whether `Repo2Docker.find_image` works
- [ ] Refactor behind an Engine class

Closes https://github.com/jupyter/repo2docker/issues/682

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ services:
 - docker
 python:
 - 3.6
+before_install:
+# https://github.com/adelton/freeipa-container/blob/podman/.travis.yml
+- sudo sh -c "add-apt-repository -y ppa:projectatomic/ppa && apt-get update -y && apt-get install -y podman"
 install:
 # Make a wheel and install it to test to catch possible
 # issues with releases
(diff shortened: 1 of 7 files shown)


betatim:
Neat! Have you thought about having a `Engine` class with an interface behind which we hide each engine? Was thinking that for two engines we can have `if` statements to select what to do but for more it would be nice to have something like the contents providers where repo2docker doesn't have to care which one is doing the work.

manics:
Definitely! This started as a "will it even work" mini project, now that it looks feasible I'll work on the design.

yuvipanda:
This is awesome and I love it! <3

manics:
@yuvipanda Thanks! I haven't had much time to work on this recently, but I've upgraded to Fedora 31 which uses cgroups2 which Docker doesn't (yet) support... so I've only got Podman now.

yuvipanda:
Ah, another casuality in the Leonard Pottering / Solomon Hykes war! Hope podman treats you well!

manics:
This has been replaced by:
- https://github.com/jupyter/repo2docker/pull/848 which adds a `ContainerEngine` interface
- https://github.com/manics/repo2docker-podman which implements the `ContainerEngine` interface using podman

meeseeksmachine:

https://discourse.jupyter.org/t/binderhub-for-hpc/143/18

PR #809, opened 2019-11-05
[MRG] Fix submodule check out

davidbrochart:
This should fix https://github.com/jupyterhub/binderhub/issues/997.
Should I add a test?

### DIFF
--- a/repo2docker/contentproviders/git.py
+++ b/repo2docker/contentproviders/git.py
@@ -34,7 +34,7 @@ class Git(ContentProvider):
                 yield line
 
         except subprocess.CalledProcessError as e:
-            msg = "Failed to clone repository from {repo}.".format(repo=repo)
+            msg = "Failed to clone repository from {repo}".format(repo=repo)
             if ref is not None:
                 msg += " (ref {ref})".format(ref=ref)
             msg += "."
--- a/repo2docker/contentproviders/git.py
+++ b/repo2docker/contentproviders/git.py
@@ -50,7 +50,7 @@ class Git(ContentProvider):
                 raise ValueError("Failed to check out ref {}".format(ref))
             # check out ref as it has not been done yet
             for line in execute_cmd(
-                ["git", "checkout", ref], cwd=output_dir, capture=yield_output
+                ["git", "checkout", hash], cwd=output_dir, capture=yield_output
             ):
                 yield line
             # If the hash is resolved above, we should be able to reset to it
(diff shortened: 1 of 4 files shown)


betatim:
We have some tests for the git content provider [here](https://github.com/jupyter/repo2docker/blob/master/tests/unit/contentproviders/test_git.py). It would be good to add a test that makes sure we don't checkout submodules from `master` when using a branch.

davidbrochart:
We can actually clone and check out a specific branch at the same time:
```bash
git clone -b <branch> <remote_repo>
```

manics:
@davidanthoff You can clone branches but not commits https://github.com/jupyterhub/binderhub/issues/997#issuecomment-549453363

betatim:
I think it was for the other David :)

manics:
Sorry, wrong David! No idea why that autocomplete came up, GitHub is normally pretty good about suggesting the most relevant user.

davidbrochart:
The errors in the CI don't seem to be related to my changes.

davidbrochart:
Do you think it can be merged?

betatim:
I'll take a look and the CI jobs have been restarted. (I was waiting for the title to change from WIP to MRG and hence had it on my low priority list.)

repo2docker/contentproviders/git.py
```diff
@@ -21,15 +21,23 @@ def fetch(self, spec, output_dir, yield_output=False):
...
+            else:
+                # don't check out HEAD, the given ref will be checked out later
+                # this prevents HEAD's submodules to be cloned if ref doesn't have them
+                cmd.extend(["--no-checkout"])
             cmd.extend([repo, output_dir])
             for line in execute_cmd(cmd, capture=yield_output):
                 yield line
 
         except subprocess.CalledProcessError as e:
             msg = "Failed to clone repository from {repo}.".format(repo=repo)
```
betatim:
```suggestion
            msg = "Failed to clone repository from {repo}".format(repo=repo)
```

betatim:
The test that fails is passing `origin/b8259dac9eb` as the ref to the checkout command which doesn't work any more so we need to improve on that.

betatim:
This is the first test in https://github.com/jupyter/repo2docker/blob/master/tests/external/reproductions.repos.yaml

You should be able to run it locally with `pytest -svx tests/external/reproductions.repos.yaml::"LIGO Gravitational Waves"` (though I think because the two tests have the same name it might select the "other one" so we need to fix that up as well)

betatim:
Merged! Should get deployed in the next few hours

davidbrochart:
Thanks!

PR #810, opened 2019-11-07
[MRG] Remove explicit checkout when ref given

betatim:
This is a follow up to #809 to see if we can remove the explicit checkout command which causes big scary warnings about "detached head" mode.


betatim:
@choldgraf WDYT?

betatim:
Or maybe @manics has a thought on this.

PR #822, opened 2019-12-01
Add support for Julia 1.0.5 and 1.3.0

davidanthoff:


betatim:
Thanks for keeping us updated!

Is there anything we should think of/keep in mind when people start using 1.3.0?

davidanthoff:
No, I can’t think of anything that would be relevant for binder. The new release has lots of great new features, but nothing that would affect how it looks from binder’s point of view, I think.

NHDaly:
🚀 ! Thanks David! :)

PR #824, opened 2019-12-04
adding instructions for contentprovider extension

choldgraf:
This does two things:

1. Adds some short instructions for extending repo2docker to a new ContentProvider
2. Updates some of the section titles in this section in order to make the sidebar easier to parse (having the word "repo2docker" in each page title seemed redundant and cluttered the sidebar)

This is related to https://github.com/jupyterhub/binderhub/pull/1017. Eventually the two docs should probably reference one another, as adding a new RepoProvider in BinderHub is often paired with adding a new ContentProvider in repo2docker. This also makes me wonder why it's called "RepoProvider" in one and "ContentProvider" in another...

### DIFF
--- a/docs/source/architecture.md
+++ b/docs/source/architecture.md
@@ -1,7 +1,11 @@
-# Architecture of repo2docker
+# Architecture
 
 This is a living document talking about the architecture of repo2docker
 from various perspectives.
+
+```eval_rst
+.. _buildpacks:
+```
 
 ## Buildpacks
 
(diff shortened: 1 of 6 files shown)


docs/source/architecture.md
```diff
@@ -1,8 +1,12 @@
-# Architecture of repo2docker
+# Architecture
 
 This is a living document talking about the architecture of repo2docker
 from various perspectives.
 
+```eval_rst
+.. _buildpacks:
+```
```
betatim: What does this do? Is this a way to smuggle RST into an otherwise MD file?

betatim:
Thanks for getting our documentation into shape. Feels like a "new years resolution to get stuff in order" spirit and it isn't even 2020 yet!

PR #829, opened 2019-12-11
[MRG] Deprecate legacy buildpack

### LINKED ISSUE #828: Legacy buildpack tests are failing in travis
In recent pull requests, we discovered that there are tests failing for the legacy build packs. For example, see this travis output:

https://travis-ci.org/jupyter/repo2docker/jobs/622090584?utm_medium=notification&utm_source=github_status

We should figure out why this is happening.

Alternatively, perhaps we can consider deprecating legacy repositories? I'm not sure what's different, but if this is purely a matter of allowing "old-style binder repositories" to work, then maybe enough time has passed that we can have a 1- or 2-month deprecation warning somewhere, and then remove those buildpacks?

betatim:
An attempt at fixing the buildpack problem. 

Fixes #828


choldgraf:
looks like the same timeout issues happened in travis too :-/

betatim:
At the moment I'm just guessing and confused :-/. Will post on the conda repository to see if they have some insight. One thing I saw is that they recently released a new version of conda, approximately around the time our builds started breaking. So maybe some feature we are relying on finally got deprecated/service turned off together with the release.

choldgraf:
agree this is super weird :-/ none of the other conda builds are unhappy, so I wonder if it something to do with andrewosh's (very outdated) base docker image, which is gonna be hard to debug

betatim:
I spent some time working on this and trying to implement the suggestion from the conda issue but haven't been getting anywhere.

What do people think of deprecating the legacy build pack? I'd propose to print an error message when people try to use it that tells them that it is no longer supported.

manics:
Sounds reasonable to me. It's been a month and no-one's complained about it being broken (compare with e.g. the broken R builds where there was a steady stream of comments).

choldgraf:
I'm also +1 on depreciating it. It feels like the legacy build pack is always going to be a challenge given that it relies on a totally unmaintained and opaque docker image, and it should be straightforward for anyone to upgrade their repo.

Note that if we depreciate it, then I believe this will only affect new builds right? Ore existing ones should still work, so we could make a little post telling folks what to do if they use a legacy pack

betatim:
I'll update this PR with a removal

choldgraf:
@betatim want me to write up a blog post or something?

betatim:
That would be cool @choldgraf !

betatim:
This now deprecates the legacy buildpack. Still needs an updated test and once there is a discourse post I'd insert a link to it into the error message.

betatim:
This is now ready for reviewing @choldgraf.

The legacy buildpack and tests of it have been removed. I added a few tests to check that the legacy buildpack raises an exception when it is used.

repo2docker/buildpacks/legacy/__init__.py
```diff
@@ -5,130 +5,34 @@
...
+        """Check if current repo should be built with the Legacy BuildPack."""
+        log = logging.getLogger("repo2docker")
         try:
             with open("Dockerfile", "r") as f:
                 for line in f:
                     if line.startswith("FROM"):
                         if "andrewosh/binder-base" in line.split("#")[0].lower():
-                            return True
+                            log.error(
+                                "The legacy buildpack was deprecated in January 2020."
```
manics:
```suggestion
                                "The legacy buildpack was removed in January 2020."
```
I've always thought of "deprecated" as meaning "no longer recommended and has limited support, will be removed in future"

betatim: Good point.

choldgraf:
This all LGTM...I am a big fan of PRs where there is more red than green :-) I'll plan on merging this in ~24h unless somebody else suggests we do otherwise!

choldgraf:
wooo! the legacy dockerfile is no more! :-)

PR #831, opened 2020-01-01
Add support for Julia 1.3.1

davidanthoff:
Always more versions :) This is a bug-fix release, so should be entirely harmless.


betatim:
Merging this despite the broken test in the legacy buildpack.

PR #832, opened 2020-01-27
[WIP] upgrade jupyterhub to 1.1.0

bitnik:
JupyterHub 1.0.0 --> 1.1.0


minrk:
The tests are failing I believe due to an update in nodejs version (v11 -> v13) in the conda-forge package. The node10 test, at least, asserts that the conda-forge nodejs version does not change, but at the same time we don't enforce that it stays the same. We can fix that by adding `nodejs=12` to our frozen environment.yml

betatim:
In the base buildpack we explicitly install a node v10 and the tests were added when we added that explicit install.

Can we drop the explicit install?

minrk:
I opened #839 to include this and update to using jupyterhub-base, which doesn't depend on nodejs, resolving the longstanding reason we used to install jupyterhub with pip instead of conda (#728)

bitnik:
thanks @minrk ! sorry that I couldn't reply on time. I am unfortunately sick since last week and can do and be online very little. I am closing this PR.

PR #836, opened 2020-01-29
try to exit on test failure on azure

minrk:
failing tests are being reported as success on azure, I assume because the last command (popd) is not exiting with a nonzero status code.


minrk:
oops, just noticed this is a duplicate of #793

PR #838, opened 2020-01-29
Start RStudio if R is installed via conda

### LINKED ISSUE #837: Start RStudio when conda environment contains R packages
### Proposed change

If the user provides an `environment.yml` file that contains an R package, we should also start RStudio via `jupyter_server_proxy`. For the start, we could determine the usage of R by checking if any of the dependencies in the `environment.yml` contains an `r-` prefix.

This would be used by people that install their R packages via `conda` as well as could it be an [alternative to MRAN](https://github.com/jupyter/repo2docker/issues/773) to get pinned versions for R packages.

This needs a bit of refactoring to expose the install commands for RStudio separately from the `RBuildPack`. Otherwise, the work will be limited to testing the setup.

### Who can do this work?

I happy to implement this incl. an example for binder.

xhochy:
Fixes #837


xhochy:
This is my current WIP state, I will have to look on how to write tests for this.

tests/conda/r/verify
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
minrk: This verify file needs to be executable for the test to run

minrk:
Great!

PR #839, opened 2020-01-29
[MRG] update base environment

minrk:
- jupyterhub 1.1 (includes #832)
- notebook 6
- nbconvert 5.6
- jupyterlab 1.2
- widgets 7.5

switch to new jupyterhub-base package which doesn't depend on unused configurable-http-proxy and node.js, and should reduce confusion about two nodes being installed. I think this fixes recent CI failures associated with a newer release of node on conda-forge.

### DIFF
--- a/repo2docker/buildpacks/conda/environment.py-3.6.yml
+++ b/repo2docker/buildpacks/conda/environment.py-3.6.yml
@@ -4,7 +4,7 @@ dependencies:
 - python=3.6.*
 - ipywidgets==7.5.1
 - jupyterlab==1.2.6
-- jupyterhub-base==1.1.0
+- jupyterhub-singleuser==1.1.0
 - nbconvert==5.6.1
 - notebook==6.0.3
 - nteract_on_jupyter==2.1.3
--- a/repo2docker/buildpacks/conda/environment.py-3.6.yml
+++ b/repo2docker/buildpacks/conda/environment.py-3.6.yml
@@ -1,10 +1,10 @@
 # AUTO GENERATED FROM environment.yml, DO NOT MANUALLY MODIFY
-# Generated on 2019-09-13 19:47:35 UTC
+# Generated on 2020-01-27 12:30:09 UTC
 dependencies:
 - python=3.6.*
 - ipywidgets==7.4.2
 - jupyterlab==0.35.4
-- jupyterhub=1.0
+- jupyterhub=1.1.0
 - nbconvert==5.4.1
 - notebook==5.7.8
 - nteract_on_jupyter==2.1.3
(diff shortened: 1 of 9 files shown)


repo2docker/buildpacks/conda/environment.py-3.6.yml
```diff
@@ -1,10 +1,10 @@
...
 dependencies:
 - python=3.6.*
-- ipywidgets==7.4.2
-- jupyterlab==0.35.4
-- jupyterhub=1.0
-- nbconvert==5.4.1
-- notebook==5.7.8
+- ipywidgets==7.5.1
+- jupyterlab==1.2.6
+- jupyterhub-base==1.1.0
```
consideRatio: Is this actually whats needed for any singleuser image when interacting with jupyterhub, as compared to the full jupyterhub? Wait what is jupyterhub-base? I didn't find it on PyPI

minrk: The `jupyterhub` conda package depends on configurable-http-proxy (and thereby nodejs) because that's what is required to run jupyterhub. `jupyterhub-base` is a *conda* package that only contains the jupyterhub Python package, and not the full dependencies of jupyterhub. Basically, this is the same thing you would get if you did `pip install jupyterhub` since pip doesn't know about non-Python dependencies. A similar pattern is followed in e.g. `matplotlib-base` conda package, which has only the matplotlib Python package, but not dependencies required for 'default' use such as Qt.

consideRatio: Ah makes sense, thank you! I've come to learn a lot about conda through you Min =)

minrk: Yeah, it can be confusing when Python packages don't map 1:1 onto conda packages. You see the same thing with linux packages, e.g. what you might think of as just `libfoo` is split into `libfoo-dev` `libfoo-static`, `libfoo`, etc. Conda can do similar things, but doesn't always. It only does it when the separate installation burden is deemed "worth it" e.g. whether to pull in large additional dependencies like nodejs. The nice thing about the conda `juptyerhub` package is it 'just works' by default, unlike the Python one, because it comes with a proxy implementation. The downside is that the same package is used to provide single-user functionality, but doesn't need the proxy. Hence the split package

minrk:
Everything appears to be okay, except for julia 0.6, which is hitting https://github.com/JuliaLang/IJulia.jl/issues/885 due to the update to jupyter-core past 4.5. Since 0.6 is old and unsupported, the simplest fix I found was to use the similarly old Python 3.5 environment when julia 0.6 is selected.

The only hiccup I saw was that when using REQUIRE, the *default* is to pick the oldest Julia, 0.6. But looking at #595, it is said that "REQUIRE is really not a thing in julia 1.0 anymore" so maybe this is fine. We could potentially bump this to 0.7, though. If REQUIRE is already old and outdated, maybe this shouldn't be a priority, and stability is preferable to up-to-date-ness for anyone using this old pattern.

repo2docker/buildpacks/conda/environment.py-3.6.yml
consideRatio: Thanks for sharing this experience Min!

PR #846, opened 2020-02-06
Install jupyter-offlinenotebook from conda-forge

xhochy:

### DIFF
--- a/repo2docker/buildpacks/conda/environment.frozen.yml
+++ b/repo2docker/buildpacks/conda/environment.frozen.yml
@@ -1,5 +1,5 @@
 # AUTO GENERATED FROM environment.py-3.7.yml, DO NOT MANUALLY MODIFY
-# Frozen on 2020-01-30 10:54:05 UTC
+# Frozen on 2020-02-06 21:01:58 UTC
 name: r2d
 channels:
   - conda-forge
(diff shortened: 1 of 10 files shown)


betatim:
Thanks for making the conda forge recipe!

I had hopes that it would save us the step of rebuilding jupyter lab (which takes a lot of time) but it doesn't :( Do you know if that is even possible?

What is the advantage to using a conda package for this? I like being able to install specific revisions from the GitHub repository so that we can update things straight after merging a PR if we want to.

xhochy:
> I had hopes that it would save us the step of rebuilding jupyter lab (which takes a lot of time) but it doesn't :( Do you know if that is even possible?

As far as I have understood the jupyter lab build system, this is currently impossible as a single JavaScript file for all extensions is produced.

> What is the advantage to using a conda package for this? I like being able to install specific revisions from the GitHub repository so that we can update things straight after merging a PR if we want to.

I prefer to have all things installed by conda when using it as an environment manager as it also makes it easier for conda to resolve version conflicts. 

As long as you iterate heavily on it, it may be good to leave this PR as it is and revisit in some months.

betatim:
Thanks for the thoughts on conda vs pip. I think I am +-0 on sticking with pip or switching to conda for this. Maybe +0 for pip until we get some feedback or time passes without feedback. Then go to using conda.

manics:
@xhochy thanks for creating the conda-forge package!
https://pypi.org/project/jupyter-offlinenotebook/ doesn't include the jupyterlab support, that's in the NPM package: https://www.npmjs.com/package/jupyter-offlinenotebook
(both share the [same source repo](https://github.com/manics/jupyter-offlinenotebook) though).

I'm not particularly happy with the JupyterLab extension though it seems to work, I think it'd be better to take advantage of the richer UI options (e.g. see https://github.com/jupyterhub/binderhub/issues/987)

manics:
This was finally done in https://github.com/jupyterhub/repo2docker/pull/996

PR #847, opened 2020-02-07
Remove nodesource' nodejs

yuvipanda:
We get node from the default JupyterLab install, since that
comes from conda-forge. This was not the case when we started -
we installed JupyterLab with pip. We no longer need the nodesource
node.

All the env vars, etc should still work!


betatim:
We used to install (an old version of) jupyterlab via pip which meant we had to provide our own install of node. Now we are installing lab from conda-forge and it is a much newer version of lab. 

Two things I wanted to check/am not sure about:
1. has node been a dependency of the lab conda-forge package for "a long time" or is this a new thing? If it is a new thing, should we keep explicitly installing nodejs (via conda-forge?) for repos that explicitly select an old version of lab in their `environment.yml`?
2. should we keep providing a nodejs version for those people who install lab via pip (e.g. in a `requirements.txt`). I think after merging this PR we will be doing that but wanted to double check.

yuvipanda:
@betatim I think explicitly asking for nodejs from conda-forge is a good idea, rather than depend on it being a dependency of JupyterLab. This means even if a jupyterlab is specified via pip, it will have access to the nodejs that was installed by conda-forge, so we don't need to do anything specific

yuvipanda:
Unfortunately my current internet connection isn't good enough to add to environment.yml and run freeze.py. Would be grateful if someone else can!

betatim:
Some of  the unit tests need updating as well as `black .` to reformat things.

yuvipanda:
Would love for someone with better internet than me to run `freeze.py` on this - I just merged in main branch.

manics:
Done!

manics:
This looks like a genuine test failure:
```
+ /srv/conda/envs/notebook/bin/npm config --global set prefix /srv/npm
/tmp/install-miniforge.bash: line 96: /srv/conda/envs/notebook/bin/npm: No such file or directory
time: 74.381
```
https://github.com/jupyterhub/repo2docker/runs/4266886501?check_suite_focus=true#step:6:2301

yuvipanda:
From looking at the logs, it looks like nodejs is just not being installed at all - and the environment.yml file looks like it is using the python 3.5 one?

manics:
It's the py35 venv test: https://github.com/jupyterhub/repo2docker/tree/main/tests/venv/py35

In case it's not obvious (took me several minutes to realise!) the `3.8` in the test matrix refers to the python used to run repo2docker, not the python inside the image that's built

minrk:
Yeah, the py35 env is not being updated since conda-forge hasn't built py35 packages in a long time. It might be hard to get a 

Changing an env that isn't meant to get updates is tricky since the switch to lockfiles. I'll see if I can figure out what to do. I'm going to try just running `conda install nodejs; conda list --explicit` and see how far that gets me.

Really, what I think we should do for deprecating version support is:

1. implement a mechanism for pinning _repo2docker_ version (and figure out all the security ramifications of letting users prevent us from running the latest version of our own tool)
2. move old versions to 'kernel only' support so they get treated like Python 2, where they can't mess with the notebook server environment, so being out of date is less of a big deal

minrk:
Updated the py35 env with:

```bash
# start with the current frozen env
python3 -m repo2docker ./tests/venv/py35 bash
# (in container)
conda install nodejs=14
# which downgraded notebook to 4.4 for some reason?
conda install notebook==5.7.0  # didn't change anything else, so unclear why it downgraded
conda list --explicit
```

and saved the result. A bit more got updated than I'd like, but I think it should be okay. I gave it some basic tests and it seemed to work.

I'm not aware of any py35 repos in the wild to test with.

yuvipanda:
Yay amazing. With this all tests pass. Anyone feel like merging? Slowly working through this for 20 months haha.

Thanks a lot @manics and @minrk

meeseeksmachine:

https://discourse.jupyter.org/t/node-14-x-bionic-release-does-not-have-a-release-file/11048/13

PR #848, opened 2020-02-11
[MRG] Define an interface for Container engines

manics:
## Summary

Closes https://github.com/jupyter/repo2docker/issues/682
- Abstract the interface that builds a Dockerfile (defined in `repo2docker/engine.py`). Engines are `LoggingConfigurable`- in future this could allow advanced configuration though the repo2docker config file. **This is the file that needs the most review** since it defines the specification for others.
- Implement docker (via docker-py) as the default implementation (see `repo2docker/docker.py`)

This is based on my attempt at a podman backend in https://github.com/jupyter/repo2docker/pull/806

In general most of the Docker Client methods could be easily implemented in a Podman wrapper.

## Details

The main exceptions are `build()` and `push_image()`:
- The Docker client can take a tar file-like-object as the Docker context. To implement it in Podman you have to untar it into a temporary directory. This is inconvenient, though not a blocker since Python makes it easy to untar.
- Build logs: The output format used by Docker is unclear, and the content of each line is specific to Docker. [In #806 I worked around it](https://github.com/jupyter/repo2docker/blob/1be717b4320c4024b4ea0f118f45e6213eefff53/repo2docker/podman.py#L227-L228
) by taking whatever podman spits out and wrapping it as `{"stream": line}` which is enough to convince repo2docker:
  - https://github.com/jupyter/repo2docker/blob/1be717b4320c4024b4ea0f118f45e6213eefff53/repo2docker/app.py#L759-L760.
  - In this PR I've decided it's easier to just return strings, and throw an exception to indicate an error. Engine classes have a flag `string_output` to indicate whether they follow the Docker API in returning objects or if they return plain strings.
- Push image logs: The docker client outputs logs and progress information as events in a way that isn't portable to podman:
  - https://github.com/jupyter/repo2docker/blob/1be717b4320c4024b4ea0f118f45e6213eefff53/repo2docker/app.py#L493-L498
  - as with build() I've switched to plain strings with the option of keeping the docker format.

Remaining todos:
- [x] Fix unit tests
- [x] Remove commented lines/in-progress notes
- [x] Update podman backend https://github.com/jupyter/repo2docker/pull/806 to implement this interface as verification of the design, either in this repo or as an external plugin with a `repo2docker.engine` entrypoint

## Other engines

[...]

PR #849, opened 2020-02-14
[MRG] Update black 19.10b0, target Python 3.5

manics:
In https://github.com/jupyter/repo2docker/pull/848 I ran into problems with black auto-formatting some new code in a way that incompatible with Python 3.5 (I think this issue is related: https://github.com/psf/black/issues/419).

It looks like this is fixed in the latest black.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,6 @@
 repos:
 - repo: https://github.com/ambv/black
-  rev: 18.9b0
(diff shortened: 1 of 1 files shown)


consideRatio:
LGTM - but i don't grasp the CI failures

betatim:
Restarted CI jobs. The failures are probably because MRAN is down.

LGTM otherwise

Issue #773, opened 2019-09-04
Alternative to MRAN


betatim:
### Proposed change
Provide an alternative way of pinning R packages to using MRAN dates. MRAN has been flaky/unavailable several times this week which means repos that use a MRAN date can't be built.

This issue is about collecting alternatives and their pros/cons. As well as maybe some information on whether the current performance of MRAN is a sign for things to come or just an intermittent problem.

### Who would use this feature?

All R users.

### Who can do this work?

To work on this you need to understand how CRAN and R packaging work. It would help if you have some understanding or connection to MRAN to judge if the current outages are connected to a policy change.

trallard:
In the past, I have successfully used `Pacman` to pin versions of R packages <https://github.com/trinker/pacman> 

That way the packages and their versions can be declared like so in the `install_packages.R` file:

```R
p_install_version(
    c("pacman", "testthat"),
    c("0.2.0", "0.9.1")
)

```
*Pros*: 
- It does what you'd expect from it, install the specific version of the package regardless of it being in CRAN, GitHub or the such
- Pacman can also be used to check dependencies of a certain R package (as in other R packages this depends as opposed to system dependencies) `p_depends(lattice) `
- It does not add additional install time 

*Cons*:
- As opposed to MRAN it is not a snapshot of the R packages in time but will install a pinned version of a package (pretty much the same as `pip install pandas==0.25.`

sje30:
MRAN seems down right now :-( which seems to be blocking my repo2docker runs

CC: @nuest 


```
Step 40/67 : RUN R --quiet -e "install.packages('devtools', repos='https://mran.microsoft.com/snapshot/2018-02-01', method='libcurl')" && R --quiet -e "devtools::install_github('IRkernel/IRkernel', ref='0.8.11')" && R --quiet -e "IRkernel::installspec(prefix='$NB_PYTHON_PREFIX')"
 ---> Running in 68d9b2021834
> install.packages('devtools', repos='https://mran.microsoft.com/snapshot/2018-02-01', method='libcurl')
Installing package into ‘/srv/rlibs’
(as ‘lib’ is unspecified)
Warning: unable to access index for repository https://mran.microsoft.com/snapshot/2018-02-01/src/contrib:
  cannot open URL 'https://mran.microsoft.com/snapshot/2018-02-01/src/contrib/PACKAGES'

```

betatim:
MRAN has been up&down several time s over the last few days :(

[...]

PR #851, opened 2020-02-15
[MRG] If looking for latest MRAN URL try earlier snapshots too

manics:
Related to https://github.com/jupyter/repo2docker/issues/773#issuecomment-586571022

The particular issue I ran into today is a missing "latest" MRAN snapshot in the Travis R tests. This PR automatically tries older snapshots in this case.

What's not clear to me from the linked issue is whether the test failures mostly occur due to a missing latest snapshot (this PR should help), or whether it's due to a broken MRAN server for a snapshot that definitely exists.

Example:
```
$ repo2docker tests/r/description-file/
Using local repo tests/r/description-file/.
Failed to get MRAN snapshot URL https://mran.microsoft.com/snapshot/2020-02-13: 404 Not Found
Using RBuildPack builder
```

### DIFF
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -1,6 +1,7 @@
 import re
 import os
 import datetime
+import requests
 
 from distutils.version import LooseVersion as V
 
@@ -135,8 +136,9 @@ class RBuildPack(PythonBuildPack):
                 # no R snapshot date set through runtime.txt
                 # set the R runtime to the latest date that is guaranteed to
                 # be on MRAN across timezones
-                self._checkpoint_date = datetime.date.today() - datetime.timedelta(
-                    days=2
+                two_days_ago = datetime.date.today() - datetime.timedelta(days=2)
+                self._checkpoint_date = self._get_latest_working_mran_date(
+                    two_days_ago, 3
                 )
                 self._runtime = "r-{}".format(str(self._checkpoint_date))
             return True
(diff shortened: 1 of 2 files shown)


betatim:
I hadn't considered the idea that the problem is missing snapshots and not MRAN being "down". So no data/expertise to offer unfortunately.

I think if the user hasn't specified a date we should try a few days into the past to increase the chances of things working.

repo2docker/buildpacks/r.py
```diff
@@ -186,6 +188,29 @@ def get_packages(self):
...
         return super().get_packages().union(packages)
 
+    def _get_latest_working_mran_date(self, startdate, max_prior):
+        """
+        Look for a working MRAN snapshot
+
+        Starts from `startdate` and tries up to `max_prior` previous days.
+        Raises `requests.HTTPError` with the last tried URL if no working snapshot found.
+        """
+        for days in range(0, max_prior + 1):
```
betatim: Why explicitly set the start of `range` to zero? As it is the default value I was expecting `range(max_prior + 1)`

betatim:
Could you add a test for the new method? Maybe something that uses a mock to remove the dependency on MRAN being available.

Our coverage has been slipping below (the arbitrary mark of) 90% so we need to start being a bit more keen to add tests to reverse that trend.

manics:
Thanks, for reviewing, I'll add a test.

manics:
Done! Though it's failing due to MRAN being down 🤣

PR #855, opened 2020-03-01
[MRG] Switch to recommending to install from master

### LINKED ISSUE #841: Should the README recommend installing from master instead of pypi?
The last release is around 6 months old and is missing several bug fixes. Since mybinder.org runs off `master` should the readme also recommend installing from master: `pip install -U https://github.com/jupyter/repo2docker/archive/master.zip` ?

betatim:
We only make a release every three months but run master on mybinder.org
which means people see bugs that were long fixed when they run locally. So
we often tell people "try master".

I can't think of a concrete reason not to recommend this. WDYT?

closes #841

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,13 +32,22 @@ Supported on Linux and macOS. [See documentation note about Windows support.](ht
 
 This a quick guide to installing `repo2docker`, see our documentation for [a full guide](https://repo2docker.readthedocs.io/en/latest/install.html).
 
-To install from PyPI:
+We recommend you install the latest and greatest version:
+```bash
+pip install -U https://github.com/jupyter/repo2docker/archive/master.zip
+```
+We frequently improve repo2docker and update the version of repo2docker on
+mybinder.org a few hours later. This means it is rare that this version
+doesn't work.
+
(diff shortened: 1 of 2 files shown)


manics:
One disadvantage of installing from the zip:
```
$ python3 -m pip install -U https://github.com/jupyter/repo2docker/archive/master.zip
...
$ repo2docker --version
0+unknown
```
Installing from git requires git to be installed but shows a version:
```
$ python3 -m pip install -U git+https://github.com/jupyter/repo2docker.git
...
$ repo2docker --version
0.11.0+13.g26542df
```

betatim:
This is annoying. I thought that one of the many things versioneer does (inside of all of its thousands of lines of code) is display a useful version string even when you install from the archive/not via git.

manics:
I think versioneer lets you install the package from a non-git URL but without a version. Compare this with e.g. [setuptools-scm](https://pypi.org/project/setuptools-scm/) which will fail if you try an installed a development version thats not a git URL.

betatim:
We now make (more) regular releases so this isn't needed any more.

PR #861, opened 2020-03-04
Use COPY --chown to prevent extra chown layer

### LINKED ISSUE #164: use COPY --chown to stage repo (docker 17.09)

yuvipanda:
COPY directive always copied files into the image owned
by root. Since we wanted this to be owned by non-root users,
we needed a RUN chown command, which added a fresh layer to
the docker image for no good reason.

This fixes that, but requires docker 19.03 to work.

Ideally, we can either:

1. Pass NB_USER to the template, so we can expand that with
   jinja2 instead of as an env. This makes it work with earlier
   versions of docker
2. Dynamically determine which version of docker we are running,
   so we can conditionally use features. Me no likey

Fixes #164

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -150,11 +150,10 @@ RUN chown -R ${NB_USER}:${NB_USER} ${REPO_DIR}
 {{ sd }}
 {% endfor %}
 
-# Copy and chown stuff. This doubles the size of the repo, because
-# you can't actually copy as USER, only as root! Thanks, Docker!
+# We need docker 19.03 for --chown to work with variable substitution
+# https://github.com/moby/moby/issues/35018
 USER root
-COPY src/ ${REPO_DIR}
-RUN chown -R ${NB_USER}:${NB_USER} ${REPO_DIR}
+COPY --chown ${NB_USER} src/ ${REPO_DIR}
 
 # Run assemble scripts! These will actually turn the specification
 # in the repository into an image.


betatim:
Do you know how to look up what the "recent" versions of docker are? I remember trying to download docker and getting lost in the need (or not if you knew the link) to register. And it kept trying to send me to some kind package thing. Basically it was annoying. Just now their website won't even tell em what version of "Docker desktop for mac" I'd be downloading :-/

On my laptop right now `docker --version` prints `Docker version 19.03.5, build 633a0ea`. Without the comparison of what is recent etc it is hard to tell how "novel" requiring 19.03 is.

It would be super to be able to use this feature and not have to do gymnastics.

minrk:
@betatim docker uses date versioning, so 19.03 means March of 2019, so ~16 months ago. That might be pushing it a little bit. `--chown` requires 17.09, so 3 years ago, but doesn't work with `${NB_USER}` substitution in the chown field. 

Note that the BinderHub *helm chart* can (and usually does, though not by default) use docker-in-docker which makes the host docker version irrelevant, so I think it would be fine from the BinderHub perspective to require use of dind if host docker is too old. I don't think that's too big of a request. So, to me, the main restriction is in linux distros for "home use". If you get docker from ubuntu (as opposed to from docker itself), then we have the following versions for currently-supported ubuntu releases:

- 16.04 LTS: 18.09.7
- 18.04 LTS: 18.09.7
- 19.10: 19.03.2
- 20.04 LTS: 19.03.8

so the 18.04 LTS won't support it, which is supported by Canonical for another 3 years. I didn't check CentOS/etc. but they tend to be way behind. For me, personally, I always install current docker from docker itself, which makes it a non-issue, but I'm sure some folks have policy reasons why they might not be able to (or want to) do that.

If we deem 19.03 too recent and still want to solve the layer-duplication issue, we can switch to using a scratch image to do the RUN permissions and then `COPY --from=scratch` which will only create one layer in the final uploaded image, but complicates the templates somewhat:

```Dockerfile
FROM scratch as src-scratch
COPY . /src
RUN chown -R ${NB_USER}:${NB_USER} /src

FROM buildpack-deps...
...
COPY --from=src-scratch /src {REPO_DIR}
```

This solves the image-size issue, but not the build-time hit of running chown in an additional step, which can be nontrivial for lots of files. Doing it in a scratch image might even make it a little worse, though it *could* mean better cache performance for the chown step, at least.

minrk:
Re-reading @yuvipanda's suggestions: I think we should do option 1: expand NB_USER in the template and require docker 17.09. That sounds easiest!

betatim:
@yuvipanda do you want to reboot this PR with Min's suggestion or should we start a new PR with a new author?

yuvipanda:
@betatim would love for someone else to take this over!

PR #863, opened 2020-03-20
Document loose conda export with --from-history

xhochy:

### DIFF
--- a/docs/source/howto/export_environment.rst
+++ b/docs/source/howto/export_environment.rst
@@ -29,9 +29,10 @@ The solution
 Export your explicit install commands
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you only want to have an ``environment.yml`` that reflects the
-``conda install`` you have run to build your conda environment, you can use
-``conda env export --from-history environment.yml``. The resulting
+To get a minimal ``environment.yml`` that only contains the packages you
+explicitly installed run
+``conda env export --from-history -f environment.yml``. We recommend that you
+use this option to create your ``environment.yml``. The resulting
 ``environment.yml`` then contains a loose pinning of the versions used, e.g.
 ``pandas=0.25`` if you explicitly requested this ``pandas`` version on
 installation. If you didn't list a version constraint during installation, it
--- a/docs/source/howto/export_environment.rst
+++ b/docs/source/howto/export_environment.rst
@@ -26,6 +26,24 @@ be resolved (``ResolvePackageNotFound``).
 The solution
 ============
 
+Export your explicit install commands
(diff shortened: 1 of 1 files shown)


xhochy:
Depends on https://github.com/jupyter/repo2docker/pull/864

docs/source/howto/export_environment.rst
```diff
@@ -26,6 +26,24 @@ be resolved (``ResolvePackageNotFound``).
 The solution
 ============
 
+Export your explicit install commands
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you only want to have an ``environment.yml`` that reflects the
```
betatim: How about something like the following? I think we should help unfamiliar users by telling them that they want to do this over other things (unless they know what they are doing, in which case they are probably not reading this).

To get a minimal ``environment.yml`` that only contains the packages you explicitly installed run ``conda env export --from-history environment.yml``. We recommend that you use this option to create your ``environment.yml``. The resulting ``environment.yml`` then contains ...

betatim:
Merged #864, rebase?

docs/source/howto/export_environment.rst
xhochy: Thanks, integrated this.

xhochy:
The Travis CI errors seem unrelated.

betatim:
> The Travis CI errors seem unrelated.

Are they ever not unrelated :D ... restarted travis. Seemed like something about running docker containers broke :-/

betatim:
Do you have an idea what the heck travis is trying to tell us? I tried googling for recent changes to how to use docker on travis and such but couldn't find anything. The libraries we install also seem to be the same versions as they were last time the build passed.

manics:
- Broke between 3 and 6 days ago: https://travis-ci.org/github/jupyter/repo2docker/builds?utm_medium=notification&utm_source=github_status
- Nothing in the Travis changelog https://changelog.travis-ci.com/
- I checked the main repo2docker Python dependencies (requests, docker, pytest, websocket-client) but couldn't see any recent releases
- Version of Docker in Travis is unchanged (verified by comparing the build logs for a failed and successful build)

Quite mysterious!

betatim:
```
docker.errors.APIError: 400 Client Error: Bad Request ("OCI runtime create failed: container_linux.go:348: starting container process caused "exec: \"/usr/local/bin/repo2docker-entrypoint\": permission denied": unknown"
```
keeps coming up as error message. In the GitHub repo the executable bit is set for the file. I wonder if it is owned by "the wrong" user??

betatim:
Travis is happy now. Not sure why it isn't reported as done though. Either way: merging this now.

PR #870, opened 2020-04-08
Add support for Julia 1.4.0

davidanthoff:
New version came out a while ago.


betatim:
Thanks :)

It is nice to see how easy it is to add a new version.

davidanthoff:
There is actually a hosted JSON file somewhere now that has all the versions that exist listed, so in theory we could even change this code so that it just looks at that file and uses the list from that file... But right now just adding these new versions here is a lot less work :)

PR #874, opened 2020-04-16
[MRG] Change --env option to work like docker's

### LINKED ISSUE #871: --env option semantics differ from docker's
### Bug description
The semantics for the `--env` option to `repo2docker` differs from the sane option in the `docker run` subcommand.

#### Expected behaviour

I expected the option to behave identically, but they do not in the case of a bare environment variable name. For example, to pass the value of a secret to a docker instance (so it does not appear on the command line), you can do:
```bash
SEKRT=fred  docker run --env SEKRT ...
```
and the value of `SEKRT` inside the container is "fred".

see also [docker documentation](https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file)

#### Actual behaviour

For `repo2docker`:
```bash
SEKRT=fred  repo2docker --env SEKRT ...
```
and `SEKRT` is undefined inside the container.
### How to reproduce

1. Download [this notebook](https://gist.github.com/34dd525848df78a5adc101b0da0ac3eb) to an empty directory
2. run `SEKRT=sam repo2docker --env SEKRT -E .`
    1. connect to server and run notebook cells
    1. note output -- this is observed
    1. stop server
3. Get docker image just created: `image=$(docker images --quiet | head -1)`
1. run `SEKRT=sam docker run --rm --env SEKRT $image`
    1. connect to server and run notebook cells
    1. note output -- this is expected
    1. stop server

### Your personal set up

 - OS: Ubuntu Ubuntu 18.04.4 LTS
 - Docker version: 18.06.1-ce, build e68fc7a
 - repo2docker version: 0.11.0

hwine:
The goal here is make `repo2docker` process the `--env` option with the same semantics as the `--env` option to `docker` itself.

Closes #871

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -560,6 +560,8 @@ class Repo2Docker(Application):
                     "mode": "rw",
                 }
 
+        # self.environment needs to be modified to match the semantics
+        # of docker's --env option
         run_kwargs = dict(
             publish_all_ports=self.all_ports,
             ports=ports,
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -560,8 +560,6 @@ class Repo2Docker(Application):
                     "mode": "rw",
                 }
 
-        # self.environment needs to be modified to match the semantics
-        # of docker's --env option
         run_kwargs = dict(
             publish_all_ports=self.all_ports,
             ports=ports,
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -39,6 +39,27 @@ def validate_image_name(image_name):
     return image_name
 
 
+# See https://github.com/jupyter/repo2docker/issues/871 for reason
+class MimicDockerEnvHandling(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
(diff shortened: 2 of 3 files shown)


betatim:
(I edited your comment (the first one) to reference the original issue so that it closes automatically if/when we merge this)

repo2docker/app.py
```diff
@@ -560,6 +560,8 @@ def start_container(self):
                     "mode": "rw",
                 }
 
+        # self.environment needs to be modified to match the semantics
+        # of docker's --env option
```
hwine, line 564: Upon further thought, I don't believe this is the correct place to fix the bug. The intention of `start_container` is just that, not start-the-container-after-mucking-with-environment-variables.

I'm investigating fixing the bug during the initial argument processing.

hwine:
Ah, great -- clean run 2nd time through!  I think this is ready for feedback.

Since it's all about behavior, I didn't see a need to modify documentation anywhere.

betatim:
Looks good to me! Merging.

Thanks for a nice PR. I learnt something :)

PR #879, opened 2020-04-17
Work around a Julia bug

davidanthoff:
This works around a bug in Julia. I expect that this will be fixed eventually, and at that point we can put this extra call behind a version check.


manics:
Do you have a link to the upstream ticket? Could be worth adding as a comment for future reference.

davidanthoff:
Yes, the upstream bug is https://github.com/JuliaLang/Pkg.jl/issues/1775. Once this is fixed I'll open a new PR that only runs this workaround for Julia versions here this is a problem.

betatim:
+- 0.06% in code coverage. Seems like "codecov go home, you are drunk" is appropriate here ;)

This looks good to me so merging it.

PR #882, opened 2020-04-23
[MRG] Workaround Docker issue impacting some tests on macOS

### LINKED ISSUE #881: failing docker tests locally on macOS (case 1)
### Bug description

With the latest versions of macOS (Catalina 10.15.4) and Docker CE (2.2.0.5), many of the tests fail to run locally.

#### Expected behaviour

Running the test suite locally on a released version of `repo2docker` should have no errors or failures.

#### Actual behaviour

A number of tests which require running docker fail. The displayed message indicates Docker was unable to complete a volume mount:
```
docker.errors.APIError: 500 Server Error: Internal Server Error ("b'Mounts denied: 
The path /var/folders/fs/fjwjlc5x3yzgbw2nnnyzfcq80000gn/T/tmpxb7ynwyc
is not shared from OS X and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> File Sharing.
See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.")
```
Following the Docker instructions to add the path to the list of allowed mounting paths does not change the behavior.

### How to reproduce

1. On a macOS box, checkout a released version of `repo2docker`
2. Setup the development environment.
3. invoke `pytest tests/unit/test_env.py` (as one example)
4. observe failure.

### Your personal set up

 - OS: macOS
 - Docker version: CE 2.2.0.5
 - repo2docker version: 0.11.4

### Workaround

While this is clearly a Docker problem, it is possible to tweak (some of) the tests to avoid needing to use the problematic directory paths.

For example, in the `test_env.py` shown above, `repo2docker` is invoked from a directory that both:
- is created in a manner that ensures the directory will be deleted at the end of the test (unless python itself crashes)
- has no content

Thus, changing the parent of the temp directory to be the git working directory allows the test to pass:
```diff
-    with tempfile.TemporaryDirectory() as tmpdir:
+    # There appear to be some odd combinations of default dir that do
+    # not work on macOS Catalina with Docker CE 2.2.0.5, so use
+    # the current dir -- it'll be deleted immediately
+
+    with tempfile.TemporaryDirectory(dir=os.path.abspath(os.curdir)) as tmpdir:
```

This approach may not be applicable for all such locations, or desired for other reasons by the team. I'm happy to prepare a PR if project folks think this is a reasonable approach.

hwine:
This works around an issue where the temp directory can not be added to
Docker. In this case, the directory is immediately deleted, and does not
occupy significant disk space. There should be no impact to the move.

Fixes #881

### DIFF
--- a/tests/unit/test_env.py
+++ b/tests/unit/test_env.py
@@ -13,7 +13,11 @@ def test_env():
     Validate that you can define environment variables
     """
     ts = str(time.time())
-    with tempfile.TemporaryDirectory() as tmpdir:
+    # There appear to be some odd combinations of default dir that do
+    # not work on macOS Catalina with Docker CE 2.2.0.5, so use
+    # the current dir -- it'll be deleted immediately
+
+    with tempfile.TemporaryDirectory(dir=os.path.abspath(os.curdir)) as tmpdir:
         username = getuser()
         subprocess.check_call(
             [


hwine:
As noted in [#881](https://github.com/jupyter/repo2docker/issues/881#issuecomment-619503153), the rest of the docker issues appear to have a different cause. Plan is to fork #881, and land this PR to address the temp directory issue.

I believe the CI failure above is not caused by this PR.

betatim:
Merged! The bot that counts if coverage of the code has gone up/down sometimes gets confused with there are small changes/high comment-to-code ratio it seems.

hwine:
Ah - thanks for that feedback on Codecov -- that has bitten me on other projects!

Issue #883, opened 2020-04-24
Make binder/start script not need to be executable


danlester:
### Proposed change

If a start script is provided by the user, ensure that the [repo2odocker-entrypoint](https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/repo2docker-entrypoint) script makes it executable before running it. Or if not possible to change file permissions, run it through the appropriate interpreter explicitly (see alternative options below).

This is similar to Issue #241 for postBuild.

### Alternative options

Instead of changing file permissions, a more complex repo2docker-entrypoint script could parse the start script (probably safely assume it's not a binary file) and explicitly run it through the specified hash-bang interpreter (e.g. #!/bin/bash).

Or the build process could copy the start script to $CONDA_DIR and run it from there - but there could be issues if the script is expecting itself to be in the repo's folder structure, e.g. in the .binder folder.

### Who would use this feature?

Mostly this isn't a problem as the start script is [given the correct permissions](https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/base.py#L185) during the Docker build process.

But in some cases, file permissions do not really exist, e.g. if the folder is mounted from a host system. E.g. building from a 'local folder' repo source when the final container is designed to mount the repo as /home/jovyan.

This happens, for example, in ContainDS software running on Windows or Mac (actually from any repo source in this case).

### How much effort will adding it take?

Estimated at 1 day.

### Who can do this work?

I would be happy to do this work myself but would like confirmation that more experienced contributors are happy with the approach.

Pointers to a relevant wide range of sample repos would be helpful too (e.g. R since I am more experienced with Python).

betatim:
Sounds like a good idea. I would avoid trying to parse scripts and extracting things to then try and locate the right interpreter, etc. Sounds easy and like something that will be super hard to get right for the last 20% of cases.

We already do `chmod +x` for `binder/start` but you mention filesystems without executable permission are a problem. I'm not sure I understand/know what the entrypoint script could do if `chmod +x` didn't work.

danlester:
Yes, you are right my original post was a bit confused: if chmod didn't work the first time around during build, we can probably assume it won't work in the entrypoint script either (although I will check this in my case.)

As a fallback I'm not personally terrified of parsing the script for clues, or maybe just running it through bash regardless. The point is we are in a case where the hoped-for behavior has failed, so risking further failure in difficult cases isn't too bad.

[...]

PR #886, opened 2020-04-29
[MRG] chmod start script from repo2docker-entrypoint

danlester:
Discussed in Issue #883 - Make binder/start script not need to be executable

Now repo2docker-entrypoint ensures 'chmod u+x' is applied to the start script specified in $R2D_ENTRYPOINT.

This helps if the chmod applied in the build process doesn't affect the final executed copy of the start script file. For example, if the repo2docker is run as a 'local folder' source which is then mounted at /home/jovyan. The image's copy of the start script had +x set, but not in the local folder itself.

It will also fix start script problems with any repo source within my [ContainDS Desktop](https://containds.com/desktop) 'local binder' software.

### DIFF
--- a/repo2docker/buildpacks/repo2docker-entrypoint
+++ b/repo2docker/buildpacks/repo2docker-entrypoint
@@ -3,6 +3,7 @@
 # we use a login shell to get a fully configured shell environment
 # (e.g. sourcing /etc/profile.d, ~/.bashrc, and friends)
 if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
+    chmod u+x "$R2D_ENTRYPOINT"
     exec "$R2D_ENTRYPOINT" "$@"
 else
     exec "$@"
--- a/repo2docker/buildpacks/repo2docker-entrypoint
+++ b/repo2docker/buildpacks/repo2docker-entrypoint
@@ -3,7 +3,9 @@
 # we use a login shell to get a fully configured shell environment
 # (e.g. sourcing /etc/profile.d, ~/.bashrc, and friends)
 if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
-    chmod u+x "$R2D_ENTRYPOINT"
+    if [[ ! -x "$R2D_ENTRYPOINT" ]]; then
+        chmod u+x "$R2D_ENTRYPOINT"
+    fi
     exec "$R2D_ENTRYPOINT" "$@"
 else
     exec "$@"


manics:
Are there any other code paths in repo2docker where external files are silently modified by the app?

danlester:
Thanks for taking a look @manics 

If mounting a local folder as described, Jupyter dumps all its config in the external folder...
i.e. you end up unexpectedly with:

.local
.cache
.bash_history
.ipynb_checkpoints

So personally I don't think the chmod on the start script is too dramatic compared to this, although arguably it is more likely to be version controlled than the files/folders listed above.

In my mind, the ability to affect the external files is 'granted' by the user mounting the folder to the Docker container. (It's already something users need to be wary of in terms of polluting the external folder.)

Not sure of other similar examples in other cases though. Please anyone let me know your thoughts!

repo2docker/buildpacks/repo2docker-entrypoint
```diff
@@ -3,6 +3,7 @@
 # we use a login shell to get a fully configured shell environment
 # (e.g. sourcing /etc/profile.d, ~/.bashrc, and friends)
 if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
+    chmod u+x "$R2D_ENTRYPOINT"
```
manics, line 6: To avoid a corner case where a file is owned by someone else but is already group/other executable how about only running `chmod` if it's not already executable by testing with something like `[ -x "$R2D_ENTRYPOINT" ]`?

manics:
Fair enough, if it already makes changes this is probably OK then!

Would be nice to have this documented somewhere but I don't know of a good place, regardless I don't think that needs to block this PR.

repo2docker/buildpacks/repo2docker-entrypoint
danlester: Good idea - I have added this to the latest commit.

repo2docker/buildpacks/repo2docker-entrypoint
```diff
@@ -3,6 +3,9 @@
 # we use a login shell to get a fully configured shell environment
 # (e.g. sourcing /etc/profile.d, ~/.bashrc, and friends)
 if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
+    if [[ ! -x "$R2D_ENTRYPOINT" ]]; then
+        chmod u+x "$R2D_ENTRYPOINT"
```
manics, line 7: I'm undecided on whether `u+x` or `a+x` is better

danlester: I specified only u+x because ultimately that is the minimum action to fix this use case.

It might also be helpful to extend to a+x but (1) I'm not sure when/why that might be useful, and (2) think it is more likely to have unforeseen side effects compared to just leaving as u+x.

I was thinking that unless anyone knows an obscure operating system or a case where a+x will succeed but u+x somehow doesn't fix the original problem, it could be left as u+x until someone discovers that it needs to be extended!

BUT for comparison, start_script is given a+x in the docker build. That is also carried out as NB_USER, although that is within the docker build which may be different.

As another comparison, in docker-stacks, the [fix-permissions](https://github.com/jupyter/docker-stacks/blob/master/base-notebook/fix-permissions) script seems to prefer permissions centred on the NB_GID group (g+rwX).

Any of these approaches works from my point of view. I would suggest keep it minimal with u+x but welcome your decision :)

manics: :+1: I'll let someone else check and merge

betatim:
Restarted the tests.

betatim:
For this discussion about modifying local files: could someone paste the exact repo2docker command they run where this happens?

Something like `repo2docker /tmp/foobar` which contains a `requirements.txt` leaves `/tmp/foobar/` unchanged as far as I can tell. However if you run `repo2docker --editable /tmp/foobar` you get extra "stuff" as well as your changes.

danlester:
@betatim Thanks for looking at this. Sorry for any confusion - by 'mount local folder' I meant editable mode. So, yes, the command would be:

```
repo2docker --editable /tmp/foobar
```

This is the scenario where the start script execute permissions are most likely to be a problem, and also the scenario where other external files may be written e.g. the jupyter hidden files (.ipynb_checkpoints, .local, .cache, etc)

I hope this makes sense.

PR #890, opened 2020-05-08
[MRG] Document test failure workarounds

### LINKED ISSUE #884: failing docker tests locally on macOS (case 2)
### Bug description

With the latest versions of macOS (Catalina 10.15.4) and Docker CE (2.2.0.5), many of the tests fail to run locally. _(This issue was split from #881 after the approach in #882 did not seem to apply to these failures.)_

#### Expected behaviour

Running the test suite locally on a released version of `repo2docker` should have no errors or failures.

#### Actual behaviour

A number of tests which require running docker fail. There appear to be a number of different failure messages, many appear to be related to failures using `conda` or `pip` to install packages inside the docker container during `self.build()` invocations. Typical message:
```
>                           raise docker.errors.BuildError(l["error"], build_log="")
E                           docker.errors.BuildError: The command '/bin/sh -c apt-get -qq update &&     apt-get -qq install --yes        libapparmo
r1        lsb-release        psmisc        sudo        > /dev/null &&     apt-get -qq purge &&     apt-get -qq clean &&     rm -rf /var/lib/apt/lis
ts/*' returned a non-zero code: 100

repo2docker/app.py:729: BuildError
__________________________________________________________________ test session ___________________________________________________________________

    def test():
        app = make_r2d(args)
        app.initialize()
        if app.run_cmd:
            # verify test, run it
            app.start()
            return
        # no run_cmd given, starting notebook server
        app.run = False
>       app.start()  # This just build the image and does not run it.

tests/conftest.py:46: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
repo2docker/app.py:743: in start
    self.build()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
```
Step 1 will be to enumerate and categorized those messages.

### How to reproduce

1. On a macOS box, checkout a released version of `repo2docker`
2. Setup the development environment.
3. invoke `pytest tests/`
4. observe failures.

### Your personal set up

 - OS: macOS
 - Docker version: CE 2.2.0.5
 - repo2docker version: 0.11.4

hwine:
Some tests will fail if the configuration of the developer's machine is not as expected. The failures have distinct message patterns, so add workarounds for each type.

Fixes #884

### DIFF
--- a/docs/source/contributing/tasks.md
+++ b/docs/source/contributing/tasks.md
@@ -23,6 +23,16 @@ If you want to run a specific test, you can do so with:
 py.test -s tests/<path-to-test>
 ```
 
+### Troubleshooting Tests
+
+Some of the tests have non-python requirements for your development machine. They are:
+
+- `git-lfs` must be installed ([instructions](https://github.com/git-lfs/git-lfs)). It need not be activated -- there is no need to run the `git lfs install` command. It just needs to be available to the test suite. 
+   - If your test failure messages include "`git-lfs filter-process: git-lfs: command not found`", this step should address the problem.
+
+- Minimum Docker Image size of 128GB is required. If you are not running docker on a linux OS, you may need to expand the runtime image size for your installation. See Docker's instructions for [macOS](https://docs.docker.com/docker-for-mac/space/) or [Windows 10](https://docs.docker.com/docker-for-windows/#resources) for more information.
+    - If your test failure messages include "`No space left on device: '/home/...`", this step should address the problem.
+
 ## Update and Freeze BuildPack Dependencies
 
 This section covers the process by which repo2docker defines and updates the


hwine:
I'm deliberately _not_ updating the section in [contributing](../tree/master/docs/source/contributing/contributing.md) dealing with dev machine setup. My reasons:

- It will duplicate information, which is evil.
- I couldn't think of a non-convoluted way to express the conditionals there (if you're not on linux or wsl2 or ...). 

Two other "it would be even better if" items I'm not handling now (perfection is the enemy of done):

- trying to figure the minimum disk space requirement (I just know it's between 64GB & 128GB)
- adding tests of the host machine that would explicitly point to these issues. This would be great if, e.g. the needed disk grew and folks who had been okay are suddenly not.

Happy to be challenged on either point, though. 😸

betatim:
Merged (after a long delay) as is. Thanks for taking the time to figure these out and write down tips for others! The disk space issue seems very weird. I am on OSX and I don't think I've had to change the disk size of my docker VM. I wonder what it could be, I have no good idea though or place to start digging -> let's keep an eye on it and maybe others come up with some new clues.

Issue #799, opened 2019-09-25
[feature request] Shiny interface deployed from RMarkdown script


choldgraf:
an issue that was opened up in BinderHub (https://github.com/jupyterhub/binderhub/issues/961) by @pablobernabeu :

---

Currently, running a Binder repo into the Shiny app interface (`?urlpath=shiny`) requires having a ui.R and a server.R script. I would like to make a feature request to enable a slightly different form of a Shiny app which is known as [Flexdashboard](https://rmarkdown.rstudio.com/flexdashboard/). 

A Flexdashboard is based on a single Rmarkdown (.rmd) script, which may have any name. In RStudio, the app is run through 'Run Document'. In the R console, it's run with `rmarkdown::run('myfile.Rmd')`.

An example may be found [in this repo](https://mybinder.org/v2/gh/pablobernabeu/Modality-exclusivity-norms-747-Dutch-English-replication/master?urlpath=rstudio), inside the 'Shiny-app' folder. Currently, the app can be run with the 'Run Document' button in RStudio, but [not through the `?urlpath=shiny` interface](https://mybinder.org/v2/gh/pablobernabeu/Modality-exclusivity-norms-747-Dutch-English-replication/master?urlpath=shiny/Shiny-app).

Thank you very much

choldgraf:
@pablobernabeu thanks for the suggestion! Could you list out the steps that would be needed to accomplish this locally?

My guess is that the change would need to be made in jupyter/repo2docker, and we'd basically need to add the logic in the R buildpack [somewhere around here](https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/r.py#L284) to decide "how" to launch a shiny server.

As with most things R-related, it'd be great if we can get some guidance from the R community on this one.

pablobernabeu:
Thank you! This would be great.

I'd like to first mention that this type of Shiny trigger is relevant when there is a .rmd script that contains the string `runtime:shiny` (`runtime: shiny` or `runtime:  shiny`).

This feature would introduce a .rmd script into the equation of the Shiny interface. Therefore it might be necessary to protect existing repos that could contain a .rmd file with `runtime:shiny` in the same folder where they have the Shiny scripts. To avoid breaking the Shiny interface of those repos, perhaps an approximate idea could be:

- when the Shiny interface is called (i.e., as determined by `?urlpath=shiny` being included in a Binder repo's URL),

[...]

PR #891, opened 2020-05-09
Install rmarkdown library

pablobernabeu:
Installing (at least trying to install) `rmarkdown` library (for background, see https://github.com/jupyter/repo2docker/issues/799, as well as Section 3.2 at https://docs.rstudio.com/shiny-server/)

Hi all (esp. @choldgraf),

I've looked back into #799. To my joy, it turns out that Shiny Server already implemented a hierarchical logic whereby R Markdown scripts are sought after pure Shiny ones, as [quoted below from RStudio](https://docs.rstudio.com/shiny-server/).

> **3.2 R Markdown**
>
> In addition to serving traditional Shiny applications out of a directory (a server.R file and an associated UI), Shiny Server now supports interactive R Markdown documents. To take advantage of this capability, you will need to make sure that R Markdown is installed and available for all users. You can do so by running the following command:
> 
> `sudo su - -c "R -e \"install.packages('rmarkdown')\""`
> 
> If a hosted directory does not include a server.R file, Shiny Server will look to see if it contains any .Rmd files. If so, Shiny Server will host that directory in "R Markdown" mode using `rmarkdown::run`.
> 
> Particular Rmd files can be accessed by referencing their full path including the filename, e.g., http://myserver.org/mydocs/hello.Rmd. If a request is made to a directory rather than to a particular Rmd file, Shiny Server will attempt to serve the file index.Rmd. If that file does not exist, the user will get an error alerting them that the file is not available.

So, I've tried to install `rmarkdown`. I would be grateful for a review of this. If this is merged, I would then follow up whether the new feature works, by looking at the document mentioned in #799.

Thank you very much

### DIFF
--- a/repo2docker/buildpacks/r.py
+++ b/repo2docker/buildpacks/r.py
@@ -314,6 +314,15 @@ class RBuildPack(PythonBuildPack):
                 ),
             ),
             (
+                "${NB_USER}",
+                # Install rmarkdown library (for background, see https://github.com/jupyter/repo2docker/issues/799, as well as Section 3.2 at https://docs.rstudio.com/shiny-server/)
+                r"""
+                R --quiet -e "install.packages('rmarkdown', repos='{}', method='libcurl')"
+                """.format(
+                    mran_url
+                ),
+            ),
+            (
                 "root",
                 # We set the default CRAN repo to the MRAN one at given date
                 # We set download method to be curl so we get HTTPS support


betatim:
Why does the `rmarkdown` package have to be installed by default/for everyone? Couldn't a repository that wants to use it install it via `install.R` or `environment.yml`?

pablobernabeu:
I don't really know, unfortunately.

In my experience, having the rmarkdown package in the `install.R` script isn't enough. My Binder repo is working well, apart from the Shiny environment.

pablobernabeu:
Hi @choldgraf and @betatim,

This is working now! 

https://mybinder.org/v2/gh/pablobernabeu/Experimental-data-simulation/master?urlpath=shiny

Thanks for your attention.

PR #896, opened 2020-05-15
Update Dockerfile to add Docker

hamelsmu:
I'm trying to add Docker into the Docker image so that I can run docker-in-docker with repo2docker which will facilitate GitHub Actions I am building at GitHub

I want to use repo2docker to power [codespaces](https://github.com/features/codespaces/) and this unblocks that for me.  I would rather depend on the official Jupyter image rather than maintaining our own.  Having this dependency will also unblock other future features. 

cc: @betatim @willingc

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,8 +16,7 @@ RUN mkdir /tmp/wheelhouse \
 FROM alpine:${ALPINE_VERSION}
 
 # install python, git, bash
-RUN apk add --no-cache git git-lfs python3 bash
-RUN apk add docker
+RUN apk add --no-cache git git-lfs python3 bash docker
 
 # install repo2docker
 COPY --from=0 /tmp/wheelhouse /tmp/wheelhouse
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@ FROM alpine:${ALPINE_VERSION}
 
 # install python, git, bash
 RUN apk add --no-cache git git-lfs python3 bash
+RUN apk add docker
 
 # install repo2docker
 COPY --from=0 /tmp/wheelhouse /tmp/wheelhouse


betatim:
(the failing codecov check is unrelated/codecov is behaving weirdly at the moment)

I think this is fine.

Trying to think through if this has any security implications for people who run repo2docker as part of something like a BinderHub. I think the answer is no. Which means we can merge this.

Looking forward to trying out codespaces with r2d support! (I'm still on the waiting list _hinthintnudgenudge_ 😀)

willingc:
Hi @hamelsmu, I love the concept of codespaces and think it will be a great benefit to education.

My one concern is security and whether the docker container opens up privileged access. https://docs.docker.com/engine/security/security/  https://snyk.io/blog/10-docker-image-security-best-practices/

Let's run this through hadolint to see if it flags anything.

Pinging @minrk as well for comment.

willingc:
FYI hadolint doesn't flag anything related to the addition. So I'm a +1 unless @minrk has additional thoughts.

Dockerfile
```diff
@@ -17,6 +17,7 @@ FROM alpine:${ALPINE_VERSION}
 
 # install python, git, bash
 RUN apk add --no-cache git git-lfs python3 bash
+RUN apk add docker
```
manics, line 20: This adds an extra Docker layer and alsos leave some apk cache files in the image. Is there a reason it can't be included in the previous line?

hamelsmu: Hi @manics great suggestion.  Made the change

betatim:
Merging this now.

One thing I'd like to understand more about is why we need to install docker. In BinderHub we run the repo2docker image as a separate Pod and provide access to a dockerd via mounting a socket into the container. The socket comes from a different Pod running DINd (docker in docker). So this question is mostly for my education/understanding of how this is being used.

Will security/snyk always add a check to PRs now Carol? If no could we make it so it does?

betatim:
The new image is about 100MB compared to 35MB before. This by itself isn't a disaster but motivates me to understand why we need to install docker/what exactly @hamelsmu wants to do inside a container based on this image :D. (I don't know if this PR is the best venue or if we should move to an issue or forum thread)

hamelsmu:
Hi Tim, the reason for the presence of Docker is so I can build a Docker
image directly from the jupyter/repo2docker image on DockerHub.

I am doing this inside GitHub Actions.  The alternative is to install
repo2docker in the Actions runner itself, and then proceed from there.
This installation takes a long time so I was trying to bundle everything
needed to use repo2docker inside the container itself.

It’s possible I’m missing something.  How are you using the repo2docker
image right now when Docker is not installed?

willingc:
@betatim snyk I think will run on each PR but if you notice it doesn't we can set it up under the Jupyter org and not my user.

@hamelsmu Let me see if I understand what your desired workflow is:

1. You have a repo that you wish to build a docker image for.
2. You want GitHub Actions on some specified condition to take the repo and use repo2docker to get a Dockerfile, build and run the created image
    - **Consideration**: To use repo2docker, you need docker running. According to the repo2docker [Doc page](https://repo2docker.readthedocs.io/en/latest/usage.html#using-repo2docker), we assume that docker is running for repo2docker to work.
3. Use docker to launch the image for the user

> It’s possible I’m missing something.  How are you using the repo2docker
image right now when Docker is not installed?

I don't think we are using repo2docker without a running docker.

hamelsmu:
Carol, yes you are correct that is exactly what I’m trying to do.  Can
someone show me how you are able to use Docker from inside the repo2docker
container without Docker being installed in that image?  If there is some
way, perhaps we can revert the change in this PR.  I wasn’t able to think
of a way but I could be overlooking something

betatim:
In a BinderHub we launch a new Pod that uses `jupyter/repo2docker:0.11.0-23.g123123` as its image. We make a docker socket available inside that Pod by mounting it from a path on the host. The command executed in the container is
```
jupyter-repo2docker
      --ref
      8d05a8b750d3d56bf80a6031768f167547564748
      --image
      gcr.io/binder-prod/r2d-f18835fd-dask-2ddask-2dexamples-49f972:8d05a8b750d3d56bf80a6031768f167547564748
      --no-clean
      --no-run
      --json-logs
      --user-name
      jovyan
      --user-id
      1000
      --push
      https://github.com/dask/dask-examples
```
which will talk to a dockerd via the socket and then push the image, etc. If you type `docker` into a shell inside that container it won't find anything.

What exactly do you mean when you say "run docker'? Is it run a docker daemon or use the docker CLI to talk to a docker daemon that is running somewhere else? The reason repo2docker can act as a client to a dockerd without docker being installed is because the Python library we use implements things itself.

I had assumed that a docker socket would "magically" appear inside a container in GH Actions, with the actual dockerd running outside of the container. Similar to what we do for BinderHub where the dockerd is running in its own container or on the host.

hamelsmu:
This is the shell script that I'm attempting to run:
https://github.com/machine-learning-apps/repo2docker-action/blob/master/create_docker_image.sh

I didn't consider the fact that you could mount the docker socket, and that
could work inside GitHub Actions.  However, there are some remaining
problems:

- I run `docker pull` so that repo2docker can use the `--cache-from`
argument by default.  This calls `docker` via the shell so it is
problematic.
- I apply multiple tags to the docker image before pushing it, which again
uses the shell command `docker tag`
- I need to package all dependencies in a single container as that is a
prerequisite for hosting a GitHub Action for other people to use.

Hope that helps, let me know if having Docker as a dependency in this
container is still reasonable.  Another route is to add the ability to do
all of the things I mentioned above via the `repo2docker` command such that
I don't have to call `docker` directly, in which case I would not need this
dependency.

Let me know your thoughts.  Thanks in advance

PR #899, opened 2020-05-25
Add Julia 1.4.2 support

davidanthoff:
Julia 1.4.2 was released today, so this adds support.


betatim:
Close and re-open to re-trigger the dockercloud CI integration. I couldn't see what was broken/the problem there but I suspect it was just flakey.

PR #908, opened 2020-06-08
[WIP] Add gitpod.io config for docs

betatim:
Getting sphinx setup and working correctly is pretty hard, especially if you just want to work on the docs (imagine you are a technical writer with a windows laptop who hasn't contributed before). This PR adds configuration that will setup and configure a work environment targets for documentation changes.

gitpod.io is like mybinder.org but focussed on "dev environments". They give you a theia (I think) UI that is very similar to VS code in your browser. With the config in `.gitpod.yml` sphinx will be installed and started to build&serve the docs as you edit them.

Because you need to login with GitHub to use it, you can then directly push to GH when you are done.


betatim:
The link to try this is https://gitpod.io/#https://github.com/betatim/repo2docker/tree/betatim-patch-1

betatim:
I think this is ready for some reviewing and feedback now. If we like it we can maybe add a link to the "final" setup in `CONTRIBUTING.md` and/or the README. I'll do that in a second PR where we can check that the links work.

betatim:
What happens when you click the link above? After authorising the gitpod app you will see a short loading screen and then this:
Screenshot 2020-06-08 at 19 22 31

The README is displayed and in the bottom tab dependencies are being installed. Eventually it'll switch to this:

Screenshot 2020-06-08 at 19 22 59

The popup tells you that you can open a new tab that will show the built docs. And if you edit something in the docs, the build with automagically update and reload (I think the kids call it "hot reloading").

Because it uses GitHub for auth you can commit, push, etc directly from there. I think it is pretty neat

manics:
I tried it, looks pretty cool! Would be good to get someone from the target audience to try it out. My suspicion is it's still tricky to navigate the editor, it took me a while to figure out how to commit and push changes from the editor, though maybe that could be handled by an addition to the docs?

choldgraf:
wow, this is really cool!

I'm also curious how this intersects with the upcoming GitHub Codespaces product...they seem quite overlapping. As a general rule I am a big fan of not depending on Microsoft for yet another part of my stack, so I'd love to give this one a shot :-)

betatim:
on the topic of codespaces: https://www.gitpod.io/blog/github-codespaces/ I think online IDE setups like this have been the future for a while (theia was the first "IDE in binder" binder I made because I had seen a demo of gitpod.io).

I just learnt from the docs that we can control which file is opened by default in the editor, so we could open one that gives people a bit of orientation.

The cost of adding this seems to be small, should we give it a try? If we could work out how to get all the dependencies to work it would be super cool to offer this as a way to work on repo2docker or BinderHub, maybe a project for "next week".

Do either of you use VS Code? I thought this was similar to it, but because of reasons I don't use VS Code so I don't actually know :D

manics:
I use VSCode, that's what I thought it was until I read your comment properly 🙂 .

choldgraf:
I agree - this sounds great and could be quite helpful in development. The only reason I haven't merged is that we have a few failing builds: https://dev.azure.com/jupyter/repo2docker/_build/results?buildId=258&view=logs&j=a9282be4-63ec-5f90-0c3f-63cd4df70f7f&t=490fe5e2-ddd3-5191-6021-7d79fc5c11d3&l=510

also FWIW, here's an interesting post on vscode vs. theia: https://blogs.eclipse.org/post/mike-milinkovich/eclipse-theia-and-vs-code-differences-explained

PR #912, opened 2020-06-12
[MRG] Make the memory limit test simpler

betatim:
Instead of checking that a build which allocates too much memory does
fail, we now only check that we pass the correct arguments to the Docker
API client. It seems reasonable to rely on the docker client working and
doing the right thing. This solves a problem where our CI is flakey
because the kernel of the VM the tests run on doesn't support this.

Hopefully this will make our CI faster and less flakey.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,9 +1,9 @@
 # Note that there are a dev-requirements.txt and a docs/doc-requirements.txt
-# files in this project, if you are updating this Pipfile make sure to 
+# files in this project, if you are updating this Pipfile make sure to
 # update these accordingly
 
 [dev-packages]
-pytest=">=3.6"
+pytest=">=4.6"
 wheel="*"
 pytest-cov="*"
 PyYAML = "*"
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,7 +1,7 @@
 # Note that there is also a Pipfile for this project if you are updating this
 # file do not forget to update the Pipfile accordingly
 pyyaml
-pytest>=3.6
+pytest>=4.6
 wheel
 pytest-cov
 pre-commit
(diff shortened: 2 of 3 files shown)


manics:
Look like the new release of pytest-cov which requires `pytest>=4.6` broke travis:
https://github.com/pytest-dev/pytest-cov/blob/master/CHANGELOG.rst#2100-2020-06-12
Though I can't see why, https://github.com/jupyter/repo2docker/blob/8d85a51cd2b8d7dee712d3c1b283de1af734396d/dev-requirements.txt#L4 only has a lower bound, not an upper bound.

betatim:
I think we got `pytest==3.6` because that got installed previously and was in the build cache. Cleared the caches and restarting the build.

betatim:
The mystery continues, the pytest-cov `setup.py` contains this:

https://github.com/pytest-dev/pytest-cov/blob/694f7fd497b07aaea76d96e484068faf440e8301/setup.py#L124-L127

so why doesn't `pip` install a newer version of pytest?

betatim:
Pushed a new commit that _removes_ the minimum version requirement for pytest that we have in `dev-requirements.txt`. If I understand the pip issue linked in the previous comment then this might lead to us getting a version of pytest that is driven by pytest-cov's requirement.

Side comment: what do I do with the `Pipfile.lock`? How does that get updated? It also makes me wonder why we have that file checked in. repo2docker is a library, so we don't need to exactly reproduce the installed versions from my laptop to all other dev's machines. Anything within the limits of our `dev-requirements.txt` should work and if not that is a bug. So maybe we shouldn't checkin the lock file? (If we were generating an env for a production system then I'd want a lockfile so that rebuilding that env across time is stable but I don't think we have the requirement for a dev environment.)

betatim:
The robots are happy 🎉! The codecov test fails but I think that is expected.

betatim:
There is one approval and the bots are happy so merging this.

PR #914, opened 2020-06-15
doc: runtime.txt installs python x.y (& concise rewording)

mdeff:


betatim:
CI is failing for the same reason it was failing in #912. We can either rebase on master or merge despite CI failures.

docs/source/config_files.rst
```diff
@@ -222,23 +222,13 @@ For these cases, we have a special file, ``runtime.txt``.
...
-
-To see an example R repository, visit our `R
-example in binder-examples <https://github.com/binder-examples/r/blob/master/runtime.txt>`_.
+Have ``python-x.y`` in ``runtime.txt`` to run the repository with Python version x.y.
+See our `Python2 example repository <https://github.com/binder-examples/python2_runtime/blob/master/runtime.txt>`_.
 
+Have ``r-<RVERSION>-<YYYY>-<MM>-<DD>`` in ``runtime.txt`` to run the repository with R version RVERSION and libraries from a YYYY-MM-DD snapshot of `MRAN <https://mran.microsoft.com/documents/rro/reproducibility>`_.
+RVERSION can be set to 3.4, 3.5, 3.6, or to patch releases for the 3.5 and 3.6 series.
+If you do not specify a version, the latest release will be used (currently R 3.6).
+See our `R example repository <https://github.com/binder-examples/r/blob/master/runtime.txt>`_.
```
betatim, line 231: What do people think of removing this section? A lot of it doesn't actually work, we thought it did but in practice it doesn't. I now recommend https://github.com/binder-examples/r-conda but that often triggers a negative response in the R community. So I am not sure what to recommend.

The reason selecting the R version via `runtime.txt` doesn't work is that the PPA we install the apt packages from only provides one version. They have multiple ones which made us think we could use them to install different versions of R but when that broke and we asked the reply was "yeah, that is off-label use. No promises that will work"

mdeff: So it's more the "way it should work" that the "way it works". What about a warning/info box with what you wrote? Encourage conda or a contribution if they know how to fix that. (I'm only an occasional R user but understands the push against conda.)

mdeff: PS: MRAN works but RVERSION doesn't right?

mdeff:
rebased on master for CI

docs/source/config_files.rst
betatim: Correct MRAN works but RVERSION doesn't.

It is more "we wish there was a way to do this, but apparently not" :(

Issue #911, opened 2020-06-11
nixpkg-unstable channel requires nix>=2.2


jboynyc:
### Bug description

I initially filed this bug [here](https://github.com/jupyterhub/binderhub/issues/1110). Trying to launch Binder with a repo containing a `default.nix` file currently fails because adding the nixpkg-unstable channel (which happens [here](https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/nix/__init__.py#L64)) requires nix>=2.2 but repo2docker's is [pinned](https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/nix/install-nix.bash#L5) to version 2.1.1.

#### Expected behaviour

Build an image that can run nix-shell.

#### Actual behaviour

Building the image fails.

### How to reproduce

Try to start Binder using the Github repository jboynyc/textnets or any other that contains `default.nix` and hasn't had an image created for it yet.

betatim:
Is there something we should consider before moving from 2.1 to 2.2? Are there things that used to work in 2.1 and now stopped working that would break repos which are currently working if we upgrade?

Otherwise this would be a good issue to get started on contributing to repo2docker by updating the nix version pin to 2.2.

Thanks for reporting this! repo2docker uses several build tools and the maintainers only use a subset of those daily. This means we rely on keen users to help us stay on top of best practices and going ons in the sub-communities.

jboynyc:
I'll try to find out!

jboynyc:
I took a swing at it @betatim. After studying previous commits and nix change logs, I couldn't find any reason for sticking with the old version.

jboynyc:
#915 closes this issue after [this commit](https://github.com/jupyter/repo2docker/commit/8bbced7ded5a21b581f1f3846ffc9f87944ba799).

PR #915, opened 2020-06-15
Bump nix version to 2.3

jboynyc:
- This addresses issue #911
- This is untested, but there is no documented reason why repo2docker
  should stick with the previously pinned version 2.1.1
- Upstream has switched from bz2 to xz, so `tar` now takes the `-J` flag
  instead of `-j`

### DIFF
--- a/repo2docker/buildpacks/nix/install-nix.bash
+++ b/repo2docker/buildpacks/nix/install-nix.bash
@@ -2,13 +2,13 @@
 # This downloads and installs a pinned version of nix
 set -ex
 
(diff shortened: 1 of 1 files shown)


betatim:
Looks good to me. Should we merge it?

jboynyc:
Cool, thanks for taking a look. Please go ahead.

FRidh:
Nix 2.3 enables sandboxing by default. Building inside a Docker container, while Nix is having sandboxing enabled is not possible. Thus, `sandbox = false` should be set in `/etc/nix/nix.conf`.

PR #918, opened 2020-06-26
allow user to specify a single port with default command

minrk:
Allows specifying `-p9999` or `-p5555:9999` without also needing to override the default command

app retrieves port arguments from length-1 port mapping for default command, instead of requiring random port OR overriding the whole command.

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -354,7 +354,7 @@ def make_r2d(argv=None):
         )
         sys.exit(1)
 
-    if args.ports and not r2d.run_cmd:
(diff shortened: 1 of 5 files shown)


minrk:
Notable difference from the docker run `-p` arg: specifying just `8000` means exposing container port 8000 *as host port 8000* rather than container port 8000 as random host port. I think this makes more sense in a repo2docker context, but it is still a deviation from the upstream argument that this inherits from.

consideRatio:
I rebased on this on current main branch and stashed away the previous state [in my fork](https://github.com/consideratio/repo2docker/tree/allow-one-port)!

consideRatio (approved):

PR #919, opened 2020-06-26
try mamba again

minrk:
might save us some time

adds timing measurement to most of the commands that take a while, not sure if we want to keep that.

Extracting some timings with mamba for `tests/conda/py37` and conda at various stages:

```csv
installer,download,bootstrap,mamba,install,lab,total,env
mamba,9.471,20.4,23.976,68.776,154.116,289.65,56.628
conda,8.33,21.946,0,111.27,153.573,307.038,61.1
```

And a plot:

Screen Shot 2020-06-26 at 15 11 25

mamba is faster than conda in this case, but not by a lot, and we have to pay some of that to get mamba installed in the first place. If it were in miniforge (https://github.com/conda-forge/miniforge/issues/23), that would save the ~20s used to get mamba. But the initial env bootstrap is completely dominated by the `jupyter lab build` call to get offline notebook. I think minimizing/avoiding `jupyter lab build` is actually one of the biggest things we should do to save time. We spend at least as much time doing lab build as conda install.

In testing, this is blocked by https://github.com/TheSnakePit/mamba/issues/360 because `mamba env update` doesn't work, which is really the main thing we want to use mamba for.

/cc @SylvainCorlay


bollwyvl:
Oooh, this would be nice. Maybe worth another look? 

If nothing else, just _having_ the `mamba` executable around would be great, if only for the runtime `!mamba install package-i-forogt` being so much faster when live coding. We were also able to [`mamba env update`](https://github.com/deathbeds/_fam/blob/07-11-2020/postBuild#L4
) a pretty [big env](https://github.com/deathbeds/_fam/blob/07-11-2020/mamba.yml) from `postBuild`. The solves and downloads were really quite snappy.

[micromamba](https://github.com/TheSnakePit/mamba#micromamba) might be a bridge too far, but might become pretty compelling, as the whole shooting match is 4.5mb... though I don't know what the "archival-grade" URL would be for downloading it just yet.

SylvainCorlay:
It seems that https://github.com/TheSnakePit/mamba/issues/360 is now fixed, although there is a follow up issue on how to behave with respect to the update of packages of the base environment.

minrk:
I think micromamba is pretty compelling, and probably something we should aspire to in this repo, especially since we don't use the root environment at all. The biggest benefit would be if we stop installing conda at all (esp. leaving out Python itself in the root env), which would be pretty interesting. I'm not certain we're ready to do that yet, though.

SylvainCorlay:
Things appear to work now!

bollwyvl:
Whoa, awesome!

While contemplating tool churn, another one to consider is [conda-lock](https://github.com/mariusvniekerk/conda-lock), which might simplify some of the freezing capabilities... it will preferentially use `mamba`, if installed, and doesn't depend on `conda` at all (`pyyaml`, `requests` and `jinja2`). It generates the `requirements.txt` flavor created by `(mamba|conda) list --explicit`, which invokes neither the solver, nor (I think) downloading any `repodata.json`, as it contains full URLs to packages. On `conda-forge`, the practice of (re)moving broken packages has been replaced by patches of `repodata.json`, so these URLs should be stable. 

A _real_ downside, though, is that the this file format still only supports the no-longer-trustworthy md5 hashes, even though most of the other tooling supports sha256.

Food for thought! Will be super excited if this lands!

SylvainCorlay:
Build is failing over

```
##[error]We stopped hearing from agent Azure Pipelines 3. Verify the agent machine is running and has a healthy network connection. Anything that terminates an agent process, starves it for CPU, or blocks its network access can cause this error. For more information, see: https://go.microsoft.com/fwlink/?linkid=846610
```

and build logs are not available. However, *during* the build, we can see the logs and so far so good, all appears to be passing.

SylvainCorlay:
OK, it might be hanging with R (conda/r/verify).

PR #921, opened 2020-06-29
Set read and execute (a+rx) permission for the repo2docker-entrypoint file

### LINKED ISSUE #775: Error after building image:  /bin/bash: /usr/local/bin/repo2docker-entrypoint: Permission denied
### Bug description
running repo2docker on an empty directory (to get a default python environment), I get:

```
jupyter-repo2docker .
Using local repo ..
Using RBuildPack builder
...
Step 44/46 : COPY /repo2docker-entrypoint /usr/local/bin/repo2docker-entrypoint
 ---> Using cache
 ---> 1a386e8b05e3
Step 45/46 : ENTRYPOINT ["/usr/local/bin/repo2docker-entrypoint"]
 ---> Using cache
 ---> dea2b546b748
Step 46/46 : CMD ["jupyter", "notebook", "--ip", "0.0.0.0"]
 ---> Using cache
 ---> 71d561c90128
{"aux": {"ID": "sha256:71d561c901289655ff31f29b3209774cae2c3853ea67e132c05a1215a47db035"}}Successfully built 71d561c90128
Successfully tagged r2d-2e1567611552:latest
/bin/bash: /usr/local/bin/repo2docker-entrypoint: Permission denied

```
#### Expected behaviour

I thought the jupyter notebook would launch.

#### Actual behaviour

Error

### Your personal set up

 - OS: manjaro linux

 - Docker version: `docker version` 
Client:
 Version:           19.03.1-ce
 API version:       1.40
 Go version:        go1.12.7
 Git commit:        74b1e89e8a
 Built:             Sat Jul 27 21:08:50 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          19.03.1-ce
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.7
  Git commit:       74b1e89e8a
  Built:            Sat Jul 27 21:08:28 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.7.m
  GitCommit:        85f6aa58b8a3170aec9824568f7a31832878b603.m
 runc:
  Version:          1.0.0-rc8
  GitCommit:        425e105d5a03fabd737a126ad93d62a9eeede87f
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683

 - repo2docker version `repo2docker --version` 
0.10.0

The only thing I did differently was to install as --user rather than root:
```
python3 -m pip install --user jupyter-repo2docker
```

giannisdoukas:
Fix for error after building image: /bin/bash: /usr/local/bin/repo2docker-entrypoint: Permission denied

Fixes #775

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -188,6 +188,10 @@ ENV R2D_ENTRYPOINT "{{ start_script }}"
 
 # Add entrypoint
 COPY /repo2docker-entrypoint /usr/local/bin/repo2docker-entrypoint
+USER root
+RUN chmod a+rx /usr/local/bin/repo2docker-entrypoint
+RUN chown ${NB_USER}:${NB_USER} /usr/local/bin/repo2docker-entrypoint
+USER ${NB_USER}
 ENTRYPOINT ["/usr/local/bin/repo2docker-entrypoint"]
 
 # Specify the default command to run
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -189,8 +189,8 @@ ENV R2D_ENTRYPOINT "{{ start_script }}"
 # Add entrypoint
 COPY /repo2docker-entrypoint /usr/local/bin/repo2docker-entrypoint
 USER root
-RUN chmod a+rx /usr/local/bin/repo2docker-entrypoint
-RUN chown ${NB_USER}:${NB_USER} /usr/local/bin/repo2docker-entrypoint
+RUN chmod a+rx /usr/local/bin/repo2docker-entrypoint \
+ && chown ${NB_USER}:${NB_USER} /usr/local/bin/repo2docker-entrypoint
 USER ${NB_USER}
 ENTRYPOINT ["/usr/local/bin/repo2docker-entrypoint"]
 


minrk:
Thanks for the PR! I'm curious how this occurs, since the repro steps given in #775 don't appear to reproduce the issue (pip-installed repo2docker has the right permissions already, since the file itself has the right permissions). Can you share a link to a travis run or similar that shows this error? I wonder if there's a pip bug (or deliberate change) stripping execute permissions somewhere.

[Here's a travis build](https://travis-ci.org/github/minrk/junk/builds/706068987) showing that current repo2docker (0.11) doesn't appear to have this issue, at least on Travis linux environments. Can you provide some environment info (or better yet, a CI run to link to)?

I have no objection to this patch, since the resulting tripling of the layer size doesn't matter for this one tiny file, but I would like to understand more about when this can happen. I suspect we may be relying on file permissions in more than one place.

Another option might be to make sure we set the permissions in the TarFile for the build context, without adding the extra directives. But since I can't manage to create an environment where this issue occurs, I can't be sure that would actually fix it.

giannisdoukas:
I don't know what happened but the day that I had the issue the owner of the entrypoint file another user but now I cannot reproduce it. I logged in with the debug and checked the permissions of the files. 
What I can show you now is that when I tried to make the patch when I changed the Dockerfile and I didn't become the root before changing the permission the CI failed. 
https://travis-ci.com/github/giannisdoukas/repo2docker/jobs/355375233

giannisdoukas:
I don't know if it is helpful but you can see that in my project I hide the issue by giving execution permission to the host machine. 

```sh
$ chmod +x $(python -c "import repo2docker; print(repo2docker.__file__[:-11])")/buildpacks/repo2docker-entrypoint
```
[With Error](https://travis-ci.com/github/giannisdoukas/ipython2cwl/builds/173487825)
[After adding chmod command](https://travis-ci.com/github/giannisdoukas/ipython2cwl/builds/173544159)

In the logs of my tests, the error message about the permission is hidden. I logged with the debug and I found out what was the issue.

westurner:
Having this issue when building on Fedora:
```bash
pip install jupyter-repo2docker
git clone https://github.com/wrdrd/docs
cd docs
repo2docker -E .
```
```bash
/bin/bash: /usr/local/bin/repo2docker-entrypoint: Permission denied
```

I'll test this patch

manics:
@westurner the pypi release is quite old, would you mind testing the current master branch with and without this PR?

manics:
Is this problem still reproducible with the latest `master` R2D? There have been some big changes to the entrypoint recently.

If so could someone please provide the exact steps to reproduce, this may need to include details of how the host OS is setup.

consideRatio:
Since we haven't reproduced in 1½ year, I'll go for a close. It seems like a relevant thing to fix if someone reproduces it though - for now I'll go for a close as part of triaging open PRs.

repo2docker/buildpacks/base.py
```diff
@@ -188,6 +188,10 @@
 
 # Add entrypoint
 COPY /repo2docker-entrypoint /usr/local/bin/repo2docker-entrypoint
+USER root
+RUN chmod a+rx /usr/local/bin/repo2docker-entrypoint
+RUN chown ${NB_USER}:${NB_USER} /usr/local/bin/repo2docker-entrypoint
```
westurner, line 193: PERF: Merge into one `RUN` image build layer with `&&`?

consideRatio, line 193:
```suggestion
RUN chmod a+rx /usr/local/bin/repo2docker-entrypoint \
 && chown ${NB_USER}:${NB_USER} /usr/local/bin/repo2docker-entrypoint
```

westurner:
- [ ] TST: add a test for this; and BLD: run it somewhere
  - [ ] BLD,TST: script to check the image
    - https://molecule.readthedocs.io/en/latest/faq.html#are-there-similar-tools-to-molecule
      - https://testinfra.readthedocs.io/en/latest/examples.html#parametrize-your-tests
    - https://testanything.org/producers.html#shell
- [ ] DOC: are there any docs needed?

minrk:
This is already tested because every test we have that runs a repo2docker image executes this file, requiring it to have execute permissions. So it's only some unusual environments where this can occur. If this were a general issue, literally all repo2docker-built images would fail to start.

> so why not also the entrypoint script?

it's not a huge deal because it's a small file, but touching a file's permissions does duplicate the file in the image layers with no actual changes, since the file is already correct with the correct permissions. This adds layers to the image with no  content.

Something weird is going on, because something is unusual about the environments where this doesn't work, since it requires the file having the wrong permissions on the host, which don't match the permissions in the repo or the package on PyPI.

That said, it's probably worth more to have the extra layers than to put time into understanding the weird environments which create the file with the wrong permissions. But since it is a workaround, it might be worth adding a comment to that effect, e.g.

```
# workaround entrypoint file sometimes losing execute permissions on the host
```

Another (better?) fix would be to ensure the execute bit is set in the [filter_tar function](https://github.com/jupyterhub/repo2docker/blob/main/repo2docker/buildpacks/base.py#L570) when adding the file in the tarball. That wouldn't require modification in the Dockerfile, extra layers, etc.

westurner:
- Dockerfile/Containerfile COPY [--chown=] [--chmod]
  https://docs.docker.com/reference/dockerfile/#copy
- [ ] BUG: allow python3-login and repo2docker-entrypoint to be executed by non-root users
  ```dockerfile

  COPY --chmod=0755 /python3-login /usr/local/bin/python3-login
  COPY --chmod=0755 /repo2docker-entrypoint /usr/local/bin/repo2docker-entrypoint
  ```

I'm having the same or a similar issue with Podman (`repo2podman`) on Fedora 40:


```
STEP 47/50: COPY /python3-login /usr/local/bin/python3-login
--> 76274f47d609
STEP 48/50: COPY /repo2docker-entrypoint /usr/local/bin/repo2docker-entrypoint
--> 36c8ff3a2e96
STEP 49/50: ENTRYPOINT ["/usr/local/bin/repo2docker-entrypoint"]
--> 52facf214f7e
STEP 50/50: CMD ["jupyter", "notebook", "--ip", "0.0.0.0"]
COMMIT r2d-2e1720530461
--> 19084b87ff9e
Successfully tagged localhost/r2d-2e1720530461:latest
19084b87ff9e216e53d027c4484d5b2483b98444cc72e1a0349ed2d4780fdde3
/bin/bash: /usr/local/bin/python3-login: Permission denied
126
CONTAINER FINISHED RUNNING.
ffae2e010a95c4bf4a1954e491a35b16509c0f8f406176187b89ecdd9bf30155
make: *** [Makefile:17: repo2docker] Error 126
```
```
/bin/bash: /usr/local/bin/python3-login: Permission denied
```
```sh
$ ls -alZ /usr/local/bin
$ ls -alZ /usr/local/bin
total 13956
drwxr-xr-x.  2 root root system_u:object_r:fusefs_t:s0     4096 Jul  9 15:45 .
drwxr-xr-x. 10 root root system_u:object_r:fusefs_t:s0     4096 Jul  9 15:45 ..
-rwxr-xr-x.  1 root root system_u:object_r:fusefs_t:s0 14272032 Mar  4 05:41 micromamba
-rwxr-x--x.  1 root root system_u:object_r:fusefs_t:s0      374 Jul  9 09:07 python3-login
-rwxr-x--x.  1 root root system_u:object_r:fusefs_t:s0     3356 Jul  9 09:07 repo2docker-entrypoint
lrwxrwxrwx.  1 root root system_u:object_r:fusefs_t:s0       28 Jul  9 15:45 xdg-open -> /usr/bin/distrobox-host-exec
```

westurner:
This machine has a default umask of 0027.

PR #924, opened 2020-07-02
avoid deprecated import of collections.abc

minrk:
abcs should be imported from collections.abc, not collections itself, which is deprecated since 3.3, due to be removed in 3.9.


betatim:
The CI failure is unrelated but looks legit. Everything ran through but in the end a package we list in `install.R` isn't installed. Wut?

betatim:
Merged despite CI failure which was/is related to R and unrelated to this PR

Issue #926, opened 2020-07-07
config dir confusion


bitnik:
### Bug description

repo2docker doesn't allow having `binder` and `.binder` folders together:

https://github.com/jupyter/repo2docker/blob/8bbced7ded5a21b581f1f3846ffc9f87944ba799/repo2docker/buildpacks/base.py#L485-L501

#### Expected behaviour
But documentation says (https://repo2docker.readthedocs.io/en/latest/usage.html#where-to-put-configuration-files):

https://github.com/jupyter/repo2docker/blob/8bbced7ded5a21b581f1f3846ffc9f87944ba799/docs/source/usage.rst#L85-L93

and I expect that having both folders is okay and when I have both, then `binder` folder is picked for configuration.

#### Actual behaviour

repo2docker throws an error: "RuntimeError: The repository contains both a 'binder' and a '.binder' directory. However they are exclusive."

### How to reproduce

1. Create both `binder` and `.binder` folders at the root of your repo
2. run repo2docker with that repo
3. then you will get the error

### Your personal set up

 - OS: linux
 - Docker version: 19.03.12
 - repo2docker version: 0.11.0+98.g8bbced7

ref: https://github.com/jupyter/repo2docker/pull/653

-------

Actually first I wanted to update the documentation about the error and make a PR. But then I thought maybe it is better to change the logic in the code in this way:

- allow to have both folders in repos
- if both exists, pick `.binder`, so the order is: first check the hidden config folder `.binder`, then the config folder `binder` and then root dir. But I dont really have reasoning for this order. 

Proposed changed shouldn't effect any already configured repos, since it was not allowed to have both folders at the same time.

What do you think?

manics:
I think it's better for users, especially those new to binder/repo2docker, to see a clear error message than an easily missed warning, and be left wondering why it's not working because they're unaware there are two directories.

betatim:
What is the use case for having both folders but ignoring one? I think the original motivation for making it an error to have both was what Simon said. The most likely reason someone would have both is that they'd expect them to "stack" or that they created the second one by accident.

bitnik:
> What is the use case for having both folders but ignoring one? 

I have no use case. I was writing a writing a script to detect if a repo has a Dockerfile and then realized there is inconsistency between docs and code. And my motivation for allowing both folders was to restrict users less, but I also understand your motivation. So I guess at least we should update the documentation, right?

manics:
[...]

PR #927, opened 2020-07-08
update docs for config dirs

bitnik:
closes https://github.com/jupyter/repo2docker/issues/926

Hopefully it is now clearer. Please feel free to make suggestions!

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -90,8 +90,8 @@ Where to put configuration files
 * A folder named ``.binder/`` in the root of the repository.
 * The root directory of the repository.
 
-`repo2docker` searches for these folders in order (``binder/``, ``.binder/``,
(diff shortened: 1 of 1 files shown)


docs/source/usage.rst
```diff
@@ -88,8 +88,8 @@ Where to put configuration files
 * A folder named ``.binder/`` in the root of the repository.
 * The root directory of the repository.
 
-`repo2docker` searches for these folders in order (``binder/``, ``.binder/``,
-root). Only configuration files in the first identified folder are considered.
+Having both ``binder/`` and ``.binder/`` folders is not allowed.
+And if one of these folders exists, configuration files in that folder are considered only.
```
manics, line 92:
```suggestion
If one of these folders exists, only configuration files in that folder are considered, configuration in the root directory will be ignored.
```

Issue #929, opened 2020-07-09
`Rscript` does not return non-zero exit code when it fails


betatim:
### Bug description
repo2docker uses `Rscript` to install R dependencies and rely on it exiting with a status code !=0 if installing a package fails. This doesn't seem to happen (any more). As a result the failure to install a dependency of a package we install in the CI goes unnoticed.

#### Expected behaviour

When installing dependencies fails the build process should exit with an error.

#### Actual behaviour

The build process completes successfully but no dependencies get installed.

### How to reproduce

1. Go to `tests/` in the repo2docker repository
2. Run `pytest -s -x -v r/simple`
3. Wait for the build to complete
4. Run `docker images` to find the name of the (latest) image that you just build for `r/simple`
5. Run `docker run -it --rm r2d-2fusers-2fthead-2fgit-2fjupyterhub-2frepo2docker-2ftests-2fr-2fsimple1594276430 /bin/bash` (replace the image name with your local one)
6. Run `Rscript install.R`
7. Immediately after the previous command run `echo $?`. This should return a value different from 0 but doesn't

### Your personal set up

 - OS: OSX
 - Docker version: 19.3.08
 - repo2docker version `8bbced7ded5a21b581f1f3846ffc9f87944ba799`

betatim:
Debugging this further:
```
thead@c8d346a6ebaf:~$ Rscript -e "sdfsdf"
Error: object 'sdfsdf' not found
Execution halted
thead@c8d346a6ebaf:~$ echo $?
1
```
does the right thing. So the problem isn't that we set an option that influences error handling.

manics:
Is this helpful? https://stackoverflow.com/questions/26244530/how-do-i-make-install-packages-return-an-error-if-an-r-package-cannot-be-install

PR #930, opened 2020-07-09
[MRG] Update snapshot date for simple R test

betatim:
This fixes the problem in #929 by updating our test so that install don't fail any more.

This isn't a fix for the underlying problem that we don't detect failed installs though, so keeping #929 open.


betatim:
Output from the Stencila test:
```
installing to /srv/rlibs/00LOCK-r-lib-roxygen2-bbf259d/00new/roxygen2/libs
** R
** tests
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
*** copying figures
** building package indices
** installing vignettes
** testing if installed package can be loaded from temporary location
** checking absolute paths in shared objects and dynamic libraries
** testing if installed package can be loaded from final location
** testing if installed package keeps a record of temporary installation path
* DONE (roxygen2)
'/usr/lib/R/bin/R' --no-site-file --no-environ --no-save --no-restore --quiet  \
  CMD INSTALL '/tmp/RtmpRQ1gl1/devtoolsd92c773cbf/stencila-r-361bbf5'  \
  --library='/srv/rlibs' --install-tests

ERROR: dependency ‘CodeDepends’ is not available for package ‘stencila’
* removing ‘/srv/rlibs/stencila’
> Installation failed: Command failed (1)
There were 50 or more warnings (use warnings() to see the first 50)

>
> stencila::register()
Error in loadNamespace(name) : there is no package called ‘stencila’
Calls: :: ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted
Removing intermediate container 26a4a342c0fe
The command '/bin/sh -c R --quiet -e "install.packages('BiocManager'); BiocManager::install(); BiocManager::install(c('graph'))" && R --quiet -e "devtools::install_github('stencila/r', ref = '361bbf560f3f0561a8612349bca66cd8978f4f24')" && R --quiet -e "stencila::register()"' returned a non-zero code: 1
```

betatim:
Does anyone use https://github.com/stencila/r ? I just updated the git commit we use to the latest available one. It still doesn't build locally :( I noticed that the last commit was in March 2019 so maybe it isn't supported/developed any more?

cc @nuest who might know more

betatim:
Maybe @nokome has some ideas why we get the above error message when trying to install Stencila for R with R 3.6

nokome:
Hi Tim. I looks like a dependency of `stencila/r`, https://github.com/duncantl/CodeDepends, is not installing:

> ERROR: dependency ‘CodeDepends’ is not available for package ‘stencila’

I'll try to release a fix for that this week.

betatim:
Cool! Thanks a lot. Do you understand how come it used to work but then broke? Did a new version of CodeDepends get released or something like that?

nokome:
I triggered a build of `stencila/r` on Travis to see if I could replicate the problem there: https://travis-ci.org/github/stencila/r/builds/512676239. It is passing on `devel` and `release` (currently 4.0.0) versions of R, but failing on `oldrelease` (currently 3.6.3) with this error.

image

The error on 3.6.3 is related to the package `XML` which is a dependency of `CodeDepends` (by the same author): https://travis-ci.org/github/stencila/r/jobs/512676240#L3398. CRAN's own package checks suggest that there is may be a problem with `XML` on `oldrel`: https://cran.r-project.org/web/checks/check_results_XML.html

> Do you understand how come it used to work but then broke? Did a new version of `CodeDepends` get released or something like that?

In short, no I don't understand why this happened. However, it seems that `stencila/r` will install under R 4.0.0. You mention it failing under R 3.6 but the commit in this PR uses a snapshot data of 2020-07-07 which I would have thought would be using R 4.0.0 (?)

Let me know if this is still a blocker for you and if you think that there is anything I can do to help.  If you need to make this work under 3.6 I might be able to pin to an older version of `CodeDepends`. In more recent work we have moved away from using `CodeDepends` mainly due to complexities of installation (it requires a non-CRAN package `graph`). Removing it as a dependency from `stencila/r` is also an option.

betatim:
Thanks for looking into this. we don't select the version of R based on the CRAN date. Pinning to an older version of `CodeDepends` would be great (when on R 3.6).

In general we need to sort out the R story in repo2docker, but first we should get stuff that used to work working again :D

betatim:
Maybe we can already merge this as it fixes some of the failures in our CI. Another step would be to mark the stencila related tests as xfail for now to avoid having our CI being red.

What do you  @manics ?

manics:
Given this bug seems to be caused by changes to an external package I think it's reasonable to mark the stencila/r tests as xfail. Do you want to add it to this PR, so we have all the changes in one place in case we want to revert this? I don't understand why updating the r runtime works, but it seems to fix the r test `¯\_(ツ)_/¯`

betatim:
Turns out I don't know how to mark a test like the stencila-r one as xfail. This is because it is a special kind of test based on a directory, not a python function. For now I deleted it.

betatim:
stencila-r tests are disabled in the CI jobs but not deleted. I don't think there is a good way to mark them as xfail because of how they are setup/custom.

I don't understand why increasing the snapshot date helps, but without it it claims that ggplot isn't a library that it knows about, which seems unlikely.

This is ready for review and merge now I think.

Issue #925, opened 2020-07-04
Rstudio package manager 


choldgraf:
Looks like rstudio now runs a package manager too 

https://mobile.twitter.com/hadleywickham/status/1279023422748659712

We should look into it as a more reliable way of getting R packages in Linux!

kkmann:
Ha, I was just gonna ask about that. Imho, it's not only about reliability but also speed. Building the tidyverse alone nowadays takes ages...

Wouldn't this just require a few changes to https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/r.py to work with a lockfile instead of install.R?

I guess there is a design choice as to how one wants to handle runtime.txt: 
1.  give it precedence over the recorded package versions in the lockfile (imho awkward)
2. use it only for infrastructure packages not recorded in the lockfile (devtools, IRKernel, ...)
I don't think it would be user friendly to require a lockfile that contains all infrastructure packages as well.

choldgraf:
I don't think that it should be too difficult to make this change. The big question is whether or not this is a "canonical packaging solution for R". The RStudio package manager is super new, and (at least for me) it is hard to tell whether RStudio basically *is* the R community, or if there are others that would suggest different packaging approaches (e.g. the `holepunch` package leverages the Rocker community stacks). Maybe either @cboettig or @karthik could advise?

betatim:
I took a look at https://packagemanager.rstudio.com/client/#/repos/1/overview to try and understand how you'd use this. It looks like this is very similar to MRAN in that it offers "frozen snapshots" and that you'd configure it in a way very similar to how we configure MRAN.

If that is true this would cool because it would be a small change and compatible/no lock-in with using other mirrors.

choldgraf:
Yeah - plus I believe they have linux binaries for everything which would be a big improvement

cboettig:
yup, we're doing this on the rocker/versioned stack now as well.  (using the binaries built for `focal` on our ubuntu 20.04 based images.  Note that most *but not all* of the `bionic` RSPM binaries will work on `focal`.

betatim:
Does a binary from the RStudio CRAN "mirror" work with all R versions or do we have to pay attention to getting a right match (or does R take care of that?)?

cboettig:
@betatim that part is no different then installing from source.  (most packages are compatible across most versions, specific packages will declare a minimum R version number in  their dependencies.  `install.packages()` checks this whether doing the source or the binary version).

[...]

PR #931, opened 2020-07-10
[WIP] Demo rstudio package manager integration

kkmann:
# Summary

Demo PR for experimenting with the new RStudio package manager R package binaries, see #925 

# (Known) open issues 
- [ ] implement cutoff date for switching from MRAN to RSPM


kkmann:
if I understand the test correct, it's failing because the snapshot date is very old (2017) and packagemanager.rstudio is not reliable - just need to switch back to MRAN before say 2020-06-01

yuvipanda:
Wah, I was just going to find and implement this. I think this is the way to go! Let's pick a cut-off date and be done.

yuvipanda:
I think this is superseeded by https://github.com/jupyterhub/repo2docker/pull/1104

manics:
As mentioned in the previous comment this is done in https://github.com/jupyterhub/repo2docker/pull/1104

PR #934, opened 2020-07-27
validate_image_name: mention lowercase, fix formatting

manics:
In https://discourse.jupyter.org/t/how-to-build-mybinder-image-automatically-at-each-commit/5381 I noticed:
- the error message doesn't mention only lower-case letters are allowed for the Docker image name
- the formatting was incorrect, lines were concatenated without a space


betatim:
Merging because the failures are due to #669

PR #938, opened 2020-08-04
Add Julia 1.5.0 support

tomyun:
Julia 1.5.0 just released yesterday. Hopefully this works for Binder.

### DIFF
--- a/repo2docker/buildpacks/julia/julia_project.py
+++ b/repo2docker/buildpacks/julia/julia_project.py
@@ -29,6 +29,7 @@ class JuliaProjectTomlBuildPack(PythonBuildPack):
         "1.4.0",
         "1.4.1",
         "1.4.2",
+        "1.5.0",
     ]
 
     @property


betatim:
Our tests fail because packages like IJulia don't precompile. Not sure if this is expected or not when we upgrade Julia version.

Do you know more about what changed between 1.5 and the previous version?

betatim:
Actually, it is the Julia legacy test ([tests/julia/julialegacy_version-1.1](https://github.com/jupyterhub/repo2docker/tree/master/tests/julia/julialegacy_version-1.1)) for v1.1. which has started failing.

@davidanthoff any ideas what recently happened in Julia world that this would stop working?

davidanthoff:
Hm, good question... Maybe the Julia definition in the `REQUIRE` file is read as a semver version, so writing `julia 1.1` in `REQUIRE` means "any version that is compatible with 1.1", and that of course would include 1.5. BUT, I think 1.5 might have removed support for REQUIRE files...

If that is so, we should just cap the Julia version one can get via a `REQUIRE` file to Julia 1.4.2. The whole `REQUIRE` stuff is completely legacy in any case.

betatim:
Modified the julia buildpack locally to print out the version with `julia --version`. Three observations: it prints 1.0.2 for the 1.0.2 test (👍), it uses the buildpack for `Project.toml` based projects, and it runs to completion. So maybe this was a glitch because ... glitches happen?

Restarted the CI jobs.

PR #940, opened 2020-08-10
stop running tests on travis

minrk:
tests are running on azure pipelines now

travis only does deploy

follow-up to #939 to complete #232

TODO:

- [x] regenerate pypi encrypted creds for new repo, but I don't seem to have this. I see that @betatim last committed them. I'm not 100% sure this is necessary, but probably good to do now rather than wait for a release to fail.

closes https://github.com/jupyterhub/team-compass/issues/322


minrk:
Hm, Azure hasn't run on the new repo, might need to poke something on pipelines

minrk:
Had to update azure devops config to create a new project for the relocated repo. I guess azure doesn't link github permissions?

It's working now, but maybe we want to try out github actions on this repo?

minrk:
closes https://github.com/jupyterhub/repo2docker/pull/940

betatim:
Which issue/PR did you want to refer to with "closes #940"? Because 940 is this PR :D noticed after adding it to the top comment so it would auto close

betatim:
After moving the pipeline do we have to re-add people so they have permissions to restart/poke around the pipeline? I can't find the button to restart the "unit" job and I think it is because I lack permissions but then I seem to always take a lot of time to find the restart button :-/

minrk:
Oops! I meant https://github.com/jupyterhub/team-compass/issues/322. Fixed.

Not sure what's up with the repo2docker builds. Maybe I didn't set up the pipeline correctly? I opened #941 to figure that out.

minrk:
> After moving the pipeline do we have to re-add people so they have permissions to restart/poke around the pipeline? 

Yes, that's why I opened #941. What email should I add for you?

betatim:
betatim@gmail.com but I think moving to GH action as in #942 is better

PR #942, opened 2020-08-11
GitHub Actions

### LINKED ISSUE #941: manage Azure Pipelines permissions and/or try github actions
In migrating to the jupyterhub org, I had to set up Azure devops anew. I ran into some weirdness because DevOps appears to not have any real GitHub integration in terms of accounts and permissions. There are two results of this:

1. I apparently didn't (and don't) have permissions on the jupyter/repo2docker pipeline, and
2. I am now the *only* person with access to the new one

I didn't realize (or maybe I just haven't found it) that Azure had this limitation, and it seems like a pretty big downside. CircleCI, Travis, etc. all support delegating permissions to the repo itself, which means we don't need to manage yet another set of accesses.

So we need to either:

1. figure out the right way to manage permissions on the Azure pipelines (I can grant permission to anyone who wants it), or
2. switch to something like GitHub actions, which I'm pretty pleased with, especially based on what @sgibson91 has been doing across various repos

My current slight inclination is to give GitHub actions a try, since I haven't had particularly good experiences with Azure pipelines (their uniquely short log retention is a common source of frustration), and several tests had to be retried in #940 (at least) three times due to warnings like:

> The remote provider has asked us to wait until 8/10/2020 9:55:13 AM before attempting to provision a resource. This will cause delays in the running of this request.

which suggests the high capacity and short queues that were once part of the reason to migrate is no longer the case.
Maybe that was because of some different configuration in the pipeline that I set up wrong, since I can't see the original. But even if that's the case, having any configuration that's not in the repo itself is a mark against Pipelines for me. I'm not opposed to sticking with Pipelines and just working out the kinks we have left, if folks are more inclined that way.

What do folks think? The main reason we moved to DevOps was the much greater concurrency available in the free tier when compared to Travis (#744) and I wonder if GitHub Actions is similar?

minrk:
You can see the results of the Action runs on this branch [here](https://github.com/minrk/repo2docker/actions?query=workflow%3A%22Continuous+Integration%22+branch%3Aactions).

Looks like this was easier than I thought (yay for copying @sgibson91's workflows in mybinder.org-deploy!). This PR trades one Microsoft CI product (Azure pipelines) for another (GitHub actions). GitHub actions have *more* concurrency and capacity at the free tier, plus we don't have to manage a separate collection of permissions, which is a big plus.

includes #940 
closes #941 

Marked WIP for discussion, since we haven't actually decided to drop Azure, but this shows what it might look like.


.github/workflows/test.yml
```diff
@@ -0,0 +1,115 @@
...
+# - https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions
+# - https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow
+# - https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions
+
+name: Continuous Integration
+
+# Only trigger the workflow when pushing to master
+on:
+  - push
+  - pull_request
```
sgibson91, line 13: So I think in this configuration, this workflow will run on _every_ push, and _every_ opened PR, regardless of destination. Just wanted to check this was intentional?

manics: I think this makes sense- it matches what travis does when you enable it on your repo, and means people can run CI locally on their own branches without opening a PR

sgibson91: That's fair, just wanted to check :)

hamelsmu: I think you can leave off -`pull_request` as that is a subset of `push` in this situation

manics: The GitHub examples include both:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-a-list-of-events

Is it related to branches in the same vs different forks?

hamelsmu: Yeah, from a practical perspective the pull_request event and push event overlap quite a bit as you can imagine.  If you leave both here in this example I think you will burn unnecessary Actions minutes from your free quota.  They are different events but folks will do a push each time they update or open a PR so I think you need only one event in this case.

minrk: That was my intention, yes. Is there a reason not to do that? Maybe I should leave the `on` blank to use GitHub's own defaults?

hamelsmu: > That was my intention, yes. Is there a reason not to do that?

why do you want to run parallel tests on every push and on every change to a PR?  I think you only need push if I am understanding correctly?  I could be wrong.  Perhaps start with just `push` and see if that gives you what you are looking for?

sgibson91: Running `pull_request` does catch `refs/pull/:prNumber/merge` though which will catch any issues caused by the branch not being up-to-date with master, which I don't think `push` will catch?

minrk: > why do you want to run parallel tests on every push and on every change to a PR?

Because that's what ~all CI services do by default, right? I just want 'normal' default behavior (i.e. what I've come to expect from Travis, Circle, etc.), which is:

- build head commit of any push to any branch on this repo (won't generally include PRs, which come from forks)
- build merge commit, as @sgibson91 mentions, of any PR on each update / open / etc.

These do tend to duplicate CI runs on PRs from branches from the same repo, which is rare. What's the event combination where `merge` is preferred to `head` in the cases where both are built? If that can't be done, duplicate tests in rare cases seem preferable to losing the merge runs.

What would be the behavior if `on` were unspecified?

From [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-multiple-events-with-activity-types-or-configuration), maybe we should have opt-in for branch builds and use:

```yaml
on:
  push:
    branches:
      - master
  pull_request:
```

so all pull requests are built but only default-branch pushes are built (this will need updating if we have backport branches, other tests that want to run without open PRs, unlike what we have now, I think).

manics: :-1: on limiting tests to branches. It's a big hindrance to anyone wanting to run tests on their own forks and branches without opening a PR. For instance if you're playing around with a few different ideas and want your commits to be CI tested you either have to edit the CI config in all of your own branches to build your other branches, or open lots of potentially pointless PRs against the parent repo.

sgibson91: Yeah, I don't think we need to change the triggers. As far as the billing goes, I find this paragraph a little confusing, but the first line is "free for public repos" which I assume is equivalent to the usual statement of "always free for public repos and pay for private repos" https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions#about-billing-for-github-actions Let's burn the minutes and see if we run into any problems 😄

hamelsmu: Good point about branches not being up to date with master!   I usually turn on branch protection rules (which I forgot to mention) to solve that issue (under settings, screenshot below) that allows you to be confident that any tests are valid.  

image

But feel free to do both incase there is a use case I am missing beyond branches being updated to master (I just couldn't think of any)!

manics: I've been tempted to suggest the latter (require branches to be up to date) but I think it imposes an unnecessary burden on external contributors, especially as we usually don't review PRs immediately. This means there are likely to be concurrent PRs, and as soon as one is merged all other PRs need to be updated. I think test with `pull_request`/`merge` (and if necessary close/open a PR to force a retest against the latest branch) is more friendly.

manics: @hamelsmu Just wanted to add that despite me being against your suggestions I think it's very valuable input, and would definitely be useful elsewhere. Would be nice if we could record this discussion somewhere instead of it disappearing into the ether when this PR is merged 😀.

hamelsmu: No problem at all!  I am by no means attached to anything and am also learning by your thoughts and comments which are very valuable.  I can see the merits of the other approaches as well which I didn't consider before!

minrk: I also agree with @manics that the require-up-to-date rule is a contributor-busywork vs test rigor trade-off that doesn't quite pay off for our projects, but I now know it's an option, which is great, thanks @hamelsmu!

Thanks for everyone's input! This is super valuable, as there's always a variety of good ways to work with these tools. If I understood the latest comments correctly, I think we want to keep both pull_request and push triggers. Does that means there isn't more to do on this front?

It would definitely be useful to jot down some of the notes about events and what they mean and what downsides there are, e.g. summarizing from this discussion:

- pull_request builds `merge` commits, which is useful as other PRs are merged
- push events can be sufficient and may eliminate the benefit of pull_request building `merge` if branches are required to be up-to-date
- requiring branches to be up-to-date adds some work for contributors, in exchange for ensuring that tests have been run with *exactly* what will result after merge. Downside: extra work for contributors if maintainers take a while to merge PRs, as authors keep needing to update their branch while waiting for maintainers to review & merge
- push may not trigger on close/reopen PR, which is a common "poke CI" practice, especially for contributors who don't have access to the "re-run CI button"  (is this correct?)
- push & pull_request together can result in fairly redundant builds (HEAD and MERGE may differ slightly) when a PR is made from within the target repo
  - avoiding redundant builds by restricting push triggers to the default branch make it harder to test one's own forks without opening PRs (I do this a lot, including in writing this PR!)

Does that sound right?

minrk:
I realized that the actions aren't running on this PR, so folks need a link to runs on my fork, which are [here](https://github.com/minrk/repo2docker/actions?query=workflow%3A%22Continuous+Integration%22+branch%3Aactions)

betatim:
I'd be happy to switch to GH Actions. The UI of the Azure pipelines confuses me (trying to find the button that lets me restart things is always a 5-10min exercise and sometimes I give up :( ) and the fact that I have to login to Azure with a separate account and don't get redirected back to where I was doesn't help. So a big 👍  from me.

.github/workflows/test.yml
manics: Yes, with the caveat that "pull_request builds merge commits, which is useful as other PRs are merged" will be a merge at the time of the build, so if another PR is later merged you need to re-trigger CI. On travis you need to close/reopen the PR to force a build of the updated "merge", it's not enough to just re-run the last job. I guess it's the same with GH actions.

README.md
```diff
@@ -1,8 +1,8 @@
 # <a href="https://github.com/jupyterhub/repo2docker"><img src="https://raw.githubusercontent.com/jupyterhub/repo2docker/3fa7444fca6ae2b51e590cbc9d83baf92738ca2a/docs/source/_static/images/repo2docker.png" height="48px" /> repo2docker</a>
 
-[![Build Status](https://travis-ci.org/jupyterhub/repo2docker.svg?branch=master)](https://travis-ci.org/jupyterhub/repo2docker)
+[![Deploy Status](https://travis-ci.org/jupyterhub/repo2docker.svg?branch=master)](https://travis-ci.org/jupyterhub/repo2docker)
+[![Build Status](https://github.com/jupyterhub/repo2docker/workflows/Continuous%20Integration/badge.svg)](![Continuous Integration](https://github.com/minrk/repo2docker/actions))
```
manics, line 4:
```suggestion
[![Build Status](https://github.com/jupyterhub/repo2docker/workflows/Continuous%20Integration/badge.svg)](https://github.com/jupyterhub/repo2docker/actions)
```

.github/workflows/test.yml
sgibson91: Yes, "opened", "closed" and "reopened" are all events that will trigger GHA under `pull_request` https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request

hamelsmu:
One thing to keep in mind re: updating branches is when you turn on branch
protection GitHub surfaces a new button on PRs “update this branch with
master” that makes rebasing easy as a click for maintainers.  But, totally
get that we don’t want this in this project!  Just letting you know!

manics:
@hamelsmu I wasn't aware of that, definitely makes it a practical option! In the interests of getting this PR merged how about we keep the PR as it currently is, and make further changes in a followup? It's something we might want to discuss as a [Team Compass issue](https://github.com/jupyterhub/team-compass/issues) so we don't end up with too many different CI and branch setups across the organisation.

hamelsmu:
By all means please don’t hold up this PR!  I was only commenting as an
FYI.  I think this PR is good to merge LGTM!

minrk:
I'm learning a lot, @hamelsmu! Keep the insights coming :)

betatim:
Merge?

PR #943, opened 2020-08-13
[MRG] Handle requirements.txt with `--pre` lines

betatim:
This is a global flag which our inspection of `requirements.txt` files
should ignore as it doesn't imply a local reference.

### DIFF
--- a/repo2docker/utils.py
+++ b/repo2docker/utils.py
@@ -469,6 +469,10 @@ def is_local_pip_requirement(line):
     if line.startswith(("--requirement", "--constraint")):
         # as above but flags are spelt out
         return True
+    # the `--pre` flag is a global flag and should appear on a line by itself
+    # we just care that this isn't a "local pip requirement"
+    if line.startswith("--pre"):
+        return False
     # strip off things like `--editable=`. Long form arguments require a =
     if line.startswith("--"):
         line = line.split("=", 1)[1]
--- a/repo2docker/utils.py
+++ b/repo2docker/utils.py
@@ -473,14 +473,29 @@ def is_local_pip_requirement(line):
     # we just care that this isn't a "local pip requirement"
     if line.startswith("--pre"):
         return False
+
     # strip off things like `--editable=`. Long form arguments require a =
     if line.startswith("--"):
-        line = line.split("=", 1)[1]
+        # if splitting doesn't work it is probably because the line contains
+        # a syntax error or our "parser" is too simplistic so we return True
+        # just in case
+        try:
+            line = line.split("=", 1)[1]
+        except IndexError:
+            return True
+
     # strip off short form arguments like `-e`. Short form arguments can be
     # followed by a space `-e foo` or use `-e=foo`. The latter is not handled
     # here. We can deal with it when we see someone using it.
     if line.startswith("-"):
-        line = line.split(None, 1)[1]
+        # if splitting doesn't work it is probably because the line contains
+        # a syntax error or our "parser" is too simplistic so we return True
+        # just in case
+        try:
+            line = line.split(None, 1)[1]
+        except IndexError:
+            return True
+
     if "file://" in line:
         # file references break isolation
         return True
--- a/repo2docker/utils.py
+++ b/repo2docker/utils.py
@@ -463,48 +463,48 @@ def is_local_pip_requirement(line):
     line = line.split("#", 1)[0].strip()
     if not line:
         return False
+
     if line.startswith(("-r", "-c")):
         # local -r or -c references break isolation
         return True
+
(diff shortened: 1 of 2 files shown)


tests/unit/test_utils.py
```diff
@@ -146,6 +146,8 @@ def test_open_guess_encoding():
         ("git+https://github.com/jupyterhub/repo2docker", False),
         ("numpy", False),
         ("# -e .", False),
+        ("--pre", False),
+        ("--pre pandas", False),
```
minrk, line 150: Can you add a test for:

    "--unrecognized"

and make sure it doesn't error (I think it will)

repo2docker/utils.py
```diff
@@ -469,6 +469,10 @@ def is_local_pip_requirement(line):
     if line.startswith(("--requirement", "--constraint")):
         # as above but flags are spelt out
         return True
+    # the `--pre` flag is a global flag and should appear on a line by itself
+    # we just care that this isn't a "local pip requirement"
+    if line.startswith("--pre"):
+        return False
     # strip off things like `--editable=`. Long form arguments require a =
     if line.startswith("--"):
         line = line.split("=", 1)[1]
```
minrk, line 478: both here and L479 we have splits without first checking if the split character (`=` here, any space using `None` on L479) is present, meaning that `split(X, 1)[1]` can raise an IndexError. We should protect any such split from raising, either with:

- try/except or
- checking for the character's presence before splitting (harder with space), or
- checking len after split before requesting `item[1]`

In general: we shouldn't assume requirements.txt is valid and such errors should be raised by pip on `pip install`, not with errors in our own code.

tests/unit/test_utils.py
minrk: also `"-e"` with nothing after it

repo2docker/utils.py
```diff
@@ -469,14 +469,33 @@ def is_local_pip_requirement(line):
...
+    # we just care that this isn't a "local pip requirement"
+    if line.startswith("--pre"):
+        return False
+
     # strip off things like `--editable=`. Long form arguments require a =
     if line.startswith("--"):
-        line = line.split("=", 1)[1]
+        # if splitting doesn't work it is probably because the line contains
+        # a syntax error or our "parser" is too simplistic so we return True
+        # just in case
```
betatim, line 481: Is there a nicer way to handle the above comment so that it is simultaneously close to the code it is about and not repeated a few lines later?

repo2docker/utils.py
betatim: Is this what you had in mind? I didn't wrap the whole function in a try/except because I thought that might swallow errors we might actually want to see, but unsure.

I also went with returning `True` in the error case because either it is something we don't understand so better safe than sorry or it will later turn out to be a syntax error in which case the return value doesn't matter. WDYT?

repo2docker/utils.py
minrk: Yeah, that works. Or use:

Or for this one, we can check for presence of `=` sign before splitting, so we know that there will be two items:

```python
if line.startswith("--") and '=' in line:
    _, line = line.split("=", 1)
```

and handle length 1 or 2 in the later split:

```python
if line.startswith("-"):
    _, *rest = line.split(None, 1)
    if not rest:
        # no argument after `--flag`, skip line
        return False
    line = rest[0]
```

...or regular expressions!

```python
# pattern to split lines that start with one or two leading `-`
# takes 'argument s' off of:
# --flag=argument s
# -f argument s
arg_pat = re.compile(r"-+[^=\s]+[=\s](.*)")
match = arg_pat.match(line)
if match:
    line = match.group(1)
...

File: tests/unit/test_utils.py [hunk 4]
```diff
@@ -146,6 +146,14 @@ def test_open_guess_encoding():
         ("git+https://github.com/jupyterhub/repo2docker", False),
         ("numpy", False),
         ("# -e .", False),
+        ("--pre", False),
+        # pip ignores the package name and treats this like `--pre` on a line
+        # by itself
+        ("--pre pandas", False),
+        # These are invalid lines as far as pip is concerned, check that our
+        # code is robust and continues running
+        ("--unrecognized", True),
```
minrk, line 155: should return False, I think? Or do we want to make any unhandled flag result in the less optimized behavior?

repo2docker/utils.py
minrk: I'm inclined to return False because if there's no argument being passed, I think it's very unlikely that it's expressing a local dependency and more likely to be something configuring pip that isn't relevant to us.

betatim:
Updated the way we split things as well as switching to `False` when we don't recognise/correctly parse a line @minrk

minrk:
Awesome, thanks!

PR #950, opened 2020-09-04
[MRG] Add a Mercurial contentprovider

paugier:
MyBinder could support Mercurial repositories

See https://github.com/jupyterhub/binderhub/issues/1148


paugier:
@gracinet you may be interested to check what I did :slightly_smiling_face:

repo2docker/contentproviders/mercurial.py
```diff
@@ -0,0 +1,79 @@
...
+
+        cmd = ["hg", "identify"]
+        cmd.extend(hg_config)
+        sha1 = subprocess.Popen(cmd, stdout=subprocess.PIPE, cwd=output_dir)
+        self._sha1 = sha1.stdout.read().decode().strip()
+
+    @property
+    def content_id(self):
+        """A unique ID to represent the version of the content.
+        Uses the first seven characters of the git commit ID of the repository.
```
gracinet, line 77: docstring still says 'git'

repo2docker/contentproviders/mercurial.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        return self._sha1[:7]
```
gracinet, line 79: This is a minor detail, but do you really need to truncate it? I believe many people in Git land are worried about 7 chars being definitely too less, but I don't know what the expectations of the callers are.

Also, I would consider that it being a SHA-1 is an implementation detail (hopefully that'll change some day). The neutral Mercurial terminology being "node id", I would simply store `self.content_id`  from the `hg identify` call, but again, that's just nitpicking.

gracinet:
@paugier I don't know about the repo2docker context, but the Mercurial part itself is simple and should do the job.

Note that Mercurial requires Python >= 3.6 (apparently only Evolve actually enfores it through its requirements).  That explains the CI failure, so you probably want to add some conditionals in your requirements and imports if repo2docker wants to keep compatibility with 3.5 (let's see maybe what project maintainers think of it first).

paugier:
Is Python 3.5 really necessary since for binderhub `python_requires='>=3.6'` (https://github.com/jupyterhub/binderhub/blob/master/setup.py#L32) ?

dev-requirements.txt
```diff
@@ -6,3 +6,5 @@ wheel
 pytest-cov
 pre-commit
 requests
+mercurial
+hg-evolve
```
betatim, line 10: These should be in `requirements.txt` not in `dev-requirements.txt` I think as they are needed by users when they want to run repo2docker

betatim:
Nice work!

As Georges pointed out there are several comments and function names that still contain "git" which should be tidied up.

The ordered list of content providers is in

https://github.com/jupyterhub/repo2docker/blob/64371cbaa7abe12a95e538bf6a4985d1423add65/repo2docker/app.py#L144

You have to add the hg provider to that list (below the Zenodo one I think) for it to become available to users.

repo2docker/contentproviders/mercurial.py
paugier: I now use the full node id (12 characters).

dev-requirements.txt
paugier: It seems that there is no `requirements.txt`. The new requirements are in the Pipfile.

paugier:
I added Mercurial to the `content_providers` list. I'm not sure that it is the right order, but I added it just before Git because for some random http addresses (not clearly Git ones), we need to call `hg id -i https://...`, which is less fast than a local check.

dev-requirements.txt
betatim: Ah, sorry, we use `setup.py` directly, so the dependencies should be listed there.

paugier: Requirements added in setup.py!

betatim:
We now use Python 3.6 as the minimal require version for repo2docker. This means if you rebase your branch on the latest `master` the checks/selectors for Python 3.5 can be removed again.

paugier:
> We now use Python 3.6 as the minimal require version for repo2docker.

Ah good! It's going to be simpler!

Another question about `setup_requires`, requirements.txt and Pipfile: since the Mercurial requirements are in `setup_requires`, I don't need to put them in requirements.txt and Pipfile?

betatim:
We don't need `requirements.txt` but I don't really know how people who use `Pipfile`s setup things. Maybe @manics knows?

manics:
I don't know why we've also got a `Pipfile` in addition to requirements... I almost opened a new issue to ask why we have it :smile:

betatim:
The reason we have a `Pipfile` is because at some point people asked if we could add it because they use it to setup their dev environments.

manics:
I found this related issue: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1221
@minrk could you give us some help please? Does `Pipfile.lock` need to be updated in sync with `Pipfile`?

paugier:
Do you understand why the tests failed in Docker Cloud? I get a 404 error when I click on Details.

minrk:
> Does Pipfile.lock need to be updated in sync with Pipfile?

It should be. If you update the Pipfile, run:

    pipenv install --dev

this will update Pipfile.lock, and you can commit the difference. Since I don't think many folks use these, there's a good chance Pipfile.lock is out of date already.

The Pipfile and requirements file only directly contain dev dependencies, so if it's a runtime dependency adding to install_requires, no additional changes are required.

The docker build is failing because the image lacks a compiler, which mercurial needs because they don't publish wheels
 except for Windows (?!). FWIW, this also means that repo2docker would not be installable without a compiler on platforms other than Windows, which is a pretty big change, and makes me think this should be an *optional* dependency.

paugier:
> The docker build is failing because the image lacks a compiler, which mercurial needs because they don't publish wheels
except for Windows (?!).

Indeed it's very strange. I'm going to ask on Mercurial mailing list.

paugier:
Regarding Pipfile, I think it doesn't need to be updated in this PR.

betatim:
Could it be that we don't get wheels because the repo2docker image uses alpine?

manics:
AFAICT `pip install mercurial` is just a convenient way to install the `hg` executable, with no coupling at the level of Python modules? If so could we treat Mercurial it like `git`, i.e. it's the user's responsibility to install the git and mercurial binaries, e.g. using a distro package?

manics:
`mercurial` is available on Alpine, but the current repo2docker Dockerfile is based on alpine:3.9.4
https://github.com/jupyterhub/repo2docker/blob/f0c6f6540a42c4ef8ce424be2c51b2cb0b97e3c2/Dockerfile#L1
and the distro version of mercurial uses python2:
```
$ podman run -it --rm alpine:3.9.4
/ # apk add --no-cache mercurial
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/community/x86_64/APKINDEX.tar.gz
(1/11) Installing libbz2 (1.0.6-r7)
(2/11) Installing expat (2.2.8-r0)
(3/11) Installing libffi (3.2.1-r6)
(4/11) Installing gdbm (1.13-r1)
(5/11) Installing ncurses-terminfo-base (6.1_p20190105-r0)
(6/11) Installing ncurses-terminfo (6.1_p20190105-r0)
(7/11) Installing ncurses-libs (6.1_p20190105-r0)
(8/11) Installing readline (7.0.003-r1)
(9/11) Installing sqlite-libs (3.28.0-r3)
(10/11) Installing python2 (2.7.18-r0)
(11/11) Installing mercurial (4.9.1-r0)
Executing busybox-1.29.3-r10.trigger
OK: 69 MiB in 25 packages
/ # 
```

In the latest `alpine:3.12` the mercurial package uses python3.
```
$ podman run -it --rm alpine:3.12.0
/ # apk add --no-cache mercurial
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/community/x86_64/APKINDEX.tar.gz
(1/11) Installing libbz2 (1.0.8-r1)
(2/11) Installing expat (2.2.9-r1)
(3/11) Installing libffi (3.3-r2)
(4/11) Installing gdbm (1.13-r1)
(5/11) Installing xz-libs (5.2.5-r0)
(6/11) Installing ncurses-terminfo-base (6.2_p20200523-r0)
(7/11) Installing ncurses-libs (6.2_p20200523-r0)
(8/11) Installing readline (8.0.4-r0)
(9/11) Installing sqlite-libs (3.32.1-r0)
(10/11) Installing python3 (3.8.5-r0)
(11/11) Installing mercurial (5.4-r0)
Executing busybox-1.31.1-r16.trigger
OK: 74 MiB in 25 packages
```

gracinet:
> AFAICT `pip install mercurial` is just a convenient way to install the `hg` executable, with no coupling at the level of Python modules? If so could we treat Mercurial like `git`, i.e. it's the user's responsibility to install the git and mercurial binaries, e.g. using a distro package?

Yes, I've been thinking about that in the background and was willing to ask you how it works for Git :-) 

To elaborate, the `hg` commands by @paugier are run in subprocesses and have been there forever. So basically any version of Mercurial would do, and in particular the ones packaged with Alpine. I believe the `>=5.2` requirement to be there only for Python 3 compatibility.

The only potential tricky part is related to the activation in [mercurial.py](https://github.com/jupyterhub/repo2docker/pull/950/files#diff-5729ec04004c0cb4cd88c54c0b3e073d) of the `evolve` and `topic` extensions. These are both provided by the `hg-evolve` Python package, and are in practice necessary for repositories that make use of the  modern features they unlock (shareable history rewriting and tools around that). If `hg-evolve` isn't installed and the repo doesn't need it, those config options will produce warnings on `stderr`, but the commands wouldn't fail.

This concern of mine if more about all-inclusive end packaging that you might be doing (Docker image?) than people that would install repo2docker on top of installations that work well for their Mercurial use cases already.

`hg-evolve` is pure Python and explicitely designed to have zero install dependencies (not even Mercurial), in order for its installation to be safe, even system-wide. Of course be installed in the same interpreter context than the `hg` executable. 

The proper way to activate it only if installed and not get warnings if it isn't needed would be to do it as part of the installation process, by dropping a rc file somewhere instead of passing `--config` options. The best place typically depends on the application. In a Docker image I would simply create a `/etc/hgrc.d/evolve.rc` file, for instance.

> 
> I found this old post which suggests Python is just incidental https://www.mercurial-scm.org/pipermail/mercurial-devel/2016-January/078773.html

Another example is that it's always be the Mercurial project stance that the stability promise is at the CLI level, [not for the inner Python API](https://www.mercurial-scm.org/wiki/MercurialApi#Why_you_shouldn.27t_use_Mercurial.27s_internal_API).

manics:
@gracinet Thanks, that's very helpful. The git provider also uses `subprocess.Popen(["git", ...` https://github.com/jupyterhub/repo2docker/blob/f0c6f6540a42c4ef8ce424be2c51b2cb0b97e3c2/repo2docker/contentproviders/git.py

Since `hg-evolve` is pure Python with no dependencies it should be safe to unconditionally install even if it's not used, then if I understand you there should be no need to deal with configuration files like `/etc/hgrc.d/evolve.rc`?

If so does this sound like a reasonable plan?
- Remove `mercurial` pypi dependency
- Keep `hg-evolve` as a required pypi dependency
- Document that `hg` should be installed as an external dependency alongside `git` (or should it be optional with an error check in `mercurial.py`?)
- Include `apk add mercurial` in the Dockerfile (to be decided: whether to bump the alpine version to avoid both Python2 and Python3 being installed and taking up more space)

What does everyone think?

paugier:
At least all checks have passed...

What I did:

- Remove `mercurial` and `hg-evolve` pypi dependencies
- Install Mercurial and hg-evolve from the Dockerfile (bump the alpine version to get Mercurial with python3, hg-evolve installed with `pip3 install --user ...` + add a hg config file to activate hg-evolve)

paugier:
The lint CI failed because I used 2 f-strings and black still uses --target-version=py35 (see https://github.com/jupyterhub/repo2docker/pull/955).

Mercurial is now optional but for the CI, I'd like to set an environment variable (`REPO2DOCKER_HG_EVOLVE_REQUIRED`) to be sure that Mercurial code is properly tested. I didn't find where I should set it up. Where would you do that?

betatim:
For my education: what are these mecurial extensions that you activate in the docker image? What happens if we don't enable them (and interact with a repo that uses them)?

One more thing we need is instructions to install mercurial in the docs. My suggestion would be to add a section like https://repo2docker.readthedocs.io/en/latest/install.html#prerequisite-docker but name it somehow "Optional: ". In particular explaining how to install `hg` and `hg-evolve` so that they can work together. It sounds like the recommended way of installing the `hg` executable is with the package manager of your OS, but then `hg-evolve` (a python package) needs to be installed into the same Python environment as the `hg` executable. Getting this right when using `brew` and/or `conda` and/or virtual envs seems tricky. Hence a short set of instructions in our docs with a link to a full guide in the mercurial docs would be great.

(We should also really have a section for `git`, but that is a new PR)

To set environment variables you need to edit the GitHub Actions config/workflows which are defined in `.github/`. This is what we use for our CI.

manics:
Personally I think it's better for tests to default to testing everything, with a flag/environment variable used to explicitly _disable_ them if required. This reduces the risk of future refactoring inadvertently disabling them (been there in other projects..... 😄).

paugier:
> Personally I think it's better for tests to default to testing everything, with a flag/environment variable used to explicitly disable them if required. 

Ok I understand! I will change that.

paugier:
> For my education: what are these mercurial extensions that you activate in the docker image? What happens if we don't enable them (and interact with a repo that uses them)?

topic and evolve are 2 extensions provided by the python package hg-evolve useful for shareable history edition. Topics are light and temporary feature branches.

If you clone a repository with topics without the topic extensions, you just don't get the commits in topics.

evolve adds many commands for shareable history and it may be that enabling evolve is not necessary for repo2docker. What do you think @gracinet? I enabled evolve + topics because it is standard for real users but actually we don't do history edition in repo2docker!

> One more thing we need is instructions to install mercurial in the docs.

Yes of course! I will add that.

>  having to edit a config file to enable the extensions seems tricky.

With Mercurial, editing config files is anyway nearly mandatory.  I think Mercurial developers really like config files :slightly_smiling_face: For example I don't think there are equivalents of `git config --global user.name "FIRST_NAME LAST_NAME"` or `git remote set-url origin https://github.com/USERNAME/REPOSITORY.git`. And to use advanced features you have to activate extensions from a config file. By default, you only have simple commands.

However, I could add `--config extensions.topic` in hg commands in `mercurial.py` so that repo2docker users wouldn't need the config file. I see no drawbacks to this solution.

Regarding having `hg-evolve` package as a dependency in setup.py, the idea may seem nice but it seems to me that it is not practically very useful. I think it is better to treat hg as git, to forget that hg is written partly in Python and to let users setup their Mercurial environment. Installing `hg-evolve` is never very complicated (depending how you install Mercurial, it can be already installed as on Windows with TortoiseHg or one uses `pip install hg-evolve --user`).

Of course, basic users should be able to use repo2docker with Mercurial without hg-evolve and topics.

For MyBinder, hg-evolve should be installed but most users don't have to know that.

paugier:
I think it is now ready for review.

Mercurial is optional but by default tested. Mercurial tests can be disable with environment variables. There is no need for any Mercurial configuration file.

For the CI, Mercurial and hg-evolve are both installed with pip (in .github/workflows/test.yml) because in Ubuntu 18.04, the Mercurial install with apt is too old for the last version of hg-evolve (Mercurial 4.5, running with Python 2.7).

manics:
Do you think we could make `hg-evolve` compulsory for Mercurial support? This will simplify the docs (especially for anyone new to Mercurial), as well as removing one of the conditionals in the tests.

minrk (approved):

minrk:
Nice! I think we can fine-tune things like requirements and installation a bit, but this looks like a great start to me!

paugier:
> Do you think we could make hg-evolve compulsory for Mercurial support? This will simplify the docs (especially for anyone new to Mercurial), as well as removing one of the conditionals in the tests.

Yes you are right. It is indeed simpler and nicer like that https://github.com/jupyterhub/repo2docker/pull/950/commits/05002a4177371ec4662eef9f1d39b29f9d047b26

manics (approved):
Great!

Do you have an example hg repo url we can test this on?

paugier:
> Do you have an example hg repo url we can test this on?

I tried for example `jupyter-repo2docker https://foss.heptapod.net/fluiddyn/fluidsim --ref phaseshift`, which works at home.

paugier:
Hi, do I need to do something more on this?

manics:
I don't think so, just need to give people time to test it.

betatim:
Thanks a lot for adding mercurial support and the patience in responding to all the comments, questions and suggestions :)

PR #957, opened 2020-09-12
Let `CMD <cmd>` in the Dockerfile template be configurable

AliMirlou:
A configurable CMD might not be something this project wants but there are many use-cases for this, like starting the notebook in a different way or even using Repo2Docker as a library to build non-jupyter Docker images. It would be nice to have this feature.

By the way, this change is completely **backward-compatible** and **does not affect** anyone who does not want to change the CMD.

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -191,7 +191,7 @@ COPY /repo2docker-entrypoint /usr/local/bin/repo2docker-entrypoint
 ENTRYPOINT ["/usr/local/bin/repo2docker-entrypoint"]
 
 # Specify the default command to run
-CMD ["jupyter", "notebook", "--ip", "0.0.0.0"]
+CMD [{% for c in command -%} "{{ c }}"{{ "," if not loop.last }} {% endfor -%}]
 
 {% if appendix -%}
 # Appendix:
@@ -482,6 +482,14 @@ class BuildPack:
         """
         return None
 
+    def get_command(self):
+        """
+        The default command to be run by docker.
+
+        This should return a list of strings to be used as Dockerfile `CMD`.
+        """
+        return ["jupyter", "notebook", "--ip", "0.0.0.0"]
+
     @property
     def binder_dir(self):
         has_binder = os.path.isdir("binder")
@@ -568,6 +576,7 @@ class BuildPack:
             base_packages=sorted(self.get_base_packages()),
             post_build_scripts=self.get_post_build_scripts(),
             start_script=self.get_start_script(),
+            command=self.get_command(),
             appendix=self.appendix,
         )
 
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -10,6 +10,7 @@ import sys
 import hashlib
 import escapism
 import xml.etree.ElementTree as ET
+from collections.abc import Iterable
 
 from traitlets import Dict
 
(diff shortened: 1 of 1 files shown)


AliMirlou:
I haven't reflected the changes in the documentations and changelog yet. Waiting for someone's approval first.

AliMirlou:
Also can anyone explain why the "lint" check is failing? Is it because of too many characters in one line?

manics:
We use pre-commit and the black autoformatter.

If you install pre-commit (e.g. with pip) and run

    pre-commit run --all-files

the code should pass the CI lint check.

AliMirlou:
@manics Can you review this PR yourself? It's not a major change.

AliMirlou:
@betatim I need some approval to continue this PR. Can you give a quick review?

AliMirlou:
Or @minrk perhaps?

AliMirlou:
@manics Can you review this PR now?

repo2docker/buildpacks/base.py
```diff
@@ -482,6 +482,14 @@ def get_start_script(self):
         """
         return None
 
+    def get_command(self):
+        """
+        The default command to be run by docker.
+
+        This should return a list of strings to be used as Dockerfile `CMD`.
```
consideRatio, line 489:
```suggestion
    def get_cmd(self):
        """
        A list of strings to be used for the Dockerfile's `CMD`.
```

repo2docker/buildpacks/base.py
```diff
@@ -554,6 +562,19 @@ def render(self):
...
+        cmd = self.get_command()
+        if isinstance(cmd, tuple):
+            cmd = list(cmd)
+        if isinstance(cmd, list):
+            cmd = str(cmd).replace("'", '"')
+        if not isinstance(cmd, str):
+            raise ValueError(
+                'Method "get_command" of buildpack "%s" must return a string, a list or a tuple.'
+                % self.__class__.__name__
+            )
```
consideRatio, line 576: - Removed validation as this isn't user facing, just for those making custom buildpacks. My goal is to minimize complexity and ease the long term maintenance and code comprehension by doing that.
- Removed tuples fixup for similar reasons.

```suggestion
        cmd = self.get_command()
        if isinstance(cmd, list):
            cmd = str(cmd).replace("'", '"')
```

I don't understand what is going on with regards to the replacement operation that remains, but I suspect its not reasonable to provide either.

repo2docker/buildpacks/base.py
```diff
@@ -568,6 +589,7 @@ def render(self):
             base_packages=sorted(self.get_base_packages()),
             post_build_scripts=self.get_post_build_scripts(),
             start_script=self.get_start_script(),
+            command=cmd,
```
consideRatio, line 592:
```suggestion
            cmd=cmd,
```

repo2docker/buildpacks/base.py
```diff
@@ -191,7 +191,7 @@
 ENTRYPOINT ["/usr/local/bin/repo2docker-entrypoint"]
 
 # Specify the default command to run
-CMD ["jupyter", "notebook", "--ip", "0.0.0.0"]
+CMD {{ command }}
```
consideRatio, line 194:
```suggestion
CMD {{ cmd }}
```

consideRatio (changes requested):
I think it could be acceptable to provide this, but I'm not sure.

I made a few suggestions towards making me feel comfortable with the implementation. I think a very strong case needs to be made for a string replacement still though, and it probably needs to be documented if a strong case for it is presented.

consideRatio:
## Related
- https://github.com/jupyterhub/repo2docker/issues/1206

manics:
Thanks for the PR. I'm closing since there's been no updates, and it's already possible to override `CMD` using appendix
https://repo2docker.readthedocs.io/en/latest/usage.html#cmdoption-jupyter-repo2docker-appendix

PR #960, opened 2020-09-14
add 4.0, 4.0.2 to list of supported R versions

### LINKED ISSUE #887: Add support for R 4.0
### Proposed change

Add support for R 4.0 to the R version map at https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/r.py#L71

**Alternatives:**

- Installation via conda, but there the latest version is also `3.6`.
- Use [`holepunch`](https://github.com/karthik/holepunch) and the [`rocker/binder`](https://hub.docker.com/r/rocker/binder/tags) image, though no `4.0.0` tag there either yet.

### Alternative options

IMO: None.

People demonstrating latest R features, or relying on latest packages requiring R 4.0, or trying to match their own environment closely.

Not much, _if/when the version is available via the default Ubuntu sources.

### Who can do this work?

Anyone with experience in running repo2docker tests and some R experience for testing.

minrk:
closes #887

At first, I tried @betatim's suggestion to use conda to get R 4, but getting R from environment.yml is incompatible with the presence of a DESCRIPTION file, which triggers the R buildpack installing a different version of R from apt as well and getting things confused.

We can definitely improve support for getting R from conda when DESCRIPTION or install.R is present, but addding R 4 to the list we get from the same source we use for >= 3.5 is a pretty small change, so I think it's the right thing to do for now.

Test included.


yuvipanda (approved):
I like! Binary packages from packagemanager.rstudio.com/ are built against the apt packages from our source, so it makes a lot of sense to get R from there than from conda - RStudio Packagemanager works with built in R install commands, and has all of CRAN. \

yuvipanda:
It's been here for a while, so gonna merge this now.

PR #962, opened 2020-09-18
Mamba number three

SylvainCorlay:
cc @minrk - all tests are passing!

This is an update on #919 (I could not push to your branch anymore).

 - The failure in #919 was due to a missing `-y` in `mamba install`, which caused the logs to overflow with repeted `Confirm changes: [Y/n]`
 - I don't know why the `-y` was not required for the `conda install` command.

I also updated mamba to the latest

### DIFF
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -79,7 +79,7 @@ class CondaBuildPack(BaseImage):
             (
                 "root",
                 r"""
-                bash /tmp/install-miniforge.bash && \
+                bash -c 'time /tmp/install-miniforge.bash' && \
                 rm /tmp/install-miniforge.bash /tmp/environment.yml
                 """,
             )
@@ -275,9 +275,10 @@ class CondaBuildPack(BaseImage):
                 (
                     "${NB_USER}",
                     r"""
-                conda env update -p {0} -f "{1}" && \
-                conda clean --all -f -y && \
-                conda list -p {0}
+                bash -c 'time mamba env update -p {0} -f "{1}" && \
+                time mamba clean --all -f -y && \
+                mamba list -p {0} \
+                '
                 """.format(
                         env_prefix, environment_yml
                     ),
@@ -293,9 +294,9 @@ class CondaBuildPack(BaseImage):
                 (
                     "${NB_USER}",
                     r"""
-                conda install -p {0} r-base{1} r-irkernel={2} r-devtools && \
-                conda clean --all -f -y && \
-                conda list -p {0}
+                time mamba install -p {0} r-base{1} r-irkernel={2} r-devtools && \
+                time mamba clean --all -f -y && \
+                mamba list -p {0}
                 """.format(
                         env_prefix, r_pin, IRKERNEL_VERSION
                     ),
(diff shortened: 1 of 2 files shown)


manics:
Have you included `time` before calling `mamba` just for comparison during development, or are you planning to keep this when it's merged? If the latter is there a more concise way to show timings, since bash normally prints out user/system/real.

SylvainCorlay:
The `time` is a leftover from #919 which was used to produce the plots in the original description.

manics:
Just a heads-up that I've opened a PR that has a minor conflict with this one: https://github.com/jupyterhub/repo2docker/pull/963 as it changes the jupyter-offlinenotebook installation step.

SylvainCorlay:
@minrk let me know if you would like me to remove the `time` calls.

repo2docker/buildpacks/conda/__init__.py
```diff
@@ -293,9 +294,9 @@ def get_env_scripts(self):
                 (
                     "${NB_USER}",
                     r"""
-                conda install -p {0} r-base{1} r-irkernel={2} r-devtools && \
-                conda clean --all -f -y && \
-                conda list -p {0}
+                mamba install -p {0} r-base{1} r-irkernel={2} r-devtools -y && \
```
SylvainCorlay, line 297: Ping @wolfv 

For some readon, `conda` did not require passing `-y` here.

wolfv: is there a conda-rc? or environment var that could enable this behavior automagically?

minrk: maybe conda detects TTY with `sys.stdin.isatty()`? That's a pretty common pattern for deciding if a program can interact with the terminal or not.

minrk:
I'd love a better mechanism for measurement, but I think leaving the times in for now is providing a bit of useful info. 👍 to merge after resolving the conflict with #963

repo2docker/buildpacks/conda/__init__.py
wolfv: yeah we do have that mechanism as well, but currently don't use it to set auto_yes = True.

SylvainCorlay:
I rebased the PR!

Notes:

 - since the installation and build of labextensions are now split, each one gets a call to `time`.
 - I had to remove two calls to `time` in https://github.com/jupyterhub/repo2docker/pull/962/commits/3e07ba2fa5da46895837479ba80c67ed6c38790b because we cannot use it in push `sh` scripts.

manics:
You've committed the backup file ` repo2docker/buildpacks/conda/install-miniforge.bash.orig`

SylvainCorlay:
> You've committed the backup file repo2docker/buildpacks/conda/install-miniforge.bash.orig

Thanks! this is fixed!

repo2docker/buildpacks/conda/__init__.py
SylvainCorlay: Are we sure this is what is going on?

minrk:
Awesome, let's give it a try!

consideRatio:
Wieee well done everyone involved! I love the benchmark @SylvainCorlay and @minrk! It is a big change if we focus on whats optimized - the red part. Awesome!!!

meeseeksmachine:

https://discourse.jupyter.org/t/previously-working-build-fails-with-environment-yml-packages-not-in-env/6410/2

PR #963, opened 2020-09-18
MRG: Disable jupyterlab extension build minimize

manics:
See https://github.com/jupyterhub/mybinder.org-deploy/issues/1607

`${NB_PYTHON_PREFIX}/bin/jupyter labextension install jupyter-offlinenotebook || true` sometimes fails. This _may_ be due to excessive memory usage by the default build minimize step:
- https://discourse.jupyter.org/t/tip-binder-jupyterlab-extension/6022
- https://discourse.jupyter.org/t/jupyter-lab-build-hangs-and-then-fails-at-webpack-config-webpack-prod-minimize-config-js/6017/3

This PR disables the minimize step.


betatim:
Sounds good to me, with my dangerous half knowledge of the jupyter lab build process.

What is the disadvantage of removing the "minimize" step?

minrk:
> What is the disadvantage of removing the "minimize" step?

Should be some increased cost of loading jupyterlab javascript

PR #967, opened 2020-10-03
WIP: s3 logs prototype

manics:
Prototype of uploading repo2docker build logs to S3 (https://github.com/jupyterhub/binderhub/issues/1156)

If you don't have access to an S3 server run a local one:

    docker run -it --rm -e MINIO_ACCESS_KEY=minio -e MINIO_SECRET_KEY=minio123 -p 9000:9000 minio/minio server /data

Download an S3 client e.g. the [minio client](https://docs.min.io/docs/minio-client-quickstart-guide.html) from https://min.io/download

Add temporary client config called `mybindertest` for the local test server

    export MC_HOST_mybindertest=http://minio:minio123@localhost:9000
    
Create a bucket called `mybinder`

    mc mb mybindertest/mybinder

Allow public downloads from the bucket

    mc policy set download mybindertest/mybinder

Repo2docker S3 logs config

    cat << EOF > s3config.py 
    c.Repo2Docker.s3_logs_endpoint = 'http://localhost:9000'
    c.Repo2Docker.s3_logs_access_key = 'minio'
    c.Repo2Docker.s3_logs_secret_key = 'minio123'
    c.Repo2Docker.s3_logs_bucket = 'mybinder'
    EOF

Build with image name `asd123`

    repo2docker --config s3config.py --image-name asd123 https://github.com/binder-examples/requirements

Logs should be available at
http://localhost:9000/mybinder/buildlogs/asd123/repo2docker.log

If you can't be bothered to setup your own Minio server you can use the public demo one at play.min.io:

    export MC_HOST_mybindertest=https://Q3AM3UQ867SPQQA43P2F:zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG@play.min.io
    mc mb mybindertest/mybinder
    mc policy set download mybindertest/mybinder

Those demo credentials are public, I've already run the above so assuming no-one else has modified it you just need:

    cat << EOF > s3config.py 
    c.Repo2Docker.s3_logs_endpoint = 'https://play.min.io'
    c.Repo2Docker.s3_logs_access_key = 'Q3AM3UQ867SPQQA43P2F'
    c.Repo2Docker.s3_logs_secret_key = 'zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG'
    c.Repo2Docker.s3_logs_bucket = 'mybinder'
    EOF

    repo2docker --config s3config.py --no-run https://github.com/binder-examples/conda
    ...
    Successfully tagged r2dhttps-3a-2f-2fgithub-2ecom-2fbinder-2dexamples-2fconda0349319:latest
    Uploading log to mybinder/buildlogs/r2dhttps-3a-2f-2fgithub-2ecom-2fbinder-2dexamples-2fconda0349319/repo2docker.log

https://play.minio.io/mybinder/buildlogs/r2dhttps-3a-2f-2fgithub-2ecom-2fbinder-2dexamples-2fconda0349319/repo2docker.log

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -11,6 +11,7 @@ import json
 import sys
 import logging
 import os
+import boto3
 import getpass
 import shutil
 import tempfile
@@ -326,6 +327,12 @@ class Repo2Docker(Application):
         config=True,
     )
 
+    s3_logs_endpoint = Unicode("", help="S3 endpoint", config=True)
+    s3_logs_access_key = Unicode("", help="S3 access key ", config=True)
+    s3_logs_secret_key = Unicode("", help="S3 secret key", config=True)
+    s3_logs_bucket = Unicode("", help="S3 bucket", config=True)
+    s3_logs_region = Unicode("", help="S3 region", config=True)
+
     all_ports = Bool(
         False,
         help="""
@@ -625,7 +632,7 @@ class Repo2Docker(Application):
                         return True
         return False
 
-    def build(self):
+    def build(self, logfile=None):
         """
         Build docker image
         """
@@ -714,6 +721,8 @@ class Repo2Docker(Application):
                         bp.__class__.__name__,
                         extra=dict(phase="building"),
                     )
+                    if logfile:
+                        logfile.write("Using %s builder\n" % bp.__class__.__name__)
 
                     for l in picked_buildpack.build(
                         docker_client,
@@ -725,8 +734,12 @@ class Repo2Docker(Application):
                     ):
                         if "stream" in l:
                             self.log.info(l["stream"], extra=dict(phase="building"))
+                            if logfile:
+                                logfile.write(l["stream"])
                         elif "error" in l:
                             self.log.info(l["error"], extra=dict(phase="failure"))
+                            if logfile:
+                                logfile.write(l["error"])
                             raise docker.errors.BuildError(l["error"], build_log="")
                         elif "status" in l:
                             self.log.info(
@@ -741,7 +754,35 @@ class Repo2Docker(Application):
                 shutil.rmtree(checkout_path, ignore_errors=True)
 
     def start(self):
-        self.build()
+        logfile = None
+        if (
+            self.s3_logs_endpoint
+            and self.s3_logs_access_key
+            and self.s3_logs_secret_key
+            and self.s3_logs_bucket
+        ):
+            logfile = tempfile.NamedTemporaryFile("w", delete=False)
+        try:
+            self.build(logfile=logfile)
+        finally:
+            if logfile:
+                logfile.close()
+                if os.stat(logfile.name).st_size:
+                    dest = f"buildlogs/{self.output_image_spec}/repo2docker.log"
+                    self.log.info(f"Uploading log to {self.s3_logs_bucket}/{dest}")
+                    s3 = boto3.resource(
+                        "s3",
+                        endpoint_url=self.s3_logs_endpoint,
+                        aws_access_key_id=self.s3_logs_access_key,
+                        aws_secret_access_key=self.s3_logs_secret_key,
+                        config=boto3.session.Config(signature_version="s3v4"),
+                        region_name=self.s3_logs_region,
+                    )
+                    s3.Bucket(self.s3_logs_bucket).upload_file(
+                        logfile.name,
+                        dest,
+                        ExtraArgs={"ContentType": "text/plain"},
+                    )
 
         if self.push:
             self.push_image()
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,7 @@ setup(
     name="jupyter-repo2docker",
     version=versioneer.get_version(),
     install_requires=[
+        "boto3",
         "docker",
         "traitlets",
         "python-json-logger",


repo2docker/app.py
```diff
@@ -741,7 +754,35 @@ def build(self):
...
     def start(self):
-        self.build()
+        logfile = None
+        if (
+            self.s3_logs_endpoint
+            and self.s3_logs_access_key
+            and self.s3_logs_secret_key
+            and self.s3_logs_bucket
+        ):
+            logfile = tempfile.NamedTemporaryFile("w", delete=False)
```
minrk, line 764: These builds logs have been useful in the past, maybe they should be kept anyway, even if not uploading to s3?

repo2docker/app.py
```diff
@@ -741,7 +754,35 @@ def build(self):
...
+                    s3 = boto3.resource(
+                        "s3",
+                        endpoint_url=self.s3_logs_endpoint,
+                        aws_access_key_id=self.s3_logs_access_key,
+                        aws_secret_access_key=self.s3_logs_secret_key,
+                        config=boto3.session.Config(signature_version="s3v4"),
+                        region_name=self.s3_logs_region,
+                    )
+                    s3.Bucket(self.s3_logs_bucket).upload_file(
+                        logfile.name,
```
minrk, line 782: It's not quite as simple as uploading the terminal output record unmodified. With this recording, the file contains a *terminal stream* with ANSI escapes, carriage-returns, and other things that need processing. If you open them in a text editor without running through this, they will be very hard to read. If we want human-readable logs, we need to process the ANSI codes before uploading the resulting file.

setup.py
```diff
@@ -47,6 +47,7 @@ def get_identifier(json):
     name="jupyter-repo2docker",
     version=versioneer.get_version(),
     install_requires=[
+        "boto3",
```
minrk, line 50: Should boto3 be a hard requirement? I would expect close to zero local users would use this feature.

minrk (commented):
Awesome! I think this is a great approach. I had been thinking we should do it at the deployment/mybinder.org level, but this does seem lots simpler.

manics:
Thanks for reviewing! I started with repo2docker for this prototype because it was easier to get the logs, but I've since figured out how to do it on the BinderHub side: https://github.com/jupyterhub/binderhub/pull/1161. Obviously all your review comments from here also apply there.

I think BinderHub is a better location since it integrates multiple components though the repo2docker is a bit simpler. Let me know which you prefer and I'll put more time into it.

betatim:
Came here to find out why "store logs in s3" needed changes in repo2docker. My guess at an implementation would have either been changes to BinderHub chart (inject a sidecar) or in BinderHub itself. Is it worth discussing the pros and cons?

Having thought about it for a bit while writing this I am starting to think that r2d is the right place (sidecar or "native"), not bhub. By putting the code into r2d we can restart the hub pod without losing logs and maintain the kind of "fire and forget" setup we have right now for builds (bhub asks for a build to start and then doesn't need to maintain any state about it).

setup.py
betatim: I don't think it should be. Wondering if there is a smaller/more lightweight library that implements the S3 part as that is all we need.

betatim:
About naming the bucket: I think there are two options:
1. use a name derived from the resolved ref (in the Git case that would be a SHA1 and URL)
2. use a name derived from the unresolved ref (in the Git case that would be a branch/tag name and URL)

(the "derived from" is meant to represent us hashing/mangling/escaping/processing the inputs to generate a bucket name)

I like naming it after the resolved ref as that generates unique names for free and we can have a history of previous builds. However I can't (yet) think of a nice way for a user to find those logs again. The scenario i am thinking about is this:
1. user builds `master` of `org/repo` thinking "let me quickly build this repo"
1. build fails with some error that looks easy to fix
1. user makes a commit in `org/repo`
1. user builds `master` of `org/repo` again
1. error persists, "ok, maybe let me look at that original log again, must have overlooked something"
1. user now has to go find the original commit's SHA1 in order to get access to the log??

Basically in a situation where you build and rebuild and rerebuild the same repo by adding commits to your repository it becomes quite tricky to (re)find the log of a previous build.

I think this is a user interface problem which we could solve in BinderHub with some GUI or by clever structuring of the bucket names or some other way. I think the right level of complexity for a user would be presenting them with a chronologically sorted list of build logs that has a link back to the source repository (in the state it was when we built it) and maybe even the human recognisable ref that they typed into the BinderHub form.

minrk:
> Basically in a situation where you build and rebuild and rerebuild the same repo by adding commits to your repository it becomes quite tricky to (re)find the log of a previous build.

I see two pieces of information useful for solving this problem:

1. We could use a [directory prefix](https://stackoverflow.com/a/35805441) (s3 doesn't *really* have directories) based on the repo only, so that it's easy to list all build logs for a repo. This may also come in handy for per-repo quotas/culling purposes, though e.g. 30-day internal bucket expiry may be sufficient for our purposes.
2. We can use metadata fields like we do [in the analytics-archive](https://github.com/jupyterhub/mybinder.org-deploy/blob/069bedbdce3d1560a9bb7ba9868ab941be8f45b8/images/analytics-publisher/archiver.py#L71-L74) to store information like resolved_ref, unresolved_ref, repo, date, etc.

Key information: prefix can be used to filter files to a short manageable list per-repo, metadata cannot, hence the prefix. bucket-storage like S3/GCS does not support things like list-files-matching-metadata.

Then we can have a BinderHub API based on this info to retrieve the build log *history* for a repo, e.g.

-  `GET /api/build-logs/$provider/$repo` -> list of build logs for repo, sorted by date with info about which ref they were building
- `GET /api/build-log/$provider/$repo/$ref` -> single log after resolving ref

setup.py
manics: I'll make it optional. I came across https://github.com/smore-inc/tinys3 but doesn't look active.

It's possible to do this using tornado/requests, the main difficulty is calculating the S3v4 signature: https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html

The dependencies of botocore don't look too bad:
https://github.com/boto/botocore/blob/717eb38a99d0f3eed9648abd62184fbc6a7947ef/setup.py#L26-L36
```
jmespath>=0.7.1,<1.0.0
python-dateutil>=2.1,<3.0.0
urllib3>=1.20,<1.26
```
It's possible to make the S3 request using botocore. For boto3 which adds a higher level wrapper add
https://github.com/boto/boto3/blob/73865126cad3938ca80a2f567a1c79cb248169a7/setup.py#L16-L20
```
s3transfer>=0.3.0,<0.4.0
```

The other library I'm aware of is Minio's python client, though I haven't used it:
https://github.com/minio/minio-py/blob/b67bba368dfc244403b73538f6a2eeafe2016155/setup.py#L40-L46
```
urllib3
pytz
certifi
python-dateutil
configparser
```

manics:
> About naming the bucket: I think there are two options:
>
>    use a name derived from the resolved ref (in the Git case that would be a SHA1 and URL)
>    use a name derived from the unresolved ref (in the Git case that would be a branch/tag name and URL)

For the benefit of anyone new to S3 I thought I'd clear up a bit of terminology. You have an S3 server, a bucket, and a key which references an object. With many public clouds you also have a region but ignore that for now.

This means the URL to your object (in this case a log-file) is something like `https://<s3-server.example.org>/<bucket>/<key>`

The `bucket` is a container for multiple objects.
The `key` is a single string, `/` has no meaning. As @minrk mentioned you can simulate a directory structure by doing a prefix search, but since it's not a real directory some you can confuse people by creating both `/bucket/foo` and `/bucket/foo/bar` as objects 😀 (some S3 implementations will prevent this).

So in my original example:
`https://play.minio.io/mybinder/buildlogs/r2dhttps-3a-2f-2fgithub-2ecom-2fbinder-2dexamples-2fconda0349319/repo2docker.log`

Server: `https://play.minio.io`
Bucket: `mybinder`
Key: `buildlogs/r2dhttps-3a-2f-2fgithub-2ecom-2fbinder-2dexamples-2fconda0349319/repo2docker.log`

manics:
I think this is quite close to being ready. I took @minrk's suggestion (thanks!) of using a Configurable which I think makes things a lot nicer, the changes to `app.py` are fairly small, most of the logic is in `logstore.S3LogStore`.

The main issue left is how to configure `S3LogStore`, in particular how to set the log name when integrated into BinderHub. At present `S3LogStore` knows nothing about the R2D configuration, so it relies on `S3LogStore.logname` to be set to the desired S3 key/path, which means it has to be set in a config file passed as `repo2docker --config my_config.py`:
```py
c.Repo2Docker.logstore = 'repo2docker.logstore.S3LogStore'
c.S3LogStore.endpoint = 'http://192.168.1.1:9000'
c.S3LogStore.access_key = 'minio'
c.S3LogStore.secret_key = 'minio123'
c.S3LogStore.bucket = 'mybinder'
c.S3LogStore.logname = 'buildlogs/repo2docker.log'
c.S3LogStore.metadata = {"R2D-Key": "value",}
```
This is probably OK for the static S3 connection parameters, but the logname will change with every build so the config would need to be rewritten. Options:
1. Don't change anything, always rewrite the config file
2. Add a new repo2docker argument to set the logname, and possibly the metadata if we want it.
3. Allow configurables to be set on the command line: `--S3LogStore.logname=foo/bar.txt`

Any thoughts? I quite like option 3 as it matches other Jupyter/Ipython applications, and hides the new feature from a typical user. It's not currently working in R2D though, was this a deliberately design choice? It seemed easy enough to add:
```diff
diff --git a/repo2docker/__main__.py b/repo2docker/__main__.py
index 907460d..5ea22d8 100644
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -232,7 +232,8 @@ def make_r2d(argv=None):
         print(__version__)
         sys.exit(0)
 
-    args = get_argparser().parse_args(argv)
+    print(argv)
+    args, unknown_args = get_argparser().parse_known_args(argv)
 
     r2d = Repo2Docker()
 
@@ -240,6 +241,9 @@ def make_r2d(argv=None):
         r2d.log_level = logging.DEBUG
 
     r2d.load_config_file(args.config)
+    if unknown_args:
+        print("Unknown args: ", unknown_args)
+        r2d.parse_command_line(unknown_args)
     if args.appendix:
         r2d.appendix = args.appendix
```

repo2docker/app.py
manics: I found a regex from https://superuser.com/questions/380772/removing-ansi-color-codes-from-text-stream/380778#380778 that should match all ANSI escape sequences

repo2docker/app.py
```diff
@@ -326,6 +327,12 @@ def _user_name_default(self):
         config=True,
     )
 
+    s3_logs_endpoint = Unicode("", help="S3 endpoint", config=True)
+    s3_logs_access_key = Unicode("", help="S3 access key ", config=True)
+    s3_logs_secret_key = Unicode("", help="S3 secret key", config=True)
+    s3_logs_bucket = Unicode("", help="S3 bucket", config=True)
+    s3_logs_region = Unicode("", help="S3 region", config=True)
```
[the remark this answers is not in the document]
manics: I've made it into a Configurable. I decided to keep the individual options because it avoids having `aws_` in the argument names when this should work for any S3 compatible store, and will also make it easier if we want to switch libraries.

manics:
On K8S sidecar vs native in BinderHub: AFAIK since the repo2docker container logs to stdout it's not possible to capture them in a sidecar without changing repo2docker to log to a location the sidecar logging agent can read, such as a file on a shared volume or a network socket. I might be wrong on this though.

BinderHub obtains the logs using the K8s API
https://github.com/jupyterhub/binderhub/blob/acdab1952385eead7650590464bd218f1892fca6/binderhub/build.py#L350

consideRatio:
I'm onboard with the idea and I love that you have created an abstract LogStore class where s3 is one implementation of it.

Happy to help with review efforts if you want to bring this to life again @manics!

PR #969, opened 2020-10-15
add chown to COPY commands to reduce layer count

### LINKED ISSUE #164: use COPY --chown to stage repo (docker 17.09)

bollwyvl:
Hi folks, thanks for repo2docker! :heart: 

In the generated Dockerfile, this combines a number of `COPY` and subsequent `USER` and `RUN chown` calls with the [`COPY --chown` syntax](https://docs.docker.com/engine/reference/builder/#copy).

Anecdotally, this seems to reduce the layer count a good deal: 45 vs 53 for one of the first tests that runs with just an `environment.yml`... my tests are still chugging, so I don't have a bunch of other data points, bless CI!

I haven't assessed the impact to size, but I would image it would result in a 2-1 reduction, as the `chown`ed files are, to my knowledge, copied in the subsequent layer. Can't hurt to look!

closes #164 
closes #861

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -13,10 +13,11 @@ import xml.etree.ElementTree as ET
 
 from traitlets import Dict
 
+# Only use syntax features supported by Docker 17.09
 TEMPLATE = r"""
 FROM buildpack-deps:bionic
 
-# avoid prompts from apt
+# Avoid prompts from apt
 ENV DEBIAN_FRONTEND=noninteractive
 
 # Set up locales properly
@@ -100,12 +101,12 @@ ENV PATH {{ ':'.join(path) }}:${PATH}
 {% if build_script_files -%}
 # If scripts required during build are present, copy them
 {% for src, dst in build_script_files|dictsort %}
-COPY --chown=${NB_USER}:${NB_USER} {{ src }} {{ dst }}
+COPY --chown={{ user }}:{{ user }} {{ src }} {{ dst }}
 {% endfor -%}
 {% endif -%}
 
 {% for sd in build_script_directives -%}
-{{sd}}
+{{ sd }}
 {% endfor %}
 
 # Allow target path repo is cloned to be configurable
@@ -137,7 +138,7 @@ ENV {{item[0]}} {{item[1]}}
 {% if preassemble_script_files -%}
 # If scripts required during build are present, copy them
 {% for src, dst in preassemble_script_files|dictsort %}
-COPY --chown=${NB_USER}:${NB_USER} src/{{ src }} ${REPO_DIR}/{{ dst }}
+COPY --chown={{ user }}:{{ user }} src/{{ src }} ${REPO_DIR}/{{ dst }}
 {% endfor -%}
 {% endif -%}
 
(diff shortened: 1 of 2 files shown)


minrk:
Thanks! Linking up to previous discussions: This closes #164 and is an expansion of #861.

The missing piece brought up in #861 was that using ARGs for username in chown means requiring docker 19.03, which was a little more recent than we wanted to force. Since we are already templating, using a template variable for NB_USER instead of env variable would let us do this in a way that works back to 17.09. Want to take a stab at that? Then I think we can land this one.

repo2docker/buildpacks/base.py
```diff
@@ -100,7 +100,7 @@
 {% if build_script_files -%}
 # If scripts required during build are present, copy them
 {% for src, dst in build_script_files|dictsort %}
-COPY {{ src }} {{ dst }}
+COPY --chown=${NB_USER}:${NB_USER} {{ src }} {{ dst }}
```
minrk, line 103: adding NB_USER as a template variable in addition to build arg will allow us to do `{{ NB_USER }}` here and support docker engine back to 17.09 instead of requiring 19.03.

bollwyvl:
Thanks for the awesome, thorough response. _of course_ y'all know about all this stuff, and :ghost:  me for not scrolling down.

this updates it to a jinja var, and also moved the magic 1000 up to one spot... as well as some notes with some of the information gleaned from the above. from a docs perspective, perhaps there is a good place I can add some docs like:

```md
## Minimum Supported Docker Engine Version
To support the current flagship deployments, `repo2docker` currently requires Docker engine 17.09+.
Newer language features should be avoided until they are generally available on <platform, platform...>
```

This could be a separate, docs only PR...

bollwyvl:
Another thing I'm thinking about while digging around in the templates... way out of scope for this PR, but...

If the script object got a little fatter, and knew about its own files, they could be added "just in time". Further, multiple files headed for the same destination could be copied in a single layer

For example, `apt.txt` can break a repo's whole little universe, and can be expensive, but if it doesn't change, everything _should_ be the same.

So something like:

```py
apt = Script(
    files={"/tmp/": ["apt.txt"]},
    directives=["RUN apt-get install /tmp/apt.txt"],
)
```

consumed like...
```Dockerfile
{% for script in stage_scripts %}

{% for file in script.files %}
{% for src, dst in script.files|dictsort %}
COPY --chown={{ user }}:{{ user }} [{% for s in src %}{{ s }},{% endfor %}] {{ dst }}
{% endfor -%}

{% for s in script.directives %}
{{ directive }}
{% endfor %}

{% endfor %}
```

betatim:
Nice work actually implementing this. We might have known about this but apparently never got around to doing something about it :D 

👍  on the docs only PR

The idea from https://github.com/jupyterhub/repo2docker/pull/969#issuecomment-709477658 also needs a new issue/PR because (at least I) need a bit more explanation to understand what and why this would work.

minrk:
Awesome, thanks for finishing this one!

bollwyvl:
Hooray! A pleasure working with you kind folks.

>
Re: that Script thing: I'll try to put something together that demonstrates
it better...

Issue #835, opened 2020-01-28
Poetry Buildpack Proposal


RobertRosca:
Hi, I recently made [a repository](https://github.com/European-XFEL-examples/jsr-operation-xgm-euxfel-paper) that runs on Binder and installs a python environment via Poetry, and I'm considering writing a buildpack for Poetry and doing a PR.

To me, Poetry is a much nicer way to specify environments than conda or pip requirement files, as it properly concretises and hashes the dependencies and also tracks the version of packages installed from git repositories via the commit.

The contribution documentation says that when looking at adding a new buildpack you should consider "How easy it is to use a given setup without support from repo2docker natively.", and the answer is that it's very easy and can be done in three or four lines.

However, given how poetry environments are completely concretised via the lock files and will not change, it would be nice to include poetry as a build pack so that the environment layer doesn't have to be rebuilt each time. Currently running the `poetry install` command in `postBuild` means that all the python packages have to be installed any time the repository updates even if the dependencies have not changed.

If the PR for this buildpack could be accepted then I'll start work on this in March.

P.S. The [feature request template](https://github.com/jupyter/repo2docker/issues/new?template=feature_request) didn't work for me, so I'm not sure what to include here.

betatim:
Sounds like a good idea! Some thoughts off the top of my head:

What is your impression of the community around poetry? Have they been around for a while and look like a project growing in support and popularity? Or something that people got excited about and now not so excited about maintaining for the long term?

We should look at the `Pipfile` build pack when thinking about this one. There are a few things which are tricky to solve (or need a nice solution) where I think we can apply the solution to both `Pipfile` and poetry.
* how to deal with people wanting to use a Python version which isn't the default we install
* can we use conda to install the Python executable instead of using pyenv?
* what happens if a poetry file doesn't/does specify a jupyter notebook version?

For Python 2 based environments we already have a nice setup of creating one environment in which the notebook server runs (with its dependencies and based on Python 3) and a second environment in which the notebook kernel runs (with the dependencies specified by the repository and using Python 2). When a repository requests Python 3 we currently don't have this two environment setup, but in the long term it would probably make sense to switch to it. There are pros and cons to doing this (I will try and find the relevant issue).

[...]

PR #972, opened 2020-10-16
WIP: Poetry buildpack

RobertRosca:
As suggested in the issue #835 this PR adds in a poetry buildpack.

This is heavily based on the implementation of the Pipenv buildpack, since they're pretty similar anyway. I've also implemented equivalent tests to what is done for pipenv.

The only large issue with this PR is how the python versioning is worked out in the buildpack. Poetry lets you specify dependency constraints [using a variety of symbols](https://python-poetry.org/docs/dependency-specification/) (e.g. `^`, `*`, `~`), and it's likely that python was specified with these requirements.

To handle these constraints poetry-core has a `semver` module which is what I use to parse the versions constraints, however this means that `poetry-core` would then be a dependency requirement for repo2docker. I really don't like adding this dependency in, but the alternative is to either re-implement this constraint system, include a copy of the `poetry.core.semver` module in the poetry buildpack, or to ignore the constraints and parse them naively. Does anybody have a better way to deal with this?


RobertRosca:
Hi @betatim, just poking you in case the PR fell a bit too far down the todo list :stuck_out_tongue: no rush though! Oh and Happy New Year!

RobertRosca:
Rebased onto main since there were some conflicts

kylebarron:
> it's likely that python was specified with these requirements.

Not just likely; specifying the Python version is mandatory. From [here](https://python-poetry.org/docs/pyproject/#dependencies-and-dev-dependencies):

> Be aware that declaring the python version for which your package is compatible is mandatory

PR #975, opened 2020-10-18
[MRG] Ensure REPO_DIR owned by NB_USER

### LINKED ISSUE #974: REPO_DIR permission denied
### Bug description
After #969 got merged recently, `REPO_DIR` itself is now owned by `root` which can cause permission errors on updating/adding files in this directory.

Before #969, `REPO_DIR` was `COPY`-ed by `root` first and then the owner of entire directory including itself was changed by `RUN chown -R ${NB_USER}:${NB_USER} ${REPO_DIR}`. Now, it is replaced by `COPY --chown={{ user }}:{{ user }} src/ ${REPO_DIR}` which leaves `REPO_DIR` under `root`.

In my case, my Julia project set aside in `REPO_DIR` different from `HOME` got a permission denied error when installing packages by Julia build pack as it was trying to create a new file `Manifest.toml` in `REPO_DIR` where it no longer had permission.

Perhaps this issue only happens when we use `--target-repo-dir` to set a directory other than `$HOME` as `REPO_DIR`.

#### Expected behaviour
`REPO_DIR` owned by `NB_USER:NB_USER`.

#### Actual behaviour
`REPO_DIR` owned by `root:root`.

### How to reproduce
1. Run `repo2docker` with `--target-repo-dir`.
2. Check permission of the repo dir inside the container.

### Your personal set up
 - OS: macOS
 - Docker version: 19.03.13
 - repo2docker version: 0.11.0+175.g8c4eaa7

tomyun:
It is my take to fix #974 by making sure `REPO_DIR` owned by `NB_USER`. It was recently changed by #969 leaving `REPO_DIR` owned by root when `REPO_DIR` was set to a directory other than `$HOME` by using `--target-repo-dir` option.

closes #974

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -113,6 +113,7 @@ COPY --chown={{ user }}:{{ user }} {{ src }} {{ dst }}
 ARG REPO_DIR=${HOME}
 ENV REPO_DIR ${REPO_DIR}
 WORKDIR ${REPO_DIR}
+RUN chown ${NB_USER}:${NB_USER} ${REPO_DIR}
 
 # We want to allow two things:
 #   1. If there's a .local/bin directory in the repo, things there
--- a/tests/conda/repo-path/verify
+++ b/tests/conda/repo-path/verify
@@ -13,5 +13,8 @@ assert sys.executable.startswith("/srv/conda/"), sys.executable
 assert os.path.exists("/srv/repo/verify")
 assert os.path.abspath(__file__) == "/srv/repo/verify"
 
+# Repo should be writable
+assert os.access("/srv/repo", os.W_OK)
+
 # We should be able to import the package in environment.yml
 import numpy


betatim:
@bollwyvl do you understand why this happens?

Adding a `RUN chown ...` back to the `Dockerfile` adds back the "duplicate layers to solve the ownership" problem that we tried to fix with #969. Would be good if we could find a way to modify the `COPY` statement so that we don't need to do this. Maybe changing if there is (or not) a trailing `/` in the paths used by `COPY` helps (a bit like in `rsync` where trailing slash changes what gets copied).

betatim:
Having looked at the diff a bit longer I think this adds the `chown` earlier in the `Dockerfile` where the directory is still empty. So duplicating the layer isn't as costly in terms of final image size. This is nice! It did make me wonder if we are sure that `${REPO_DIR}` will always exist by this point? I think so but I am not sure, does anyone know?

It would be good to add a test case for the situation that got broken by #969 so that this bug doesn't come back in the future.

tomyun:
I was wondering if `COPY` could support alternative behaviors depending on trailing `/`, but couldn't find any information at least from Docker document. I also had an impression that running `chown` earlier shouldn't be too costly since the directory must be empty at this point as you mentioned.

In regards with the existence of `${REPO_DIR}`, IIUC, when it was not manually set, it should be the same as `$HOME` which already exists with proper ownership. If it's set to something else, an empty directory owned by root should have been just created by `WORKDIR`, then its permission needs to be taken care of by extra `chown` as suggested in this PR.

bollwyvl:
Yeah, `COPY` does weird stuff, but them's the layer brakes. Creating an empty `${REPO_DIR}` _really_ early with the proper perms would solve it, though, and an empty layer isn't going to add too much weight, if it can't be piggybacked off a previous layer. So worst case (untested):

```Dockerfile
ARG REPO_DIR=${HOME}
ENV REPO_DIR ${REPO_DIR}  # i think you get a free ENV from ARG, so can maybe get rid of this
RUN [ ! -d ${REPO_DIR} ] || mkdir -p ${REPO_DIR} && chown -R ${NB_USER}:${NB_USER} ${REPO_DIR}
WORKDIR ${REPO_DIR}       # this was probably creating it
```

It looks like this might have to be the way, as it next gets used (maybe) `COPY`ing the `preassemble_script_files`.

And yes, the "big copy" was probably missing the trailing slash:
```Dockerfile
COPY --chown={{ user }}:{{ user }} src/ ${REPO_DIR}/
```

>  add a test case for the situation that got broken by #969 so that this bug doesn't come back in the future.

Agreed, that would help clarify the issue!

bollwyvl:
Regarding test case: is there an existing `dir/test_foo.py` into which it could go? I see a `test_args.py` in `unit`, but i'd think if it actually turns docker on, it's not a unit test anymore...

tomyun:
Just added a minimal test case for checking write permission on `${REPO_DIR}`.

Seems like #976 is being worked on as an expanded solution to this PR and I appreciate that. However, in case review process takes longer since #976 also includes other changes not solely related to this issue, my hope is that we first review and merge this minimal PR first and continue working on the enhanced solution with more time for review.

This issue is currently holding back our pipeline for building docker images used in a class and we're in need of a new update published for this week ASAP. Thanks in advance for your understanding!

betatim:
I'd be good on merging this PR first and then iterating. Thanks a lot for adding the test case as well (thanks for looking for a test to add it to!).

Running the test locally without your patch to the code to see it fail. Seeing is believing :D

betatim:
Should we deployed now on mybinder.org

bollwyvl:
> Should be deployed now on mybinder.org

Hooray, shipping something that fixed the issue! :shipit: 

> holding back our pipeline for building docker images

Hm... seems like relying on `latest` in production for basically anything docker-related is asking for trouble...

> continue working

There's probably still some docs value to #976... I need to assess where it has diverged, and address some additional review comments, so have busted it back down to `wip`...

tomyun:
> > holding back our pipeline for building docker images
> 
> Hm... seems like relying on `latest` in production for basically anything docker-related is asking for trouble...

I totally agree with that sentiment although sometimes it's not avoidable when we have to rely on external tools, in our case, including [repo2docker-action](https://github.com/jupyterhub/repo2docker-action) which relies on `master` image of repo2docker. In addition to that, Julia support in the current stable version of repo2docker is way behind (1.3.1 vs. 1.5.2) and we can't use it.

PR #976, opened 2020-10-19
Allow REPO_DIR to be a non-existing folder by creating it and providing the user with permissions

### LINKED ISSUE #974: REPO_DIR permission denied

bollwyvl:
## PR summary by Erik 2022-10-31

- __Non-existing REPO_DIR folder__
  If users provide a REPO_DIR as a build arg to a folder that doesn't exist, the root user will create it but leave the non-root user without permissions to access it. This PR resolves that and adds a test to verify it by adding some `mkdir` and `chown` statements.
- __Dockerfile refactoring__
  It includes basic refactoring of the Dockerfile to reduce the layers by coalescing ENV statements and making use of the recommended syntax with explicit `=` in ENV statements.
- __Refined test documentation__
  It updates a comment to clarify how the test suite works.

---

**Updated**

This PR (originally attempted to try out some alternatives to #974 and (#975) now:
- > changes the target dir TBD (might not, if #975 changes are sufficient: @manics and @tomyun offered some additional approaches to be considered/discussed)
- adds some additional read-/write-ability checks to the custom target dir test case
  - in the process of figuring out where to land it... 
    - expands the docs in `conftest.py` for providing extra arguments via `extra-args.yml`
- partially to offset layer addition for path checking, where possible... 
  - updates all uses of `ENV` to the [recommended `=` syntax](https://docs.docker.com/engine/reference/builder/#env), replacing use of the "alternative" (space-based) syntax:
    > The alternative syntax is supported for backward compatibility, but discouraged for the reasons outlined above, and may be removed in a future release.
  - combines multiple consecutive `ENV` calls to a single, multi-line directive

--- 
> original

> - fixes #974 (attempted)
> - alternative to #975

> ~~Some guidance on _where_ to test this would be helpful... looking through the examples, I didn't see a way to add cli args, e.g. a `repo2docker-cli-args` file, perhaps that would be useful? this could be piggybacked off any of the existing ones, then...~~

> I found :point_up: and added some docs in `conftest.py`. Since the conda example was already testing repo-dir, i just added the test in that place.

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -106,13 +106,10 @@ USER root
 # Allow target path repo is cloned to be configurable
 ARG REPO_DIR=${HOME}
 ENV REPO_DIR=${REPO_DIR}
-RUN if [ ! -d "${REPO_DIR}" ] \
-    ; then \
-        mkdir -p "${REPO_DIR}" \
-        && chown -R ${NB_USER}:${NB_USER} "${REPO_DIR}" \
-    ; else \
-        echo "${REPO_DIR} already exists..." \
-    ; fi
+# Create a folder and grant the user permissions if it doesn't exist
+RUN if [ ! -d "${REPO_DIR}" ]; then \
+        /usr/bin/install -o ${NB_USER} -g ${NB_USER} -d "${REPO_DIR}"; \
+    fi
 
 WORKDIR ${REPO_DIR}
 RUN chown ${NB_USER}:${NB_USER} ${REPO_DIR}
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -105,7 +105,9 @@ USER root
 
 # Allow target path repo is cloned to be configurable
 ARG REPO_DIR=${HOME}
-ENV REPO_DIR ${REPO_DIR}
+RUN [ ! -d ${REPO_DIR} ] \
+    && mkdir -p ${REPO_DIR} \
+    && chown -R ${NB_USER}:${NB_USER} ${REPO_DIR}
 WORKDIR ${REPO_DIR}
 RUN chown ${NB_USER}:${NB_USER} ${REPO_DIR}
 
@@ -144,7 +146,7 @@ COPY --chown={{ user }}:{{ user }} src/{{ src }} ${REPO_DIR}/{{ dst }}
 USER root
 
 # Copy stuff.
-COPY --chown={{ user }}:{{ user }} src/ ${REPO_DIR}
+COPY --chown={{ user }}:{{ user }} src/ ${REPO_DIR}/
 
 # Run assemble scripts! These will actually turn the specification
 # in the repository into an image.
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -27,18 +27,18 @@ RUN apt-get -qq update && \
 RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \
     locale-gen
 
-ENV LC_ALL en_US.UTF-8
-ENV LANG en_US.UTF-8
-ENV LANGUAGE en_US.UTF-8
+ENV LC_ALL=en_US.UTF-8 \
+    LANG=en_US.UTF-8 \
+    LANGUAGE=en_US.UTF-8
 
 # Use bash as default shell, rather than sh
-ENV SHELL /bin/bash
+ENV SHELL=/bin/bash
 
 # Set up user
 ARG NB_USER
 ARG NB_UID
-ENV USER ${NB_USER}
-ENV HOME /home/${NB_USER}
+ENV USER=${NB_USER} \
+    HOME=/home/${NB_USER}
 
 RUN groupadd \
         --gid ${NB_UID} \
@@ -81,13 +81,13 @@ EXPOSE 8888
 {% if build_env -%}
 # Environment variables required for build
 {% for item in build_env -%}
-ENV {{item[0]}} {{item[1]}}
+ENV {{item[0]}}={{item[1]}}
 {% endfor -%}
 {% endif -%}
 
 {% if path -%}
 # Special case PATH
-ENV PATH {{ ':'.join(path) }}:${PATH}
+ENV PATH={{ ':'.join(path) }}:${PATH}
 {% endif -%}
 
 {% if build_script_files -%}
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -1,11 +1,19 @@
...
-a test. jupyter-repo2docker is run on that directory,
-and then ./verify is run inside the built container. It should
-return a non-zero exit code for the test to be considered a
-success.
+
+Test lifecycle:
+- Find all directories that contain `verify` or `*.repos.yaml`
+- If `verify` is found:
+    - Run `jupyter-repo2docker` on the test directory.
+      - Extra arguments may be added as YAML list of strings in `extra-args.yaml`.
```
bollwyvl, line 9: took me a while to find this, hence updating this section

tests/conda/repo-path/verify
```diff
@@ -13,5 +14,14 @@ assert sys.executable.startswith("/srv/conda/"), sys.executable
...
 assert os.path.abspath(__file__) == "/srv/repo/verify"
 
+# All files in repo dir should be readable and writeable
+for path in glob("/src/repo/**/*", recursive=True):
+    assert os.access(path, os.R_OK)
+    assert os.access(path, os.W_OK)
+
+# Should be able to make a new file
+with open("/srv/repo/writeable", "w") as fp:
+    fp.write("writeable")
```
bollwyvl, line 24: seems sufficiently aggressive

repo2docker/buildpacks/base.py
```diff
@@ -111,7 +111,15 @@
...
-ENV REPO_DIR ${REPO_DIR}
+ENV REPO_DIR=${REPO_DIR}
+RUN if [ ! -d "${REPO_DIR}" ] \
+    ; then \
+        mkdir -p "${REPO_DIR}" \
+        && chown -R ${NB_USER}:${NB_USER} "${REPO_DIR}" \
+    ; else \
+        echo "${REPO_DIR} already exists..." \
+    ; fi
+
```
bollwyvl, line 122: So, at present, this would fail if you tried to target `/tmp` or `/home` or something, which already exists, but that you didn't (and shouldn't) own.

repo2docker/buildpacks/base.py
```diff
@@ -30,18 +30,18 @@
 RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \
     locale-gen
 
-ENV LC_ALL en_US.UTF-8
-ENV LANG en_US.UTF-8
-ENV LANGUAGE en_US.UTF-8
+ENV LC_ALL=en_US.UTF-8 \
```
bollwyvl, line 33: Keep keep working that layer count down...

repo2docker/buildpacks/base.py
```diff
@@ -111,7 +111,15 @@
 
 # Allow target path repo is cloned to be configurable
 ARG REPO_DIR=${HOME}
-ENV REPO_DIR ${REPO_DIR}
+ENV REPO_DIR=${REPO_DIR}
+RUN if [ ! -d "${REPO_DIR}" ] \
+    ; then \
+        mkdir -p "${REPO_DIR}" \
+        && chown -R ${NB_USER}:${NB_USER} "${REPO_DIR}" \
```
tomyun, line 118: Perhaps we can remove `-R` as it's an empty directory just created.

bollwyvl: yeah, it probably doesn't magically do `umask` or anything (though maybe we should think about that)...

bollwyvl: will wait to get any other feedback, it's a lot of coal to burn!

manics: FYI you can use something like `/usr/bin/install -d <directory> -m <mode> -o <owner> -g <group>` to do everything in one go if you want, doesn't really matter though.

tests/conftest.py
minrk: Nice clarification!

repo2docker/buildpacks/base.py
yuvipanda: Yeah, can we use this instead? `if` in bash always makes me feel like I'm too dumb to make it work

consideRatio: I'll go for this as resolved, allowing anyone to refactor this further later.

repo2docker/buildpacks/base.py
```diff
@@ -111,7 +111,15 @@
...
+    ; else \
+        echo "${REPO_DIR} already exists..." \
+    ; fi
```
consideRatio, line 121:
```suggestion
# Create a folder and grant the user permissions if it doesn't exist
RUN if [ ! -d "${REPO_DIR}" ]; then \
        /usr/bin/install -o ${NB_USER} -g ${NB_USER} -d "${REPO_DIR}"; \
    fi
```

consideRatio:
I applied a suggestion and rebased this PR on the main branch to get the test suite to run.

repo2docker/buildpacks/base.py
consideRatio: Resolved it by applying my own suggestion, based on `/usr/bin/install -d <directory> -m <mode> -o <owner> -g <group>`

consideRatio:
Tests pass, I'll make another rebase to get a bit cleaner git history before merge.

image

PR #977, opened 2020-10-22
[MRG] Remove reference to `master` branch from CLI doc

### LINKED ISSUE #971: Detect default branch instead of assuming master (at least for GitHub repos)
### Proposed change

GitHub (at least) now supports setting your default branch to something other than `master`, the recommended option being `main`. https://www.zdnet.com/article/github-to-replace-master-with-main-starting-next-month/
This can be confusing for new users of mybinder.org, e.g. https://discourse.jupyter.org/t/cant-solve-could-not-resolve-ref-for-gh-fredericfoulonlycee-math-spe-python-master/6301
We should change repo2docker's behaviour to search for/fetch the default branch name of the git repository, rather than assuming/hard code it to being `master`.

I don't know if r2d assumes everything is a git repository or if this is something we'll specifically have to build into the GitHub repo handling section.

### Alternative options

We continue to wait until this is more common before carrying out this work. Perhaps adding an information box to the front page of mybinder.org so folks are aware.

Everybody who has created a new GitHub repository since 1st October 2020 or have manually renamed their default branches.

I imagine there's a git command we can run to fetch the default branch name and then parse that into our standard branch handling code for non-`master` branches.

### Who can do this work?

Useful skills:

- Knowledge of the r2d codebase: where things are being parsed to, etc
- Some git fu to detect the default branch name

betatim:
Make repo2docker ready for the new default branch name on GitHub: `main`.

This already worked before and still works (bokeh renamed their default branch from `master` to `main` a while back):
```
# check out default branch (main)
$ repo2docker --no-build https://github.com/bokeh/bokeh-notebooks/
# uses old name, is redirected and checks out new default branch (main)
$ repo2docker --no-build --ref master https://github.com/bokeh/bokeh-notebooks/
```

A newly created repository that uses main as default branch name works:
```
repo2docker --no-build https://github.com/betatim/new-default-branch-name
# Newly created repo with main as default branch, raises exception
repo2docker --no-build --ref master https://github.com/betatim/new-default-branch-name
```

This PR only changes an error messages and a help text. I couldn't find any other explicit references to `master` as default branch name.

closes #971

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -96,7 +96,8 @@ def get_argparser():
     argparser.add_argument(
         "--ref",
         help=(
-            "If building a git url, which reference to check out. " "E.g., `master`."
+            "Reference to build instead of default reference. For example"
+            " branch name or commit for a Git repository."
         ),
     )
 
--- a/repo2docker/contentproviders/git.py
+++ b/repo2docker/contentproviders/git.py
@@ -47,9 +47,18 @@ class Git(ContentProvider):
                 self.log.error(
                     "Failed to check out ref %s", ref, extra=dict(phase="failed")
                 )
-                raise ValueError("Failed to check out ref {}".format(ref))
+                if ref == "master":
+                    msg = (
+                        "Failed to check out the 'master' branch. "
+                        "Maybe the default branch is not named 'master' "
+                        "for this repository.\n\nTry not explicitly "
+                        "specifying `--ref`."
+                    )
+                else:
+                    msg = "Failed to check out ref {}".format(ref)
+                raise ValueError(msg)
             # We don't need to explicitly checkout things as the reset will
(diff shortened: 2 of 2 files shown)


betatim:
A thing I noticed while working on this is that a command like:
```
repo2docker --no-build bokeh
```
will lead to the following labels being added to the container image:
```
LABEL repo2docker.ref="None"
LABEL repo2docker.repo="https://github.com/bokeh/bokeh-notebooks/"
LABEL repo2docker.version="0.11.0+179.g19c698e.dirty"
```
Does anyone remember if using `None` for the ref is on purpose? I was somehow expecting the resolved ref to appear there, not the one the user inputted (for which `None` is the right value because it signals "user didn't specify anything").

minrk:
> Does anyone remember if using None for the ref is on purpose?

I don't think so. It should probably be the resolved ref

PR #981, opened 2020-10-29
[MRG] Set JUPYTER_CONFIG_DIR

tomyun:
This is for Julia build pack. For consistency with `JUPYTER_DATA_DIR` which is already set to our custom non-home directory (`/srv/conda/envs/notebook/share/jupyter`), this PR sets `JUPYTER_CONFIG_DIR` to a similar location (`/srv/conda/envs/notebook/etc/jupyter`). These settings are listed in the [Jupyter documentation](https://jupyter.readthedocs.io/en/latest/use/jupyter-directories.html).

It would be particularly useful in the scenario when user home is mounted to a persistent volume as with a typical JupyterHub environment. Without this `JUPYTER_CONFIG_DIR`, any settings, especially for custom notebook/lab extensions, would be stored in the home directory by default and they would not be available on runtime when the home is replaced by another volume.

### DIFF
--- a/repo2docker/buildpacks/julia/julia_project.py
+++ b/repo2docker/buildpacks/julia/julia_project.py
@@ -78,6 +78,7 @@ class JuliaProjectTomlBuildPack(PythonBuildPack):
             ("JULIA_DEPOT_PATH", "${JULIA_PATH}/pkg"),
             ("JULIA_VERSION", self.julia_version),
             ("JUPYTER", "${NB_PYTHON_PREFIX}/bin/jupyter"),
+            ("JUPYTER_CONFIG_DIR", "${NB_PYTHON_PREFIX}/etc/jupyter"),
             ("JUPYTER_DATA_DIR", "${NB_PYTHON_PREFIX}/share/jupyter"),
         ]
 


betatim:
Reading the documentation makes me think that Jupyter will search several paths for configuration files. Do you know if it merges the config from all directories or if it stops once it finds one config file?

Right now I see the following when I start a repository on mybinder.org:
```
$ jupyter --paths
config:
    /home/jovyan/.jupyter
    /srv/conda/envs/notebook/etc/jupyter
    /usr/local/etc/jupyter
    /etc/jupyter
```
Would it work for you to store your config files in one of the other directories?

The reason I'm asking is that there are repos which get built with repo2docker that include a `.jupyter/something-something.py` file in order to adjust the configuration. I think we'd break those by setting `JUPYTER_CONFIG_DIR` as that removes `/home/jovyan/.jupyter` from the search path.

tomyun:
I thought this would be useful since some extensions like [WebIO.jl](https://github.com/JuliaGizmos/WebIO.jl) detects installation path depending on these variables. But then as you said there would be other extensions doing the other way around and we may not want to break them. Also there may be other ways to control them through `postBuild` or `start` scripts if needed.

With that, this now makes me think we shouldn't touch `JUPYTER_DATA_DIR` in Julia build pack either especially when it's not something exclusive to Julia and could potentially impact packages/extensions for other languages like Python.

PR #983, opened 2020-11-06
[MRG] Update to node 14

jtpio:
According to the Releases page, node 10 will soon be out of maintenance:

https://nodejs.org/en/about/releases/

image

This change bumps the node version to 14.

Also for extra context: starting from https://github.com/jupyterlab/jupyterlab/pull/9282, JupyterLab requires node>=12, which currently makes the Binder preview builds fail: https://github.com/jupyterlab/jupyterlab/pull/9296#issuecomment-723226359

PR #985, opened 2020-11-11
Drop support for stencila

minrk:
The stencila components that nbstencilaproxy relied upon have been abandoned, and stencila has changed in ways that apear to make it incompatible with nbstencilaproxy.

As a result, remove all support for stencila, and leave a note that folks are welcome to try to restore it.

This should unblock #983


betatim:
I support dropping support. As well as adding it back if someone comes asking for it and knows how to upgrade the components.

PR #987, opened 2020-11-25
[MRG] Stream jupyter server logs to a file

betatim:
By streaming the logs also to a file users can access them from inside
the container which helps them debug issues without having to ask an
admin.

Follow up from https://github.com/jupyterhub/binderhub/issues/1156#issuecomment-703448118

### DIFF
--- a/repo2docker/buildpacks/repo2docker-entrypoint
+++ b/repo2docker/buildpacks/repo2docker-entrypoint
@@ -19,3 +19,6 @@ if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
 else
     exec "$@" >&"$log_fd" 2>&1
 fi
+
+# Close the loggging output aggain
+exec {log_fd}>&-
--- a/repo2docker/buildpacks/repo2docker-entrypoint
+++ b/repo2docker/buildpacks/repo2docker-entrypoint
@@ -15,10 +15,10 @@ if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
     if [[ ! -x "$R2D_ENTRYPOINT" ]]; then
         chmod u+x "$R2D_ENTRYPOINT"
     fi
-    exec "$R2D_ENTRYPOINT" "$@" >&"$log_fd" 2>&1
+    exec "$R2D_ENTRYPOINT" "$@" >&"$log_fd"
 else
-    exec "$@" >&"$log_fd" 2>&1
+    exec "$@" >&"$log_fd"
 fi
 
-# Close the loggging output aggain
-exec {log_fd}>&-
+# Close the logging output again
+#exec {log_fd}>&-
--- a/repo2docker/buildpacks/repo2docker-entrypoint
+++ b/repo2docker/buildpacks/repo2docker-entrypoint
@@ -2,11 +2,19 @@
 # lightest possible entrypoint that ensures that
 # we use a login shell to get a fully configured shell environment
 # (e.g. sourcing /etc/profile.d, ~/.bashrc, and friends)
+
+# Setup a file descriptor (FD) that is connected to a tee process which
+# writes its input to $REPO_DIR/.jupyter-server-log.txt
+# We later use this FD as a place to redirect the output of the actual
+# command to. We can't add `tee` to the command directly as that will prevent
+# the container from exiting when `docker stop` is run.
+exec {log_fd}> >(exec tee $REPO_DIR/.jupyter-server-log.txt)
+
 if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
     if [[ ! -x "$R2D_ENTRYPOINT" ]]; then
         chmod u+x "$R2D_ENTRYPOINT"
     fi
-    exec "$R2D_ENTRYPOINT" "$@"
+    exec "$R2D_ENTRYPOINT" "$@" >&"$log_fd" 2>&1
 else
-    exec "$@"
+    exec "$@" >&"$log_fd" 2>&1
 fi
--- a/repo2docker/buildpacks/repo2docker-entrypoint
+++ b/repo2docker/buildpacks/repo2docker-entrypoint
@@ -8,6 +8,7 @@
 # We later use this FD as a place to redirect the output of the actual
 # command to. We can't add `tee` to the command directly as that will prevent
 # the container from exiting when `docker stop` is run.
+# See https://stackoverflow.com/a/55678435
 exec {log_fd}> >(exec tee $REPO_DIR/.jupyter-server-log.txt)
 
 if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
--- a/tests/unit/test_env.py
+++ b/tests/unit/test_env.py
@@ -58,6 +58,10 @@ def test_env():
 
     # all docker output is returned by repo2docker on stderr
     # extract just the declare for better failure message formatting
+    # stdout should be empty
+    assert not result.stdout
+
+    # stderr should contain lines of output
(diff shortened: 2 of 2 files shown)


repo2docker/buildpacks/repo2docker-entrypoint
```diff
@@ -6,7 +6,7 @@ if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
     if [[ ! -x "$R2D_ENTRYPOINT" ]]; then
         chmod u+x "$R2D_ENTRYPOINT"
     fi
-    exec "$R2D_ENTRYPOINT" "$@"
+    exec "$R2D_ENTRYPOINT" "$@" 2>&1 | tee /jupyter-server-log.txt
```
manics, line 9: Won't this require the entrypoint to be run as `root` in order to write to `/`?

betatim: Good question. I hadn't thought that far yet, was still at the "does this even work" stage :D

betatim: You were right and our testing robots found it as well :D

betatim:
When we add `| tee some-log.txt` to the end of our command like I did in the previous commits then the SIGTERM that `docker stop <containerID>` sends gets lost. As a result `docker stop` then sends a SIGKILL which means our container exits with a status != 0 (actually with 137 = 128 +9, where 9 is the signal that got sent). I couldn't work out why/how/where the SIGTERM gets "lost".

The solution implemented now is from https://stackoverflow.com/a/55678435 which suggests using process substitution. As far as I understand it we create a new file descriptor which is connected to a process that runs `tee`. Then for our actual command we redirect output to that special FD. Because it is "just" output redirection and no pipes are involved `docker stop` does the right thing.

I don't fully understand what we are doing so I think the comment in the entrypoint script could be improved in terms of what it is we are actually doing and why we need to do this. If someone has some spare brain cycles (or is an expert on this :D) all ideas would be welcome.

betatim:
@manics if you want to learn more about output redirection and bash than I ever wanted to know this is the PR for you :D

repo2docker/buildpacks/repo2docker-entrypoint
```diff
@@ -2,11 +2,19 @@
...
+# command to. We can't add `tee` to the command directly as that will prevent
+# the container from exiting when `docker stop` is run.
+exec {log_fd}> >(exec tee $REPO_DIR/.jupyter-server-log.txt)
+
 if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
     if [[ ! -x "$R2D_ENTRYPOINT" ]]; then
         chmod u+x "$R2D_ENTRYPOINT"
     fi
-    exec "$R2D_ENTRYPOINT" "$@"
+    exec "$R2D_ENTRYPOINT" "$@" >&"$log_fd" 2>&1
```
minrk, line 17: is

```suggestion
    exec "$R2D_ENTRYPOINT" "$@" 2>&1 | tee $REPO_DIR/.jupyter-server-log.txt
```

not enough?

repo2docker/buildpacks/repo2docker-entrypoint
```diff
@@ -2,11 +2,19 @@
 # lightest possible entrypoint that ensures that
 # we use a login shell to get a fully configured shell environment
 # (e.g. sourcing /etc/profile.d, ~/.bashrc, and friends)
+
+# Setup a file descriptor (FD) that is connected to a tee process which
+# writes its input to $REPO_DIR/.jupyter-server-log.txt
+# We later use this FD as a place to redirect the output of the actual
+# command to. We can't add `tee` to the command directly as that will prevent
+# the container from exiting when `docker stop` is run.
+exec {log_fd}> >(exec tee $REPO_DIR/.jupyter-server-log.txt)
```
minrk, line 11: This is a very mysterious line! I have no idea what the syntax of `{name}>` means! But clearly it results in the desired effect - `$log_fd` is defined as an integer, e.g. 10. Where did you learn this?

minrk (approved):

repo2docker/buildpacks/repo2docker-entrypoint
betatim: The all knowing SO: https://stackoverflow.com/a/55678435

repo2docker/buildpacks/repo2docker-entrypoint
```diff
@@ -2,11 +2,19 @@
 # lightest possible entrypoint that ensures that
 # we use a login shell to get a fully configured shell environment
 # (e.g. sourcing /etc/profile.d, ~/.bashrc, and friends)
+
+# Setup a file descriptor (FD) that is connected to a tee process which
+# writes its input to $REPO_DIR/.jupyter-server-log.txt
+# We later use this FD as a place to redirect the output of the actual
+# command to. We can't add `tee` to the command directly as that will prevent
+# the container from exiting when `docker stop` is run.
```
manics, line 10:
```suggestion
# the container from exiting when `docker stop` is run.
# See https://stackoverflow.com/a/55678435
```
I still don't understand exactly what this does, but it's probably worth including that link since you found it useful!

betatim:
Ready to review/go.

manics:
Let's merge this 😄

manics:
Merging this might've broken a test https://github.com/jupyterhub/repo2docker/runs/1778527705
```
=================================== FAILURES ===================================
___________________________________ test_env ___________________________________

    def test_env():
        """
        Validate that you can define environment variables
    
        See https://gist.github.com/hwine/9f5b02c894427324fafcf12f772b27b7
        for how docker handles its -e & --env argument values
        """
        ts = str(time.time())
        # There appear to be some odd combinations of default dir that do
        # not work on macOS Catalina with Docker CE 2.2.0.5, so use
        # the current dir -- it'll be deleted immediately
    
        with tempfile.TemporaryDirectory(dir=os.path.abspath(os.curdir)) as tmpdir:
            username = getuser()
            os.environ["SPAM"] = "eggs"
            os.environ["SPAM_2"] = "ham"
            result = subprocess.run(
                [
                    "repo2docker",
                    # 'key=value' are exported as is in docker
                    "-e",
                    "FOO={}".format(ts),
                    "--env",
                    "BAR=baz",
                    # 'key' is exported with the currently exported value
                    "--env",
                    "SPAM",
                    # 'key' is not exported if it is not exported.
                    "-e",
                    "NO_SPAM",
                    # 'key=' is exported in docker with an empty string as
                    # value
                    "--env",
                    "SPAM_2=",
                    "--",
                    tmpdir,
                    "/bin/bash",
                    "-c",
                    # Docker exports all passed env variables, so we can
                    # just look at exported variables.
                    "export",
                ],
                universal_newlines=True,
                stdout=subprocess.PIPE,
                stderr=subprocess.PIPE,
            )
        assert result.returncode == 0
    
        # all docker output is returned by repo2docker on stderr
        # extract just the declare for better failure message formatting
        # stdout should be empty
        assert not result.stdout
    
        # stderr should contain lines of output
        declares = [x for x in result.stderr.split("\n") if x.startswith("declare")]
>       assert 'declare -x FOO="{}"'.format(ts) in declares
E       assert 'declare -x FOO="1611767496.4447188"' in []
E        +  where 'declare -x FOO="1611767496.4447188"' = <built-in method format of str object at 0x7f0b5975b2b0>('1611767496.4447188')
E        +    where <built-in method format of str object at 0x7f0b5975b2b0> = 'declare -x FOO="{}"'.format

/home/runner/work/repo2docker/repo2docker/tests/unit/test_env.py:66: AssertionError
```

betatim:
This is annoying :-/ I thought we had it fixed and locally it works and previously the tests passed :-/

PR #989, opened 2020-11-26
Delegate jupyter execution to pipenv virtualenv if it is found among the dependencies

### LINKED ISSUE #827: Pipenv that includes Jupyter can’t be installed
### Bug description

Pipenv is run with the `--system` cli switch, so any dependence inside the lock filter that clash with the system packages cause the build to fail.

E.g. https://mybinder.org/v2/gh/mjpieters/adventofcode/b7f80b433311f035916a0bd74fae818b9c231105 fails to install as pip isn’t allowed to uninstall `terminado`:
 
```none
[pipenv.exceptions.InstallError]: ['Collecting terminado==0.8.3 (from 
-r /tmp/pipenv-jtgp9b6o-requirements/pipenv-y1rwqxuh-requirement.txt (
line 1))', '  Using cached https://files.pythonhosted.org/packages/ff/
96/1d9a2c23990aea8f8e0b5c3b6627d03196a73771a17a2d9860bbe9823ab6/termin
ado-0.8.3-py2.py3-none-any.whl', 'Installing collected packages: termi
nado', '  Found existing installation: terminado 0.8.2']              
[pipenv.exceptions.InstallError]: ["ERROR: Cannot uninstall 'terminado
'. It is a distutils installed project and thus we cannot accurately d
etermine which files belong to it which would lead to only a partial u
ninstall."]                                                           
ERROR: ERROR: Package installation failed...                          
Removing intermediate container 689fa80c20bf
```

I’ve picked a specific revision for the above mybinder URL to ensure reproducibility.

Perhaps the bulls should be retried without the `--system` switch in such cases?

rubendibattista:
Hello, this implements a workaround to fix #827. Basically it parses the `Pipfile`, and if it finds `jupyter` in it, it installs the `Pipfile.lock` stuff in a virtualenv and it delegates the execution of Jupyter to it. This is achieved adding some level of customisation of the default command of the generated `Dockerfile` adding the `get_cmd` method. This method is then used in the `pipfile` buildpack to override the command in the case `jupyter` is found among the dependencies. 

Additionally, this does install the latest version of `pipenv` available on `pypi` in the container (instead of the pinned one).


rubendibattista:
Ping? 👉👈

minrk:
Thanks for the patch! I don't think this is how we should do it, though.

The bug in #827 is because a conda package is missing what should be considered required metadata (it is required for pip to work right). Fixing that conda package should be the right fix, as done in https://github.com/conda-forge/terminado-feedstock/pull/16 rathre than working around it.

The second point is that `jupyter` should ~never be in a Pipfile consumed by repo2docker. It's a metapackage that brings in lots of things that can't work in docker, like the qtconsole. This also won't catch most instances affected, which would mean any package that depends on terminado (namely `notebook` or `jupyter_server`, but could be arbitrarily many packages).

rubendibattista:
Thanks @minrk. Ok, let's keep the discussion on #827 because I still have the problem.

PR #990, opened 2020-11-28
buildpacks/nix: disable sandboxing (bugfix)

FRidh:
The version bump #915 broke the Nix buildpack in case one does a build.

Nix 2.3 enables sandboxing by default. Building inside a Docker
container, while Nix is having sandboxing enabled is not possible. Thus,
sandbox = false should be set in /etc/nix/nix.conf.

cc @costrouc

### DIFF
--- a/repo2docker/buildpacks/nix/__init__.py
+++ b/repo2docker/buildpacks/nix/__init__.py
@@ -23,17 +23,22 @@ class NixBuildPack(BaseImage):
         """
         Return series of build-steps common to all nix repositories.
         Notice how only root privileges are needed for creating nix
-        directory.
+        directory and a nix.conf file.
 
          - create nix directory for user nix installation
+         - disable sandboxing because its unsupported inside a Docker container
          - install nix package manager for user
+
         """
         return super().get_build_scripts() + [
             (
                 "root",
                 """
             mkdir -m 0755 /nix && \
-            chown -R ${NB_USER}:${NB_USER} /nix /usr/local/bin/nix-shell-wrapper /home/${NB_USER}
+            chown -R ${NB_USER}:${NB_USER} /nix /usr/local/bin/nix-shell-wrapper /home/${NB_USER} && \
+            mkdir -p /etc/nix && \
+            touch /etc/nix/nix.conf && \
+            echo "sandbox = false" >> /etc/nix/nix.conf
             """,
             ),
             (
--- a/tests/nix/test-building/default.nix
+++ b/tests/nix/test-building/default.nix
@@ -1,0 +1,27 @@
+let
+  # Pinning nixpkgs to specific release
+  # To get sha256 use "nix-prefetch-git <url> --rev <commit>"
+  commitRev="5574b6a152b1b3ae5f93ba37c4ffd1981f62bf5a";
+  nixpkgs = builtins.fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/archive/${commitRev}.tar.gz";
+    sha256 = "1pqdddp4aiz726c7qs1dwyfzixi14shp0mbzi1jhapl9hrajfsjg";
+  };
+  pkgs = import nixpkgs { config = { allowUnfree = true; }; };
+
+  # Test that we can actually build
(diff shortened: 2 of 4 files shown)


FRidh:
Moved the version bump to https://github.com/jupyterhub/repo2docker/pull/991.

docs/source/changelog.rst
```diff
@@ -40,6 +40,7 @@ Bug fixes
   :user:`manics.`
 - Add help text for commandline arguments in :pr:`517` by :user:`yuvipanda`.
 - Fix submodule checkout in :pr:`809` by :user:`davidbrochart`.
+- Fix Nix builds in :pr:`990` by :user:`FRidh`.
```
FRidh, line 43: wrong release

betatim: This changelog is for a past release so I don't think it makes sense to add to them. What was your thinking here?

FRidh: Didn't check the name or number, just assumed the one at the top is the unreleased one, considering project often have that.

I tried following the guidelines here, but its actually not possible to follow them regarding the changelog part because there is no section for them.

tests/nix/test-building/default.nix
```diff
@@ -0,0 +1,27 @@
...
+
+  # Test that we can actually build
+  test-build = pkgs.runCommand "test-build" { } ''
+    touch $out
+  '';
+
+in
+pkgs.mkShell {
+  buildInputs = with pkgs; [
+    python36Packages.numpy
```
betatim, line 19: I think this means we are building numpy from scratch. This can take a bit of time. We are trying to speed up/keep at current level the runtime of our tests (as it is already pretty long).

I don't think there is a way to test for this without having to build a container image. Maybe we can think of one together? Basically more of a unit test type test than an "end to end" test like this.

If we can't replace this by a unit test it would be good to replace `numpy` with something like [`there`](https://pypi.org/project/there/) which is super quick to install. Or maybe we don't even have to install any additional Python packages?

FRidh: > I think this means we are building numpy from scratch. This can take a bit of time. We are trying to speed up/keep at current level the runtime of our tests (as it is already pretty long).

No, it does not. It is cached by the NixOS organization.

betatim:
Thanks for finding and fixing this! I left a few small comments, mostly about trying to speed up the tests.

repo2docker/buildpacks/nix/__init__.py
```diff
@@ -23,17 +23,20 @@ def get_build_scripts(self):
...
+
         """
         return super().get_build_scripts() + [
             (
                 "root",
                 """
             mkdir -m 0755 /nix && \
-            chown -R ${NB_USER}:${NB_USER} /nix /usr/local/bin/nix-shell-wrapper /home/${NB_USER}
+            chown -R ${NB_USER}:${NB_USER} /nix /usr/local/bin/nix-shell-wrapper /home/${NB_USER} && \
+            mkdir -p /etc/nix && touch /etc/nix/nix.conf && echo "sandbox = false" >> /etc/nix/nix.conf
```
betatim, line 39: nitpick: can you split this command into multiple lines like he test? Otherwise it is easy to overlook that this line also does `touch` and `echo`. Mostly because almost all(?) of the rest of the repo2docker code base splits commands into multiple lines connected with `&& \`

FRidh: I did that on purpose, because these three parts absolutely needs to be done together, but sure, I can do that

docs/source/changelog.rst
betatim: Sorry about the misleading instructions. My recommendation would be to remove this line.

Or if you have the time start a new section for the next release. Though I am undecided if we will keep having manually made change logs as we are moving towards automatically tagging and releasing repo2docker once a month.

FRidh:
Changelog removed.

FRidh:
Figshare failure is unrelated.

FRidh:
As far as I can tell this is ready to go.

FRidh:
Bugfix is still needed.

PR #992, opened 2020-12-06
Use mambaforge instead of miniforge

SylvainCorlay:
Using mambaforge instead of miniforge, and saving some time by removing the call to `conda install mamba`.

### DIFF
--- a/repo2docker/buildpacks/conda/install-miniforge.bash
+++ b/repo2docker/buildpacks/conda/install-miniforge.bash
@@ -5,12 +5,11 @@ set -ex
 
 
 cd $(dirname $0)
-MINIFORGE_VERSION=4.8.2-1
-MAMBA_VERSION=0.6.1
+MINIFORGE_VERSION=4.9.2-2
 # SHA256 for installers can be obtained from https://github.com/conda-forge/miniforge/releases
-SHA256SUM="4f897e503bd0edfb277524ca5b6a5b14ad818b3198c2f07a36858b7d88c928db"
+SHA256SUM="7a7bfaff87680298304a97ba69bcf92f66c810995a7155a2918b99fafb8ca1dc"
 
-URL="https://github.com/conda-forge/miniforge/releases/download/${MINIFORGE_VERSION}/Miniforge3-${MINIFORGE_VERSION}-Linux-x86_64.sh"
+URL="https://github.com/conda-forge/miniforge/releases/download/${MINIFORGE_VERSION}/Mambaforge-${MINIFORGE_VERSION}-Linux-x86_64.sh"
 INSTALLER_PATH=/tmp/miniforge-installer.sh
 
 # make sure we don't do anything funky with user's $HOME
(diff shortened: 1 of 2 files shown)


betatim:
Converted this PR to "draft mode" which is what others have started using to signal "please don't merge yet, still working on this". An alternative is to put "[WIP]" at the start of the title but I think that is going out of fashion.

SylvainCorlay:
Thanks @betatim, the pre-release bacame a release in the meantime. This is ready to go!

minrk (approved):

repo2docker/buildpacks/conda/install-miniforge.bash
```diff
@@ -42,11 +41,6 @@ echo 'update_dependencies: false' >> ${CONDA_DIR}/.condarc
 # avoid future changes to default channel_priority behavior
 conda config --system --set channel_priority "flexible"
 
-# do all installation with mamba
-time conda install -y mamba==${MAMBA_VERSION}
```
minrk, line 46: Maybe we want to keep `mamba install -y mamba==${MAMBA_VERSION}` to allow pinning versions more recent than what's in the mambaforge release? We always had this for conda before adopting mamba, because we could update conda to be more recent than a miniconda release.

SylvainCorlay: Why not, although it may defeat the purpose of saving this call to the solver.

minrk: But at least it's mamba instead of conda, and should be quick if the version matches what is already isstalled, right?

SylvainCorlay: Ah yes I can use mamba!

minrk: use:

```bash
mamba install -y -S mamba==${MAMBA_VERSION}
```

and it shouldn't download repodata or anything if the version is what's already installed

minrk: Hm, nevermind. My env is weird and mamba does seem to fetch repodata even if specs are already satisfied. We could add checks beforehand, but I don't think there's much point since repodata will be fetched immediately on the next line anyway

SylvainCorlay: I have a reservation about this: what if the set mamba version causes a downgrade of conda? I think that conda downgrades are only allowed if a certain configuration parameter is set.

SylvainCorlay: @minrk let me know what you think!

minrk: Realistically, we are never going to request mamba older than what's in mambaforge, so it should only cause upgrades, not downgrades. We always used this in the past to get a conda update that's not in miniconda yet, and would only use this if there's a mamba update (or conda even!) that's *not* in the latest mambaforge, but we still want.

This could be moot if you think mambaforge will be updated on ~every conda or mamba release. Miniconda was pretty far from that, though.

SylvainCorlay: OK, I re-added the update of mamba to the version specified.

minrk:
Woo!

PR #994, opened 2020-12-19
[MRG] Fetch available Julia versions from hosted json

tomyun:
Instead of adding hard-coded version strings whenever a new release comes out, this PR fetches a list of currently available Julia versions from officially hosted [JSON](https://julialang-s3.julialang.org/bin/versions.json) file (JuliaLang/www.julialang.org#1101).

1. The resulting list `all_julias` should cover the same range we used to have, except it now contains more versions starting from 0.1.2, not 0.7.0 as we have now. I don't think this would cause a problem though as any projects relying on old 0.x versions would use REQUIRE file covered by julia_require.py.

2. Setting `default_julia_version` from `all_julias` is pushed back to the end of `julia_version` property to minimize json requests.

3. I noticed that our `find_semver_match()` is currently not fully compatible with [Pkg.jl compat specifiers](https://julialang.github.io/Pkg.jl/v1.5/compatibility/). For example,  `"= 1.2.3"` with space would result into an error with repo2docker. We also don't support pre-release strings like `"1.5.0-rc1"`. We may address these issues in a separate PR later on.

### DIFF
--- a/repo2docker/buildpacks/julia/julia_project.py
+++ b/repo2docker/buildpacks/julia/julia_project.py
@@ -18,9 +18,12 @@ class JuliaProjectTomlBuildPack(PythonBuildPack):
     @property
     @functools.lru_cache(maxsize=1)
     def all_julias(self):
-        json = requests.get(
-            "https://julialang-s3.julialang.org/bin/versions.json"
-        ).json()
+        try:
+            json = requests.get(
+                "https://julialang-s3.julialang.org/bin/versions.json"
+            ).json()
+        except Exception as e:
+            raise RuntimeError("Failed to fetch available Julia versions: {e}")
         vers = [semver.VersionInfo.parse(v) for v in json.keys()]
         # filter out pre-release versions not supported by find_semver_match()
         filtered_vers = [v for v in vers if v.prerelease is None]
--- a/repo2docker/buildpacks/julia/julia_project.py
+++ b/repo2docker/buildpacks/julia/julia_project.py
@@ -1,8 +1,10 @@
 """Generates a Dockerfile based on an input matrix for Julia"""
+import functools
 import os
+import requests
 import toml
 from ..python import PythonBuildPack
-from .semver import find_semver_match
+from .semver import find_semver_match, semver
 
 
 class JuliaProjectTomlBuildPack(PythonBuildPack):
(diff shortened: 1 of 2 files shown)


tomyun:
Can we get a review on this? Next major release (1.6) is currently on beta and we're also expecting a minor release (1.5.4) coming out soon.

repo2docker/buildpacks/julia/julia_project.py
```diff
@@ -1,5 +1,8 @@
 """Generates a Dockerfile based on an input matrix for Julia"""
+from functools import cmp_to_key
 import os
+import requests
+import semver
```
betatim, line 5: Should this be `from . import semver` and a few lines further down with the other imports from within the package?

repo2docker/buildpacks/julia/julia_project.py
```diff
@@ -13,32 +16,20 @@ class JuliaProjectTomlBuildPack(PythonBuildPack):
...
-        "1.3.1",
-        "1.4.0",
-        "1.4.1",
-        "1.4.2",
-        "1.5.0",
-        "1.5.1",
-        "1.5.2",
-        "1.5.3",
-    ]
+    @property
```
betatim, line 19: What about combining this with `@functools.lru_cache(maxsize=1)`? It would skip HTTP requests on repeated look up. We can't use `functools.cache` because it is too new :(

betatim:
I left two review comments about details for this implementation.

One other "big picture" comment: should we do this at release time instead of at runtime? This would mean running this code in `setup.py` and packaging the available versions when it runs. I think I'd prefer that slightly over this. It reduces the number of requests made at runtime, less hassle if running r2d in an environment without (good) public internet access, no leaking of user details like IP address to the julia server and for support cases we know what the versions connected to a particular version of repo2docker are.

repo2docker/buildpacks/julia/julia_project.py
tomyun: I changed it to import python package `semver` from local module `semver`.

tomyun:
Thanks for the review!

I also had a similar idea at first then realized users need to download a much larger archive of Julia binary from the same server anyways, so benefit would be minimal. We can't use r2d for Julia if public internet access is not available.

Tying a particular version of Julia and r2d sounds like a good idea for support, but I would personally prefer to always get the latest stable version unless there was a particular requirement coded in Project.toml. Otherwise we would still need to wait until a new release of r2d for whatever other updates in order to have a new version of Julia available in the pipeline.

tomyun:
Rebased on master to get rid of unrelated test errors. Code also got some changes due to fixing deprecation warnings with semver package (python-semver/python-semver#233).

minrk:
>  I would personally prefer to always get the latest stable version unless there was a particular requirement coded in Project.toml

That does make sense as a desire, though it isn't how repo2docker works for anything else, and I'm reticent to make Julia behave differently from every other interpreter. All 'default' versions of all interpreters are pinned for a given repo2docker version. With Julia, it happens to be the most recent available version in the repo2docker list, but since that's static it has a similar behavior.

How would you feel about specifying a default minor version in repo2docker, say `1.5` and installing the latest matching patch release? Then minor releases are opt-in and patch releases are automatic.

repo2docker/buildpacks/julia/julia_project.py
```diff
@@ -54,6 +48,7 @@ def julia_version(self):
                 if _julia_version is not None:
                     return _julia_version
 
+        default_julia_version = self.all_julias[-1]
```
minrk, line 51: might be worth catching errors here:

```suggestion
        try:
            default_julia_version = self.all_julias[-1]
        except Exception as e:
            raise RuntimeError("Failed to fetch available Julia versions: {e}")
```

tomyun: Good idea. They're now handled around the actual `requests.get()` call.

tomyun:
Sounds like a fair point. I've added a default fallback compat version string which is currently set to "1.5".

minrk:
Great, thanks!

PR #995, opened 2020-12-30
[WIP] freeze.py: Support podman and selinux

manics:
Adds two arguments to `freeze.py` so it can optionally be run with Podman and/or SELinux.

I also ran `freeze.py` to update existing unpinned dependencies.

### DIFF
--- a/repo2docker/buildpacks/conda/freeze.py
+++ b/repo2docker/buildpacks/conda/freeze.py
@@ -128,7 +128,7 @@ def set_python(py_env_file, py):
 if __name__ == "__main__":
     parser = ArgumentParser()
     parser.add_argument("--containerrun", default="docker", help="Container runtime")
-    parser.add_argument("--volopts", default="Host volume mount options")
+    parser.add_argument("--volopts", help="Host volume mount options")
     parser.add_argument(
         "pys",
         nargs="*",
--- a/repo2docker/buildpacks/conda/freeze.py
+++ b/repo2docker/buildpacks/conda/freeze.py
@@ -10,7 +10,7 @@ Usage:
 
 python freeze.py [3.8]
 
-If you want to run this with Podman add '--containerrun podman'.
+If you want to run this with Podman add '--containerruntime podman'.
 If you need additional host volume options, e.g. due to SELinux, add '--volopts z'
 """
 
@@ -41,7 +41,7 @@ FROZEN_FILE_T = os.path.splitext(ENV_FILE_T)[0] + ".frozen.yml"
 yaml = YAML(typ="rt")
 
 
-def freeze(env_file, frozen_file, containerruntime="docker", hostmountopts=""):
+def freeze(env_file, frozen_file, containerruntime="docker", hostmountopts=None):
     """Freeze a conda environment.yml
 
     By running in docker:
@@ -50,6 +50,9 @@ def freeze(env_file, frozen_file, containerruntime="docker", hostmountopts=""):
         conda env export
 
     Result will be stored in frozen_file
+
+    containerruntime is the executable used to run a container,
+    for example docker or podman
     """
     frozen_dest = HERE / frozen_file
 
@@ -71,6 +74,8 @@ def freeze(env_file, frozen_file, containerruntime="docker", hostmountopts=""):
 
     if hostmountopts and hostmountopts[0] != ":":
         hostmountopts = f":{hostmountopts}"
+    else:
+        hostmountopts = ""
     check_call(
         [
             containerruntime,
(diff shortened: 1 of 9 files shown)


repo2docker/buildpacks/conda/freeze.py
```diff
@@ -35,7 +41,7 @@
 yaml = YAML(typ="rt")
 
 
-def freeze(env_file, frozen_file):
+def freeze(env_file, frozen_file, containerruntime="docker", hostmountopts=""):
```
betatim, line 44: Can we come up with a more descriptive name for `containerruntime`? When I started reading the code I was preparing myself for witnessing some compatibility magic dance but after reading it I think this is "just" the name of the executable to use. However `containerruntime_executable` is even worse than the current name, so I have no good alternative suggestion. But think it is worth word smithing for a few minutes to see if we can find a nice name.

betatim, line 44: Is the `hostmountopts=""` on purpose? Asking because the 'pythonic' thing is to use `None` as the empty default value instead of `[]` or `""`. That way `None` is the "I am not passing anything, please use the default" value and `""` is the "I (the caller) passed `""` on purpose, please use it and not the default" value.

For my education: what are host mount options?

repo2docker/buildpacks/conda/freeze.py
```diff
@@ -118,13 +126,26 @@ def set_python(py_env_file, py):
 
 
 if __name__ == "__main__":
-    # allow specifying which Pythons to update on argv
-    pys = sys.argv[1:] or ("2.7", "3.6", "3.7", "3.8")
+    parser = ArgumentParser()
+    parser.add_argument("--containerrun", default="docker", help="Container runtime")
+    parser.add_argument("--volopts", default="Host volume mount options")
```
betatim, line 131:
```suggestion
    parser.add_argument("--volopts", help="Host volume mount options")
```
should it also have a default value?

repo2docker/buildpacks/conda/freeze.py
manics: I'm not a big fan of the `containerruntime` name either, but I already went through a few other options before choosing this and I can't think of anything better. Anyone else have a suggestion?

manics: I chose `""` because the default value should be no additional options to maintain the existing behaviour. I can change it to `None`, and add another statement `if hostmountopts is None: hostmountopts = ""` if that's preferable?

Thinking about it maybe `mountopts` is a better name the `hostmountopts` since in theory you could pass other Docker mount options though I can't think why (for example readonly `ro` is one option but that's obviously not much use for this script :smile: ).

If you have SELinux enabled by default it prevent containers mounting arbitrary filesystem directories such as home directories. To give it permissions you have to relabel (set soem SELinux attributes) on the directory you want to mount, either manually, or by passing the `z` mount option to tell docker or podman to automatically relabel the directory.

repo2docker/buildpacks/conda/freeze.py
manics: I thought argparse defaults to `""`? I can make it explicit though.

manics:
Thanks for reviewing! There's a mysterious failing test....

repo2docker/buildpacks/conda/freeze.py
betatim: Maybe we stick with it and add a comment that is a sentence that says something like "This name is a bit weird. The variable contains the name of the container runtime executable."

Or we leave it as is :-/

betatim: Ja dunno. The `if hostmounopts is None:` pattern is more typing but maybe also more  common" No strong feelings really.

repo2docker/buildpacks/conda/freeze.py
betatim: I'm confused. Currently you are setting the default value of this argument to `"Host volume mount options"`. This seems like it should be the help text, not the value. I hadn't thought of the default value of `default=`

manics: Sorry, you're right! I made an error, and also managed to misread your comment.

betatim:
I think the failing test is related to figshare for which you made a PR (🎉).

The other one (dockercloud-stage) is even more mysterious and we need @minrk to take a look (I think). At least I don't seem to be able to see anything but a 404 when I click the "Details" link. The reason to care about this is that we rely on dockerhub to build our images. If that is broken then the next time henchbot proposes a PR and it gets merged it will try to deploy an image that doesn't exist. Which isn't good :D

minrk:
docker-cloud was a timeout. I've disabled the docker-cloud CI status (I think) since it's redundant with the builds now added to github actions, and opaque to ~everyone so not a super useful tool!

manics:
https://github.com/jupyterhub/repo2docker/pull/1024 is a better solution that gets rid of the need for a container.

Issue #961, opened 2020-09-15
jupyter_offlinenotebook missing for custom environment


tomyun:
### Bug description

When using a custom version of python by setting up a dependency in `environment.yml`, `jupyter_offlinenotebook` is installed in a directory not picked up by the custom environment, leading to an error logged by Jupyter.

Here is the error shown up when launching Jupyter for a docker image with Python 3.8:
```
[W 16:53:26.843 NotebookApp] Error loading server extension jupyter_offlinenotebook
    Traceback (most recent call last):
      File "/srv/conda/envs/notebook/lib/python3.8/site-packages/notebook/notebookapp.py", line 1942, in init_server_extensions
        mod = importlib.import_module(modulename)
      File "/srv/conda/envs/notebook/lib/python3.8/importlib/__init__.py", line 127, in import_module
        return _bootstrap._gcd_import(name[level:], package, level)
      File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
      File "<frozen importlib._bootstrap>", line 991, in _find_and_load
      File "<frozen importlib._bootstrap>", line 973, in _find_and_load_unlocked
    ModuleNotFoundError: No module named 'jupyter_offlinenotebook'
```

#### Expected behaviour
`jupyter_offlinenotebook` should be installed in a directory used by custom environment (i.e. `/srv/conda/envs/notebook/lib/python3.8/site-packages` in this case).

#### Actual behaviour
`jupyter_offlinenotebook` is installed in the default directory (`/srv/conda/envs/notebook/lib/python3.7/site-packages`).

### How to reproduce
1. Define a custom Python version in `environment.yml`:
```yml
dependencies:
  - python=3.8
```
2. Build a docker image and run.
3. Error shown up in the log.

### Your personal set up
 - OS: macOS 10.15
 - Docker version: 19.03.12
 - repo2docker version: 0.11.0+134.g64371cb

manics:
I wasn't able to reproduce this:
https://gist.github.com/manics/e16a6d746862015c2f92930e092a5b5f

Note there was a warning about needing to rebuild jupyterlab (https://github.com/jupyterhub/mybinder.org-deploy/issues/1607), but I think that's a separate issue.

Can you provide more details on your setup?

tomyun:
I should have mentioned that I was using JuliaProjectTomlBuildPack. You could trigger this by having an empty `Project.toml` in `binder` directory where `environment.yml` is located.

PR #997, opened 2021-01-04
Update jupyter-offlinenotebook to 0.2.1

manics:
This includes the JupyterLab 3 extension that does not require a build step. It remains compatible with JupyterLab 2 (but the extension must be installed separately).

As far as JupyterLab 2 is concerned there is no change. For JupyterLab3 the extension is effectively pre-installed since it's part of the PyPI package, so it should "just work" if someone choose to install JupyterLab 3.

See also:
- https://github.com/jupyterlab/jupyterlab-demo/pull/102#issuecomment-754188038
- https://github.com/jupyterhub/repo2docker/pull/996

### DIFF
--- a/repo2docker/buildpacks/conda/install-miniforge.bash
+++ b/repo2docker/buildpacks/conda/install-miniforge.bash
@@ -51,7 +51,7 @@ time mamba env create -p ${NB_PYTHON_PREFIX} -f /tmp/environment.yml
 # Install jupyter-offline-notebook to allow users to download notebooks
 # after the server connection has been lost
 # This will install and enable the extension for jupyter notebook and JupyterLab 3
-time ${NB_PYTHON_PREFIX}/bin/python -m pip install jupyter-offlinenotebook==0.2.0
+time ${NB_PYTHON_PREFIX}/bin/python -m pip install jupyter-offlinenotebook==0.2.1
 # and this installs it for lab 2. Keep going if the lab version is incompatible
 # with the extension.
 # Don't minimize build as it may fail, possibly due to excessive resource usage
--- a/repo2docker/buildpacks/conda/install-miniforge.bash
+++ b/repo2docker/buildpacks/conda/install-miniforge.bash
@@ -50,9 +50,9 @@ time mamba env create -p ${NB_PYTHON_PREFIX} -f /tmp/environment.yml
 
 # Install jupyter-offline-notebook to allow users to download notebooks
 # after the server connection has been lost
-# This will install and enable the extension for jupyter notebook
-time ${NB_PYTHON_PREFIX}/bin/python -m pip install jupyter-offlinenotebook==0.1.0
-# and this installs it for lab. Keep going if the lab version is incompatible
+# This will install and enable the extension for jupyter notebook and JupyterLab 3
+time ${NB_PYTHON_PREFIX}/bin/python -m pip install jupyter-offlinenotebook==0.2.0
+# and this installs it for lab 2. Keep going if the lab version is incompatible
 # with the extension.
 # Don't minimize build as it may fail, possibly due to excessive resource usage
 # https://discourse.jupyter.org/t/tip-binder-jupyterlab-extension/6022


jtpio:
Thanks!

Will the `jupyter lab build --minimize=False` command below still trigger a new build? Maybe it's possible to not run it if the JupyterLab version is >= 3.0?

manics:
I _think_ that repo2docker:
1. Installs miniforge
2. Installs a default set of packages defined from the `environment*.yml` in https://github.com/jupyterhub/repo2docker/tree/d8a3d7fa0ac8b1ec151d473a86ca813535369a54/repo2docker/buildpacks/conda (which includes JupyterLab 2)
3. Installs the user environment, which is where JupyterLab would be upgraded

Hopefully someone will correct me if I'm wrong!

manics:
The conda-forge package has also been updated to `0.2.0`: https://anaconda.org/conda-forge/jupyter-offlinenotebook/files

So if we want we could switch to that as suggested in https://github.com/jupyterhub/repo2docker/pull/846 https://github.com/jupyterhub/repo2docker/pull/996
I'm happy to make that change in this PR if that's the preferred option.

tomyun:
IIUC, switching to use conda-forge package in `environment.yml` may help solving #961.

betatim:
> I _think_ that repo2docker:
> 
>     1. Installs miniforge
> 
>     2. Installs a default set of packages defined from the `environment*.yml` in [`d8a3d7f`/repo2docker/buildpacks/conda](https://github.com/jupyterhub/repo2docker/tree/d8a3d7fa0ac8b1ec151d473a86ca813535369a54/repo2docker/buildpacks/conda) (which includes JupyterLab 2)
> 
>     3. Installs the user environment, which is where JupyterLab would be upgraded
> 

I also think this is what happens.

That means that by default repo2docker runs with jupyterlab 2 and in order to install offline-notebook we have to run the build step. At a later point in the build process the user might install lab 3. The new thing here is that the user won't have to run the build step again. Right now lots of repos use a `postBuild` to re-run the build. This is what can be removed after we merge this PR.

betatim:
Any idea why the (3.8, conda) tests fail with `No matching distribution found for jupyter-offlinenotebook==0.2.0`?

manics:
My first guess was a stale PyPI cache, but I've re-run the tests and they're still failing so I'll need to investigate further.

manics:
The failing test uses a Python 3.5 environment, but I set the minimum Python version to 3.6
https://github.com/manics/jupyter-offlinenotebook/blob/24c3429d06660a21f05c912a041479a33db66a77/setup.py#L96
I forgot that repo2docker mixes the Jupyter front end and the execution environment for Python 3.*.

I can think of a few options:
1. Use same strategy as installing the JupyterLab extension and allow it to fail: `pip install jupyter-offlinenotebook==0.2.0 || true`.
2. Change https://github.com/manics/jupyter-offlinenotebook/blob/24c3429d06660a21f05c912a041479a33db66a77/setup.py#L96 to `python_requires>=3.5` and release `0.2.1`. The previous release was missing a `python_requires` statement which is why it previously worked.
3. Switch to the conda package, which means different versions of jupyter-offlinenotebook could be used. The [3.5 environment](https://github.com/jupyterhub/repo2docker/blob/560b1d96a0e39cb8de53cb41a7c2d8d23384eb82/repo2docker/buildpacks/conda/environment.py-3.5.frozen.yml) could either use version 0.1.0, or we could just omit the extension since any Python 3.5 repositories will be quite old, and users probably won't miss this extension. The other [3.* environments](https://github.com/jupyterhub/repo2docker/tree/560b1d96a0e39cb8de53cb41a7c2d8d23384eb82/repo2docker/buildpacks/conda) would use 0.2.0.

1 is the least work. 3 is the most work, but I think it's worth doing if everyone's agreed. What do you think?

manics:
Does anyone have a preference for any of the three options in https://github.com/jupyterhub/repo2docker/pull/997#issuecomment-755698303 ? Otherwise I'll go with (3), switching to conda

betatim:
I think using a conda package is good and the only downside is that it is more work to setup now. I think providing an older version is fine for Python 3.5.

manics:
I'm closing this as it looks like https://github.com/jupyterhub/repo2docker/pull/996 will be merged soon

PR #998, opened 2021-01-04
[MRG] Weekly test of master to check for external failures

manics:
Test failures in PRs can sometimes be quite hard to debug. By adding a weekly automated test run of the default branch we'll have a bit more information on whether a PR is failing due to a change in that PR or due to some external factor. I've chosen Sunday, the time `10:36` was picked using `$RANDOM`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,10 +7,13 @@
 
 name: Continuous Integration
 
-# Only trigger the workflow when pushing to master
 on:
-  - push
-  - pull_request
+  push:
+  pull_request:
+  schedule:
+    # Weekly test so we know if tests break for external reasons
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -7,10 +7,13 @@
 
 name: Continuous Integration
 
-# Only trigger the workflow when pushing to master
 on:
```
manics, line 10: I've changed this from a list to a dict to follow the schema from https://json.schemastore.org/github-workflow (otherwise VSCode complains).

betatim:
Do you know where we will be able to see the status of this run? Will it make the status of the last commit on the default branch red if it fails?

manics:
It'll appear under the actions tab, e.g. https://github.com/ome/ansible-role-nginx-proxy/actions
though it looks like the last commit status is linked to the original run, not the scheduled repeats.

PR #1003, opened 2021-01-13
Add entrypoint script which automatically propagates *_PROXY env vars…

g-braeunlich:
… to docker config

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,5 +30,10 @@ RUN pip3 install --no-cache-dir /tmp/wheelhouse/*.whl \
 COPY ./docker/git-credential-env /usr/local/bin/git-credential-env
 RUN git config --system credential.helper env
 
+# add entrypoint
+COPY ./docker/entrypoint /usr/local/bin/entrypoint
+RUN chmod +x /usr/local/bin/entrypoint
+ENTRYPOINT ["/usr/local/bin/entrypoint"]
+
 # Used for testing purpose in ports.py
 EXPOSE 52000
--- a/docker/entrypoint
+++ b/docker/entrypoint
@@ -1,0 +1,35 @@
+#!/bin/sh
+set -e
+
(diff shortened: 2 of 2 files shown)


manics (approved):
This looks good to me- I tested it in a VM.

One thing I noticed is that this configures the proxy for Dockerfile `RUN` commands during docker build, but not for pulling the `FROM` image- I think this can only be changed by configuring the parent docker daemon outside the container which makes sense when you think about it, but it might make the corresponding BinderHub configuration more complicated.

@betatim You're [making a release soon](https://github.com/jupyterhub/repo2docker/pull/1004), are you happy for this to be merged or do you want to wait until after the release?

meeseeksmachine:

https://discourse.jupyter.org/t/binder-behind-outbound-proxy/7428/5

manics:
@betatim Do you have any concerns with this? Otherwise let's merge it.

minrk (approved):
Happy to have this change!

I'll just comment for anyone who has concerns about maintaining the generation of valid JSON config produced by `echo`ing lines, repo2docker is in Python so we could write this config file with Python and make it a little simpler, e.g.:

```python
config = {"proxies": {"default": {}}
for env, key in {"HTTP_PROXY": "httpProxy"...}.items():
    value = os.environ.get(env)
    if value:
        config["proxies"]["default"][key] = value

if config["proxies"]["default"]:
    with open(config_path, "w") as f:
        json.dump(config, f)

os.execv(...)
```

g-braeunlich:
Ah, I see. Would you recommend to integrate this into the repo2docker script rather than a entrypoint?

minrk:
> Would you recommend to integrate this into the repo2docker script rather than a entrypoint?

No, I meant that the entrypoint script itself could be in Python if we have issues with this. It's fine as-is, though!

PR #1008, opened 2021-01-28
Investigating the missing logs

betatim:
This is a follow up to #987 

As we observed there the `test_env.py` tests break. Locally this also happens, but only sometimes. After about half a day of messing around and trying to figure out where the logs go I am resigning myself to not being able to figure it out :-/

It seems even with somewhat "extreme" measures like in the PR you either get the logs from the command executed (both in the "normal" logs and the newly added "late logs" or you don't get them in either).

A thing that seems to help is to add a `; sleep 1` to the command the test runs.

This makes me think that the issue is related to buffering of some kind, some where. I am tempted to suggest we add the `sleep` to the test and assume that the flakeyness is due to the very short execution time of the container, which isn't typical for repo2docker containers.

Unfortunately I don't understand enough about how the logging in docker works to know if changing config options in the logging driver https://docs.docker.com/config/containers/logging/configure/ would help or not :-/

A popular google result regarding "my docker logs are missing" is a note that journald rate limits log producers but I think this doesn't apply to us because I am running OSX. This means I don't have journald and as far as I understand it the default is logging driver is the `json-file` driver which doesn't use journald.

Thoughts welcome on how to fix this properly or the workaround.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -605,12 +605,24 @@ class Repo2Docker(Application):
         try:
             for line in container.logs(stream=True):
                 self.log.info(line.decode("utf-8"), extra=dict(phase="running"))
+
         finally:
             container.reload()
             if container.status == "running":
                 self.log.info("Stopping container...\n", extra=dict(phase="running"))
                 container.kill()
             exit_code = container.attrs["State"]["ExitCode"]
+
+            container.wait()
+
+            self.log.info(
+                "Container finished running.\n".upper(), extra=dict(phase="running")
+            )
+            # are there more logs? Let's send them back too
+            late_logs = container.logs().decode("utf-8")
+            for line in late_logs.split("\n"):
+                self.log.info(line + "\n", extra=dict(phase="running"))
+
             container.remove()
             if exit_code:
                 sys.exit(exit_code)
--- a/repo2docker/buildpacks/repo2docker-entrypoint
+++ b/repo2docker/buildpacks/repo2docker-entrypoint
@@ -15,10 +15,10 @@ if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
     if [[ ! -x "$R2D_ENTRYPOINT" ]]; then
         chmod u+x "$R2D_ENTRYPOINT"
     fi
-    exec "$R2D_ENTRYPOINT" "$@" >&"$log_fd"
+    exec "$R2D_ENTRYPOINT" "$@" 2>&1 >&"$log_fd"
 else
-    exec "$@" >&"$log_fd"
+    exec "$@" 2>&1 >&"$log_fd"
 fi
 
 # Close the logging output again
-#exec {log_fd}>&-
+exec {log_fd}>&-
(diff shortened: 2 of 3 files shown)


minrk:
I poked around for a bit and was able to reproduce it consistently until I got my debugging in place, then it promptly started succeeding 100% of the time and hasn't failed since (even after reverting changes).

Buffer ideas:

- PYTHONUNBUFFERED might help
- maybe calling sys.stdout / sys.stderr.flush()?
- [stdbuf -o0](https://www.xspdf.com/resolution/50725719.html) might be the equivalent for bash?

minrk:
Given that failure sems to have become more frequent, I think maybe we should merge this one as-is to unblock folks and proceed with.

Maybe the above `stdbuf -o{log_fd}`  should go in the entrypoint in the right place?

repo2docker/app.py
```diff
@@ -605,12 +605,24 @@ def wait_for_container(self, container):
...
             exit_code = container.attrs["State"]["ExitCode"]
+
+            container.wait()
+
+            self.log.info(
+                "Container finished running.\n".upper(), extra=dict(phase="running")
+            )
+            # are there more logs? Let's send them back too
+            late_logs = container.logs().decode("utf-8")
+            for line in late_logs.split("\n"):
```
minrk, line 623: In my runs, late_logs is never empty, it's always the complete logs of the container. So this is always duplicating the logs and *also* missed the same logs.

repo2docker/buildpacks/repo2docker-entrypoint
```diff
@@ -15,10 +15,10 @@ if [[ ! -z "${R2D_ENTRYPOINT:-}" ]]; then
     if [[ ! -x "$R2D_ENTRYPOINT" ]]; then
         chmod u+x "$R2D_ENTRYPOINT"
     fi
-    exec "$R2D_ENTRYPOINT" "$@" >&"$log_fd"
+    exec "$R2D_ENTRYPOINT" "$@" 2>&1 >&"$log_fd"
```
minrk, line 18: I think this merge is a good thing to do, regardless

manics:
> Unfortunately I don't understand enough about how the logging in docker works to know if changing config options in the logging driver https://docs.docker.com/config/containers/logging/configure/ would help or not :-/

A bit late... but I don't think the Docker daemon logging options will make a difference as they only have an effect _after_ the logs have been emitted from the container on stdout/stderr. They're more to do with how the Docker daemon stores or forwards them.

I'm also in favour of merging this to fix the logs, and to see if it somehow makes the logs appear on mybinder (so far in all containers I've launched `~/.jupyter-server-log.txt` is always empty).

minrk:
@manics good call. I'll go ahead and merge, and try to refine a bit as I attempted (unsuccessfully) in #1014

repo2docker/app.py
betatim: Absolutely right. I added this thinking that maybe we missed some logs. But that isn't the problem.

If I remove/comment out the "delete stopped container" part of r2d and then run `docker logs <container>` afterwards the logs are also empty (when this problems occurs). So I think your ideas about `PYTHONUNBUFFERED` and `stdbuf` are the way to go investigating.

I think, the "late log" lines should be removed

Issue #1009, opened 2021-01-28
Add support for postBuildAdmin


g-braeunlich:
### Proposed change
In addition to `postBuild` script (userspace) also process `postBuildAdmin` scripts (system space).

### Alternative options / Who would use this feature?
See discussions in #487 and https://gitter.im/jupyterhub/binder?at=5fce35b6fb7f155587ad481a

betatim:
My initial reaction is that we shouldn't support this because it is not a frequently used feature and it adds another "magic" file.

repo2docker itself tries to support frequently used community practice workflows. And making it as effortless as possible for the user if they follow the community practices. This means we choose the trade-off in favour of the frequent and against the rare. For rare or expert use cases there is the escape hatch of providing your own `Dockerfile`.

The magic files we already have in repo2docker are one of the things we regret the most because they are totally custom. Which means they make things more complicated and are less well thought out. So we are trying to not add new ones unless there really, really is no other way.

This means: can we find more use cases where a user needs this feature and there is no alternative? For example the user in the linked gitter chat seemed happy to use a `Dockerfile`. Are there some specific cases in #487 that we could cover with this? Asking because in my mind "I want to change the base image" is often about "I don't want Ubuntu but X" which we can't solve with a `postAdminBuild`.

trybik:
This is thel use-case that started this thread:

> We want users (effectively jovian user), to be able to modify runtime environment; in particular, to run sudo apt-get update/install ... over Terminal in the Notebook Server session but also maybe install other scientific libraries which don't really support easily user installation.

https://github.com/jupyterhub/repo2docker/issues/487#issuecomment-698409277

The `postBuildAdmin` solution was suggested and briefly discussed w/ @manics in the #487 thread (cf. https://github.com/jupyterhub/repo2docker/issues/487#issuecomment-698473158 ).

In both this and ImageMagic cases it's not about change of the actual base image but about admin-only config.

manics:
I think part of the issue is that repo2docker allows you to install operating system packages using `apt`, and since it's at the OS level that means it's often only configurable by root. As an aside, there's another future problem here- if the base image is upgraded from
https://github.com/jupyterhub/repo2docker/blob/cb4621f2544efba496f6cf58f6684846d7c49643/repo2docker/buildpacks/base.py#L18
to the next LTS then since packages may have been dropped or undergone breaking changes this could result in binder repositories failing to build or run.

[...]

PR #1010, opened 2021-01-28
Add postPostBuildAdmin configuration file

g-braeunlich:
Closes: #1009


betatim:
A note on form: Please don't create an issue and a PR that closes it at the same time. It tends to split up discussions and creates twice as many notifications. If there is something you have code for it is better to open a PR directly and discuss there. If there is no code yet (maybe you first want to build consensus before coding) an issue is better.

I'll join the discussion in the issues as that has more details than the top comment here.

g-braeunlich:
OK. Sorry. Shall I remove the `Closes` link?

ctr26:
Would love to see this

PR #1011, opened 2021-02-11
fix dataverse regression introduced in last release

MridulS:
A regression was introduced in https://github.com/jupyterhub/repo2docker/pull/993 which broke the Dataverse build process.

This should fix https://github.com/jupyterhub/binderhub/issues/1263 (the repo builds locally now).

also should there be a test for this?

### DIFF
--- a/repo2docker/contentproviders/dataverse.py
+++ b/repo2docker/contentproviders/dataverse.py
@@ -73,7 +73,7 @@ class Dataverse(DoiProvider):
                 parsed_url._replace(path="/api/search", query=search_query)
             )
             self.log.debug("Querying Dataverse: " + search_url)
-            data = self.urlopen(search_url).json()
+            data = self.urlopen(search_url).json()["data"]
             if data["count_in_response"] != 1:
                 self.log.debug(
                     "Dataverse search query failed!\n - doi: {}\n - url: {}\n - resp: {}\n".format(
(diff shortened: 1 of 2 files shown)


minrk:
I believe this is tested, but the response is mocked and the same mistake was introduced to the test as well,. so the code matches the mock, but not the real thing. The `'data'` that was [here](https://github.com/jupyterhub/repo2docker/pull/993/files#diff-d9e0e0e09a494af3bf644359610a37465e54ffd4e74fb41dd1600655a20d52b8L47) is missing [here](https://github.com/jupyterhub/repo2docker/pull/993/files#diff-d9e0e0e09a494af3bf644359610a37465e54ffd4e74fb41dd1600655a20d52b8R63). I think if you update the mocks with the same patch, this should be passing agiain.

This is the trouble with mocks in tests! They only test that your code works with your mocks, not whether your mocks are accurate :/

sgibson91:
> This is the trouble with mocks in tests! They only test that your code works with your mocks, not whether your mocks are accurate :/

I literally suffered a whole new wave of imposter syndrome when I first started learning about mocking!

We have CI repos on GitHub and GitLab, could we have one on Dataverse too or is it a "host your own server" thing?

MridulS:
Ah yeah, thanks for the catch!

The test_env seems like an unrelated test fail.

pdurbin:
First of all, thanks for fixing this!! 🎉 

Does repo2docker need real DOIs? If not, you're welcome to use https://demo.dataverse.org for testing. 😄

minrk:
Merging since this fixes the issue. The env test is unrelated, we'll need to figure out if it was just a flaky run or will keep failing.

PR #1016, opened 2021-02-17
Refine buffered output debugging

minrk:
Shouldn't change much, but a refinement of #1008

- use `since` when fetching late logs to reduce duplicated output (does not eliminate it due to required integer-second rounding of timestamps; requires UTC timestamp parsing not in the Python stdlib. :sigh:)
- demote late logs to debug-level, since they may not include any new output at all, and always include redundant output
- use capfd to capture output in test_env for easier debugging

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -19,7 +19,6 @@ import time
 from urllib.parse import urlparse
 
 import escapism
-from iso8601 import parse_date
 from pythonjsonlogger import jsonlogger
 
 from traitlets import Any, Dict, Int, List, Unicode, Bool, default
@@ -650,13 +649,6 @@ class Repo2Docker(Application):
                 "Container finished running.\n".upper(), extra=dict(phase="running")
             )
             # are there more logs? Let's send them back too
-            if last_timestamp:
-                # docker only accepts integer timestamps
-                # this means we will usually replay logs from the last second
-                # of the container
-                # we should check if this ever returns anything new,
-                # since we know it ~always returns something redundant
-                last_timestamp = int(parse_date(last_timestamp).timestamp())
             late_logs = container.logs(since=last_timestamp).decode("utf-8")
             for line in late_logs.split("\n"):
                 self.log.debug(line + "\n", extra=dict(phase="running"))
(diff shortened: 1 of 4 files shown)


yuvipanda:
@minrk happy to merge if you resolve the merge conflicts

minrk:
Rebased

yuvipanda:
@minrk the test failures seem related...

repo2docker/app.py
```diff
@@ -629,9 +630,12 @@ def wait_for_container(self, container):
         Displaying logs while it's running
         """
 
+        last_timestamp = None
         try:
-            for line in container.logs(stream=True):
-                self.log.info(line.decode("utf-8"), extra=dict(phase="running"))
+            for line in container.logs(stream=True, timestamps=True):
```
manics, line 635: This will need an addition to
https://github.com/jupyterhub/repo2docker/blob/eb7520901913d61e7745d136ce127142d5fa8506/repo2docker/engine.py#L24-L36
https://github.com/jupyterhub/repo2docker/blob/eb7520901913d61e7745d136ce127142d5fa8506/repo2docker/docker.py#L16-L17

and a spec of what it should return. Logging was one of the big problems I had in https://github.com/jupyterhub/repo2docker/pull/848, see the comment about `Build logs` in the PR description.

minrk: Thanks! Exposed `timestamps` and `since` which are required to avoid duplication of output.

minrk: Docker `logs(since)` has a quirk that it only seems to accept integer timestamps. Since we now have a wrapper layer, what do you think we should accept here? A datetime object? An integer timestamp like docker itself? A timestamp string as prefixed when `timestamps=True`?

manics: Not sure.... They all seem reasonable, so go with whatever you want.

minrk: Went with symmetry, so our `since` accepts timestamps exactly as they appear, and moved the cast-to-int quirk inside the Docker implementation

manics (approved):

PR #1017, opened 2021-02-18
freeze with mamba, add 3.9

minrk:
- use mamba to install in freeze.py to save time
- add 3.9 to frozen environment list
- bump notebook, nbconvert, jupyterhub-singleuser to current

### DIFF
--- a/repo2docker/buildpacks/conda/environment.frozen.yml
+++ b/repo2docker/buildpacks/conda/environment.frozen.yml
@@ -1,5 +1,5 @@
 # AUTO GENERATED FROM environment.py-3.7.yml, DO NOT MANUALLY MODIFY
-# Frozen on 2021-02-22 10:03:32 UTC
+# Frozen on 2021-02-22 12:26:46 UTC
 name: r2d
 channels:
   - conda-forge
@@ -21,7 +21,7 @@ dependencies:
   - blinker=1.4=py_1
   - brotlipy=0.7.0=py37hb5d75c8_1001
   - c-ares=1.17.1=h36c2ea0_0
-  - ca-certificates=2021.1.19=h06a4308_0
+  - ca-certificates=2020.12.5=ha878542_0
   - certifi=2020.12.5=py37h89c1867_1
   - certipy=0.1.3=py_0
   - cffi=1.14.5=py37hc58025e_0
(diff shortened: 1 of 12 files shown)


repo2docker/buildpacks/conda/environment.frozen.yml
```diff
@@ -101,31 +104,30 @@ dependencies:
   - python-json-logger=2.0.1=pyh9f0ad1d_0
   - python_abi=3.7=1_cp37m
   - pytz=2021.1=pyhd8ed1ab_0
-  - pyzmq=22.0.3=py37h499b945_0
-  - readline=8.0=he28a2e2_2
+  - pyzmq=20.0.0=py37h499b945_1
```
manics, line 107: A downgrade is unusual, is this expected?

repo2docker/buildpacks/conda/environment.py-2.7.frozen.yml
```diff
@@ -8,53 +8,50 @@ channels:
...
   - jupyter_core=4.6.3=py27h8c360ce_1
-  - ld_impl_linux-64=2.35.1=hea4e1c9_2
-  - libffi=3.2.1=he1b5a44_1007
+  - libffi=3.3=h58526e2_2
   - libgcc-ng=9.3.0=h2828fa1_18
   - libgomp=9.3.0=h2828fa1_18
   - libsodium=1.0.17=h516909a_0
   - libstdcxx-ng=9.3.0=h6de172a_18
   - ncurses=6.2=h58526e2_4
-  - openssl=1.1.1i=h7f98852_0
```
manics, line 33: Any idea why `openssl` is removed?

manics (commented):
The switch to mamba looks good! I had a couple of inline queries about some of the dependency changes.

repo2docker/buildpacks/conda/environment.frozen.yml
minrk: This would be due to pinning added in https://github.com/conda-forge/ipykernel-feedstock/pull/74 to block a failing pyzmq release, and should be removed if I refreeze after https://github.com/conda-forge/ipykernel-feedstock/pull/75 . Let's find out...

repo2docker/buildpacks/conda/environment.py-2.7.frozen.yml
minrk: This was a symptom of a real issue! Mamba defaults to prioritizing latest version over channel, which meant that this env was getting a lot of packages from the defaults channel (which is still updated, unlike conda-forge py27 for the most part, so it has more recent versions packaged for py2.7), whereas conda has switched strict channel priority to the default.

Setting strict channel priority explicitly results in matching behavior for mamba and conda-forge, and now openssl is installed again.

I don't know if we want to add validation to avoid this in the future? We specifically don't export envs with the channel name in them because that would prevent our `conda-forge/label/broken` strategy from working (i.e. if the same package is marked as broken, what we do now will still work, whereas specifying channel name in the frozen env would cause a package being marked as broken to prevent all future installations).

manics (approved):
LGTM. I think it'd be good to have some sanity checks in the script to check for package removals or downgrades, but that can be discussed separately.

PR #1018, opened 2021-02-18
bump docker action version

### LINKED ISSUE #1005: Docker Action has a warning about a deprecated GHA config
Sounds like we need to update something in github actions:

image

minrk:
<del>No relevant [changes](https://github.com/docker/build-push-action/blob/v2/UPGRADE.md) for us, I think</del> had to set tags, context

closes #1005


minrk:
oop, that was wrong.

manics (approved):

PR #1019, opened 2021-02-18
buildpacks.r: dont use apt-key directly to respect *_proxy env vars

g-braeunlich:
Another patch to improve working behind a proxy.
This time the problem is, that gpg does not always respect the *_proxy env vars.


minrk:
Thanks for tracking all these down!

PR #1024, opened 2021-03-03
Refreeze with conda-lock

### LINKED ISSUE #895: Replace `freeze.py` with `conda-lock`
### Proposed change

We use [freeze.py](https://github.com/jupyter/repo2docker/blob/master/repo2docker/buildpacks/conda/freeze.py) to generate fully specified `environment.yml` files out of our `environment.yml` files. It creates a docker image, installs everything, and then lists out the installed packages.

Instead, we could use [conda-lock](https://pypi.org/project/conda-lock/). This resolves `environment.yml` files to fully specified lock files we can use. It doesn't install the packages either, so should be much faster.

### Alternative options

Do nothing! Our current solution works well.

Developers of repo2docker who modify `environment.yml` files. Currently the freezing process can take a while - this would make it faster.

We need to investigate if conda-lock will actually do what we want. After that, implementing it shouldn't be *too* difficult - a couple hours of work for someone very familiar with the codebase, maybe 8-12h for someone who isn't.

### Who can do this work?

1. Decent python skills
2. A vague understanding of how conda's environment.yml files work

minrk:
closes #895

Positive sides:

- Freeze is way faster (90 seconds to freeze all 5 envs vs...too long, I don't want to wait for that again)
- Install should be faster as well, due to the explicit install (will measure a bit later) which means no solve step

Possible downsides and caveats:
- We cannot have any pip-installed dependencies (or rather, we have to explicitly install them with pip ourselves) in the base env. We don't currently, but we have in the past. not insurmountable, but we would have to make this an extra explicit stage, rather than relying on conda to call pip (https://github.com/conda/conda/issues/8372)
- Our conda-forge/label/broken strategy is not possible, but it is *also* obsolete due to [changes](https://github.com/jupyterhub/repo2docker/issues/895#issuecomment-780766182) in how conda-forge deals with broken packages

Todo:

- [x] update freeze.py to use conda-lock
- [x] refreeze
- [x] measure installation time
- [x] compare packages since the changed format makes it hard to see what packages may have changed. I'll do this with `conda list --explicit` for each of the frozen envs from d24d7e4645eca7b26d4d8487c8057ead7e7a6365
- [ ] Decide on Python 3.5 case, where some packages are coming from pip, which means either two-stage install or preserve support for `environment.yml`. This PR currently supports a separate requirements.txt for pip-installed packages after the fact, which seems the simplest to me.


minrk:
I found the best way to view the diff is to make fake [branches to compare](https://github.com/minrk/repo2docker/compare/before-explicit...after-explicit). This caught a few things I needed to fix, but the results look right now.

minrk:
Apparently the explicit install doesn't take any less time. I guess the solve is already fast when a complete env is already specified.

minrk:
I think the main thing left to decide here is what to do about Python 3.5 / pip installs in the base environment. Conda lockfiles do not support pip-dependencies, but `environment.yml` files do.

Because we have no pip-installed packages in the base env anymore (a good thing), it was relatively simple to switch from `mamba env create -f environment.yml` to `mamba create -f environment.lock`. *However*, since Python 3.5 was frozen a long time ago and doesn't get updates anymore, it's still an environment.yml with pip dependencies. That means it cannot be replaced with a single lockfile.

There are two options for pip-installs:

- leave it as an environment.yml (simpler, no change, but requires swtiches for calling different install commands for the two input file formats based on the file extension)
- make pip-install an explicit separate step with an explicit, separate requirements.txt

Neither of these are perfect. The main downside of requiring environment.yml for pip is that adding a single pip dependency to the base env means discarding the lockfile and going back to environment.yml, which `freeze.py` no longer produces. It also doesn't let us control the pip install command (e.g. `--no-cache` or `--no-deps`).

I think explicit pip-install with a separate file following unconditional lockfile install is much better *at install time*. The snag is how to maintain the frozen pip requirements files over time, i.e. given direct `requirements.in` and `conda.lock` (or original environment.yml), how do we calculate the frozen `requirements.txt` that won't mess with conda-installed versions? `pip-compile` *may* work *some* of the time, but only generally when conda is getting the latest of everything. but if e.g. conda packages delay updating, this will get multiple versions of the same package. Since pip-installed packages should be minimal, the only feasible option might be to manually maintain frozen requirements.txt. This will quickly be a pain if there are nontrivial dependencies, though.

I've updated this PR to handle this separate pip stage for Python 3.5 in a way that should work if we ever want to add pip-only dependencies to the other base envs, so folks can see what that looks like and review. This means that every version always has a conda lockfile. The previous commit has the version supporting lock *or* environment.yml, but not explicit pip.

repo2docker/buildpacks/conda/__init__.py
```diff
@@ -81,7 +109,7 @@ def get_build_scripts(self):
                 r"""
                 TIMEFORMAT='time: %3R' \
                 bash -c 'time /tmp/install-miniforge.bash' && \
-                rm /tmp/install-miniforge.bash /tmp/environment.yml
+                rm -rf /tmp/install-miniforge.bash /tmp/env
```
minrk, line 112: Question: is it worth removing the environment spec? These are quite small files and might be nice to have as remnants in the images for debugging.

repo2docker/buildpacks/conda/requirements.py-3.5.txt
```diff
@@ -0,0 +1,17 @@
+# pip-installed packages for Python 3.5
+# these should only be packages
+# whose versions are too new to have builds available for Python 3.5
+alembic==1.0.3
```
minrk, line 4: This is exactly the `pip:...` lines from the previous environment.py-3.5.frozen.yml

repo2docker/buildpacks/conda/freeze.py
```diff
@@ -19,30 +19,22 @@
...
 
 ENV_FILE_T = HERE / "environment.py-{py}.yml"
-FROZEN_FILE_T = os.path.splitext(ENV_FILE_T)[0] + ".frozen.yml"
 
 yaml = YAML(typ="rt")
 
 
-def freeze(env_file, frozen_file):
-    """Freeze a conda environment.yml
+def freeze(env_file, frozen_file, platform="linux-64"):
```
minrk, line 32: This switch to conda-lock removes a significant barrier to supporting arm via setting this to `platform=linux-aarch64`

repo2docker/buildpacks/conda/freeze.py
```diff
@@ -2,7 +2,7 @@
 """
 Freeze the conda environment.yml
 
-It runs the freeze in a continuumio/miniconda3 image to ensure portability
+Using conda-lock
```
manics, line 5: Is `conda-lock` the only dependency? The command `conda-lock --mamba` implies mamba may be needed, though it ran fine for me without.

repo2docker/buildpacks/conda/freeze.py
```diff
@@ -58,43 +50,32 @@ def freeze(env_file, frozen_file):
...
+    # FIXME: conda-lock 0.8 requires {platform} in template
+    # https://github.com/conda-incubator/conda-lock/pull/78
+    frozen_template = str(frozen_dest) + ".{platform}"
+    frozen_tempfile = pathlib.Path(frozen_template.format(platform=platform))
+
+    check_call(
+        [
+            "conda-lock",
+            # FIXME: adopt micromamba after ordering is fixed
+            # https://github.com/conda-incubator/conda-lock/issues/79
```
manics, line 62: https://github.com/conda-incubator/conda-lock/issues/79 is fixed, though to make review easier this could be done in a follow-up PR so it's clear if there are other changes

repo2docker/buildpacks/conda/freeze.py
```diff
@@ -58,43 +50,32 @@ def freeze(env_file, frozen_file):
                 return
     print(f"Freezing {env_file} -> {frozen_file}")
 
+    # FIXME: conda-lock 0.8 requires {platform} in template
+    # https://github.com/conda-incubator/conda-lock/pull/78
```
manics, line 54: https://github.com/conda-incubator/conda-lock/pull/78 is merged but this can be updated in a follow-up PR.

manics (commented):
Nice work, this is a definite improvement over the docker freeze script.

repo2docker/buildpacks/conda/freeze.py
minrk: conda-lock can install mamba itself if it's not present

repo2docker/buildpacks/conda/__init__.py
minrk: Good call. Let's clean them up for now since they are in tmp, and we can open a separate PR to persist them to a non-tmp location if we decide that makes senes.

repo2docker/buildpacks/conda/__init__.py
```diff
@@ -23,19 +23,47 @@ class CondaBuildPack(BaseImage):
...
     def get_build_env(self):
         """Return environment variables to be set.
 
         We set `CONDA_DIR` to the conda install directory and
         the `NB_PYTHON_PREFIX` to the location of the jupyter binary.
 
         """
+        if not self._nb_environment_file:
+            # get_build_scripts locates requirements/environment files
+            self.get_build_scripts()
```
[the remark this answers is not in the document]
minrk: You're absolutely right, it should be `get_build_script_files()`. If it's passing, I think that probably means `get_build_script_files()` is reliably being called before `get_build_env()` and this branch is never needed. I've fixed it with a more detailed comment.

manics (approved):
Awesome!
@betatim any comments, otherwise I'll merge.

manics:
Merging, thanks!

PR #1027, opened 2021-03-23
Fix logo URL in README

betatim:
For some reason the URL we use to get the logo has stopped working. This updates the URL to a recent commit.

Out of curiosity it would be cool to understand why this broke.


manics (approved):
The original URL was set in https://github.com/jupyterhub/repo2docker/commit/0c70513f1d5f36592539d95314bdafa2e2bedf0a but looking at the PR https://github.com/jupyterhub/repo2docker/pull/463 the branch was subsequently force pushed and commit 3fa7444fca6ae2b51e590cbc9d83baf92738ca2a no longer exists. My guess is the commit was replaced by a force push, and the old commit has been garbage collected.

PR #1034, opened 2021-04-01
Fix regression in hydroshare introduced after moving to requests

MridulS:
The build process was broken by https://github.com/jupyterhub/repo2docker/pull/993, binderhub wasn't able to build hydroshare repos.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,7 +110,8 @@ jobs:
           pip freeze
           # hg-evolve pinned to 9.2 because hg-evolve dropped support for
           # hg 4.5, installed with apt in Ubuntu 18.04
-          $(hg debuginstall --template "{pythonexe}") -m pip install hg-evolve==9.2 --user
+          pip install mercurial
+          $(hg debuginstall --template "{pythonexe}") -m pip install hg-evolve --user
 
       - name: "Run tests"
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,8 +110,8 @@ jobs:
           pip freeze
           # hg-evolve pinned to 9.2 because hg-evolve dropped support for
           # hg 4.5, installed with apt in Ubuntu 18.04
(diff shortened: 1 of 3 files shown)


MridulS:
`pip install hg-evolve` seems to be breaking on GitHub Actions https://github.com/jupyterhub/repo2docker/runs/2246541424

I have updated the action to use latest `mercurial` and `hg-evolve`, any reason why repo2docker isn't using the latest release of `mercurial` available via `pip`?

manics:
> I have updated the action to use latest mercurial and hg-evolve, any reason why repo2docker isn't using the latest release of mercurial available via pip?

@paugier added the mercurial provider in https://github.com/jupyterhub/repo2docker/pull/950

paugier:
> any reason why repo2docker isn't using the latest release of mercurial available via pip?

Because there are no Mercurial wheels for Linux on PyPI: https://pypi.org/project/mercurial/#files. So installing Mercurial via pip triggers a Mercurial build and we don't want to do that here.

I don't understand what has changed and why installing the same version of the same package on the same Python is now broken. Anyway, maybe another quick way to fix this error ("ImportError: No module named setuptools") would be to install setuptools for the Python used by Mercurial by adding:

`$(hg debuginstall --template "{pythonexe}") -m pip install setuptools --user`

before installing hg-evolve 9.2.

paugier:
Or one could use Ubuntu 20.04 ?

MridulS:
Thanks @paugier!

Updated the CI to install setup tools for mercurial.
Actions are happy now.

sblack-usu:
I'm from HydroShare and just checking in to see if any help is needed to get this merged.  I'm also curious to see if there is anything we could do to catch this earlier.  It looks like it wasn't caught here because requests are mocked and it was caught in the binderhub repoproviders test after binderhub picked up the latest repo2docker.

betatim:
LGTM.

I think your diagnosis is right about why it broke and no one noticed.

Issue #1026, opened 2021-03-22
Default to JupyterLab single document mode (/doc)


manics:
### Proposed change
Switch the default repo2docker interface from jupyter-notebook to JupyterLab's single document mode.

I'd like to gauge everyone's feelings on if/when we'd make the switch.

### Background
The trigger for this comes from the decision to switch Z2JH to JupyterLab: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/776#issuecomment-797276517. I asked whether we could change Z2JH to not set `singleuser.cmd` and instead default to whatever the default entrypoint/command is in the Docker image and instead change the default in the base docker-stacks images https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/776#issuecomment-804063583, an advantage being we would no longer ignore the default start-up scripts that lead to problems such as https://discourse.jupyter.org/t/pyspark-library-is-missing-from-jupyter-pyspark-notebook-when-running-with-jupyterhub-zero-to-jupyterhub-k8s/8450/2, and also backwards compatibility would hopefully be easier to handle.

@parente pointed me to an existing docker-stacks discussion on switching to JupyterLab https://github.com/jupyter/docker-stacks/issues/1217 where one of the suggestions is to wait for other projects such as Z2JH and BinderHub (i.e. repo2docker) to switch. What does everyone think? If you can envisage switching R2D in the next few months then it might make sense to hold off on the Z2JH switch, and co-ordinate with docker-stacks instead to switch everything in one go. On the other hand if it's longer then we should probably figure out how to switch Z2JH and ensure [older images continue to work](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/776#issuecomment-797276517).

SylvainCorlay:
Thanks! +1000 on this! This is long overdue, as lab is where most of the team has been putting work (and love) for so long. This is also where all the current efforts are going with respect to accessibility, internationalization, collaborative editing, and the 3.x series of lab resolves the main pain points for extension authors.

minrk:
I think we should make this switch now. Summarizing the separate transitions we can make [from discussion in the meeting](https://github.com/jupyterhub/team-compass/blob/master/docs/monthly-meeting/2021-02-18.md#agenda-items) and [pr](https://github.com/jupyterhub/repo2docker/pull/996):

1. default UI to lab in repo2docker (NotebookApp.default_url)
[...]

PR #1035, opened 2021-04-09
Default UI to JupyterLab

SylvainCorlay:
Cf #1026

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -582,6 +582,7 @@ class Repo2Docker(Application):
                 "--port",
                 port,
                 "--NotebookApp.custom_display_url=http://{}:{}".format(host_name, port),
+                "--NotebookApp.default_url=/lab",
             ]
             ports = {"%s/tcp" % port: port}
         else:


manics:
I haven't tried this locally yet, but if I go to `/doc` in mybinder I see just the launch tab, the file browser is closed by default:
image

Can we have it expanded by default? Or would it be better to use `/lab` which automatically displays the file browser?

SylvainCorlay:
We could use `lab` when no filepath is provided and `doc` otherwise? I understand that the demand was to use `doc` (the simple mode) when possible.

minrk:
When I visited `/doc` it opened a notebook, but maybe that was leftover state in my browser. Yeah, if that's the case, maybe let's use `/lab` here and keep `/doc` for filepath in binderhub.

manics:
@SylvainCorlay Sorry for the confusion. I suggested the single document mode in https://github.com/jupyterhub/repo2docker/issues/1026 because I thought it'd be similar to the old notebook: show a single document when you're open a file, show a list of files if not, but that's obviously not how it behaves.

If we're happy with `/lab` are there any benefits to setting `default_url`, or could we change the command to
```py
run_cmd = [
                "jupyter",
                "lab",
...
```
(or `jupyterlab`) instead?

minrk:
> are there any benefits to setting default_url, 

Setting default_url is switching the default *ui* without switching the default *server* implementation to jupyter_server, which starting `jupyter lab` would do. See [this comment](https://github.com/jupyterhub/repo2docker/issues/1026#issuecomment-816454222) for details. It's still something we should do, but has a larger impact on things like extensions.

SylvainCorlay:
Just following up on this. Is there anything blocking this PR ?

minrk (approved):

SylvainCorlay:
:tada:

PR #1038, opened 2021-04-29
replace mamba & conda with micromamba

madhur-tandon:

### DIFF
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -34,6 +34,7 @@ class CondaBuildPack(BaseImage):
             ("CONDA_DIR", "${APP_BASE}/conda"),
             ("NB_PYTHON_PREFIX", "${CONDA_DIR}/envs/notebook"),
             ("MAMBA_ROOT_PREFIX", "${CONDA_DIR}"),
+            ("MAMBA_EXE", "/tmp/bin/micromamba"),
         ]
         if self.py2:
             env.append(("KERNEL_PYTHON_PREFIX", "${CONDA_DIR}/envs/kernel"))
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -33,6 +33,7 @@ class CondaBuildPack(BaseImage):
         env = super().get_build_env() + [
             ("CONDA_DIR", "${APP_BASE}/conda"),
             ("NB_PYTHON_PREFIX", "${CONDA_DIR}/envs/notebook"),
+            ("MAMBA_ROOT_PREFIX", "${CONDA_DIR}"),
         ]
         if self.py2:
             env.append(("KERNEL_PYTHON_PREFIX", "${CONDA_DIR}/envs/kernel"))
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -279,7 +279,7 @@ class CondaBuildPack(BaseImage):
                     "${NB_USER}",
                     r"""
                 TIMEFORMAT='time: %3R' \
-                bash -c 'time /tmp/bin/micromamba update -p {0} -f "{1}" && \
+                bash -c 'time /tmp/bin/micromamba install -p {0} -f "{1}" && \
                 # time /tmp/bin/micromamba clean --all -y && \
                 /tmp/bin/micromamba list -p {0} \
                 '
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -279,7 +279,7 @@ class CondaBuildPack(BaseImage):
                     "${NB_USER}",
                     r"""
                 TIMEFORMAT='time: %3R' \
-                bash -c 'time /tmp/bin/micromamba env update -p {0} -f "{1}" && \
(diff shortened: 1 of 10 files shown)


minrk:
Awesome, thanks for the PR!

I think it might be good to land #1024 first, since I think it touches several of the same bits.

Things to think about:

- what is r2d's commitment to having `conda` and/or `mamba` available in the base env? Should we have alias wrappers for conda/mamba->micromamba? How many things would we expect to work and not if we did that?
- should we be bootstrapping an env with conda installed (i.e. add conda itself to environment.yml, as I do [over here](https://github.com/minrk/simula-summer-school/blob/00a9f724230c68a45b497f3219d1a545a6850c80/image/environment.yml#L8)?

One nice thing about using micromamba, I believe, is that we can bootstrap the root env itself. This can include conda/mamba or not. So we don't need two envs - root + notebook, we can go back to just one (root) with an optional extra env for the kernel.

minrk:
How well defined is the micromamba functionality subset? I see https://mamba.readthedocs.io/en/latest/micromamba.html#available-commands but is every option on every one of those commands supported?

wolfv:
Hi @minrk we're gearing up `micromamba` for interactive use, and a lot of the existing 'conda / mamba' functionality is supported. Also, we often follow conda for the CLI flags, with one major exception for the `conda env create` command that we always found confusing (and it's actually a historical artefact in the conda codebase). The documentation page is also slightly outdated: we recently added support for `update` and `remove`, as well as reading `.condarc` / `.mambarc` files.

Running the repo2docker tests has actually forced us to add a couple extra pieces that were previously missing:

- `micromamba env list` (although we're undecided if we should move that to `micromamba list --envs` or something similar
- added the ability to specify `pip` packages in environment files (however, we haven't yet implemented Conda's fallback behavior when `pip` isn't explicitly mentioned in the environment.yml file)
- added the ability to add channels from specs (e.g. `micromamba install bioconda::mypkg` will automatically "add" `-c bioconda`

I have been exclusively using micromamba interactively on my laptop for a while now and it seems to hold up :) We'll soon find the time to add more details to our docs and write a blogpost detailing some of the recent improvements.

minrk:
Great, thanks!

> added the ability to specify pip packages in environment files

FWIW, #1024 removes repo2docker's reliance on conda calling pip, but I think it is still probably good to match the spec.

SylvainCorlay:
Something that may be interesting in that regard is madhur's PR to grayskull automatically turning pip wheels into conda packages. This could be another way to handle that in the future while avoiding version inconsistencies between pip and conda.

minrk:
Oop, I forgot that *users'*  env files could depend on pip dependencies, so that's definitely a requirement for use with r2d.

wolfv:
yeah sure :) I mean, users _can_ fix it by adding `- pip` in the line above. I saw that at some point, conda also printed a warning message about that.

But we'll follow conda's behavior here and automagically add `pip` to the deps.

minrk:
Right. Pip ought to be explicitly in our base env, at least, which is created before running the `env update` with the user's environment file.

repo2docker/buildpacks/conda/__init__.py
```diff
@@ -63,7 +63,7 @@ def get_build_scripts(self):
 
         All scripts here should be independent of contents of the repository.
 
-        This sets up through `install-miniforge.bash` (found in this directory):
+        This sets up through `install-micromamba.bash` (found in this directory):
```
minrk, line 66: We've done this before, and should have fixed it then - it doesn't really make sense to keep renaming this file as we change some implementation details inside it. Maybe change it to `install-conda.bash` or `install-base-env.bash`? Whether we use conda, miniconda, micromamba, etc. should not mean renaming this file anymore.

madhur-tandon: Done!

wolfv:
Hmm, I think there is actually an additional issue with pypy coming up now. We need to fix the de-prioritization with track features

repo2docker/buildpacks/conda/__init__.py
```diff
@@ -279,7 +279,7 @@ def get_env_scripts(self):
                     "${NB_USER}",
                     r"""
                 TIMEFORMAT='time: %3R' \
-                bash -c 'time /tmp/bin/micromamba install -p {0} -f "{1}" && \
+                bash -c 'time /tmp/bin/micromamba env update -p {0} -f "{1}" && \
```
wolfv, line 282: I think we only have the `update` command (not `env update`).

madhur-tandon: Yeah, I tried both, both of them fail locally for me.
The only thing that works is using `install` along with specifying `- python` in the `environment.yml` file

madhur-tandon: I have still pushed the changes to see if it works with the CI, though it fails locally for me

wolfv: I am pushing a build with my changes for libsolv. Let's see if that will fix the issues here!

wolfv:
In the Python 3.5 test the selected `libstdcxx-ng==7.2.0=hdf63c60_3` is only available from the defaults channel. It's marked broken on conda-forge:

Screenshot 2021-06-02 at 09 42 23

So it's correct that micromamba chooses the package from the defaults channel.

However, this test seems to indicate some issue with the new-style `.conda` packages that I'll look into now.

SylvainCorlay:
Closing in favor of #1062 which has been rebased.

PR #1043, opened 2021-05-27
always unpack a single zenodo zip

### LINKED ISSUE #1042: Always unpack single zip files from Zenodo
### Proposed change

Uploading a .zip is the only way to represent folders on zenodo. It is also what zenodo does on github imports.

[Right now](https://github.com/jupyterhub/repo2docker/blob/master/repo2docker/contentproviders/zenodo.py#L71) repo2docker will only unpack the zip if DOI is marked as being software. I propose to unpack a single file unconditionally instead. The motivation is as follows:

- Some uploads containing both software and data (e.g. processing scripts) may classify themselves as datasets.
- If the user is facing a binder with a single zip file, their likely first step is to unzip said file.

The only downside I can think of is hitting disk space limits, however the workaround is trivial: the repository owner then needs to upload more than a single zip file. At the same time, currently the repository owner needs to choose any 2 out of the following 3 options:

- Having folders
- Having a functional binder
- Not labeling the repo as software

### Alternative options

Keep everything as it is...

Zenodo users

No effort (editing 2 lines), I'm about to open a PR.

### Who can do this work?

I can!

akhmerov:
Closes #1042 (see discussion there).

### DIFF
--- a/repo2docker/contentproviders/zenodo.py
+++ b/repo2docker/contentproviders/zenodo.py
@@ -63,12 +63,11 @@ class Zenodo(DoiProvider):
 
         record = resp.json()
 
-        is_software = deep_get(record, host["type"]).lower() == "software"
         files = deep_get(record, host["filepath"])
         only_one_file = len(files) == 1
         for file_ref in files:
             for line in self.fetch_file(
-                file_ref, host, output_dir, is_software and only_one_file
+                file_ref, host, output_dir, unzip=only_one_file
             ):
                 yield line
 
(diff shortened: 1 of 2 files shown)


akhmerov:
Docs failure is unrelated :(

yuvipanda:
If you add a test pointing to an appropriate zenodo DOI that will be extracted, am happy to merge this!

If you rebase, the docs failure should be gone now

akhmerov:
Please take a look, I've rebased and adjusted an existing test.

yuvipanda:
\o/ thanks a lot, @akhmerov! I'll try deploy this later today

yuvipanda:
@akhmerov deployed now! Can you test it out?

akhmerov:
Yep, works just as expected! Thanks!

yuvipanda:
Thanks a lot, @akhmerov!

PR #1048, opened 2021-06-21
Add Guix buildpack

hulecomte:
Hi!  This patch adds a backend that uses the GNU Guix package manager.  Guix provides reproducible software environments, transparency and provenance tracking.  Coupled with a channels.scm file, it allows users to pin to a specific Guix revision from which the set of packages listed in their manifest.scm is deployed, thereby supporting very fine-grain reproducibility.

Let me know what you think

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -37,6 +37,7 @@ from .buildpacks import (
     PipfileBuildPack,
     PythonBuildPack,
     RBuildPack,
+    GuixBuildPack,
 )
 from . import contentproviders
 from .utils import ByteSpecification, chdir
@@ -95,6 +96,7 @@ class Repo2Docker(Application):
             JuliaProjectTomlBuildPack,
             JuliaRequireBuildPack,
             NixBuildPack,
+            GuixBuildPack,
             RBuildPack,
             CondaBuildPack,
             PipfileBuildPack,
--- a/repo2docker/buildpacks/guix/__init__.py
+++ b/repo2docker/buildpacks/guix/__init__.py
@@ -1,0 +1,83 @@
+"""BuildPack for guix environments"""
+import os
+
+from ..base import BuildPack, BaseImage
+
+
+class GuixBuildPack(BaseImage):
+    """A Guix Package Manager BuildPack"""
+
+    def get_build_env(self):
+        """"""
+        env = super().get_build_env() + [
+            ("GUIX_DIR", "${APP_BASE}/guix")
+        ]
+        return env
+
+    def get_path(self):
+        """Return paths to be added to PATH environment variable"""
+        path = super().get_path()
+        path.insert(0, "${GUIX_DIR}/bin")
+        return path
+
+
+
+    def get_build_scripts(self):
+        """
+        Install a pinned version of Guix package manager,
+        precised in guix-install.bash.
+        """
+        return super().get_build_scripts() + [
+            (
+                "root",
+                """
+                bash /tmp/.local/bin/guix-install.bash
+                """,
+            ),
+
+        ]
+
+    def get_build_script_files(self):
+
+        """Copying guix installation script on the image"""
+        return {
+            "guix/guix-install.bash":
+                "/tmp/.local/bin/guix-install.bash",
+        }
+
+    def get_assemble_scripts(self):
+        """
+        Launch Guix daemon with --disable-chroot to avoid the need
+        of root privileges for the user.
+        Make sure we never use Debian's python by error by renaming it
+        then, as an user install packages listed in manifest.scm,
+        use guix time-machine if channels.scm file exists.
+        Finally set Guix environment variables.
+        """
+        assemble_script ="""
+                 /var/guix/profiles/per-user/root/current-guix/bin/guix-daemon \
+                 --build-users-group=guixbuild --disable-chroot & \
+                 su - $NB_USER -c '{}' && \
+                 echo 'GUIX_PROFILE="/var/guix/profiles/per-user/'$NB_USER'/guix-profile" ; \
+                 source "$GUIX_PROFILE/etc/profile"'>> /etc/profile.d/99-guix.sh
+                 """
+
+        if os.path.exists(self.binder_path("channels.scm")):
+            assemble_script = assemble_script.format(
+                "guix time-machine -C  " + self.binder_path("channels.scm") +
+                " -- package -m "  + self.binder_path("manifest.scm")
+            )
+        else:
+            assemble_script = assemble_script.format(
+                "guix package -m " + self.binder_path("manifest.scm")
+            )
+        return super().get_assemble_scripts() + [
+            ( "root",
+              assemble_script,
+             )
+        ]
+
+    def detect(self):
+        """Check if current repo should be built with the guix BuildPack"""
+        return os.path.exists(self.binder_path("manifest.scm"))
+    
--- a/repo2docker/buildpacks/guix/guix-install.bash
+++ b/repo2docker/buildpacks/guix/guix-install.bash
@@ -1,0 +1,14 @@
+#!/bin/bash
+# This downloads and installs a pinned version of Guix using a pinned version of the installation script.
+# Here is the commit associated with Guix version 1.3.0 the latest version when this script was written.
+set -euxo pipefail
+
+GUIX_COMMIT="a0178d34f582b50e9bdbb0403943129ae5b560ff"
+BIN_VER="1.3.0x86_64-linux"
+GUIX_SHA256="bcdeaa757cd42d2c9de4791272737e9ee0d518398403955f113611f4a893380a"
+
+wget "https://git.savannah.gnu.org/cgit/guix.git/plain/etc/guix-install.sh?id=$GUIX_COMMIT"
+
+echo "$GUIX_SHA256  guix-install.sh?id=$GUIX_COMMIT" | sha256sum -c
+
+(yes || true) | BIN_VER=$BIN_VER bash guix-install.sh?id=$GUIX_COMMIT
--- a/docs/source/config_files.rst
+++ b/docs/source/config_files.rst
@@ -231,6 +231,27 @@ to produce a reproducible environment.
 To see an example repository visit
 `nix binder example <https://github.com/binder-examples/nix>`_.
 
+.. _manifest.scm:
+
+``manifest.scm`` - the Guix package manager
+===========================================
+
+Specify packages to be installed by the `Guix package manager <https://guix.gnu.org/>`_.
+All packages specified in |manifest|_ will be installed in a container using |guix_package|_.If you do not install Jupyter explicitly, Repo2docker will no be able to start your container. In addition, you can use different `channels <https://guix.gnu.org/manual/en/html_node/Channels.html>`_ rather
+than the ones available by default (official channels of GNU Guix 1.3.0).
+You must describe such channels in a ``channels.scm`` file which will be used
+alongside ``manifest.scm`` with the |guix_time-machine|_ command. Furthermore, using a ``channels.scm`` file lets you `pin a specific revision <https://guix.gnu.org/manual/en/html_node/Replicating-Guix.html>`_ of Guix, allowing you to unambiguously specific the software environment to reproduce.
+
+For more information about Guix please read the `manual <https://guix.gnu.org/manual/en/guix.html>`_.
+
+.. |manifest| replace:: ``manifest.scm``
+.. _manifest: https://guix.gnu.org/manual/en/html_node/Invoking-guix-package.html#index-profile-manifesthy
+
+.. |guix_package| replace:: ``guix package``
+.. _guix_package: https://guix.gnu.org/manual/en/html_node/Invoking-guix-package.html#Invoking-guix-package
+
+.. |guix_time-machine| replace:: ``guix time-machine``
+.. _guix_time-machine: https://guix.gnu.org/manual/en/html_node/Invoking-guix-time_002dmachine.html
 
 ``Dockerfile`` - Advanced environments
 ======================================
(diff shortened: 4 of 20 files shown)


repo2docker/buildpacks/guix/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+        Install guix package manager version 1.3.0.x86_64-linux, using
+        an unmodified installation script found at
+        https://git.savannah.gnu.org/cgit/guix.git/plain/etc/guix-install.sh
+        """
+        return super().get_build_scripts() + [
+            (
+                "root",
+                """
+                yes | BIN_VER=1.3.0.x86_64-linux \
+                bash /home/${NB_USER}/.local/bin/guix-install.bash
```
yuvipanda, line 26: Unfortunately this file appears to be GPL, and so can not be included in the repo. Can we fetch it at build time, just like we do for nix?

repo2docker/buildpacks/guix/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+            ),
+
+        ]
+
+    def get_build_script_files(self):
+
+        """Copying guix installation script on the image"""
+        return {
+            "guix/guix-install.bash":
+                "/home/${NB_USER}/.local/bin/guix-install.bash",
```
yuvipanda, line 37: If the built image is used with a JupyterHub (as it often is), then everything under $HOME will often instead be replaced with a persistent directory. So ideally buildpacks should leave $HOME clean. Other buildpacks seem to put their libraries under `${APP_BASE}/<something>` (see https://github.com/jupyterhub/repo2docker/blob/de7468d78f6309f6eb7fbd80f2e1758e0efc6b52/repo2docker/buildpacks/conda/__init__.py#L34).

Can we do the same here? Where does everything install to?

yuvipanda (changes requested):
Thank you so much for adding this.

We already support nix here - can you tell me just a little bit about how they relate to each other?

Also, the current maintainers of repo2docker don't have a lot of guix experience. Do you think there will be enough interest from the guix community itself to keep this up to date? 

I've left general comments but haven't managed to test it.

I look forward to working on getting this merged!

civodul:
Hi @yuvipanda.  I can reply on the Nix vs. Guix bit (I'm a Guix co-maintainer).  :-)

Guix and Nix are independent projects with different code bases.  The tools are both "functional package managers", which sets them apart from other tools out there; you'll find some of the core concepts are the same, and Guix borrowed them from Nix (Guix's build daemon is based on that of Nix).

Apart from that, the user interfaces are different (in Guix we try hard to make the CLI accessible to non-experts, to offer customization options, and to provide good integration for a variety of software deployment tools), the packages they provide are different (e.g., Guix insists on free software built from source as a way to improve transparency and security), and so on.

There's a growing [scientific computing community around Guix](https://hpc.guix.info) with a focus on high-performance computing (HPC) and reproducible science.  I'm confident there'll be interest in keeping Guix support afloat in repo2docker.

Let me know if you need more info!

yuvipanda:
@civodul Thank you for your explanations - that sounds good enough to me. Let's working on getting stuff outside of $HOME and work towards getting this deployed! \o/

If you rebase your PR, it should also fix the failing docker build and build docs as well.

repo2docker/buildpacks/guix/__init__.py
hulecomte: Done, guix-install.bash is not under GPL anymore, a pinned installation script is now fetched at build time!

repo2docker/buildpacks/guix/__init__.py
hulecomte: Done, $HOME should be clear now, let me now what you think!

hulecomte:
Thanks for your feedback @yuvipanda, what do you think now ?

repo2docker/buildpacks/guix/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+        renaming it, then, as an user install packages listed in
+        manifest.scm, use guix time-machine if channels.scm file exists
+        """
+        assemble_script ="""
+                 /var/guix/profiles/per-user/root/current-guix/bin/guix-daemon \
+                 --build-users-group=guixbuild --disable-chroot & \
+                 mv /usr/bin/python /usr/bin/python.debian && \
+                 su - $NB_USER -c '{}' && \
+                 echo 'GUIX_PROFILE="/var/guix/profiles/per-user/$NB_USER/.guix-profile" ; \
+                 source "$GUIX_PROFILE/etc/profile"'>> ~/.bash_profile
```
yuvipanda, line 54: This puts it under `$HOME` and so won't be present when used with a JupyterHub.  Maybe somewhere in `/etc/profile.d`?

repo2docker/app.py
```diff
@@ -99,6 +100,7 @@ def _default_log_level(self):
             CondaBuildPack,
             PipfileBuildPack,
             PythonBuildPack,
+            GuixBuildPack,
```
yuvipanda, line 103: I think the ordering here matters - repo2docker will go from top  to bottom, stopping at the first one. For example, by being below `CondaBuildPack`, if any `environment.yml` file is found in the git repo, `GuixBuildPack` is never called. Let's move this up, just behind the Nix one maybe?

yuvipanda (changes requested):
Thanks, @hulecomte! I've left some more comments. I haven't tested it locally yet, though.

Another question I have is - if a user sets up a repo to use guix, then they will need to make sure that jupyter is installed inside the container. the Nix buildbpack has a [nice note](https://github.com/jupyterhub/repo2docker/blob/master/docs/source/config_files.rst#defaultnix---the-nix-package-manager), can you add something similar for Guix too?

I'm not sure why, but it looks like the lint failure must be fixed first before the integration tests will run (not sure why?!). Can you take a shot at it?

Thank you for all your work! \o/

repo2docker/buildpacks/guix/guix-install.bash
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# This downloads and installs a pinned version of Guix
+set -ex
+
+wget https://git.savannah.gnu.org/cgit/guix.git/plain/etc/guix-install.sh?id=a0178d34f582b50e9bdbb0403943129ae5b560ff
```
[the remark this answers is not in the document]
hulecomte: I edited this file according to your comment, I added some more comments about why I choose this hash, I created variables at the beginning of the file making it easier to upgrade the script when needed. 
I tried to keep it simple though.
WDYT?

repo2docker/buildpacks/guix/guix-install.bash
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# This downloads and installs a pinned version of Guix
+set -ex
```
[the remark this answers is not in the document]
hulecomte: I edited the script according to that. I add to fix one of my piped command (last command with `yes`) but everything work fine here now.

repo2docker/buildpacks/guix/__init__.py
hulecomte: Done.

repo2docker/app.py
hulecomte: Right, done.

repo2docker/buildpacks/guix/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        Wake up the guix daemon with root permission, set guix environnement
+        variables, make sure we never use debian's python by error by
+        renaming it, then, as an user install packages listed in
+        manifest.scm, use guix time-machine if channels.scm file exists
+        """
+        assemble_script ="""
+                 /var/guix/profiles/per-user/root/current-guix/bin/guix-daemon \
+                 --build-users-group=guixbuild --disable-chroot & \
+                 mv /usr/bin/python /usr/bin/python.debian && \
```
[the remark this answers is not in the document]
hulecomte: I added some comments about this steps in the header of the function, let me now what you think.

  The point of using Guix is to avoid any confusions about the version or the provenance of the software you want to use and therefore guarantee and allow reproducibility of a software environment. Therefore having a python that we now nothing about could jeopardize that. Also counting on the path precedence to deal with that give us no guarantee of software environment reproducibility and could influence our results by making things work when they shouldn't for example.

While testing the Guix buildpack, I tried installing and using numpy, it take me a long time to realize that I where trying to do so but with the python already installed on the docker image, wich could not work with numpy (versions conflict). Also it can be confusing when the software use environment variables and you are not aware of all the other software present on the system.

yuvipanda (changes requested):
Thanks for addressing these, @hulecomte! I still haven't managed to test this locally, but hope this is all still useful.

To respond to https://github.com/jupyterhub/repo2docker/pull/1048#discussion_r664601423, IMO we shouldn't hand-move things installed with another package manager (in this case, apt). If you try to use apt to remove it, I'm sure it'll take a bunch of other packages with it. IMO, this means we shouldn't try to do that, since it will just silently break *other things*. We already do this with nix, conda, etc - IMO that's what we should do here too.

When the container runs and the user tries to do `pip install numpy`, it should automatically use the 'correct' python installed - in this case, the guix one. For example, in the conda buildpack (https://github.com/jupyterhub/repo2docker/blob/f07111c4a58220263562f726fb55288da57fe71f/repo2docker/buildpacks/conda/__init__.py#L54), just the `PATH` is enough to 'do the right thing'. And `/usr/bin/python3` is never used.

Let's try do the same here?

Thanks for working with me through this!

repo2docker/buildpacks/guix/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+        Launch Guix daemon with --disable-chroot to avoid the need
+        of root privileges for the user.
+        Make sure we never use Debian's python by error by renaming it
+        then, as an user install packages listed in manifest.scm,
+        use guix time-machine if channels.scm file exists.
+        Finally set guix environment variables.
+        """
+        assemble_script ="""
+                 /var/guix/profiles/per-user/root/current-guix/bin/guix-daemon \
+                 --build-users-group=guixbuild --disable-chroot & \
```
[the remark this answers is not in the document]
hulecomte: Indeed, you need the guix-daemon running to be able to use guix!

repo2docker/buildpacks/guix/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+                 mv /usr/bin/python /usr/bin/python.debian && \
+                 su - $NB_USER -c '{}' && \
```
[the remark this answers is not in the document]
hulecomte: Hey, so Im struggling a bit with this part. It seems I can't keep the guix-daemon running between distinct steps. In some cases I have some issues when the docker images is not completely rebuilt, steps by steps but reusing some steps already executed before. That's why I did that here, I launch the guix-daemon just before running the Guix command I need, therefore I'm absolutely sure guix-daemon is running. Maybe you have some clues or ideas about a better way to do so ?

repo2docker/buildpacks/guix/guix-install.bash
```diff
@@ -0,0 +1,13 @@
+# This downloads and installs a pinned version of Guix using a pinned version of the installation script.
```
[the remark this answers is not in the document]
hulecomte: Hey thanks for that, I will add it in the next rebase!

hulecomte:
> Thanks for addressing these, @hulecomte! I still haven't managed to test this locally, but hope this is all still useful.
> 
> To respond to [#1048 (comment)](https://github.com/jupyterhub/repo2docker/pull/1048#discussion_r664601423), IMO we shouldn't hand-move things installed with another package manager (in this case, apt). If you try to use apt to remove it, I'm sure it'll take a bunch of other packages with it. IMO, this means we shouldn't try to do that, since it will just silently break _other things_. We already do this with nix, conda, etc - IMO that's what we should do here too.
> 
> When the container runs and the user tries to do `pip install numpy`, it should automatically use the 'correct' python installed - in this case, the guix one. For example, in the conda buildpack (
> 
> https://github.com/jupyterhub/repo2docker/blob/f07111c4a58220263562f726fb55288da57fe71f/repo2docker/buildpacks/conda/__init__.py#L54
> ), just the `PATH` is enough to 'do the right thing'. And `/usr/bin/python3` is never used.
> 
> Let's try do the same here?
> 
> Thanks for working with me through this!

Hey yuvipanda! I'm currently trying that but it seems I still can use python while using bash script to test the buildpack...Maybe I am not doing it well enough? What do you think ?

Anyway, I don't now why lint test is failing and I don't know how it works too. How can I reproduce it locally ?

Thanks!

hulecomte:
Hey @yuvipanda here are some updates and news:
-I modified the doc and the guix-install.bash as you suggested.
-Added a PATH 'do to the right thing' as we discussed it previously.

What do you think?

repo2docker/buildpacks/guix/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+            "guix/guix-install.bash":
+                "/tmp/.local/bin/guix-install.bash",
+        }
+
+    def get_assemble_scripts(self):
+        """
+        Launch Guix daemon with --disable-chroot to avoid the need
+        of root privileges for the user.
+        Make sure we never use Debian's python by error by renaming it
+        then, as an user install packages listed in manifest.scm,
```
jgarte, line 53: @hulecomte What did you mean at this line?

```suggestion
Make sure we never use Debian's python by error by renaming it
then, as an user installs packages listed in manifest.scm,
```
Maybe, state it again and I can help with clarifying it.

civodul:
Hi @yuvipanda & all! Would be great if @hulecomte's work could get in.  :-)

Is there anything left that you think we should do?

Thanks in advance!

civodul:
Hi @yuvipanda, could you take another look at this PR?

Thanks in advance.

rekado:
I'm very excited about this addition.  Are there any other obstacles to getting this merged?

civodul:
Hi! Any update here?

Kabouik:
Bumping that because I'm interested too.

PR #1050, opened 2021-06-29
Stop using deprecated add_stylesheet in sphinx

yuvipanda:
add_stylesheet has been deprecated for a while.

This should hopefully fix the CI failures in circleci

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -47,7 +47,7 @@ from recommonmark.transform import AutoStructify
 
 
 def setup(app):
-    app.add_stylesheet("custom.css")  # may also be a URL
+    app.add_css_file("custom.css")  # may also be a URL
     app.add_config_value(
         "recommonmark_config", {"auto_toc_tree_section": "Contents"}, True
     )


yuvipanda:
A different CI step is failing, but at least CircleCI is fixed!

betatim:
This is actually a fairly annoying failure. If we merge this PR without fixing it I think we won't get a new image pushed to docker hub, but we wil lget an automated PR for mybinder.org-deploy, which if we merge it will lead to errors because the builder image doesn't exist.

It looks like the error is around ruamel.yaml for which there are no wheels that work on alpine. I have a vague memory of us having a problem like this before, but I don't remember the solution. The only thing I remember is to be wary of ruamel.yaml updates/changes :-/

betatim:
The obvious fix is to add `gcc` or "build essentials" (?) to the build stage docker image. However I have a feeling that the fact that the build doesn't just work is a sign of something else being broken. So we can fix the symptom by adding build tools but we won't be fixing the underlying issue.

At some point we discussed using alpine vs a slim ubuntu/debian image. A switch would probably also ease the pain here, because debian/ubuntu is less "outlandish" than alpine. However, I think we decided that the added pain of using alpine was worth it because we get significantly smaller images. Even if you consider layer sharing between the repo2docker image and the images built by repo2docker. This means we should "stick" with Alpine.

betatim:
@yuvipanda can you add `build-base` to the `apk` command for the build stage of `Dockerfile` in the root directory? I think this will fix the build

yuvipanda:
@betatim I already made a separate PR addressing it :) https://github.com/jupyterhub/repo2docker/pull/1051

yuvipanda:
I've rebased this, hopefully that should make it pass all the tests

manics:
Does the minimum version of Sphinx need to be bumped, currently we claim almost any version is supported:
https://github.com/jupyterhub/repo2docker/blob/ea90ae23d363d600ee172c52da01268d76cc8ebf/docs/doc-requirements.txt#L3

In https://github.com/jupyterhub/mybinder.org-deploy/pull/1914 I pinned all versions exactly and now dependabot takes care of testing upgrades, would it be worth doing the same here?

yuvipanda:
@manics full pinning + dependabot seems pretty awesome. However, for this PR, I have just changed the pin to a minimum of 1.8 - that's when `add_css_file` was introduced - this lets us unbreak our green CI check marks!

yuvipanda:
Would you be willing to do the pinning + dependabot setup here too? Would be sweet! Will provide review + merges if necessary!

manics:
I opened another PR with the pinning, then realised you'd already opened this one :smile: https://github.com/jupyterhub/repo2docker/pull/1052

I'll merge this, and open a follow-up PR instead.

PR #1053, opened 2021-07-03
Remove CircleCI docs build

yuvipanda:
Same as our RTD build, and the previews there are
nicer

Follow-up https://github.com/jupyterhub/repo2docker/pull/1052


manics (approved):
We may need to disable the job in circle-ci too, but I don't have permissions:
image

yuvipanda:
@manics I halted build!

PR #1054, opened 2021-07-03
Remove Pipfile & Pipfile.lock

yuvipanda:
- Seem unused - definitely not updated.
- I think pipenv has also lost a lot of its popularity
  over the last year or two
- This does not affect the Pipfile *buildpack*, so repos we
  build that have a Pipfile will not see any difference

Fixes https://github.com/jupyterhub/repo2docker/issues/1032


manics (approved):
Three months since https://github.com/jupyterhub/repo2docker/issues/1032 was opened and no-one's objected

PR #1055, opened 2021-07-04
Build ARM64 Docker images

manics:
We're doing it in JupyterHub and Z2JH, why not here too?

This is a modification of the [JupyterHub workflow](https://github.com/jupyterhub/jupyterhub/blob/c0b9250376a3c5919da8304b33c5124697996cec/.github/workflows/release.yml#L72-L145). The main difference is the Docker image tags are calculated using the method from the existing Docker Hub automated build hook:
https://github.com/jupyterhub/repo2docker/blob/eb7520901913d61e7745d136ce127142d5fa8506/hooks/post_push#L3-L6

Prerequisites:
- [ ] Disable automated Docker Hub build
- [ ] Add `secrets.DOCKERHUB_USERNAME` `secrets.DOCKERHUB_TOKEN` to push to Docker Hub
- [ ] Figure out how to freeze packages for arm64 (https://github.com/jupyterhub/repo2docker/pull/1024 makes it a lot easier)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,3 +32,66 @@ jobs:
         uses: pypa/gh-action-pypi-publish@master
         with:
           password: ${{ secrets.PYPI_API_TOKEN }}
+
+  publish-docker:
+    runs-on: ubuntu-20.04
+
+    services:
+      # So that we can test this in PRs/branches
+      local-registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+
+    steps:
+      - name: Should we push this image to a public registry?
+        run: |
+          if [ "${{ startsWith(github.ref, 'refs/tags/') || (github.ref == 'refs/heads/master') }}" = "true" ]; then
+              # Empty => Docker Hub
+              echo "REGISTRY=" >> $GITHUB_ENV
+          else
+              echo "REGISTRY=localhost:5000/" >> $GITHUB_ENV
+          fi
+
+      # versioneer requires the full git history for non-tags
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -32,3 +32,66 @@ jobs:
...
+          # Allows pushing to registry on localhost:5000
+          driver-opts: network=host
+
+      - name: Setup push rights to Docker Hub
+        if: env.REGISTRY != 'localhost:5000/'
+        run: |
+          docker login -u "${{ secrets.DOCKERHUB_USERNAME }}" -p "${{ secrets.DOCKERHUB_TOKEN }}"
+
+      # when building jupyter/repo2docker:master
+      # also push jupyter/repo2docker:1.2.3-3.abcd1234 (replace + with -)
```
yuvipanda, line 80: Looking at https://hub.docker.com/r/jupyter/repo2docker/tags?page=1&ordering=last_updated, looks like the current tags are like `2021.03.0-49.gde7468d`. Since we're picking it up from versioneer, I'm guessing the new tags will also be similar?

manics: It should be, there was some trial and error to get this version to match the one from `hooks/post_push`

minrk:
> Figure out how to freeze packages for arm64 (Refreeze with conda-lock #1024 makes it a lot easier)

I *believe* this will mostly be adding `--platform linux-aarch64 --platform linux-64` to the lock call (and dealing with any issues if there are unsupported packages or versions).

Then it's mostly picking the right lockfile for the architecture in the CondaBuildPack. I'm not sure where that should come from. Explicit arg only, or detect platform? I'm not sure where else we assume x64.

consideRatio:
This is a PR related to #1180

PR #1061, opened 2021-07-09
Rename requirements.py-3.5.txt to requirements.py-3.5.pip

manics:
#1024 moved some pip dependencies out of the conda environment file, but this means dependabot is now trying to bump them.

Since these dependencies are for Python 3.5 having them bumped by dependabot _may_ break something. Dependabot doesn't have a way to exclude files/directories, so this renames the file instead.


consideRatio (approved):
Makes sense to me!

consideRatio:
@manics since we are low on maintainers for this repo, i suggest you do a self-review at this point in time and self-merge if this still make sense to you! I'm not feeling confident about pressing merge in general as I don't feel confident about the code base.

minrk:
Unfortunate, but without a dependabot configuration to ignore files, no real choice, I suppose. Let's keep an eye on it and switch back when we can configure dependabot in a sensible way.

Issue #1063, opened 2021-07-13
ModuleNotFoundError: No module named 'chardet'


johnhoman:
### Bug description
Package dependency `chardet` not declared in setup file.

#### Expected behaviour
```bash
➜  repo2docker git:(master) mkvirtualenv -q repo2docker
(repo2docker) ➜  repo2docker git:(master) python -m pip install -e . -q
WARNING: You are using pip version 21.0.1; however, version 21.1.3 is available.
You should consider upgrading via the '/Users/jhoman/.virtualenvs/repo2docker/bin/python -m pip install --upgrade pip' command.
(repo2docker) ➜  repo2docker git:(master) python -c "import repo2docker.utils"
(repo2docker) ➜  repo2docker git:(master) echo $?
0
```
#### Actual behaviour
```sh
➜  repo2docker git:(master) mkvirtualenv -q repo2docker
(repo2docker) ➜  repo2docker git:(master) python -m pip install -e . -q
WARNING: You are using pip version 21.0.1; however, version 21.1.3 is available.
You should consider upgrading via the '/Users/jhoman/.virtualenvs/repo2docker/bin/python -m pip install --upgrade pip' command.
(repo2docker) ➜  repo2docker git:(master) python -c "import repo2docker.utils"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/jhoman/github.com/jupyterhub/repo2docker/repo2docker/__init__.py", line 6, in <module>
    from .app import Repo2Docker
  File "/Users/jhoman/github.com/jupyterhub/repo2docker/repo2docker/app.py", line 28, in <module>
    from .buildpacks import (
  File "/Users/jhoman/github.com/jupyterhub/repo2docker/repo2docker/buildpacks/__init__.py", line 2, in <module>
    from .python import PythonBuildPack
  File "/Users/jhoman/github.com/jupyterhub/repo2docker/repo2docker/buildpacks/python/__init__.py", line 4, in <module>
    from ..conda import CondaBuildPack
  File "/Users/jhoman/github.com/jupyterhub/repo2docker/repo2docker/buildpacks/conda/__init__.py", line 10, in <module>
    from ...utils import is_local_pip_requirement
  File "/Users/jhoman/github.com/jupyterhub/repo2docker/repo2docker/utils.py", line 6, in <module>
    import chardet
ModuleNotFoundError: No module named 'chardet'
(repo2docker) ➜  repo2docker git:(master) echo $?
1
```

[...]

PR #1064, opened 2021-07-14
fix: add chardet, a not explicitly declared dependency

johnhoman:
The `setup.py` file is missing the imported dependency [chardet](https://pypi.org/project/chardet/). This PR adds that dependency 

PR will fix issue #1063

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,7 @@ setup(
     name="jupyter-repo2docker",
     version=versioneer.get_version(),
     install_requires=[
+        "chardet",
         "docker",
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
Thanks this LGTM! I ran into the same issue and could indeed find that `repo2docker` in the `utils.py` file imports chardet explicitly, which makes it a bug to be fixed.

I believe this happens now but not later because a failure of `docker-py` we have as a dependency in repo2docker is no longer depending `six` which in turn would have required `chardet`.

PR #1066, opened 2021-07-15
Workaround docker-py dependency's failure to import six

consideRatio:
`repo2docker` depends on `docker`, but `docker==5.0.0` is failing to import its dependency `six` and isn't getting patched because of a lack of maintainers it seems.

This is a PR similar to one we have done for `chartpress` as well (https://github.com/jupyterhub/chartpress/commit/8b486732f4ea518f65f4538dbb4e334f5e8aa5c6). The reason for repo2docker tests succeeds anyhow without this workaround is because six is available due to other reasons in the environment where tests run apparently.

PS: The Dockerfile was updated as installing a six wheel would fail as it would uninstall an existing six first but run into a failure doing so, so I adjusted the Dockerfile to not try to install a `six` wheel from the previous Dockerfile build step.

Closes #1039

PR #1067, opened 2021-07-24
[MRG] Release 2021.08.0

manics:
It's about time we had a release! This also adds changelogs for the previous two releases (`2021.01.0` and `2021.03.0`)

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -24,6 +24,7 @@ API changes
 Bug fixes
 ---------
 
+- Workaround docker-py dependency's failure to import six :pr:`1066:` by :user:`consideratio`
 - fix: add chardet, a not explicitly declared dependency :pr:`1064` by :user:`johnhoman`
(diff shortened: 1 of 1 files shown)


consideRatio:
repo2docker fails to install because of #1064 isn't released currently. Further in some context where `six` insn't available, it will also fail by being dependant on docker where docker in turn has a critical regression in 5.0.0 where that I workaround in https://github.com/jupyterhub/repo2docker/pull/1066.

I think we should workaround that issue by merging #1066 before this release, because docker-py is so under-maintained that no release seem to happen causing many dependent projects to fail.

manics:
We need https://github.com/jupyterhub/repo2docker/pull/1071 or something similar to fix the Docker build, then I think we can release!

manics:
Any last minute issues? Otherwise I'll tag!

consideRatio:
Excellent work on this @manics!!! :heart:!

This LGTM! How did you do the changelog in RST format?

manics:
`github-activity` followed by a regex search and replace with backreferences 😃

PR #1071, opened 2021-08-04
GitHub workflow build and push to Docker hub

manics:
The `jupyter/repo2docker` Docker Hub image [hasn't been updated for 20 days (last tag is 16 July 2021)](https://hub.docker.com/r/jupyter/repo2docker/tags?page=1&ordering=last_updated).

[Two PRs have been merged](https://github.com/jupyterhub/repo2docker/pulls?q=is%3Apr+merged%3A%3E2021-07-16+) since that last build and those should've triggered a new Docker image.

Automated Docker Hub builds were meant to be [discontinued from 18 June](https://www.docker.com/blog/changes-to-docker-hub-autobuilds/) which is before the last build, but perhaps they kept going for a bit longer?

This PR is a copy of https://github.com/jupyterhub/repo2docker/pull/1055 but for amd64 only.
Since this is building for the native architecture there should be no time penalty in using docker buildx.

Required:
- [x] Add secrets for publishing to [`quay.io/jupyterhub/repo2docker`](https://quay.io/repository/jupyterhub/repo2docker?namespace=jupyterhub): `secrets.DOCKER_REGISTRY_USERNAME` `secrets.DOCKER_REGISTRY_TOKEN`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,3 +32,66 @@ jobs:
         uses: pypa/gh-action-pypi-publish@master
         with:
           password: ${{ secrets.PYPI_API_TOKEN }}
+
+  publish-docker:
+    runs-on: ubuntu-20.04
+
+    services:
+      # So that we can test this in PRs/branches
+      local-registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+
+    steps:
+      - name: Should we push this image to a public registry?
+        run: |
+          if [ "${{ startsWith(github.ref, 'refs/tags/') || (github.ref == 'refs/heads/master') }}" = "true" ]; then
+              # Empty => Docker Hub
+              echo "REGISTRY=" >> $GITHUB_ENV
+          else
+              echo "REGISTRY=localhost:5000/" >> $GITHUB_ENV
+          fi
+
+      # versioneer requires the full git history for non-tags
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      # Setup docker to build for multiple platforms, see:
+      # https://github.com/docker/build-push-action/tree/v2.4.0#usage
+      # https://github.com/docker/build-push-action/blob/v2.4.0/docs/advanced/multi-platform.md
+
+      - name: Set up QEMU (for docker buildx)
+        uses: docker/setup-qemu-action@25f0500ff22e406f7191a2a8ba8cda16901ca018
+
+      - name: Set up Docker Buildx (for multi-arch builds)
+        uses: docker/setup-buildx-action@2a4b53665e15ce7d7049afb11ff1f70ff1610609
+        with:
+          # Allows pushing to registry on localhost:5000
+          driver-opts: network=host
+
+      - name: Setup push rights to Docker Hub
+        if: env.REGISTRY != 'localhost:5000/'
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -32,3 +32,66 @@ jobs:
...
+          docker login -u "${{ secrets.DOCKERHUB_USERNAME }}" -p "${{ secrets.DOCKERHUB_TOKEN }}"
+
+      # when building jupyter/repo2docker:master
+      # also push jupyter/repo2docker:1.2.3-3.abcd1234 (replace + with -)
+      - name: Get list of repo2docker docker tags
+        run: |
+          VERSION=$(python3 -c 'import versioneer; print(versioneer.get_version().replace("+", "-"))')
+          TAGS="${{ env.REGISTRY }}jupyter/repo2docker:$VERSION"
+          if [ "${{ github.ref }}" == "refs/heads/master" ]; then
+            TAGS="$TAGS,${{ env.REGISTRY }}jupyter/repo2docker:master"
```
manics, line 86:
```suggestion
          if [ "${{ github.ref }}" == "refs/heads/main" ]; then
            TAGS="$TAGS,${{ env.REGISTRY }}jupyter/repo2docker:main"
```
Depending on whether this is merged before or after https://github.com/jupyterhub/repo2docker/pull/1068

consideRatio:
Hmmm... we must use https://docs.docker.com/docker-hub/access-tokens/ for this, so we should add a bot user to a team with permissions to this repo. I don't have permissions to do that.

I could pay money to become dockerhub pro user and be able to registered another docker personal access token and expose all repo's i have access to via that token. Aaaargh... PyPI have it perfect, they allow a user to create a token specific to a repo they have access to --- that is what I'd like here as well. Well...

Okay here are the options as I see them:
1. One member of the jupyter dockerhub org with push access currently exposes a PAT (personal access token), but only one can be created for free.
2. An owner of the jupyter dockerhub org invites the jupyterhubbot account, which in turn currently have 2 PATs but only 1 is free, so it would need to delete both and create a new that is used for all the three repo's it is configured for use with currently.
  image
3. We abandon docker hub in favor of quay.io
4. We do something else.

manics:
I think moving to quay.io is a good long term solution and provides more stability than Docker Hub. If we have a solution for Docker Hub I'd support also pushing to quay.io as it means we're no longer tied to a single registry.

If we decided to move to quay.io and to drop automatic pushes to Docker Hub we could reduce inconvenience to users by manually pushing just releases to Docker Hub for a year which is hopefully enough time to communicate the change in registry.

manics:
I've rebased to resolve a conflict.

I've tested the changes required for quay.io here: https://github.com/manics/repo2docker/commit/2898409b659abc8d6c343ecaa168389155020ad2
- https://quay.io/repository/manics/repo2docker?tag=latest&tab=tags

manics:
I've:
- updated the workflow for quay.io
- created https://quay.io/repository/jupyterhub/repo2docker?namespace=jupyterhub
- added a bit to the readme

I don't have permissions on the GitHub repo to add the secrets though.

consideRatio:
@manics nice work! I've given you full access to this repo, can you setup the container registry credentials yourself?

I have been locked out of quay.io as part of when they required everyone to sign in using redhat accounts. My coupling with the github account etc failed somehow and now I have an open support ticket to get back in control of my quay.io account.

manics:
Thanks, added! Let's see if it works.....

PR #1072, opened 2021-08-14
Typo fix in utils docstring

jgarte:


yuvipanda:
Thank you very much for the patch, @jgarte!

PR #1073, opened 2021-08-15
Set default, type, and help method for engine in argparse

jgarte:


minrk (approved):

jgarte (commented):
Thank you!

betatim:
The formatting needs fixing for the CI to pass. You can use `black .` in the repositories top level directory to have the formatting fixed automatically.

manics:
The only disadvantage I can think of is it hard-codes the CLI to explicitly use `docker`. This means we might introduce a breaking change if we want to switch to "autodetect" in future.

minrk:
> The only disadvantage I can think of is it hard-codes the CLI to explicitly use docker. This means we might introduce a breaking change if we want to switch to "autodetect" in future.

This is already the case. The only difference now is the default is now _communicated_ in the CLI, it is still unchanged. Makign it `autodect` will require a change in two places, but will not have any differences wrt backward compatibility.

PR #1079, opened 2021-09-07
Add `--gpus` flag and config to start built container with nvidia Driver.Capabilities

adriendelsalle:
Description
---

Add a `--gpus` CLI flag to enable gpu support when running a container from the CLI
Add a `gpus`  config on the app to store CLI value, default it to `False` to preserve current behavior
Pass a device request for gpu to the engine API

This assumes that the host is well configured as stated in [docker documentation](https://docs.docker.com/config/containers/resource_constraints/#gpu): both NVIDIA drivers and container runtime have to be installed.

Now, running `repo2docker --gpus <some-repo>` makes GPU(s) accessible from the container:

```
adrien@71ab26d163ff:~$ nvidia-smi
Tue Sep  7 10:00:55 2021       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 470.57.02    Driver Version: 470.57.02    CUDA Version: 11.4     |
(...)
+-----------------------------------------------------------------------------+
```

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -219,6 +219,13 @@ def get_argparser():
 
     argparser.add_argument("--engine", help="Name of the container engine")
 
+    argparser.add_argument(
+        "--gpus",
+        dest="gpus",
+        action="store_true",
+        help="Expose the GPU device(s) to the containers",
+    )
+
     return argparser
 
 
@@ -280,6 +287,8 @@ def make_r2d(argv=None):
         # Can't push nor run if we aren't building
         args.run = False
         args.push = False
+
+    r2d.gpus = args.gpus
 
     r2d.run = args.run
     r2d.push = args.push
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -381,6 +381,14 @@ class Repo2Docker(Application):
         config=True,
     )
 
+    gpus = Bool(
+        False,
+        help="""
+        Expose GPU device(s) when running a container
+        """,
+        config=False,
+    )
+
     engine = Unicode(
         "docker",
         config=True,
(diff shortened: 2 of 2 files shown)


repo2docker/__main__.py
```diff
@@ -281,6 +288,8 @@ def make_r2d(argv=None):
         args.run = False
         args.push = False
 
+    r2d.gpus = args.gpus
```
betatim, line 291:
```suggestion
    r2d.expose_gpus = args.gpus
```
what do you think of this name? The generic `gpus` is a bit generic. As in, when you see it later in the code you could think that it contains a list of GPU devices or references to GPUs or that it is a flag.

repo2docker/__main__.py
```diff
@@ -219,6 +219,13 @@ def get_argparser():
 
     argparser.add_argument("--engine", help="Name of the container engine")
 
+    argparser.add_argument(
+        "--gpus",
+        dest="gpus",
+        action="store_true",
+        help="Expose the GPU device(s) to the containers",
```
betatim, line 226:
```suggestion
        help="Expose the GPU device(s) to the container",
```

betatim:
I think this is a good thing to add. Two things I've been thinking about and don't know the answer to:

1. is there something like this when we use podman as container runtime? There has been some work to make steps towards being able to use backends other than docker to run containers so we should think about how adding CLI flags like this interacts with that effort. Maybe @manics can help us answer this/give his perspective
2. in your comment in https://github.com/jupyterhub/repo2docker/issues/471 you mentioned that some images already configure things so that no extra parameters are needed. Is there the possibility of things breaking if you use an image like that and set the extra capabilities here? I think no but wanted to check with you

consideRatio:
I think the approach for this must be more generic.

1. More may be required by the container to have this work properly in my experience, even though this may be one of the things of relevance. Having `--gpus` declare it enables it without further detail is hiding complexity making us as maintainers responsible to help users that assume what we wrote was sufficient when its not always is.
2. This is probably nvidia specific and won't work well if we would want to support another kind of GPU if we just name it `--gpus`

I'll make a judgement call and close this PR that is adding a flag to configure things automatically, but think a PR to document how to use `extra_run_kwargs` for this purpose would be reasonable.

PR #1081, opened 2021-09-12
Bump JupyterLab to 3.1.11

choldgraf:
This bumps JupyterLab to the latest patch version, to remove some inappropriate language (yes I can't believe I just said that).

I'm not sure if this is all that's needed to fix the problem in Binder more generally, but it seemed reasonable? It's been a while since I've looked at the buildpack mechanisms. Would be happy to try making fixes elsewhere if they need to happen, or others feel free to push to this branch!

closes https://github.com/jupyterhub/repo2docker/issues/1080


betatim:
I think you also need to run `repo2docker/buildpacks/conda/freeze.py` to refreeze the environments. Some documentation on doing that https://repo2docker.readthedocs.io/en/latest/contributing/tasks.html#update-and-freeze-buildpack-dependencies

choldgraf:
OK I think that I've done that properly 😬 - two quick things:

- I needed to install something called `conda-lock` but this wasn't in the `dev-requirements.txt`. Should I add it there or at least add step on that page to install it?
- I don't _think_ I needed Docker in order to do this, even though it's the first step in the instructions you gave (I don't have Docker installed). I just want to confirm that this is true haha

betatim:
The change to `conda-lock` happened in https://github.com/jupyterhub/repo2docker/pull/1024, so I think it makes sense that the instructions and what actually happens differ a bit. We should have remembered to update the instructions :-/

Adding it to `dev-requirements.txt` makes sense to me.

Given our docs are slightly. out of sync, I think the heuristic to use is "is this file totally new to the repo?" to decide which need to be committed and which shouldn't be committed. Can you double check that? (I looked via the GH PR "diff" view" but It is a bit tricky to tell which file is entirely new and which just has a huge diff).

choldgraf:
Also just to confirm for @betatim - yep I followed the approach you described. Since the files didn't match up, I figured I should only commit anything that was _modified_ but not added.

manics:
@choldgraf can you list the files that were _added_? Last time I ran the script it only modified existing files.

choldgraf:
I don't think that anything was added actually - if I recall it was just an addition because of the `venv` command.

PR #1085, opened 2021-09-16
update user_interface doc to reflect that lab is default

minrk:
reverse docs since now classic notebook is opt-in and lab is default, instead of vice versa. Everything else is the same.


docs/source/howto/user_interface.rst
```diff
@@ -38,6 +25,27 @@ or folder.
 To learn more about URLs in JupyterLab and Jupyter Notebook, visit
 `starting JupyterLab <http://jupyterlab.readthedocs.io/en/latest/getting_started/starting.html>`_.
 
+Classic Notebook
+================
```
choldgraf, line 29:
```suggestion
Classic Notebook Interface
==========================
```

just to disambiguate a bit, since "Notebook" is already such an overloaded term.

choldgraf (approved):
LGTM

PR #1086, opened 2021-09-19
[MRG] Test R minor version and add more tests

aplamada:
Update the R tests and add more tests to address the issues noticed in #1077 .

Feedback is welcome.


aplamada:
@yuvipanda When I noticed the problem with the R version I had in mind at least to systematically improve the tests for the minor version, and I thought to do it similar to the Python part. 
In the meantime the master evolved and I noticed also some improvements in this direction. 
What do you recommend me to do with this PR?
I understand that what I did: just improving the tests and not solving the problem is not the most constructive approach ...

yuvipanda:
I didn't see this PR properly before (sorry @aplamada) and I think I've added very similar tests to #1104. Can you see if any of these tests would be helpful there? If not, let's close this PR?

I'm really sorry this wasn't reviewed sooner. I think added tests are always useful, wether they solve the underlying problem or not!

aplamada:
@yuvipanda Thanks for the reply.
The tests are very similar. 
`tests/conda/r` could be improved similar to `tests/r/` to include tests for the versions `3.6`, `4.0` and `4.1`.

I think it makes sense to close this PR.

yuvipanda:
@aplamada yeah, that would be great! I haven't really paid any attention to the conda/r stuff at all. Would you be interested in working on that? I can help with more prompt review this time.

aplamada:
Yes. I am thinking to close this PR and make a new one for conda tests only.
When I will close it I will mention  #1104 and the new one. What do you think?

yuvipanda:
@aplamada that sounds great.

aplamada:
The PR is closed since the test are covered in #1104 and #1108 .

PR #1093, opened 2021-10-19
Add option `--group-id` to specify numerical group id

davidjsherman:
Add an option `--group-id` to specify **numerical group id** separately from the numerical user id. If a group id is not specified, maintain the legacy behavior of using the numerical user id for the group.

Images created with `--group-id=0` follow [Openshift best practices](https://docs.openshift.com/container-platform/4.9/openshift_images/create-images.html#images-create-guide-openshift_create-images) for container images, see [discussion](https://developers.redhat.com/blog/2020/10/26/adapting-docker-and-kubernetes-containers-to-run-on-red-hat-openshift-container-platform). Such images will work in the default unprivileged  [restricted security context constraint](https://docs.openshift.com/container-platform/4.9/authentication/managing-security-context-constraints.html#default-sccs_configuring-internal-oauth) on Openshift and OKD Kubernetes platforms, where a non-root arbitrary user id is allocated on the fly to containers.

The new test *test_user_groups* in [`tests/unit/test_users.py`](tests/unit/test_users.py) verifies the group id and checks file ownership. The existing test *test_users* now also tests for the legacy behavior. I admit that these tests are very similar and could perhaps be combined.

### DIFF
--- a/tests/unit/test_users.py
+++ b/tests/unit/test_users.py
@@ -89,7 +89,7 @@ def test_user_groups():
                 "--",
                 "/bin/bash",
                 "-c",
-                "id -u > id && id -g > grp && pwd > pwd && whoami > name && echo -n $USER > env_user".format(
+                "id -u > id && id -g > grp && stat --format %u:%g grp > id_grp && pwd > pwd && whoami > name && echo -n $USER > env_user".format(
                     ts
                 ),
             ]
@@ -106,3 +106,5 @@ def test_user_groups():
         # When group-id specified, group id in container is same as specified
         with open(os.path.join(tmpdir, "grp")) as f:
             assert f.read().strip() == groupid
+        with open(os.path.join(tmpdir, "id_grp")) as f:
+            assert f.read().strip() == userid + ":" + groupid


yuvipanda (changes requested):
Thanks a lot for opening this, @davidjsherman. Apologies for the late review.

https://github.com/jupyterhub/repo2docker/search?q=chown shows there's a bunch of `chown` calls that currently rely on the assumption that the gid is the same as uid, and will need to be modified to take the gid into account properly. Once that's done, I think this is good to be merged.

davidjsherman:
I can do that. Do you by chance have a general idea of the tests that need to be made? I can verify that the uids and gids are as specified, but it would be better to check functional dependencies, if there are any.

yuvipanda:
@davidjsherman not sure of the specifics, but perhaps look at the paths that are being chowned, and see what tests already cover them? Then, a gid variant can be added there.

tests/unit/test_users.py
```diff
@@ -58,3 +58,53 @@ def test_user():
...
+                username,
+                "--group-id",
+                groupid,
+                tmpdir,
+                "--",
+                "/bin/bash",
+                "-c",
+                "id -u > id && id -g > grp && stat --format %u:%g grp > id_grp && pwd > pwd && whoami > name && echo -n $USER > env_user".format(
+                    ts
+                ),
```
westurner, line 94: Why does this need `.format(ts)`?

westurner:
- [ ] add `--group-name` too for completeness
  - Though that isn't necessary to merge this

PR #1097, opened 2021-11-19
Set labels when building image from Dockerfile

TimoRoth:
We'd like to build our own custom images based on our own Docker Base-Images.
It works mostly fine, but the one big issue I ran into is that our Jupyter Hub does not show the images in its UI after successfully building them.

This is due to the labels it set on the command line not actually making it into the image. Adding them is easily achieved by adding this single line.

I also noticed that tljh makes use of an appendix to set some more stuff. So I extended this PR to have support for it as well.
For appendices to work, I needed to modify the Dockerfile on the fly, so I added that to the render method of the Docker BuildPack.
While making the build method based on that, I eventually realized that I had effectively re-implemented the base classes build method, with next to no differences.
And sure enough, just removing the build method entirely works like a charm.

This in turn meant that the label change had to move to the base class as well, where it could work in the exact same way, while simplifying the embedded Dockerfile a bit.

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -204,6 +204,14 @@ def get_argparser():
 
     argparser.add_argument("--appendix", type=str, help=Repo2Docker.appendix.help)
 
+    argparser.add_argument(
+        "--label",
+        dest="labels",
+        action="append",
+        help="Extra label to set on the image, in form name=value",
+        default=[],
+    )
+
     argparser.add_argument("--subdir", type=str, help=Repo2Docker.subdir.help)
 
     argparser.add_argument(
@@ -245,6 +253,13 @@ def make_r2d(argv=None):
     r2d.load_config_file(args.config)
     if args.appendix:
         r2d.appendix = args.appendix
+
+    for l in args.labels:
+        if "=" in l:
+            key, val = l.split("=", 1)
+            r2d.labels[key] = val
+        else:
+            r2d.labels[l] = ""
 
     r2d.repo = args.repo
     r2d.ref = args.ref
--- a/repo2docker/docker.py
+++ b/repo2docker/docker.py
@@ -73,6 +73,7 @@ class DockerEngine(ContainerEngine):
         dockerfile="",
         fileobj=None,
         path="",
+        labels=None,
         **kwargs,
     ):
         return self._apiclient.build(
@@ -87,6 +88,7 @@ class DockerEngine(ContainerEngine):
             dockerfile=dockerfile,
             fileobj=fileobj,
             path=path,
+            labels=labels,
             **kwargs,
         )
 
--- a/repo2docker/engine.py
+++ b/repo2docker/engine.py
@@ -176,6 +176,7 @@ class ContainerEngine(LoggingConfigurable):
         dockerfile="",
         fileobj=None,
         path="",
+        labels=None,
         **kwargs,
     ):
         """
@@ -204,6 +205,8 @@ class ContainerEngine(LoggingConfigurable):
             A tar file-like object containing the build context
         path : str
             path to the Dockerfile
+        labels : dict
+            Dictionary of labels to set on the image
 
         Returns
         -------
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -237,6 +237,17 @@ class Repo2Docker(Application):
         """,
     )
 
+    labels = Dict(
+        {},
+        help="""
+        Extra labels to set on the final image.
+
+        Each Label is a key-value pair, with the key being the name of the label
+        and the value its value.
+        """,
+        config=True,
+    )
+
     json_logs = Bool(
         False,
         help="""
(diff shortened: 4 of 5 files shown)


manics:
This works because we pass `**kwargs` straight through to the build client:
- https://github.com/jupyterhub/repo2docker/blob/4da2e1e71565b26a4bf1d0f0d26ae7c2373a1fd7/repo2docker/engine.py#L179
- https://github.com/jupyterhub/repo2docker/blob/4da2e1e71565b26a4bf1d0f0d26ae7c2373a1fd7/repo2docker/docker.py#L76
- https://github.com/jupyterhub/repo2docker/blob/4da2e1e71565b26a4bf1d0f0d26ae7c2373a1fd7/repo2docker/docker.py#L90

Should we tighten this up and make it a requirement for the builder to support labels, so `**kwargs` is only used for additional args?

TimoRoth:
Made the labels parameter a proper argument on the engine, as suggested.
Also extended the scope of the PR a bit, since I noticed that tljh also relies on --appendix to work, which was so far not the case for Dockerfile based builds.
This somehow ended up simplifying the code a whole bunch.

The main Problem with that is that it's a breaking chance for Dockerfiles which rely on ADD'ind or COPY'in in files. Since those will now be one level deeper, in /src/, in the build context.
As a bonus, the two provided wrapper scripts become available to Dockerfile builds, which at lest for our needs is quite handy, as we will no longer need to duplicate those into our r2d repo.

minrk:
👍 to adding labels, but the appendix was removed from custom dockerfiles on purpose in #250. It seems to be pretty context-dependent whether adding an appendix is appropriate for custom dockerfiles. I do like that it's very easy to understand what repo2docker does with custom Dockerfiles - no modifications. What assumptions can an appendix safely make of an arbitrary Dockerfile? It's a lot easier to write an appendix if you know it will be extending a generated Dockerfile from the standard templates, rather than requiring that it work for every possible Dockerfile. At least for BinderHub, I suspect the right thing to do is _no_ appendix, so maybe this should be an off-by-default flag that custom Dockerfiles can be modified?

TimoRoth:
On 23.11.2021 16:41, Min RK wrote:
> 👍 to adding labels, but the appendix was removed from custom 
> dockerfiles on purpose in #250 
> <https://github.com/jupyterhub/repo2docker/pull/250>. It seems to be 
> pretty context-dependent whether adding an appendix is appropriate for 
> custom dockerfiles. I do like that it's very easy to understand what 
> repo2docker does with custom Dockerfiles - no modifications. What 
> assumptions can an appendix safely make of an arbitrary Dockerfile?

tljh uses it to add labels via cli.
There does not seem to be another way to do that via CLI.
So adding a way to add labels via CLI would also be fine in my book, but 
it would need changes in tljh, which right now relies purely on 
--appendix for that.

> It's 
> a lot easier to write an appendix if you know it will be extending a 
> generated Dockerfile from the standard templates, rather than requiring 
> that it work for every possible Dockerfile. At least for BinderHub, I 
> suspect the right thing to do is /no/ appendix, so maybe this should be 
> an off-by-default flag that custom Dockerfiles can be modified?

I don't really see a downside to supporting the appendix with custom 
dockerfiles. It adds more flexibility for sure, and it seems more 
unexpected to me for the argument to just not do anything than for it to 
do something unexpected.

TimoRoth:
Changed this to expose the labels as CLI options instead, hope that's more acceptable.
Just needs tljh adjusted to actually use it.

repo2docker/docker.py
```diff
@@ -73,6 +73,7 @@ def build(
         dockerfile="",
         fileobj=None,
         path="",
+        labels={},
```
minrk, line 76: Good to avoid mutables in signatures:

```suggestion
        labels=None,
```

repo2docker/engine.py
```diff
@@ -176,6 +176,7 @@ def build(
         dockerfile="",
         fileobj=None,
         path="",
+        labels={},
```
minrk, line 179: Same here

```suggestion
        labels=None,
```

minrk (approved):
Great! I definitely think it's reasonable to expect the appendix to be applied, especially for single-calls to `jupyter-repo2docker`. I _also_ think it's reasonable as a repo author to expect your Dockerfile to be fully respected. Because both make sense, I think applying the appendix to the Dockerfie buildpack according to an option makes sense to me (and also warning if it's defined and not applied).

TimoRoth:
With a way to externally set labels, I can resolve our main problem.
So I'll see if I can come up with a clean implementation for the appendix situation and open a separate PR.

repo2docker/__main__.py
```diff
@@ -246,6 +254,13 @@ def make_r2d(argv=None):
     if args.appendix:
         r2d.appendix = args.appendix
 
+    for l in args.labels:
+        if "=" in l:
```
minrk, line 258: FWIW, this is a perfect application of [str.partition](https://docs.python.org/3/library/stdtypes.html#str.partition):

```python
key, _, value = l.partition("=")
```

minrk:
Thanks!

PR #1100, opened 2021-11-30
Add command line option to pass extra build args

TimoRoth:
This is effectively the second part of https://github.com/jupyterhub/repo2docker/pull/1097

Instead of allowing an appendix for custom Dockerfiles, this adds the ability to pass arbitrary build args to the build process, which is much cleaner and more predictable than appending lines to the Dockerfile.

For the normal buildpacks this does generally nothing, since they don't expect any buildargs you can't already set via other switches, but for custom Dockerfiles it adds a range of new possibilities.

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -263,11 +263,8 @@ def make_r2d(argv=None):
         r2d.appendix = args.appendix
 
     for l in args.labels:
-        if "=" in l:
-            key, val = l.split("=", 1)
-            r2d.labels[key] = val
-        else:
-            r2d.labels[l] = ""
+        key, _, val = l.partition("=")
+        r2d.labels[key] = val
 
     for a in args.build_args:
         key, _, val = a.partition("=")
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -212,6 +212,14 @@ def get_argparser():
         default=[],
     )
 
+    argparser.add_argument(
+        "--build-arg",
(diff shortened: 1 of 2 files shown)


repo2docker/__main__.py
```diff
@@ -246,6 +254,13 @@ def make_r2d(argv=None):
     if args.appendix:
         r2d.appendix = args.appendix
 
+    for a in args.build_args:
+        if "=" in a:
+            key, val = a.split("=", 1)
+            r2d.extra_build_args[key] = val
+        else:
+            r2d.extra_build_args[a] = ""
+
```
minrk, line 263:
```suggestion
        key, _, value = a.partition("=")
        r2d.extra_build_args[key] = val
```

minrk (approved):

yuvipanda:
Thanks a lot, @TimoRoth! Would love for a few example use cases!

TimoRoth:
Our setup is at https://github.com/OGGM/r2d
Currently the custom Dockerfile setup is in the test-docker branch, but will eventually be merged into master now that this is merged.
We have a lot of intricate binary dependencies, and also profit a lot from using Pyston.

When trying to set this up with conda, we end up with a horrendously broken environment that fails most of our tests with obscure errors(see the chain of failed actions GH Actions builds for the master branch on that same repo) most of the time. While with our docker images it's no huge issue to get this working at all times, since we have very precise control about how everything fits together.

PR #1102, opened 2021-12-16
Support R 4.1

yuvipanda:


tests/r/r4/verify
```diff
@@ -2,6 +2,6 @@
 library('ggplot2')
 
 print(version)
-if (version$major != "4") {
+if (version$major != "4" && as.double(version$minor) >= 1) {
   quit("yes", 1)
```
minrk, line 6: I think this change only affects 3.x >= 3.1, and 5.x >= 5.1. Is that intentional?

yuvipanda: @minrk I updates this line to more explicitly just check for 4.1

PR #1104, opened 2021-12-17
Get binary R packages from packagemanager.rstudio.com

### LINKED ISSUE #1041: Current RStudio version does not support R 4.1.0 graphics engine
### Bug description

[RStudio v1.2.5001](https://github.com/jupyterhub/repo2docker/blob/81e1e39/repo2docker/buildpacks/_r_base.py#L7-L10) does not support the R graphics engine v14 that comes with R v4.1.0 ("Camp Pontanezen") (see [release notes](https://stat.ethz.ch/pipermail/r-announce/2021/000670.html)).

#### Expected behaviour

No warning message.

#### Actual behaviour

```
Warning message:
R graphics engine version 14 is not supported by this version of RStudio. The Plots tab will be disabled until a newer version of RStudio is installed. 
```

### How to reproduce

1. [Binder](https://gke.mybinder.org/v2/gl/fkohrt%2FRMarkdown-sandbox/f5d8582b?urlpath=rstudio) ([source repo](https://gitlab.com/fkohrt/RMarkdown-sandbox/-/tree/f5d8582b))
2. Enter `R.version.string` (leading `R version 4.1.0 (2021-05-18)`)
3. Get warning message

### Your personal set up

Using `mybinder.org`.

### Related information

See also rstudio/rstudio#8383.

yuvipanda:
packagemanager.rstudio.com is a CRAN mirror provided
by rstudio, with *binary packages* prebuilt for many Linux
Distributions! https://www.rstudio.com/blog/announcing-public-package-manager/
has more excellent detail. It cuts down install times for R packages
by almost 90% in some cases!

Like MRAN (which we use now), they also provide a daily snapshot
of CRAN at that date
(https://docs.rstudio.com/rspm/news/#rstudio-package-manager-2021090).
The URL for CRAN for a particular date can be fetched via an API
call. We call that API, and we retry for earlier dates if we can't find one
for that date. However, note that rspm seems to do serverside magic
to give us packages from the earlier date anyway, so we don't need to
do the MRAN backoff behavior yet.

One possible issue about changing existing binder repos to use binary
builds rather than source builds is that the binary builds sometimes
require you have an apt package installed, and will fail if it is
not. We had to install the zmq library apt package for example -
source installs compile zmq from source, which is where the speedup
comes from. But unlike python wheels or conda packages, these binary
builds are not self-contained - they are linked to apt packages from
the specific distros. So some repos that worked before might fail now.
Due to this, we default to RSPM only if one of the following conditions are true:

1. The snapshot date is recent - After Jan 1 2022. I picked this arbitrarily.
2. R 4.1+ is being requested, as [mran](https://mran.microsoft.com/timemachine) doesn't
    seem to support R 4.1?

We also bring in newer versions of RStudio based on what R version they support,
and a matching jupyter-rsession-proxy. Fixes #1041

A bug where asking for R 4.0 gave us R 4.1 is also fixed, and we add a separate test for
that as well. Fixes https://github.com/jupyterhub/repo2docker/issues/1077

TODO:
- [x] Make sure fallback MRAN date works by checking to see if the URL exists still
- [x] Modify retry logic to look for snapshots in packagemanager.rstudio.com CRAN
        before looking into MRAN
- [x] Pin our devtools installs to dates and version that we know work for a given
       R version
- [x] Decide what the cutoff date for switchover to binary packages is
- [x] Fix failing unit test


manics:
Is it guaranteed that CRAN will have a daily snapshot for each day after `<YYYY-MM-DD>` that is well defined? If not then would it be worth looking backwards for the closest CRAN snapshot to avoid a situation where someone has a repo whose `runtime.txt` references a date in CRAN, who then subsequently updates it to a more recent date that isn't in CRAN, causing it to switch to MRAN?

repo2docker/buildpacks/r.py
```diff
@@ -186,28 +185,41 @@ def get_packages(self):
 
         return super().get_packages().union(packages)
 
-    def _get_latest_working_mran_date(self, startdate, max_prior):
+    def get_cran_mirror_url(self, snapshot_date):
+        # Call the API to find out if we have a snapshot available for the given date.
+        # If so, use the URL for that snapshot. If not, fall back to MRAN.
+        snapshots = requests.post(
+            "https://packagemanager.rstudio.com/__api__/url",
```
manics, line 192: How did you find out about this API? It's not in https://packagemanager.rstudio.com/__api__/swagger/index.html

yuvipanda:
@manics I had initially read https://docs.rstudio.com/rspm/news/#rstudio-package-manager-2021090 and came to the conclusion they'll have a snapshot for all days, but on re-reading it I'm not sure. I'll set the MRAN / packagemanager.rstudio.com cutoff be based on dates, and retry to slightly older snapshots if it can't find one for that date.

repo2docker/buildpacks/r.py
yuvipanda: @manics hmm, I found it by looking at network tab in https://packagemanager.rstudio.com/client/#/repos/1/overview

RaoOfPhysics:
Looking at https://packagemanager.rstudio.com/client/#/repos/1/overview it appears as if they've gone and snapshotted all of the days (with two exceptions in October 2017)?

yuvipanda:
@RaoOfPhysics ah, glad you found the holes in October!!! Will help me test and make sure we cover those cases.

RaoOfPhysics:
Here you go, @yuvipanda.

Screenshot 2021-12-20 at 12 14 24

yuvipanda:
Ok, I've changed the logic for when rspm is used as default to either be a snapshot request date in 2022+ or asking for R4.1. I think with these two, we shouldn't break any old repos that were dependent on source builds.

RaoOfPhysics:
I haven’t thought _too_ carefully about it, but I think the logic makes sense. Why R4.1+ and not R4.0+ though?

yuvipanda:
@RaoOfPhysics The R4.1 decision is because R4.0 is the latest I see in MRAN (https://mran.microsoft.com/timemachine). I'm mostly trying to make sure we break as little existing repositories as possible...

yuvipanda:
Unfortunately it looks like R 4.1 is being installed even when we ask for R 4.0. Trying to figure out why.

yuvipanda:
Looking at `apt list --installed` shows that `r-base-core` was not pinned by us, so ended up picking the latest version. I fixed that.

yuvipanda:
I've fixed the tests as well now!

repo2docker/buildpacks/_r_base.py
```diff
@@ -13,56 +14,62 @@
 SHINY_URL = "https://download3.rstudio.org/ubuntu-14.04/x86_64/shiny-server-1.5.12.933-amd64.deb"
```
minrk, line 14: These constants are unused now?

minrk (approved):

repo2docker/buildpacks/_r_base.py
yuvipanda: @minrk removed!

repo2docker/buildpacks/r.py
```diff
@@ -295,33 +327,30 @@ def get_build_scripts(self):
                 """,
             ),
             (
-                "${NB_USER}",
-                # Install a pinned version of IRKernel and set it up for use!
+                "root",
+                # RStudio's CRAN mirror needs this to figure out which binary package to serve.
+                # If not set properly, it will just serve up source packages
+                # Quite hilarious, IMO.
+                # See https://docs.rstudio.com/rspm/admin/binaries.html
```
manics, line 334: https://docs.rstudio.com/rspm/admin/binaries.html is a 404 at the moment :disappointed:

repo2docker/buildpacks/r.py
```diff
@@ -295,33 +327,30 @@ def get_build_scripts(self):
...
+                echo 'r-cran-repos={cran_mirror_url}' > /etc/rstudio/rsession.conf
                 """.format(
-                    devtools_version=DEVTOOLS_VERSION, irkernel_version=IRKERNEL_VERSION
+                    cran_mirror_url=cran_mirror_url
                 ),
             ),
             (
                 "${NB_USER}",
-                # Install shiny library
+                # Install a pinner version of devtools, IRKernel and shiny
```
manics, line 348: `pinner`, do you mean `pinned`?

tests/unit/test_r.py
```diff
@@ -64,28 +64,47 @@ def test_mran_date(tmpdir, runtime, expected):
...
+        # No snapshot for this date from
+        date(2021, 10, 22): date(2021, 10, 22),
+        # Snapshot exists for this date
+        date(2022, 1, 1): date(2022, 1, 1),
+    }
+
+    r = buildpacks.RBuildPack()
+    for requested, expected in test_dates.items():
+        snapshot_url = r.get_rspm_snapshot_url(requested)
+        print(snapshot_url)
```
manics, line 81: Is this `print` statement intentional?

repo2docker/buildpacks/r.py
```diff
@@ -186,28 +185,64 @@ def get_packages(self):
...
+
+    def get_devtools_snapshot_url(self):
+        """
+        Return url of snapshot to use for getting devtools install
+
+        devtools is part of our 'core' base install, so we should have some
+        control over what version we install here.
+        """
+        # IRKernel gets into CRAN on Nov 16 2018 (https://packagemanager.rstudio.com/client/#/repos/1/packages/IRkernel),
+        # so we try snapshot to just after that. We use rspm as we want binary packages all the time
```
manics, line 243: I'm afraid I don't understand what you mean by "we try snapshot to just after that". Do you mean "anytime after that"?

manics (commented):
A few minor comments, once those are answered this looks good to merge!

tests/unit/test_r.py
yuvipanda: Removed!

repo2docker/buildpacks/r.py
yuvipanda: I just cleaned up this function to remove the FIXME and be clearer.

repo2docker/buildpacks/r.py
yuvipanda: Found a replacement link!

repo2docker/buildpacks/r.py
```diff
@@ -72,10 +73,10 @@ def r_version(self):
             "3.5.1": "3.5.1-2bionic",
             "3.5.2": "3.5.2-1bionic",
             "3.5.3": "3.5.3-1bionic",
-            "3.6": "3.6.1-3bionic",
+            "3.6": "3.6.3-1bionic",
             "3.6.0": "3.6.0-2bionic",
-            "3.6.1": "3.6.1-3bionic",
-            "4.0": "4.0.2-1.1804.0",
+            "3.6.1": "3.6.1-1bionic",
```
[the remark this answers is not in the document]
yuvipanda: I thought I didn't see it in apt-cache madison r-base-core (added in this PR) but I was wrong. Must've gotten confused when I was upgrading 3.6 to 3.6.3-1bionic. Fixed.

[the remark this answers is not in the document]
yuvipanda: Thanks for catching this!

repo2docker/buildpacks/r.py
yuvipanda: Yep! Fixed!

manics (approved):
:partying_face:

yuvipanda:
@manics awesome! THANK YOU SO MUCH!

PR #1106, opened 2021-12-23
Add help message to freeze.py

manics:
I usually run `python freeze.py --help` to remind myself of how it works..... but instead it creates a `environment.py---help.yml` file


betatim:
Whoop!

PR #1107, opened 2022-01-05
Bump default R version to 4.1

yuvipanda:
3.6 is almost 4 years old now.

Related to https://github.com/jupyterhub/repo2docker/pull/1104 which
brings in a lot of other R improvements.


RaoOfPhysics:
Yay! Thanks for this, @yuvipanda! 🎉

manics (approved):
LGTM!

Have we done anything special in the past when bumping major R versions? Or do we just merge and deploy?

yuvipanda:
@manics I think we just merge and deploy. I'd want to merge #1104 as well before we deploy this, so people get faster builds

PR #1108, opened 2022-01-07
[MRG] add tests for R conda

aplamada:
It simply adds tests for R 3.6, 4.0 and 4.1 using conda.

This is a follow up of the closed PR #1086 .
The structure for the tests is as in #1104 .

@yuvipanda Can you please have a look at this new PR?


yuvipanda:
@aplamada thanks for making these! Any idea why they are failing?

aplamada:
@yuvipanda Thanks!  
It is not possible to satisfy all dependencies (see below). 


```bash
Looking for: ['r-base=4.1', 'r-irkernel=1.1', 'r-devtools']
  
Encountered problems while solving.
Problem: package r-irkernel-1.1-r35h6115d3f_0 requires r-base >=3.5,<3.6.0a0, but none of the providers can be installed
```

Previously it used to downgrade the R version (see [here](https://github.com/jupyterhub/repo2docker/issues/1077#issuecomment-909247900)).

yuvipanda:
ahhh. I don't really know enough about conda to know how to fix it tho :|

manics:
@yuvipanda Do we need to pin the version of `r-irkernel`? Can we rely on the conda solver to automatically install a compatible version of `r-irkernel` for what `r-base` version is requested?
https://github.com/jupyterhub/repo2docker/blob/43525352a7baf6c5a16c618d4dd4edf550511de6/repo2docker/buildpacks/conda/__init__.py#L359

aplamada:
@manics @yuvipanda  the new version `r-irkernel=1.2` works fine with all conda R tests.

manics:
Thanks!

PR #1110, opened 2022-01-08
Revert "Bump default R version to 4.1"

manics:
Reverts jupyterhub/repo2docker#1107 as the tests need to be updated:
https://github.com/jupyterhub/repo2docker/issues/1109


yuvipanda:
The tests for this now pass, but so does main.

PR #1118, opened 2022-01-25
Use a smaller R library in our tests

yuvipanda:
ggplot2 is a big library with a bunch of dependencies,
making our already slow R tests even slower. This PR
switches them to using the tinytest library, which has
no dependencies and is much smaller.

### DIFF
--- a/tests/conda/r3.6/environment.yml
+++ b/tests/conda/r3.6/environment.yml
@@ -1,3 +1,3 @@
 dependencies:
   - r-base=3.6
-  - r-ggplot2
+  - r-tinytest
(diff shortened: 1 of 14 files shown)


minrk (approved):

yuvipanda:
Looks like tinytest doesn't exist on old MRAN.

minrk:
@yuvipanda do you know why `build` succeeded, but `verify` failed? That means that the install is failing to install the package, but the `install.R` step in docker is being treated as a success.

yuvipanda:
@minrk I'm guessing the install.R failed, but was / is being ignored, while the `library` call trying to load it actually fails.

minrk:
Does seem to save at least 10 minutes total. 🎉

yuvipanda:
Thanks a lot, @minrk

PR #1119, opened 2022-01-25
Delete /tmp/downloaded_packages after running install.R

yuvipanda:
This can often leave ~100-200MB in /tmp that isn't re-used
after the image is set up.


minrk:
Should we have a test that verifies we aren't leaving files in likely cache places (tmp, .cache, apt places)?

minrk (approved):

yuvipanda:
@minrk I opened a new issue for that https://github.com/jupyterhub/repo2docker/issues/1121

PR #1122, opened 2022-01-26
Remove deprecated calls to distutils

minrk:
Python is removing all version-string parsing from the standard library, so use the semver utilities we already have for Julia.

Had to add some zero-padding, because `semver.VersionInfo.parse` is unnecessarily pedantic and doesn't accept '1' or '1.0', only '1.0.0'.

### DIFF
--- a/tests/unit/test_semver.py
+++ b/tests/unit/test_semver.py
@@ -1,5 +1,6 @@
 import pytest
-from repo2docker.buildpacks.julia import semver
+from semver import VersionInfo
+from repo2docker import semver
 
 
 @pytest.mark.parametrize("test_input, expected", [("1.5.2", (1, 5, 2)), ("1", (1,))])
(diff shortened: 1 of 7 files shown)


yuvipanda (approved):
Happy to merge after tests pass.

yuvipanda:
Looks like a real failure!

tests/unit/test_semver.py
```diff
@@ -145,3 +146,20 @@ def test_largerthan_equal():
         semver.create_semver_matcher(">=1.2.3").match(semver.str_to_version("1.2.2"))
         == False
     )
+
+
+@pytest.mark.parametrize(
+    "vstr, expected",
+    [
```
manics, line 153: Do we need a test case for suffixes (`1.2.3-rc4`)?

minrk: Don't think so, since we only use this for exact or truncated versions.

PR #1123, opened 2022-01-26
Allow passing in traitlets via commandline

yuvipanda:
Without this, you *always* needed a repo2docker_config.py
file to configure anything. This PR makes r2d match the
behavior of most traitlets based applications (like jupyter_server,
jupyterhub, nbconvert, etc)

Fixes https://github.com/jupyterhub/repo2docker/issues/1112

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -67,6 +67,20 @@ def get_argparser():
     )
 
     argparser.add_argument(
+        "--help-all",
+        dest="help_all",
+        action="store_true",
+        help="Display all configurable options and exit.",
+    )
+
+    argparser.add_argument(
+        "--version",
+        dest="version",
+        action="store_true",
+        help="Print the repo2docker version and exit.",
+    )
+
+    argparser.add_argument(
         "--config",
         default="repo2docker_config.py",
         help="Path to config file for repo2docker",
@@ -223,13 +237,6 @@ def get_argparser():
     argparser.add_argument("--subdir", type=str, help=Repo2Docker.subdir.help)
 
     argparser.add_argument(
-        "--version",
-        dest="version",
-        action="store_true",
-        help="Print the repo2docker version and exit.",
-    )
-
-    argparser.add_argument(
         "--cache-from", action="append", default=[], help=Repo2Docker.cache_from.help
     )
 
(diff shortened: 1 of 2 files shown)


manics:
Do you know how to get the `--help-all` argument working? E.g. for jupyterhub

```
$ jupyterhub --help
...
Options
=======
The options below are convenience aliases to configurable class-options,
as listed in the "Equivalent to" description-line of the aliases.
To see all configurable class-options for some <cmd>, use:
    <cmd> --help-all

...
```
```
$ jupyterhub --help-all
...
Class options
=============
The command-line option below sets the respective configurable class-parameter:
    --Class.parameter=value
This line is evaluated in Python, so simple expressions are allowed.
For instance, to set `C.a=[0,1,2]`, you may type this:
    --C.a='range(3)'

Application(SingletonConfigurable) options
------------------------------------------
--Application.log_datefmt=<Unicode>
    The date format used by logging formatters for %(asctime)s
    Default: '%Y-%m-%d %H:%M:%S'

...
```

yuvipanda:
@manics so we're mixing argparse and traitlets here, which is causing some of these issues. Ideally we would *just* be using traitlets and this will go away... I'm not entirely sure how to get `--help-all` working, other than to try get rid of argparse.

manics (approved):
@minrk Do you know of an easy fix? If not let's merge!

minrk:
I added `--help-all`. `Application().print_help()` is the method to call (`classes=True` is `--help-all`).

 It does get a little funky, since there are two full argparse CLI instances running now (the one generated by traitlets, and our handcrafted one). I think if we want to have full traitlets CLI parsing, perhaps we should _use_ traitlets CLI parsing. I understand not wanting to do that (it's simpler and nicer to write your own parser), but if we want the `--Class.trait=value magic`, it might be better to go all-in. We ought to get things like `--show-config` and friends for free, too.

manics:
CI failure is unrelated
```
  File "/usr/local/bin/jupyter-notebook", line 5, in <module>
    from notebook.notebookapp import main
  File "/usr/local/lib/python3.5/site-packages/notebook/notebookapp.py", line 134, in <module>
    from .terminal import TerminalManager
  File "/usr/local/lib/python3.5/site-packages/notebook/terminal/__init__.py", line 4, in <module>
    import terminado
  File "/usr/local/lib/python3.5/site-packages/terminado/__init__.py", line 7, in <module>
    from .websocket import TermSocket
  File "/usr/local/lib/python3.5/site-packages/terminado/websocket.py", line 87
    self.log_terminal_output(f'STDOUT: {content[1]}')
```
Also occurred on `main` this morning 😞 https://github.com/jupyterhub/repo2docker/runs/4963525952?check_suite_focus=true

yuvipanda:
@minrk agree going all inn is probably a good idea. Do you want me to do that to move this PR forward? Or can / should it be deferred?

minrk:
> Or can / should it be deferred?

It's fine to defer it.

> CI failure is unrelated

The test failure is a metadata bug in terminado 0.13. Should be fixed by https://github.com/jupyter/terminado/pull/140 once 0.13.0 is yanked.

minrk (approved):

manics:
Green! :rocket: 

Thanks all!

manics:
Follow-up issue here: https://github.com/jupyterhub/repo2docker/issues/1129

PR #1124, opened 2022-01-26
Allow passing in extra args to Docker initialization

yuvipanda:
Puts it in the ContainerEngine interface specific to Docker,
so PodMan can implement its own when necessary.

Fixes https://github.com/jupyterhub/repo2docker/issues/711

### DIFF
--- a/repo2docker/docker.py
+++ b/repo2docker/docker.py
@@ -73,7 +73,8 @@ class DockerEngine(ContainerEngine):
         try:
             kwargs = docker.utils.kwargs_from_env()
             kwargs.update(self.extra_init_args)
-            self._apiclient = docker.APIClient(version="auto", **kwargs)
+            kwargs.setdefault("version", "auto")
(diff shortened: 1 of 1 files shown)


repo2docker/docker.py
```diff
@@ -53,12 +54,26 @@ class DockerEngine(ContainerEngine):
...
+
     def __init__(self, *, parent):
         super().__init__(parent=parent)
         try:
-            self._apiclient = docker.APIClient(
-                version="auto", **docker.utils.kwargs_from_env()
-            )
+            kwargs = docker.utils.kwargs_from_env()
+            kwargs.update(self.extra_init_args)
+            self._apiclient = docker.APIClient(version="auto", **kwargs)
```
minrk, line 76: avoid error where `version` could be set multiple times:

```suggestion
            kwargs.setdefault("version", "auto")
            self._apiclient = docker.APIClient(**kwargs)
```

PR #1128, opened 2022-01-27
put micromamba in /usr/local/bin and use mamba for installs

minrk:
instead of tmp, since we want it to stick around.


minrk:
@SylvainCorlay any idea what's up [here](https://github.com/jupyterhub/repo2docker/runs/4965996930?check_suite_focus=true#step:6:1117)?
```
Error opening for reading "/tmp/mambafwypUVpTrZx": No such file or directory
Could not write out repodata file '"/srv/conda/pkgs/cache/b556ea5a.json"': No such file or directory
```

minrk:
Looks like the CLI compatibility isn't _quite_ there because there are a couple of places where `micromamba install` takes an environment.yml file, while mamba/conda must still use `env update --file`.

manics (approved):
Shall we get this in and test on mybinder.org?

minrk:
Let's give it a go, thanks for reviewing!

PR #1141, opened 2022-02-25
handle r version being unspecified in environment.yml

### LINKED ISSUE #1140: "Error during build: .0.0 is not valid SemVer string"
When I try to start mybinder: https://mybinder.org/v2/gh/ngs-docs/2020-ggg-201b-rnaseq/latest?urlpath=rstudio

I get the following error message upon build:

```
Waiting for build to start...
Picked Git content provider.
Cloning into '/tmp/repo2docker9q9ck4ah'...
HEAD is now at bfa8ad2 update README to point to 201b branch
Error during build: .0.0 is not valid SemVer string
```

From random googling I'm seeing this pop up occasionally in some docker-related places, but nothing clearly connected to my issue. Any ideas?

minrk:
when getting r-base from conda, `self.r_version` may be a falsy empty string. We can't do version comparisons when we don't have a version to compare!

closes #1140

### DIFF
--- a/repo2docker/buildpacks/_r_base.py
+++ b/repo2docker/buildpacks/_r_base.py
@@ -14,7 +14,7 @@ def rstudio_base_scripts(r_version):
     shiny_proxy_version = "1.1"
     shiny_sha256sum = "80f1e48f6c824be7ef9c843bb7911d4981ac7e8a963e0eff823936a8b28476ee"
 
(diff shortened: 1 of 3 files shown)


krassowski:
Looking forward to this getting merged :) Is there a way to disable rstudio installation as a workaround?

manics (approved):

PR #1144, opened 2022-03-15
Support setting appendix on Dockerfiles too

yuvipanda:
Until now, 'appendix' was ignored for repos built with
their own Dockerfile. This fixes that, and sets the appendix
on repos built with Dockerfiles too.

I want to set a REPO_URL env var for all images built with
repo2docker-action
(https://github.com/jupyterhub/repo2docker-action/issues/86)
and this is needed for that.


tests/dockerfile/simple/verify
```diff
@@ -1,3 +1,6 @@
 #!/bin/bash
 set -euo pipefail
 /usr/local/bin/sayhi.sh
+
+# Test that the appendix worked, even though this is a custom Dockerfile
+test $(cat /tmp/appendix) == "appendix"
```
consideRatio, line 6:
```suggestion
test $(cat /tmp/appendix) == "appendix"

```

yuvipanda:
https://github.com/jupyterhub/binderhub/blob/e04a8c3cec71ae8a1bf984da8623ecf8d81398bb/binderhub/binderspawner_mixin.py#L102 is how BINDER_REPO_URL is set. Instead, I just want to set REPO_URL in repo2docker itself.

repo2docker/buildpacks/docker.py
```diff
@@ -18,7 +19,15 @@ def render(self, build_args=None):
...
         Dockerfile = self.binder_path("Dockerfile")
         with open(Dockerfile) as f:
-            return f.read()
+            content = f.read()
+
+        if self.appendix:
+            content += "\n"
+            content += self.appendix
+
+        print(content)
```
consideRatio, line 28: Should this print statement be added here?

repo2docker/buildpacks/docker.py
```diff
@@ -18,7 +19,15 @@ def render(self, build_args=None):
         """Render the Dockerfile using by reading it from the source repo"""
```
consideRatio, line 19: Docstring needs updating, it would also help my review as I'm not confident about anything about a `appendix` relates to.

consideRatio (commented):
My understanding made explicit:
- `repo2docker` can be run with an [`--appendix` flag](https://repo2docker.readthedocs.io/en/latest/usage.html#cmdoption-jupyter-repo2docker-appendix), but when a Dockerfile was created, it wasn't honored.
- `REPO_URL` is the proposed name of a new environment variable to solve a problem.

It is my understand that a discussion about an env called `REPO_URL` would be a new environment variable. If that is correct, I'd very much like that `REPO_URL` instead is named `REPO2DOCKER_REPO_URL` or similar. Environment variables with a prefix related to the source code that defines them is a major help to avoid confusion, not only the the project that defines them but in other projects where the confusion otherwise propagates. I'm thinking for example on the amount of questions jupyterhub/z2jh/binderhub could get about REPO_URL that really was about repo2docker if we would leave out a REPO2DOCKER prefix.

yuvipanda:
@consideRatio ah sure re: naming of the env var. I'll open a separate PR to deal with that, this PR doesn't touch that.

repo2docker/buildpacks/docker.py
yuvipanda: Ah, removed!

repo2docker/buildpacks/docker.py
yuvipanda: Augmented.

repo2docker/buildpacks/docker.py
```diff
@@ -15,10 +16,21 @@ def detect(self):
...
 
     def render(self, build_args=None):
-        """Render the Dockerfile using by reading it from the source repo"""
+        """
+        Render the Dockerfile used by reading it from the source repo
+
+        If an appendix is configured, it will be appended to the Dockerfile
+        found in the source repo.
+        """
         Dockerfile = self.binder_path("Dockerfile")
```
consideRatio, line 25: Trying to understand if there was an issue with this code, I reacted to seeing the reference `self.dockerfile` be removed and a hardcoded mention of `"Dockerfile"` remain. In practice, `self.dockerfile` is hardcoded to "Dockerfile"` and only some references to that string relied on the placeholder name.

If this refactoring had been made, or `"Dockerfile"` would been consistently used as a hardcoded string, I wouldn't been confused.

```suggestion
        Dockerfile = self.binder_path(self.dockerfile)
```

consideRatio (approved):
This LGTM!

consideRatio:
I labelled this as a bug because it wasn't documented that `--appendix` would only work for non-Dockerfile situations.

repo2docker/buildpacks/docker.py
yuvipanda: @consideRatio i didn't even notice I replaced that! I just removed it, no reason to keep that.

manics:
Note this was discussed in a previous PR https://github.com/jupyterhub/repo2docker/pull/1097#issuecomment-976739820

betatim:
I still like the logic/argument that manics linked to. This means "no appendix for customer `Dockerfile` based builds is a feature, not a bug".

consideRatio:
Ah hmmm...

I haven't formed a clear opinion if it is a feature or a bug, but right now it is an undocumented exception as I understand it - which i consider a documentation bug at least.

For now I'll leave a blank vote with regards to making it supported or explicitly fail/emit warning, and a strong vote for any of those options is better than undocumented and silently ignored situation we currently have. Ping me for further input, I feel I'd need to onboard myself a bit further to be able to think clearly what I consider to make sense about supporting it or not.

manics:
I've made an alternative proposal in https://github.com/jupyterhub/repo2docker-action/issues/86#issuecomment-1069394867

yuvipanda:
@manics I like that :)

yuvipanda:
Going to close this as this was considered and rejected. I do agree with @consideRatio this should be documented though.

PR #1149, opened 2022-04-08
Update base notebook packages

yuvipanda:
- We should stop auto updating 3.6 now, so we can move on to
  newer notebook packages - a new notebook version ought to fix
  issues like https://github.com/2i2c-org/infrastructure/issues/1170
  for example.
- Update py3.6 environment one last time based on previous 
  contents of environment.yml file. From now on, 3.6 will only get security updates.
  This is the same way we treat 3.5
- No longer explicitly pin nbconvert, as whatever version notebook
  brings in should be sufficient.
- Bump up ipywidgets and jupyterlab while we're at it.
- No longer pin jupyterhub-singleuser either, as that too works
  properly with 3.7


yuvipanda:
https://github.com/jupyterhub/repo2docker/pull/1150 fixes the lint failure

minrk:
Probably worth exploring: conda-lock now has more conventional 'lockfile' file format, which allows updating individual packages (unlike the explicit conda export we currently use, which is harder to update manually). Only downside: the lockfile export is not (yet) installable by micromamba, so we still need the explicit format to use for the bootstrap installation.

Perhaps simplest option:

1. generate the lockfile from environment.yml, and
2. immediately generate the (redundant) `--kind explicit` file to use for installation


Then (I think) packages can still be updated more easily in the 'frozen' envs (mostly for security updates).

yuvipanda:
@minrk do you think that can be decoupled from this PR?

minrk:
Sorry, I swear I started that comment with "(not for this PR)". Definitely don't need to address that here. Just something that occurred to me because shipping security updates to the py35 env is a pain, and now will be for py36.

minrk (approved):

yuvipanda:
ty, @minrk :) Think you can hit the merge button too? :D

yuvipanda:
Thanks a lot, @minrk!

PR #1156, opened 2022-05-03
Fix Read-Only filesystem permission issue for log file

### LINKED ISSUE #1155: Error opening .jupyter-server-log.txt in singularity container
### Bug description

Starting a singularity container of an image that was converted from a repo2docker generated docker image will fail with following message:

```
  File "/usr/local/bin/repo2docker-entrypoint", line 110, in <module>
    main()
  File "/usr/local/bin/repo2docker-entrypoint", line 38, in main
    log_file = open(log_path, "ab")
OSError: [Errno 30] Read-only file system: '/srv/data/.jupyter-server-log.txt'
```
Singularity containers are by default immutable and thus /srv/data is a read-only filesystem. There was a PR ( #1151 ) that catches the "Permissioned Denied" case but probably should be extended to the generic OSError case (Read-Only filesystem)

#### Expected behaviour

repo2docker-entrypoint run sucessfully

#### Actual behaviour

repo2docker-entrypoint fails with above error message

### How to reproduce

1. Create a singularity image form a repo2docker generated docker image: `singularity build jupyterhub-server.sif docker-daemon://...`
2. Run singularity image: `singularity run  jupyterhub-server.sif`

### Your personal set up

 - OS: Centos7
 - Singularity version: 3.8.7-1.el7
 - repo2docker version: latest master

timeu:
This fixes #1155

### DIFF
--- a/repo2docker/buildpacks/repo2docker-entrypoint
+++ b/repo2docker/buildpacks/repo2docker-entrypoint
@@ -30,17 +30,21 @@ def main():
     #  2. current working directory: "."
     #  3. default temp directory for the OS (e.g. /tmp for linux)
     log_dirs = [".", tempfile.gettempdir()]
+    log_file = None
     if "REPO_DIR" in os.environ:
         log_dirs.insert(0, os.environ["REPO_DIR"])
     for d in log_dirs:
         log_path = os.path.join(d, ".jupyter-server-log.txt")
         try:
             log_file = open(log_path, "ab")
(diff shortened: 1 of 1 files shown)


manics:
Thanks, your explanation makes sense. Do you think we could use this opportunity to check that `log_file` was opened after the loop is exited, since in the rare event that the log file couldn't be opened in the temp directory the error would be hidden?

Issue #1159, opened 2022-06-03
Perform shallow fetches instead of clones


daradib:
### Current status

repo2docker performs shallow clones when refspec is not provided (#420). BinderHub always specifies a refspec, so this does not speed up any BinderHub builds.

    # When $ref is None
    git clone --depth 1 "$repo"

    # Otherwise
    git clone --no-checkout "$repo"
    hash="$(git rev-parse --quiet "ref")" # check_ref function
    git reset --hard "$hash"

A previous attempt was made to limit clone depth (#120), but was reverted in #130 and #131.

### Proposed change

Perform shallow fetches, which support commits/branches/tags, unlike `git clone -b`.

    git init -q
    git remote add origin "$repo"
    git fetch --depth 1 origin "$ref"
    git reset --hard FETCH_HEAD

This should also obviate the need for `check_ref`.

### Alternative option

I will submit a PR to shallow clone when refspec is explicitly set to `HEAD`, not just when it is not specified as it does currently. This should speed up BinderHub builds pointing to HEAD.

PR #1160, opened 2022-06-03
[MRG] Shallow clone HEAD

daradib:
Previously repo2docker only performed shallow clones when refspec was not provided (#420). We now also perform shallow clones if refspec is explicitly set to `HEAD`. BinderHub always specifies a refspec, often `HEAD`, so this should improve some build times.

See #1159 for extending shallow clones/fetches to other refspecs.

### DIFF
--- a/repo2docker/contentproviders/git.py
+++ b/repo2docker/contentproviders/git.py
@@ -1,5 +1,4 @@
 import subprocess
-import sys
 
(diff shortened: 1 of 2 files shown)


yuvipanda:
yay! I think not specifying a ref is same as HEAD so this is <3

yuvipanda:
Thank you for the PR, @daradib! Hope to see more :D

PR #1161, opened 2022-06-03
Get R from RStudio provided apt packages (.deb files)

yuvipanda:
- Instructions at https://docs.rstudio.com/resources/install-r/
- Simpler than getting it from CRAN apt repositories, as it's a single deb that is
   downloaded on a per-version basis rather than an entire apt repository with
   all the quirks that entails.
- RStudio provides debs *all* R versions for Ubuntu 20.04, 18.04 and 22.04 - 
  making it much easier for us to switch to a different base Ubuntu image (see
  #909)
- The packages are built from https://github.com/rstudio/r-builds, so even if RStudio
   stops providing them in the future, they can be built.
- Better compatibility with binary builds from packagemanager.rstudio.com (compared to
   R from conda), as these are appropriate deb builds for particular Ubuntu versions that 
   also come from rstudio.
- We just pass through full versions to the RStudio CDN - so new R versions will now
   be available as soon as RStudio hosts them! No more changes required in repo2docker
   when a new R version is available. This means this PR also adds support for R 4.2.
   This is awesome!
- We no longer hardcode *any* ubuntu or distro identifiers in the code!


yuvipanda:
Nah, RStudio < 4.0 was already broken lol. Fixed by https://github.com/jupyterhub/repo2docker/pull/1148.

Had to tell rstudio exactly where R is by setting R_HOME: https://github.com/jupyterhub/repo2docker/pull/1161/commits/0fed94cd6a4cbac973723cd8f165f7e35344e92e

Also, it rstudio will fail to start if the uid is < 1000. And on my mac, my uid was 502! So I had to test repo2docker with a `--user-id=1000` flag to get it to work. Took me a long time.

tests/r/simple/runtime.txt
```diff
@@ -1 +1 @@
-r-2020-07-07
+r-2022-01-07
```
manics, line 1: Does this need to be bumped? Or could we add it as a new test case?

tests/unit/test_r.py
```diff
@@ -7,20 +7,6 @@
 from repo2docker import buildpacks
 
 
-def test_unsupported_version(tmpdir):
```
manics, line 10: Does the removal of this test mean all versions are supported?

repo2docker/buildpacks/r.py
```diff
@@ -141,6 +133,19 @@ def detect(self):
...
+    def get_env(self):
+        """
+        Set custom env vars needed for RStudio to load
+        """
+        return super().get_env() + [
+            # rstudio (rsession) can't seem to find R unless we explicitly tell it where
+            # it is - just $PATH isn't enough. I discovered these are the env vars it
+            # looks for by digging through RStudio source and finding https://github.com/rstudio/rstudio/blob/main/src/cpp/r/session/RDiscovery.cpp
+            ("R_HOME", f"/opt/R/{self.r_version}/lib/R"),
+            ("R_DOC_DIR", "${R_HOME}/doc"),
```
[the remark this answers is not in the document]
yuvipanda: @manics it gets expanded at runtime. I think it's actually clearer to let it expand at runtime, as you can see that these are dependent if you run `env` while you are using the image.

tests/unit/test_r.py
yuvipanda: YESSS! basically anything that has a deb produced by RStudio is now supported

tests/r/simple/runtime.txt
yuvipanda: I unbumped it!

yuvipanda:
Thanks a lot for the review, @manics! I think I've addressed your comments now

repo2docker/buildpacks/r.py
```diff
@@ -141,6 +133,19 @@ def detect(self):
...
             return True
 
+    def get_env(self):
+        """
+        Set custom env vars needed for RStudio to load
+        """
+        return super().get_env() + [
+            # rstudio (rsession) can't seem to find R unless we explicitly tell it where
+            # it is - just $PATH isn't enough. I discovered these are the env vars it
+            # looks for by digging through RStudio source and finding https://github.com/rstudio/rstudio/blob/main/src/cpp/r/session/RDiscovery.cpp
```
consideRatio:
```suggestion
            # looks for by digging through RStudio source and finding https://github.com/rstudio/rstudio/blob/54f1090ab32d9670f075a76bd575a8e0e2c26e25/src/cpp/r/session/RDiscovery.cpp#L57-L74
```

If this changes, for example the code logic is moved, its helpful to know how the logic looked historically so you can search for its removal or relocation. So, I suggest adjusting this link to a permalink.

repo2docker/buildpacks/r.py
```diff
@@ -68,33 +64,29 @@ def r_version(self):
...
+            "4.0": "4.0.5",
+            "3.6": "3.6.3",
+            "3.5": "3.5.3",
+            "3.4": "3.4.0",
+            "3.3": "3.3.3",
         }
+
         # the default if nothing is specified
-        r_version = "4.1"
+        r_version = "4.1.0"
```
consideRatio, line 77: Should this be 4.1.3? The version map for 4.1 points to 4.1.3 now and previously pointed to 4.1.2. The previous version map used the latest available 4.1 but now the latest available version is 4.2, so if the policy is to use the latest, we could consider bumping it to 4.2.0 instead as well.

One change should be done here though, not using an older patch version for 4.1

consideRatio: Hmmm, below there is logic related to `self._r_version` and `r_version`, I think that makes us use the version map if you set it to 4.1, so maybe stop providing the patch version here and rely on the version map?

consideRatio (commented):
Some minor comments to be addressed, but otherwise looks amazing!!

repo2docker/buildpacks/r.py
yuvipanda: done!

I want to bump the default to 4.2 too, but in another PR

repo2docker/buildpacks/r.py
yuvipanda: @consideRatio github seems to have fucked this up somehow, the latest code already points to a git tag here.

yuvipanda:
@consideRatio ty, I've accounted for both your suggestions i think.

consideRatio (approved):
An update of the title to reflect using downloaded deb files and this is ready for merge i think. (I've not tested this, just reviewed code.)

manics:
How risky do you think this change is to users? Are you happy for it to go into a tag/release in the near future, or do you think it needs extended testing on mybinder.org first?

yuvipanda:
@consideRatio done!

@manics in general i always like testing in repo2docker for a week or two before a release :D

repo2docker/buildpacks/r.py
```diff
@@ -74,7 +74,8 @@ def r_version(self):
         }
 
         # the default if nothing is specified
-        r_version = "4.1"
+        # Use full version is needed here, so it a valid semver
+        r_version = version_map["4.1"]
```
consideRatio, line 78: Hmm actually, looking at this code, it seems that r_version is pointless to set here. Whenever it used referenced later in this function, it is always set explicitly first independent of its previous value.

I think this assignment is irrelevant, and we missed that because its name is too similar to `_r_version`.

consideRatio: Hmm I don't understand what you mean yet Yuvi

What I see is the function scoped new variable `r_version` is set on line 78. To me it seems that the value assigned is never read from the variable until the variable scope expires, making the assignment irrelevant.

Can you pinpoint the location in the code the assigned value on line 78 is influencing something?

yuvipanda: @consideRatio read in line 92 here: https://github.com/jupyterhub/repo2docker/pull/1161/files#diff-3fa4ebd097693f90be6a3f6fce113df19b43c5c3efcb9c5b0070432c09249b2fR92

yuvipanda: you'll note that it's outside the if defined in line 84, so if that if branch is not taken, this value is read and used.

consideRatio: Ah sorry for the noise!

yuvipanda: No worries, @consideRatio! I hope the comment helps clarify it!

yuvipanda:
@consideRatio @manics anything else i can do to help get this merged? :D

consideRatio:
I saw no unresolved comments and this has been open for a long time so i figure a merge is in order! Thanks Yuvi and good that you pinged!!

yuvipanda:
Yay thanks a lot for the merge, @consideRatio!

PR #1162, opened 2022-06-23
Remove outdated comment about Docker 17.09 syntax

yuvipanda:
We've already been using newer syntax for a while (--chown)

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -10,7 +10,6 @@ import sys
 import hashlib
 import escapism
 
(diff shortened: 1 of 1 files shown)


manics:
Based on https://github.com/jupyterhub/repo2docker/pull/969#issuecomment-708991908 I think the comment is still correct, and we don't want to bring in any newer features. Or are we already using other features that rely on a Docker version newer than 17.09?

yuvipanda:
aaah, silly me. thanks for the catch, @manics

PR #1163, opened 2022-06-29
Add JPL Dataverse

foobarbecue:
NASA's Jet Propulsion Laboratory recently implemented a dataverse installation and would like to be able to use it with repo2docker / binderhub.


yuvipanda:
Thanks a lot, @foobarbecue!

PR #1166, opened 2022-07-08
ci: switch to using a 2fa enabled accounts pypi api-token

consideRatio:
In the settings for this project, there were two separate api tokens setup for this project. One stemming from `mybinderteam` account in PyPI, and one for `jupyterhub-bot`. Only the latter has 2FA enabled, so in this PR I'm switching to using that PyPI API token instead and have removed `mybinderteam` on PyPI to have access to this package.

### Related
- https://github.com/jupyterhub/repo2docker/settings/secrets/actions
- https://github.com/jupyterhub/team-compass/issues/539


yuvipanda:
Thanks @consideRatio

PR #1168, opened 2022-07-24
Use full path to `buildpack-deps:bionic`

manics:
Non-default engines may require the full path including registry

Closes https://github.com/jupyterhub/repo2docker/issues/1146


manics:
Will be included in https://github.com/jupyterhub/repo2docker/pull/909

PR #1169, opened 2022-07-28
Support pulling from zenodo sandbox too

yuvipanda:
Test uploads should go to the sandbox, rather than to
main zenodo. The installation is exactly the same.


sgibson91 (approved):
LGTM!

PR #1177, opened 2022-08-22
consistent log handling when not using JSON loggers

minrk:
our logger was never being quite hooked up when not using JSON loggers, meaning that log messages (such as those fixed here to include newlines) were never shown.

### DIFF
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -154,7 +154,7 @@ class CondaBuildPack(BaseImage):
             "conda/activate-conda.sh": "/etc/profile.d/activate-conda.sh",
         }
         py_version = self.python_version
-        self.log.info("Building conda environment for python=%s" % py_version)
+        self.log.info("Building conda environment for python={py_version}\n")
         # Select the frozen base environment based on Python version.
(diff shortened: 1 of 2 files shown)


repo2docker/buildpacks/conda/__init__.py
```diff
@@ -154,7 +154,7 @@ def get_build_script_files(self):
             "conda/activate-conda.sh": "/etc/profile.d/activate-conda.sh",
         }
         py_version = self.python_version
-        self.log.info("Building conda environment for python=%s" % py_version)
+        self.log.info("Building conda environment for python={py_version}\n")
```
manics, line 157:
```suggestion
        self.log.info(f"Building conda environment for python={py_version}\n")
```

minrk: Good catch! This is probably the feature I want most from linters after 'does it compile' syntax checking. F541 (f-string without any placeholders) doesn't improve anything, but its inverse is super useful _and_ not checked-for.

manics: It'd be useful, but also tricky to avoid incorrect warnings where `{...}` isn't for formatting.

PR #1178, opened 2022-08-22
get CI working again

minrk:
Started as one fix, but now it's a couple:

- Avoid redundant resolution in Dockerfile pip install

  avoids error attempting to 'upgrade' a package installed by apk (a dependency of mercurial, perhaps), which pip refuses.

  Since dependencies are resolved already, adding `--ignore-installed --no-deps` just installs all the wheels as-is, skipping the redundant solve (already done in `pip wheel`) and upgrade.
- Update GHA runners to 20.04, since 18.04 is shutting down. This, in turn, changes how mercurial is installed to be a simple pip install, since `hg` is using Python 2.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   docker:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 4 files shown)


minrk:
self-merging since everything's failing without this

PR #1182, opened 2022-09-02
mount wheels from build stage instead of copying them

minrk:
avoids extra layer of unused wheel files

use mounted pip cache instead of disabling cache - all the benefits of no cache (size), and all the benefits of cache (local rebuild time).

I think we should consider enabling cache mounts instead of `--no-cache-dir` in repo2docker builds themselves, but that means requiring buildkit, which affects users. Requiring buildkit in our own image doesn't affect users at all.

see also : https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2856


yuvipanda:
Thanks, @minrk!

PR #1184, opened 2022-09-06
fail on unsupported Python

### LINKED ISSUE #552: Should requesting an unsupported version of Python be an error
With a repository like https://github.com/betatim/sdfg/tree/2cc8b8784862bdff252adbca785bb16c9fd291cd the user might expect to get Python 3.4 installed. However that version isn't supported by repo2docker.

On the following line we emit a warning in the logs

https://github.com/jupyter/repo2docker/blob/abe256d25a60179addff3a87aa5a0148e94a1aca/repo2docker/buildpacks/conda/__init__.py#L119

however it is easy to miss it as it comes before all the docker commands. Maybe we should make it an error to request a version of Python that is older than the oldest version we support. Having a fall through/default for other cases makes sense, I think.

### LINKED ISSUE #1183: When setting Python version to 3.10 using runtime.txt, you actually get v3.7
### Bug description

_Note: This bug was originally surfaced via a help request in gitter._

When you use a `runtime.txt` file to set the Python version, and request v3.10, you actually get version 3.7 (I suspect the default version r2d is configured to install?)

#### Expected behaviour

The Python version stated in the `runtime.txt` file is respected and Python version 3.10 is installed.

#### Actual behaviour

Instead of 3.10, Python 3.7 is available.

### How to reproduce

I created an MWE repo to demonstrate this: https://github.com/sgibson91/pythonv3.10-binder-mwe

1. Launch the above repo in binder: [Binder](https://mybinder.org/v2/gh/sgibson91/pythonv3.10-binder-mwe/HEAD)
2. Open a Terminal
3. Execute the command `python --version`
4. See result is `3.7.*`, not `3.10.*`

Alternatively:

2. Open a new Notebook
3. Execute the below code in a cell:

```
import sys
print(sys.version)
```

4. See result is `3.7.*`, not `3.10.*`

### Your personal set up

See files in https://github.com/sgibson91/pythonv3.10-binder-mwe for setup details.

minrk:
rather than warning (which users usually won't notice) and building with the wrong version of Python

closes #1183 (also closes #552 which is the same)

I think this was originally a warning because at some point we had a second step to explicitly install the requested Python version _after_ setting up the base env, so it still _could_ work, but without a fully frozen env. Then that step could potentially fail informatively if e.g. the requested Python didn't exist. But we don't do that anymore (if we ever did, I don't trust my memory), so failing seems better than the current behavior of just using the default Python instead.

### DIFF
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -163,7 +163,7 @@ class CondaBuildPack(BaseImage):
         frozen_name = "environment.lock"
         pip_frozen_name = "requirements.txt"
         if py_version:
-            if self.py2:
+            if self.python_version == "2.7":
                 # python 2 goes in a different env
                 files[
                     "conda/environment.py-2.7.lock"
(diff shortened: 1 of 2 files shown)


betatim:
Makes sense to me

PR #1185, opened 2022-10-04
Use enum to standardise `phase`

manics:
Replace the phase strings used in logs with an Enum.

This is a follow-up to https://github.com/jupyterhub/binderhub/pull/1525#issuecomment-1222036439 and replaces the inconsistent use of `failed`/`failure` with `failed`


consideRatio (approved):
LGTM!

PR #1188, opened 2022-10-09
ci: refactor julia/r/conda tests - now ~25 min instead of ~50 min

consideRatio:
Previously the CI system took ~50 minutes. The tests in the folders `tests/conda`, `tests/julia`, and `tests/r` all took longer than 25 minutes but are now reduced to about ~25 minutes or less.

There has been very little compromise on existing test coverage, and new test coverage has been gained against new language versions etc.

## Change description

- **Deleted and merged tests**
  Test coverage is retained well I think, even expanded towards modern versions. Overall the following principles guided what I retained. 
  - *oldest* - py2, py3.5, r3.6?, julia 1.0.2? (Project.toml buildpack), julia 0.6.4? (REQUIRE buildpack)
    See #1192
  - *default* - py3.7, r4.2, julia latest (Project.toml buildpack), julia 0.6.4 (REQUIRE buildpack)
    See #1190
  - *latest* - py3.10, r4.2, julia latest (Project.toml buildpack)
  - *before and after* - R with mran and R with rspm, julia environment variables after 1.0+
  - *pinnability* - ensure language versions can be pinned
  - *functionality* - binder folder, description file, requirements file, etc.
- **Renamed test folders, updated and coalesced READMEs**

### Changes to `tests/conda` tests

After | Before | Note
-|-|-
`py35-binder-dir` | `binder-dir` `py37` | Makes an old version the oldest version, and merges in the binder directory test
`py310-requirements-file` | `py38` `requirements` | Makes a less old version the latest version, and merges in the requirements file test
`py2` | `simple-py2` | Renamed
`r-postbuild-file` | `r-unspecified` `default-env` `simple` `r4.0` `r4.1` | Merges in the postBuild file test (`default-env`), merges tests of modern R versions into the latest version (`r4.0`, `r4.1`), merges tests of the default Python version (`simple`)
`r3.6-target-repo-dir-flag` | `repo-path` `r3.6` | Merges extra-args.yml file test with r3.6 pinning test
-| `channel-dep` | A deleted test, testing functionality part of mamba since 0.6 introduced more than two years ago

### Changes to `tests/r` tests

After | Before | Note
-|-|-
`r3.6-mran` | `r3.6` | Renamed
`r4.0-mran` | `mran`, `r4.0` | Merges similar tests
`r4.0-rspm` | `r4.0-rspm`, `r4.1` | Merges `r4.0` and `r4.1`, 
`r-rspm-apt-file` | `apt`, `simple` | Merges in the apt.txt file test
[...]

PR #1191, opened 2022-10-09
Remove conda buildpacks pin of r-irkernel to 1.2

consideRatio:
If r-base is pinned, r-irkernel will resolve to a version that is compatible with it. But if we pin r-irkernel and not r-base, the opposite will happen and we will end up with an older version of r-base than is supported by r-irkernel's modern versions.

I find debugging how versions resolve with conda is really tricky, so unless we have clear principles of what the pin should be and why, I strongly advocate we don't have it pinned here.

In this case, having r-irkernel pinned to 1.2 caused us the conda buildpack to get stuck at R version 4.1 instead of going to R 4.2 that is now available. Using a runtime.txt without R pinned will result in 4.2 though, but the conda buildpack is stuck at 4.1 without this.

## Related

- A test to catch this is available in #1188. If this is merged, we can remove a commit about this from there and it can be made a ci only PR
- The R buildpack defaults to 4.2 as of #1188

### DIFF
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -362,7 +362,7 @@ class CondaBuildPack(BaseImage):
                 (
                     "${NB_USER}",
                     r"""
-                ${{MAMBA_EXE}} install -p {0} r-base{1} r-irkernel=1.2 r-devtools -y && \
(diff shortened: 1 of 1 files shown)


yuvipanda:
Given r-devtools is not pinned, this seems fine to me.

PR #1196, opened 2022-10-19
Jupyterhub 3, separate Python kernel env from notebook/jupyterlab env

manics:
This is an experimental PR to add JupyterHub 3.0.0 for all Python versions by separating the kernel environment from the notebook/lab environment. This is already done for Python 2.7, and this PR extends it to all Python versions so we don't have to worry about compatibility between JupyerLab/JupyterHub and the kernel environments.

I've no idea how usable this will be in practice....

- Closes https://github.com/jupyterhub/repo2docker/issues/741
- Closes https://github.com/jupyterhub/repo2docker/issues/1195
- Closes https://github.com/jupyterhub/repo2docker/issues/956


minrk:
There are some tricky issues with separating the server env from the kernel env, especially involving things like widgets that include both a kernel and frontend component. For that reason, I don't think we should have separate envs by default, or even in most cases.

Separating the envs solve a lot of problems for us on mybinder.org, because we want to control some aspects of the server env, but don't care much about the kernel env, but this is in conflict with the folks who use binder to e.g. demo prototype JupyterLab features, widgets, UI extensions, etc., which must be in the server env.

I _do_ think better support for separation of the server and kernel env will let us be more proactive in upgrading the server env, e.g. having the kernel-only env for 'less supported' runtime versions. We already do this for py2, but I think it would be appropriate to be doing the same with 3.5 and 3.6 - i.e. saying you can still get a 3.6 _kernel_, but we can't keep supporting a 3.6 _server_. That would mean folks sticking to old Python for a UI demo will break sooner than folks sticking to old Python for kernel-only behavior, which I think reflects the reality of targeting those two areas.

But even doing that, folks with older kernels will not have working widgets (eventually, if not immediately) because the versions won't match.

That would mean generalizing the 'separate kernel from notebook' behavior a bit beyond py2 (and applying it to Python <=3.6 from now on), but still not doing it by default.

What do folks think?

manics:
A third option, though it's a lot more work: split jupyterhub-singleuser into a separate package, and add support for older python versions?

minrk:
Yeah, I think we'll probably split jupyterhub-singleuser out _someday_, but I think the development is still too tightly integrated with hub at this point.

Yet another option is to keep the older versions as single envs, but explicitly exclude `jupyterhub-singleuser` from the older versions (or pin older versions). The freeze machinery does already have the infrastructure for different versions to have different base envs, I believe. This same issue is going to come up eventually for old Pythons for every package, not just jupyterhub-singleuser.

PR #1199, opened 2022-10-23
Create wheels with the build package, stop calling setup.py directly

consideRatio:
I understand it as using setup.py directly is being phased out in favor of using for example `build` that I think can handle building wheels for packages with `setup.py` or `pyproject.toml`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,6 +44,13 @@ jobs:
           python -m build --sdist --wheel .
           ls -l dist
 
+      # ref: https://github.com/actions/upload-artifact#readme
+      - uses: actions/upload-artifact@v3
+        with:
+          name: repo2docker_service-${{ github.sha }}
+          path: "dist/*"
+          if-no-files-found: error
+
       # This step is only run when a new tag is pushed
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -34,14 +34,22 @@ jobs:
...
+      - name: build release
         run: |
-          python setup.py sdist bdist_wheel
+          python -m build --sdist --wheel .
+          ls -l dist
+
+      # ref: https://github.com/actions/upload-artifact#readme
+      - uses: actions/upload-artifact@v3
+        with:
+          name: repo2docker_service-${{ github.sha }}
```
minrk, line 50: Curios, why `_service` instead of just `repo2docker-`?

consideRatio: Copy paste mistake of mine related to work on providing a jupyterhub service exposing repo2docker functionality! Nice catch

consideRatio, line 50:
```suggestion
          name: repo2docker-${{ github.sha }}
```

consideRatio:
Going for a merge to rebase #1200 on top of this!

PR #1200, opened 2022-10-23
ci: stop running pre-commit in gha, rely on pre-commit.ci

consideRatio:
I've activate pre-commit.ci to run against this github repo already and it works. This PR updates the pre-commit job that also acted as a job populating a pip cache as just a job that populates the pip cache, and does so a bit more completely.

This PR includes #1199 that introduced the `build` package into dev-requirements.txt.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,47 +39,7 @@ env:
   GIT_AUTHOR_NAME: CI User
 
 jobs:
-  populate-pip-cache:
-    runs-on: ubuntu-22.04
-
-    strategy:
-      fail-fast: false
-      matrix:
-        python_version: ["3.9"]
-
-    steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
-        with:
-          python-version: "${{ matrix.python_version }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -39,7 +39,7 @@ env:
   GIT_AUTHOR_NAME: CI User
 
 jobs:
-  pre-commit:
+  populate-pip-cache:
```
minrk, line 42: Is it worth it to keep this vs remove the job? We don't have similar jobs anywhere else.

consideRatio: We could remove it, but I think I'd like for it to stay for now to avoid adding further changes while trying to get PRs triaged and resolved.

I think we could gain some seconds by not having it, but we would also burden pypi servers ~10 times more.

consideRatio:
I rebased this again on #1199 that was included before

.github/workflows/test.yml
consideRatio: Ah lets remove it!

minrk (approved):

PR #1205, opened 2022-10-26
Implement support for dockerignore and containerignore

### LINKED ISSUE #937: Excluding files from the repository
Is there a means to exclude files of the git repository from the docker image?

The use case is that of a (private) repository including solution notebooks to an assignment.

I presume this could be achieved with a post-build or so, but simply asking if we could have a configuration thing to include e.g. directories from the build (as well as the .git folder I guess).

sgaist:
Currently repo2docker creates a context object that includes the whole content of the repository it builds an image for. Thus it includes folders like .git which is usually something that has no interest in the final image, can take quite a lot of space and most importantly, kills the caching of that layer.

This patch adds support for reading `dockerignore` and `containerignore` files that are used to ensure only the relevant data are used to build the image.

By default it also excludes the .git folder if neither of these files are provided.

It makes use of docker-py's own file discovery mechanism to ensure coherence and avoid re-inventing a working wheel.

Note: `containerignore` comes from podman that reads either file. See the [podman-build documentation](https://github.com/containers/podman/blob/717a8dfb477d20e6a88157fa0716efa998d46e04/docs/source/markdown/podman-build.1.md.in).

Fixes #937
Related to #268 and #269 final discussions.


sgaist:
I just thought about something to discuss based on the default "ignore .git" behaviour introduced by this patch.

Since the goal of repo2docker is to provide a ready to use work environment with only the relevant pieces from a repo, should we have a configurable list of paths that we could ignore by default so people don't need to introduce another configuration file ?

I am currently mainly thinking of git and forge specific files and folders like `.github`, `.gitlab-ci.yml`, etc.

sgaist:
@manics I follow you on the reproducibility side but I thought that the following labels were there for that:
```
LABEL repo2docker.ref="<commit hash>"
LABEL repo2docker.repo="git@gitlab.example.org/mygroup/myprojectgit"
LABEL repo2docker.version="2022.10.0"
```
?

I just saw that `repo2docker.ref` could contain `HEAD` which does not provide a useful information for reproducibility. The information would only be retrievable if one I explicitly runs the image to go into the git clone or based on the image generated name. May I suggest that HEAD should be replaced by the corresponding hash at image build time ?

In case we keep the .git folder by default, what are you thinking about my suggestion of a new argument that can be used with a list of files/folders to ignore ?

consideRatio:
I rebased this PR by mistake, sorry for the noise!

sgaist:
@consideRatio no worries !

On that matter, something I am unsure of, it it globally preferred to rebase or to merge main back to the branch ? I am used to do the former to have a "semi-linear" history as they say in GitLab's configuration. I thought I have read something about it in the contributor documentation but I can't find it anymore.

consideRatio:
> On that matter, something I am unsure of, it it globally preferred to rebase or to merge main back to the branch ?

There is clear no consensus, but I think of two situations. When a PR is accepted an merged/rebased/squashed, and how to handle PR development before that.

- We almost always use the merge button when merging PRs in jupyterhub org to retain commit authorship as I understand it.
- I appreciate rebasing my PRs before merging them to have an easier to follow git history, but I dislike if rebasing makes a reviewer uncertain what is already reviewed once and what are new changes etc. Due to this, I'm often rebasing PRs before review efforts are put in, but not after.

sgaist:
Thanks @consideRatio, sounds good to me.

I haven't yet had a really big review on GitHub were I could see that rebasing did break the process for reviewers. Some times I have encountered policies that would be "we use merge from main" when updating a pull request but that is all. If it could help shape a consensus for the Jupyter project, we can use one of my contribution to do some testing in that regard.

sgaist:
@manics a small ping to get your opinion an my reproducibility related questions above as well as `.git` handling.

sgaist:
Having re-read #269, the ignore `.git` by default would be "out of spec" with regard to what docker does so I will remove it.

I'll open tickets for the other points as they are not directly related to this patch.

repo2docker/buildpacks/base.py
```diff
@@ -608,7 +609,33 @@ def _filter_tar(tar):
         for fname in ("repo2docker-entrypoint", "python3-login"):
             tar.add(os.path.join(HERE, fname), fname, filter=_filter_tar)
 
-        tar.add(".", "src/", filter=_filter_tar)
+        exclude = []
+
+        for ignore_file_name in [".dockerignore", ".containerignore"]:
+            if os.path.exists(ignore_file_name):
```
minrk, line 615: Let's make sure we respect the `binder` dir overrides by using `self.binder_path(ignore_file_name)`, sot he top-level file can be ignored.

minrk, line 615:
```suggestion 
            ignore_file_name = self.binder_path(ignore_file_name)
            if os.path.exists(ignore_file_name):
```

and it's worth a test to show that top-level .dockerignore is ignored if there's a `binder` directory. This can be added to `tests/venv/binder-dir`

sgaist: Thanks for pointing that out ! I had forgotten about that configurable part.

sgaist: Done

minrk (approved):

PR #1211, opened 2022-11-02
reconcile CLI/config priority

### LINKED ISSUE #1210: `--config <my-config-file>` largly ignored as defaults in argparse take precedence
If `repo2docker --config <my-config-file>` is used, boolean configuration will be ignored.

__Example configuration in `repo2docker_config.json`__

```json
{
  "Repo2Docker": {
    "run": false,
    "push": true
  }
}
```

__Example way to execute repo2docker where the configuration is ignored__

```shell
repo2docker --config=repo2docker_config.json <some repo>
```

## Background

repo2docker has traitlets based configuration and is derived from the traitlets library's Application class. It can be provided both .py and .json file configuration. See [`Application.load_config.file()`](https://traitlets.readthedocs.io/en/stable/config-api.html#traitlets.config.Application.load_config_file) and note that it is used early.

https://github.com/jupyterhub/repo2docker/blob/b6098c02f315a37d690a0c538973de53dd0631a3/repo2docker/__main__.py#L282-L282

The issue is that a config file that provides a boolean value for `Repo2Docker.run` for example will be ignored when the repo2docker CLI is parsing its flags.

https://github.com/jupyterhub/repo2docker/blob/b6098c02f315a37d690a0c538973de53dd0631a3/repo2docker/__main__.py#L329-L330

This could be resolved if there was a way in argparse to check if a flag had been passed or not, or if a default value had been provided. I can imagine an ugly workaround like having a default value of "this-flag-was-not-set" and then check for that or similar, but I'd like to avoid that.

Is there a way to let the traitlets configured defaults be respected instead of beign overruled by the argparse provided defaults?

https://github.com/jupyterhub/repo2docker/blob/b6098c02f315a37d690a0c538973de53dd0631a3/repo2docker/app.py#L326-L332

## Two sets of defaults?

When using the CLI, the defaultof `run` is True (becomes false if `--no-run` is passed), but when using Repo2Docker the application, it will be False by default. If we reconcile these to one default it would become a breaking change, either for the CLI or the library.

## What to do?

I'm looking to make use of the `repo2docker` CLI, but relying on as much as possible to be provided via the provided configuration file. Like that, a repo2docker-service that is supposed to run repo2docker can accept a config file and execute repo2docker based on that config, but if values are ignored by default values, that is a problem.

minrk:
- avoid CLI defaults clobbering values from config. Mostly by using `default=None`, and adding `if args.name is not None` checks before assigning attributes, to avoid setting any values from argparse defaults that would override values set by config.
- add inverse `--build` and `--run` flags for boolean options that could be overridden in config
- fix `--Class.trait=value` priority relative to config files (config file should be loaded before loading values from CLI)

Draft for discussion of #1210

EDIT: Closes #1210


consideRatio (commented):
Wieee thanks for the very thourough followup @minrk!!!

Before anything, can you validate my understanding that of `config=True` is correct that not having it means it can't be managed via traitlets configuration files or traitlet flags like `--Repo2Docker.run=True` any more?

I ended up trialing a strategy for a repo2docker-service to help users run repo2docker that may stop working if `config=True` now is removed on some options though. Is it reasonable to retain `config=True` to support this kind of config-only startup of repo2docker?

Practically, I made a custom Repo2Docker CLI to accept `-f/--config` and nothing more besides traitlets based flags.

```python
import argparse
import sys

from repo2docker import Repo2Docker

argparser = argparse.ArgumentParser()
argparser.add_argument(
    "--config",
    "-f",
    dest="config_files",
    action="append",
    default=[],
)
args, traitlet_args = argparser.parse_known_args(sys.argv[1:])

r2d = Repo2Docker()
for config_file in args.config_files:
    r2d.load_config_file(config_file)
r2d.parse_command_line(traitlet_args)
r2d.initialize()
r2d.start()
```

I then made use of the entrypoint script like this...

```
                    # Build entrypoint script and user provided repo2docker
                    # config is injected
                    'echo "$REPO2DOCKER_SERVICE_BUILD_ENTRYPOINT" > /tmp/build_entrypoint.py &&',
                    'echo "$REPO2DOCKER_SERVICE_REPO2DOCKER_CONFIG" > /tmp/repo2docker_config.json &&',
                    # Entrypoint is run
                    "exec python3 /tmp/build_entrypoint.py",
                    # ... with user managed config
                    "--config=/tmp/repo2docker_config.json",
                    f"--Repo2Docker.output_image_spec={build.image_name}",
                    f"--Repo2Docker.ref={build.ref}",
                    f"--Repo2Docker.repo={build.repo}",
                    # ... and builder managed config
                    "--Repo2Docker.run=False",
                    "--Repo2Docker.push=True",
                    "--Repo2Docker.user_id=65534",
                    "--Repo2Docker.user_name=nobody",
```

Maybe I could revert to a Repo2Docker native CLI based approach again though with this, but even if I did, it would feel good to be consistent and go for 100% traitlets based configuration instead of a mix.

Overall besides my wish to be able to use traitlets all the way still, this LGTM!

minrk:
>  it means it can't be managed via traitlets configuration files or traitlet flags like --Repo2Docker.run=True any more?

correct. I think that's a good thing? Avoiding more than one way to set the same value makes things easier to follow, I think. The main reason for CLI config=True is that _everything_ can be set on the CLI, even things we haven't exposed explicitly. I'd argue the current argparse approach is deliberately choosing to _select_ what's on the CLI, rather than exposing everything traitlets-style.

> Is it reasonable to retain config=True to support this kind of config-only startup of repo2docker?

For the second class (run, push, etc.), I think it is reasonable, but I don't think we should do that without _also_ making sure we have both directions for all such boolean flags. I _don't_ think it makes sense to allow the repo/ref/image to build, for instance, to be set via config files. I think the only reason to have those as `config=True` is that it is _required_ for traitlets CLI. Since we don't use traitlets CLI, that should be irrelevant, and I don't think facilitating third-party autogenerated CLI should be something we support.

Since your use case seems to be fully covered by the existing CLI, I'm not sure why generating a CLI that behaves differently is desirable.

But we can definitely go all the way back to everything-is-configurable, we just need to add more handling of each value being set via config without overriding from cli, etc.

consideRatio:
Thanks for your work on this @minrk!!! :heart: :tada: I think the changes look good and address relevant issues and are now only thinking about one aspect - the removal of `config=True` from some configuration.

Consider a service exposing repo2docker to end users via a REST API ([I'm working on one](https://github.com/consideratio/repo2docker-service)). Wouldn't it be neat if it could just accept a blob of repo2docker config in json format in order to launch a build, instead of flags or a mix of flags and config?

I'm currently planning for the repo2docker-service I work on to function like that, thinking it could be an implementation that is sustainable to maintain, document, and build upon. Perhaps even accompanied with traitlets generated JSONSchema or similar.

If you are open to retaining `config=True` for such reason, I could offload you of implementation work and push commits to take this PR all the way. I'm eager to keep developing on repo2docker-service :)

minrk:
The config-as-JSON-REST case is an interesting one. I'm definitely open to it! I'm not sure quite where I land on it. I think supporting it is to the detriment of repo2docker itself, because it means that generated documentation of config files / help output produces documented options that should never be set and may not work (the current state of things, fixed by this PR).  I think it _doesn't_ make sense to specify repo or ref in a config file, so adding some logic to the CLI to handle a config case that shouldn't ever happen doesn't feel great to me, where clearly specifying 'repo and ref cannot be set via a config file' seems like the right behavior.

None of the options I've unmarked would logically be exported in a schema of configuration for repo2docker (e.g. into a helm chart like BinderHub). That's a benefit of the change, as I see it. That's _not_ the same as a schema of inputs, which could take all _traits_ which seems it would be unaffected by this change. But those non-config options could not be loaded _via a config object_, so they would have to be handled separately, which is why I see the appeal.

Do you think that a repo2docker service should not expose inputs such as `image` and `ref` at a level more prominent than detailed config like cache_from, etc.? I think a service should fail early and prominently if those are unspecified, so they _should_ get some special handling. Your service also needs to not expose some options like `run` which should always be False for a service, so exposing it in config means you need to have special-handling to override the input config (not hard - a last, highest priority config object works fine). All of those suggest to me that it is a _benefit_ to your service to not have these exposed in config, even if it means a few more lines of implementation.

I think a REST body that is like:

```python
"image": "...",
"ref": "...",
"config": { additional_options_here }
```

is clearer than a single passthrough. WDYT?

consideRatio:
For reference, flags that no longer gets `config=True` are the following: `repo`, `ref`, `output_image_spec`, `push`, `run`, `dry_run`.

- `push` - repo2docker-service run on a single computer / k8s cluster probably want `push` coerced falsy / truthy.
- `run`, `dry_run` - repo2docker-service would probably always want to coerce them falsy
- `repo` - required by repo2docker itself
- `ref` - optional by repo2docker itself
- `output_image_spec` - optional by repo2docker itself, repo2docker-service may have validation or coercion of the value

In my mind atm, I'm thinking the ideal outcome is that these are all configurable, and a REST API could accept a pure JSON config. repo2docker-service could then perform validation and/or coercion also itself relying on JSON config. For example, users of repo2docker-service could provide their own `repo2docker_config_coerced.json`, or their own `repo2docker_config_schema.json` for what criteria should be met before and/or after config coercion.

Back to the bigger picture again: we are considering if repo2docker should retain `config=True` for everything that the CLI can configure, or if repo2docker should be opinionated on what must be explicitly passed via CLI flags/params.

> Do you think that a repo2docker service should not expose inputs such as image and ref at a level more prominent than detailed config like cache_from, etc.?

I still value repo2docker-service interacting with repo2docker in a config native way. I see it as prominent features would be prominent by repo2docker-service specific schema validation config coercion - also config native without coupling to the CLI.

> I think it doesn't make sense to specify repo or ref in a config file, so adding some logic to the CLI to handle a config case **that shouldn't ever happen** doesn't feel great to me, where clearly specifying 'repo and ref cannot be set via a config file' seems like the right behavior.

This seems like the key question: is supporting "config native execution" acceptable from repo2docker's perspective?

I'm inclined to say yes atm.

- I greatly value working with config data directly and exclusively over working with a mix of pure config and config influencing CLI flags/args.
- A downside for repo2docker is that the argparse implementation complexity grows, but at the same time I see how the consistency of the logic improves. There is no longer a need to keep track of what is given special treatment or not.

manics:
It sounds like there a conflict between using repo2docker as a backend where having everything configurable through a standard mechanism (`config=True`) is helpful, and repo2docker as an interactive tool where you want a friendlier interface. Is it worth separating these two use-cases so in the default CLI argparse has priority, and anyone wanting Traitlets to have priority uses a different entrypoint with no argparse support?

minrk:
@manics yeah, that's basically the situation we have now with this PR. Everything is still configurable so if you build a default traitlets CLI, it will expose every input via config files, CLI, etc. (this is not changed in this PR, after reverting the removal of `config=True` for options where that doesn't make sense with our own CLI). But the CLI we use for repo2docker itself has higher priority options, only one of which (repo) cannot actually be specified via config when using the CLI (because it's always overridden). Everything else can be specified by config files _and_ overridden by the CLI. So you get standard priority of CLI > config > defaults, with every option being settable at every level (except repo, which can only be set by CLI because it's a required argument). the main change is that every option in the CLI has to be recognizable as 'specified' or not, so that we don't get the situation of 'cli > cli _default_ > config,' which several options have prior to this PR (#1210)

repo2docker/__main__.py
```diff
@@ -181,6 +198,13 @@ def get_argparser():
         action="store_true",
         help="Push docker image to repository",
     )
+    argparser.add_argument(
+        "--no-push",
+        dest="push",
+        action="store_false",
+        help="Don't push docker image to repository (default).",
+    )
+    argparser.set_defaults(push=None)
```
consideRatio, line 207: `--no-clean` and `--no-push` differ in a way I don't understand, by having `argparser.set_defaults(push=None)` compared to having `default=None,`. Is this intented or does it make sense to do it the same way?

consideRatio (approved):
@minrk thank you for thinking and working on this so thoroughly, I'm sorry for being pushy about the `config=True` part and thank you for compromising for now.

I've left a comment on one part that just looked a bit odd but not wrong - this LGTM!

The PR title is outdated now right, still being in WIP?

repo2docker/__main__.py
minrk: They aren't different on purpose, I just didn't implement both directions for every boolean flag. Failing to implement both directions only means one case isn't covered: when config inverts the default for a bool, there's no flag for setting it back to the default on a per-call basis. This isn't _strictly_ required, but it seems relevant to cover for at least some flags, hence the `--[no-]run`, etc.

The reason for `argparser.set_defaults(push=None)` is that there are now _two_ arguments defining the `push` argument, so it's ambiguous which one should set the default (or maybe both need to?), so setting the default outside the two argument definitions seemed to me to be the clearest.

consideRatio: Ah, makes sense to have it set once outside both of them then - good point!

consideRatio (approved):

PR #1216, opened 2022-11-07
Create Julia projects based on binder/Project.toml when found

frankier:
Closes https://github.com/jupyterhub/repo2docker/issues/1215


frankier:
Okay I've added a test now. I'm not really sure why coverage has gone down, but maybe this can be merged anyway since it fixes a real bug and has a regression test.

consideRatio:
Nice this LGTM, but I pushed a commit to coalesce the test cases first - was that okay with you?

Our test suite is slow and each added test folder adds ~5 minutes or test execution time or so. In this case, the Julia tests ended up also being the slowest, so by coalescing the tests to one single we ended up with a reduced total duration.

frankier:
Fine by me! Yes I noticed the slowness. Neither Docker nor Julia come for free startup time wise unfortunately!

consideRatio:
It was recently taking ~50 minutes, and work to coalesce tests like this made it go to ~25 minutes :) Thanks for your understanding!

PR #1218, opened 2022-11-11
ci: cleanup no longer used test logic related to memlimit

consideRatio:
It seems that as part of #912, the memlimit files became unused. This PR removes them so that people like me don't end up confused about them or debugging why they aren't used etc.

Pinging @yuvipanda as the author of the test scripts being cleaned away from this repo.


yuvipanda:
thanks @consideRatio

PR #1234, opened 2023-01-18
Bump to Node 18

jtpio:
Similar to the previous bump: https://github.com/jupyterhub/repo2docker/pull/1172

Node 18 is now the current LTS: https://nodejs.dev/en/about/releases/

image

Also JupyterLab 4 will likely require Node 18 for those who want to to build it with `jupyter lab build` or install extensions the deprecated way with `jupyter labextension install`: https://github.com/jupyterlab/jupyterlab/pull/13722


manics:
I presume this is fine.... does anyone know of a reason not to upgrade to node 18?

minrk:
> does anyone know of a reason not to upgrade to node 18?

No, this should be fine.

jtpio:
Just rebased and refreezed to fix the conflicts.

minrk (approved):

PR #1235, opened 2023-01-20
Also install dev requirements on Gitpod

jtpio:
Gitpod is really convenient for updating packages and running `python freeze.py`. However one has to install the dev dependencies to be able to lock the environments.

This PR automatically installs the dev dependencies on Gitpod so the env is ready on startup.

This also follows the documentation for setting up a dev environment: https://repo2docker.readthedocs.io/en/latest/contributing/contributing.html#set-up-a-local-virtual-environment

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,10 +1,12 @@
 tasks:
   - init: |
       pip3 install sphinx-autobuild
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #1236, opened 2023-01-20
Update to `jupyter-resource-usage==0.7.0`

jtpio:
This should fix the following error when trying to use Jupyter Server 2 on Binder:

image

`jupyter-resource-usage==0.7.0` has support for Jupyter Server 2 now.


manics (approved):
Thanks!

Issue #1153, opened 2022-04-20
tidyverse loading difference between JupyterLab and RStudio


pierrepo:
### Bug description

The R package `tidyverse` loads successfully in JupyterLab but fails in RStudio.

We use as an example [this repository](https://github.com/plasmabio/template-r/) that contains the description of a R environment with the `tidyverse` package installed with conda.

#### Expected behaviour

The R instruction

```r
library(tidyverse)
```
should work without any error.

#### Actual behaviour

In JupyterLab, we can load `tidyverse` both in a R notebook:

![image](https://user-images.githubusercontent.com/733861/164316464-d38ae2c9-5add-4c1d-92f2-a68d7b8d205a.png)

and in a terminal:

![image](https://user-images.githubusercontent.com/733861/164316671-9d6369b7-2895-4c68-9127-a9a5c4f0acf4.png)

In RStudio, this fails:

![image](https://user-images.githubusercontent.com/733861/164316990-f7c71761-ae6b-494b-8871-64b350f6228f.png)

For debugging purpose, here is the output of the `sessionInfo()` instruction:
```r
R version 4.1.3 (2022-03-10)
Platform: x86_64-conda-linux-gnu (64-bit)
Running under: Ubuntu 18.04.6 LTS

Matrix products: default
BLAS/LAPACK: /srv/conda/envs/notebook/lib/libopenblasp-r0.3.20.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] fansi_1.0.3     utf8_1.2.2      digest_0.6.29   crayon_1.5.1   
 [5] IRdisplay_1.1   repr_1.1.4      lifecycle_1.0.1 jsonlite_1.8.0 
 [9] evaluate_0.15   pillar_1.7.0    rlang_1.0.2     cli_3.2.0      
[...]

PR #1237, opened 2023-01-22
Add Conda env library path to RStudio configuration

TimStewartJ:
Currently RStudio packages installed in the environment do not show up in RStudio. The current workaround is to set the `LD_LIBRARY_PATH` environment variable in the `start` file. This change applies a similar fix during the build, adding the proper library path to `rserver.conf`.

Relevant discussion can be found in #1153 and here:

https://discourse.jupyter.org/t/glibcxx-3-4-26-not-found-from-rstudio/7778/5

### DIFF
--- a/repo2docker/buildpacks/conda/__init__.py
+++ b/repo2docker/buildpacks/conda/__init__.py
@@ -396,6 +396,7 @@ class CondaBuildPack(BaseImage):
                     echo auth-none=1 >> /etc/rstudio/rserver.conf && \
                     echo auth-minimum-user-id=0 >> /etc/rstudio/rserver.conf && \
                     echo "rsession-which-r={env_prefix}/bin/R" >> /etc/rstudio/rserver.conf && \
+                    echo "rsession-ld-library-path={env_prefix}/lib" >> /etc/rstudio/rserver.conf && \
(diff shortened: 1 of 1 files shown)


repo2docker/buildpacks/conda/__init__.py
```diff
@@ -396,6 +396,7 @@ def get_env_scripts(self):
                     echo auth-none=1 >> /etc/rstudio/rserver.conf && \
                     echo auth-minimum-user-id=0 >> /etc/rstudio/rserver.conf && \
                     echo "rsession-which-r={env_prefix}/bin/R" >> /etc/rstudio/rserver.conf && \
+                    echo "rsession-ld-library-path={env_prefix}/lib" >> /etc/rstudio/rserver.conf && \
```
yuvipanda, line 399: Looking at the docs in https://docs.posit.co/ide/server-pro/1.0.25/r-sessions.html (ctrl-f 'rsession-ld-library-path'), this *adds* rather than replaces elements in `LD_LIBRARY_PATH` so this is good to go it think.

yuvipanda (approved):

yuvipanda:
Thanks @TimStewartJ and apologies for the late review!

TimStewartJ:
No worries! Thank you!

PR #1239, opened 2023-02-03
Python 3.11 / JupyterHub 3, take 3

### LINKED ISSUE #741: Fresh Python kernels for reliable builds.
Programs managing python packages made available through configuration files with associated _lock files_, will probably regularly run into issues with conflicting dependencies that cannot be resolved. An example of such conflict arose around `certifi` (#725), a package that could not be successfully uninstalled by `pip` to be upgraded to the pinned version by the lock file that typically ends up doing these things.

I suggested a primitive workaround in #726, but this workaround had no guarantees to work and would certainly fail in a very complicated manner if it did fail. In short, it would pretend like the package was uninstalled before installing the new package, even though it wasn't, and override old files.

The most reliable solution I think, with performance drawbacks, is to always provide a fresh Python environment independent of the notebook server.

### LINKED ISSUE #956: Switch to using jupyter_server to launch singleuser servers
### Proposed change
This is a follow-up from https://github.com/jupyterhub/repo2docker/issues/945

`jupyter_server` is a new way of launching singleuser apps such as Jupyter notebook and JupyterLab. Pre-releases of JupyterLab 3 used this new class but they [failed to launch with repo2docker](https://github.com/jupyterhub/repo2docker/issues/945). The ability to run JupyterLab using the old method was restored, and JupyterLab 3.0.0b3 now works with repo2docker.

Repo2docker should be updated to work with the new method.

### Alternative options
If we do nothing JupyterLab 4 will fail to launch.

Existing users. Also anyone who want to launch an alternative singleuser server on repo2docker.

I don't know, perhaps someone who understands what's changed could outline what needs to be done?

### Who can do this work?
See previous

### LINKED ISSUE #1195: Supporting JupyterHub singleuser 3.0.0
I had a look at what's required for this to bump `jupyterhub-singleuser` to `3.0.0` (https://github.com/jupyterhub/binderhub/pull/1544#issuecomment-1278971454).

JupyterHub 3.0.0 requires Python 3.7+

## Python 3.7, 3.8, 3.9, 3.10 environments
This is easy, just update
https://github.com/jupyterhub/repo2docker/blob/d7be04efb590a2a1227614ccaf9b504ad9333470/repo2docker/buildpacks/conda/environment.yml#L11
and run [`freeze.py`](https://github.com/jupyterhub/repo2docker/blob/d7be04efb590a2a1227614ccaf9b504ad9333470/repo2docker/buildpacks/conda/freeze.py) to update Python 3.7, 3.8, 3.9 and 3.10 environments.

## Python 2.7 environment

Also easy, since the kernel environment is decoupled from the notebook/JupyterLab environment.

## Python 3.5 environment
Not automatically updated:
https://github.com/jupyterhub/repo2docker/blob/d7be04efb590a2a1227614ccaf9b504ad9333470/repo2docker/buildpacks/conda/environment.py-3.5.yml#L1-L2
Stuck on a very old version, too old for 3.0.0:
https://github.com/jupyterhub/repo2docker/blob/d7be04efb590a2a1227614ccaf9b504ad9333470/repo2docker/buildpacks/conda/requirements.py-3.5.pip#L8

## Python 3.6 environment
Not automatically updated:
https://github.com/jupyterhub/repo2docker/blob/d7be04efb590a2a1227614ccaf9b504ad9333470/repo2docker/buildpacks/conda/environment.py-3.6.yml#L1-L2
Stuck on an old version, too old for 3.0.0:
https://github.com/jupyterhub/repo2docker/blob/d7be04efb590a2a1227614ccaf9b504ad9333470/repo2docker/buildpacks/conda/environment.py-3.6.lock#L131

## Options

- Separate Python 3.5 and 3.6 environments from the notebook/JupyterLab environment, similar to Python 2.7
- Separate all Python environments from the notebook/JupyterLab environment. If we're going to migrate 3.5 and 3.6 environments to the decoupled setup, it probably makes sense to do it for all Python versions to make things easier in the future, and also solves some potential package conflicts: https://github.com/jupyterhub/repo2docker/issues/741
- Separate all environments for all languages from the notebook/JupyterLab environment. Architecturally I think this makes sense, but is more work: https://github.com/jupyterhub/repo2docker/issues/868

minrk:
continuation of #1196, implementing [this suggestion](https://github.com/jupyterhub/repo2docker/pull/1196#issuecomment-1299774835).

Changes:

- Python 3.11 env is added
- JupyterHub is updated to 3.1.1
- JupyterHub/notebook are no longer installed via pip in the Python 3.5 env, since it's kernel-only (updating 3.6 is hard, since we only have a lockfile!)
- separate kernel env logic is generalized to a `.separate_kernel_env` flag, instead of `.py2` special case, currently triggered by `python < 3.7` and easy to update in the future.

In testing likely user surprise, I ran a test with `postBuild` and Python 3.6, and noticed that $NB_PYTHON_PREFIX comes _before_ $KERNEL_PYTHON_PREFIX on PATH. I know this is intentional, but it has consequences, because it decreases the reproducibility of certain postBuild actions (likely the _least_ stable part of repo2docker) because things like `pip` and `jupyter` refer to the _server_ environment, though `conda install` is managed to install in the _kernel_ environment. But changing it could mean common postbuild actions like `jupyter labextension build` stop working because they are run against the wrong env. So there's not an obvious solution.

I think _probably_ a better and more expected experience for most users would be for $KERNEL_PYTHON_PREFIX to be first on PATH, and the server (when it's different) to be last. But we also need to make sure kernel-first doesn't cause problems when we are trying to launch the server, which is kinda the main thing we do here. I'd _like_ to save that for a later discussion, if we can, but I understand if folks would rather solve it. As it is now, this PR is _preserving_ behavior of server-first, only raising the question of whether that _should_ change, not changing it.

I'd rather unblock long-delayed JupyterHub and Python updates soon, which unfortunately requires a solution to this 3.5/3.6 problem, and I think this is the best compromise.

- closes #741
- closes #1195
- closes #956
- closes #1196


consideRatio:
Thank you @minrk for working this! This LGTM from what I can tell but I struggle with understanding the consequences of things in repo2docker. There are consequences of staying behind as well though and would very much like to get this merged due to them.

I've seen some test failures, so I restarted the test run in the PR and started new one in the master branch as well to compare. Still waiting for the outcome.

- [this PRs Test workflow execution](https://github.com/jupyterhub/repo2docker/actions/runs/4166516433/jobs/7210949331)
- [main branch Test workflow execution](https://github.com/jupyterhub/repo2docker/actions/runs/4166439606)

yuvipanda:
This looks ok to me, as it primarily just adopts the mechanism we have for py2/3 and updates it to be 'is python new enough?'. I am not sure about the failing tests but otherwise good to me!

minrk:
Most of the failures were due to updated expectations of py35/36, expecting kernel Python to be first on $PATH, which is not the case when the two are separate. I've updated the tests to match the new behavior (explicitly test with kernel Python, rather than `/usr/bin/env python` which is _server_ python).

The Julia failure is not merely one of test expectation, because Julia 0.6 pins Python 3.5 due to compatibility issues between the associated old version of IJulia and updates to jupyter-core. I've set the $JUPYTER env to the kernel env when running with 0.6, which seems to fix the problem (I ran the test, as well as manually building the repo and launching a Julia notebook, which worked).

0.6 is also still our _default_ Julia (when used with REQUIRE), which has been outdated now since the release of 0.7, _in 2018_. I don't want to change defaults in this PR, but maybe we should drop support for Julia 0.6 and/or deprecate REQUIRE altogether.

I think it all works now (we'll see what CI says).

minrk:
one last typo. optimistic 🤞 for CI this time

consideRatio (approved):
:heart: :tada: nice work!!! I've reviewed the recent changes, and they LGTM!

I understand this as possibly breaking, but not so much, and that we don't overview the various situations so well. I figure its mainly an issue for users of an image with jupyterhub if so, where things starts working better if the server is jupyterhub 3 based suddenly, while it may also degrade if the jupyterhub is jupyterhub 1 based for exmaple?

I suggest we call this breaking.

minrk:
I think it's all set from my perspective.

It's definitely breaking for Python 3.5/3.6 repos. Also, I guess the JupyterHub version upgrade is technically breaking, too, for folks using r2d to build JupyterHub images. Python 2.7, 3.7-3.9 users shouldn't see a difference.

PR #1245, opened 2023-02-12
Add optional registry credentials to push()

manics:
This is to help with setting dynamic registry credentials when pushing in BinderHub

Example using traitlets on the command line:

`repo2docker --push --no-run --image-name docker.io/USERNAME/image:test --ContainerEngine.registry_credentials=registry=docker.io --ContainerEngine.registry_credentials=username=USERNAME --ContainerEngine.registry_credentials=password=PASSWORD https://github.com/binderhub-ci-repos/cached-minimal-dockerfile`

Example using environment variable:
`CONTAINER_ENGINE_REGISTRY_CREDENTIALS='{"registry":"docker.io","username":"USERNAME","password":"PASSWORD"}' repo2docker --push --no-run --image-name docker.io/USERNAME/image:test https://github.com/binderhub-ci-repos/cached-minimal-dockerfile`

### DIFF
--- a/repo2docker/docker.py
+++ b/repo2docker/docker.py
@@ -120,6 +120,8 @@ class DockerEngine(ContainerEngine):
         return Image(tags=image["RepoTags"], config=image["ContainerConfig"])
 
     def push(self, image_spec):
+        if self.registry_credentials:
+            self._apiclient.login(**self.registry_credentials)
         return self._apiclient.push(image_spec, stream=True)
 
     def run(
--- a/repo2docker/engine.py
+++ b/repo2docker/engine.py
@@ -2,8 +2,11 @@
 Interface for a repo2docker container engine
 """
 
+import json
+import os
 from abc import ABC, abstractmethod
 
+from traitlets import Dict, default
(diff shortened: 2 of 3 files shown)


minrk:
LGTM! My only question, when thinking about https://github.com/jupyterhub/binderhub/pull/1637 / secrets, etc. is: is there value in making it easier to pass _just_ the password via a distinct variable? i.e. in the token case, not requiring rebuilding the container dictionary, but pass the single key that actually changes each time in an environment variable? If it doesn't make a difference, then this looks AOK to me.

minrk (approved):

manics:
I can see the value in a separate env var for just the password, the decision is whether to add it now, or leave it for when there's demand?

Probably the main decision is whether passing the registry parameters as a JSON blob is acceptable, or whether a more structured approach with separate `registry:str` `username:str` `password:str` properties is better. I like the flexibility of the JSON blob, and it means a separate password field can be added later as a non-breaking change, but I'm also happy to switch to the individual fields if that's preferred.

PR #1248, opened 2023-02-16
avoid duplicate log statements by memoizing getters

### LINKED ISSUE #1246: duplicate logs about environment
### Bug description

When building an environment, the logs about the Python version and buildpack display multiple times because `get_build_script_files()` ultimately gets called multiple times:

```
python -m repo2docker tests/pipfile/py36/
Using local repo tests/pipfile/py36/.
Building conda environment for python=3.6
User-requested packages for legacy Python version 3.6 will be installed in a separate kernel environment.
Using PipfileBuildPack builder
Building conda environment for python=3.6
User-requested packages for legacy Python version 3.6 will be installed in a separate kernel environment.
Building conda environment for python=3.6
User-requested packages for legacy Python version 3.6 will be installed in a separate kernel environment.
```

#### Expected behaviour

Log statements about selecting the environment version and other warnings should occur only once.

#### Actual behaviour

Log statements occur 2-3 times, depending on the buildpack and version
### How to reproduce

1. `python -m repo2docker tests/pipfile/py36/`

#### Propose solution

I see two options:

1. move these log statements to somewhere that is guaranteed to be called only once. Maybe `detect()`?
2. ensure these methods currently being called multiple times are called only once (i.e. remove calls like [this](https://github.com/jupyterhub/repo2docker/blob/1e61d4f1aeec9a6c0e1dc2fedd237f9002e2c36c/repo2docker/buildpacks/conda/__init__.py#L57))
3. memoize the functions with `lru_cache`, so they only execute once, even if they are called multiple times. This might be the simplest, and should do the right thing.

minrk:
ensures `get_` functions are called only once, avoiding duplicate log statements. Closes #1246 .

This may be too clever to do it magically on methods that start with `get_`, but it works and is the _simplest_ if we want to clearly always cache all these methods (I think we do).

The alternatives are:

1. more standard, far more changes (both now and over time) - apply `@lru_cache()` on every method, including every subclass method override (easy to miss, as the only downside of forgetting it is duplicate log statements and a negligible amount of wasted time)
2. to ensure these methods are called only once on the parent class, e.g. by assigning the results to variables, or calling directly via an additional private method `self._get_...`. This is ultimately the same as the memoize here, but implemented with explicit private methods instead of caching all public methods based on their name.
3. like 2., but only on the functions we expect to call more than once (`get_..._files()` are called in build and again in render)

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -7,6 +7,7 @@ import string
 import sys
 import tarfile
 import textwrap
+from functools import lru_cache
 
 import escapism
 import jinja2
@@ -230,6 +231,18 @@ class BuildPack:
                 "support is experimental in repo2docker."
             )
         self.platform = ""
+        self._memoize()
+
+    def _memoize(self):
+        """Memoize `get_foo` methods with lru_cache()
+
+        Avoids duplicate log statements when calling what should be idempotent methods more than once
+        """
+        for method_name in dir(self):
+            if method_name.startswith("get_"):
+                method = getattr(self, method_name)
+                if callable(method):
+                    setattr(self, method_name, lru_cache()(method))
 
     def get_packages(self):
         """
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -232,7 +232,7 @@ class BuildPack:
             )
         self.platform = ""
 
-    @lru_cache
+    @lru_cache()
     def get_packages(self):
         """
         List of packages that are installed in this BuildPack.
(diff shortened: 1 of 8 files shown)


yuvipanda (changes requested):
While this is super clever, this makes me somewhat uncomfortable - I can imagine myself up at 3am some day, looking for reasons why my `get_somethingelse` method is just not being called. I'd much prefer alternative 1, with lru being decorated for each of them...

minrk:
Fair enough! Now `@lru_cache` is applied to all versions

yuvipanda:
Thanks @minrk

PR #1249, opened 2023-02-17
Ensure `BuildPack.python_version` is specified

### LINKED ISSUE #1243: There's more than one way to get the default Python version
### Proposed change

Make sure there's only one path to a 'default' python version when Python is unspecified.

Currently CondaBuildPack allows `python_version` to be an empty string, while PythonBuildPack does not.

### Alternative options

Leave it as-is, which means more work to update the default Python version (missed in #1219, suggesting it's a problem).

Maintainers updating the default Python version, users trying to understand how a default Python is chosen

Not too much, I think. It should mostly involve ensuring a Python version is set to a default value, rather than allowing it to be an empty string. It should be:

1. remove the `environment.lock` that lacks a Python version, and
2. whenever `python_version` would be empty, follow `major_pythons.get["3"]`, as is done [here](https://github.com/jupyterhub/repo2docker/blob/12f94dc69e58e7d17b273d49334b74a02a00d39f/repo2docker/buildpacks/python/__init__.py#L25), but not currently in every case
3. assert/assume that python_version is never empty once resolved

this will allow us to assume/ensure that `python_version` is never empty.

### Who can do this work?

Anyone who wants to read through the Python logic of the base conda buildpack and freeze.py.

minrk:
- removes lockfile copies without Python version number
- enforces that `buildpack.python_version` is always specified (major_pythons['3'] in cases where it could have been falsy before)
- warns when Python version is unspecified, which ensures future reproducibility failures

closes #1243


minrk:
I don't trust the coverage report (and don't think we should have any coverage failure conditions), because it [says](https://app.codecov.io/github/jupyterhub/repo2docker/commit/66b8765c6993ddfb2ed61d6154862246045880d9/blob/repo2docker/buildpacks/conda/__init__.py#L178) `if self.separate_kernel_env` is never True, but it clearly is in e.g. `tests/conda/py2`.

yuvipanda:
good enough for me, @minrk :)

PR #1269, opened 2023-04-26
BLD: FROM docker.io/buildpack-deps:bionic for manics/repo2podman#53

westurner:
- https://github.com/manics/repo2podman/issues/53
  - [ ] #1269, at least

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -14,7 +14,7 @@ import jinja2
 
 # Only use syntax features supported by Docker 17.09
 TEMPLATE = r"""
-FROM buildpack-deps:bionic
+FROM docker.io/buildpack-deps:bionic
 
 # Avoid prompts from apt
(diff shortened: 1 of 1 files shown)


repo2docker/buildpacks/base.py
```diff
@@ -14,7 +14,7 @@
 
 # Only use syntax features supported by Docker 17.09
 TEMPLATE = r"""
-FROM buildpack-deps:bionic
+FROM docker.io/buildpack-deps:bionic
```
manics, line 17:
```suggestion
FROM docker.io/library/buildpack-deps:bionic
```
I thought `/library/` was always required when using the full name, but the tests seem to be passing without. However https://github.com/docker-library/official-images links to https://hub.docker.com/u/library

yuvipanda:
Fixed now with https://github.com/jupyterhub/repo2docker/pull/909

PR #1271, opened 2023-05-18
[MRG] Improve docs on Dockerfile escape hatch

MattF-NSIDC:
I was struggling to find how to get the generated Dockerfile contents. When I found it, I thought I'd open a PR to add a link from the first place I looked for the next person who struggles :)

### DIFF
--- a/docs/source/config_files.rst
+++ b/docs/source/config_files.rst
@@ -245,5 +245,9 @@ With Dockerfiles, a regular Docker build will be performed.
 .. note::
     If a Dockerfile is present, all other configuration files will be ignored.
 
+If you're working on an existing project and need the `Dockerfile` escape
+hatch, you can :ref:`view the generated Dockerfile
+<usage-debugging-with-debug-and-no-build>`.
+
 See the `Advanced Binder Documentation <https://mybinder.readthedocs.io/en/latest/tutorials/dockerfile.html>`_ for
 best-practices with Dockerfiles.
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -29,9 +29,9 @@ default_role = "literal"
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
 templates_path = ["_templates"]
 extlinks = {
-    "issue": ("https://github.com/jupyterhub/repo2docker/issues/%s", "Issue #"),
-    "pr": ("https://github.com/jupyterhub/repo2docker/pull/%s", "PR #"),
-    "user": ("https://github.com/%s", "@"),
+    "issue": ("https://github.com/jupyterhub/repo2docker/issues/%s", "Issue #%s"),
+    "pr": ("https://github.com/jupyterhub/repo2docker/pull/%s", "PR #%s"),
+    "user": ("https://github.com/%s", "@%s"),
 }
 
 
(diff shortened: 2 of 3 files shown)


MattF-NSIDC:
I experienced some unexpected behavior and added a warning to the docs :point_up:

MattF-NSIDC:
I feel like more documentation is needed. When I create a `Dockerfile` following these instructions and attempt to build, there are errors because the Dockerfile references paths that aren't in the build context:

```
Step 26/60 : COPY --chown=1002:1002 build_script_files/-2fhome-2fmfisher-2f-2eminiconda3-2fenvs-2frepo2docker-2flib-2fpython3-2e11-2fsite-2dpackages-2frepo2docker-2fbuildpacks-2fconda-2factivate-2dconda-2esh-cf5cc0 /etc/profile.d/activate-conda.sh
COPY failed: file not found in build context or excluded by .dockerignore: stat build_script_files/-2fhome-2fmfisher-2f-2eminiconda3-2fenvs-2frepo2docker-2flib-2fpython3-2e11-2fsite-2dpackages-2frepo2docker-2fbuildpacks-2fconda-2factivate-2dconda-2esh-cf5cc0: file does not exist(repo2docker)
```

When I remove those steps, later steps fail because `mamba` and `conda` are both unavailable. What else can I do to document the process of switching from a non-Dockerfile config to a Dockerfile-based config?

manics:
The debug output is just for debugging, it's not intended to be buildable since you also need the full context that's created on the fly, and I don't think we should recommend it as a way to create a Dockerfile.

I've got a playground/proof-of-concept in https://github.com/manics/repo2shellscript if you're interested in seeing the full complexity.

MattF-NSIDC:
Thanks @manics! Is there anything that can be documented about migrating from a config-file based build to a Dockerfile-based build? Does `repo2docker` provide any value add for that use case? I'm struggling with a [build that has broken man pages](https://github.com/CryoInTheCloud/hub-image/issues/63), and I'm not seeing a `repo2docker` path to fixing it right now (since I think I need to do stuff as root to fix the issue).

MattF-NSIDC:
https://github.com/jupyterhub/repo2docker/issues/700

MattF-NSIDC:
Here's how I solved my problem: https://github.com/CryoInTheCloud/hub-image/pull/65

As you know, I totally missed the "Can I use repo2docker to bootstrap my own Dockerfile?" FAQ on my first time through :laughing: I've added a link to it where I would have seen it my first time through and replaced my bad suggestions with hopefully better ones, in addition to a new FAQ to resolve the specific problem with `man` pages.

docs/source/config_files.rst
```diff
@@ -240,10 +240,13 @@ In the majority of cases, providing your own Dockerfile is not necessary as the
...
 
-With Dockerfiles, a regular Docker build will be performed.
+Try the ``--appendix`` :ref:`CLI flag <usage-cli>` as a last resort. If you
+still can't achieve your goals, you should create a Dockerfile from scratch to
+define your image.
 
 .. note::
-    If a Dockerfile is present, all other configuration files will be ignored.
+   With Dockerfiles, a regular Docker build will be performed. All other
+   configuration files will be ignored.
```
MattF-NSIDC, line 249: @manics I'd like to add a sentence here clarifying what value `repo2docker` is providing to the user in this case. e.g. `All other configuration files will be ignored, but \`repo2docker\` will still ___`? Thoughts?

Thanks for your input on this change :)

PR #1281, opened 2023-06-05
Recreate changelog as markdown

yuvipanda:
github-activity, used for other JupyterHub release processes, now outputs markdown. This PR moves the existing changelog to use markdown, and provides a more complete list of changes too.

Ref https://github.com/jupyterhub/repo2docker/issues/1280


minrk:
Looks like most of the PR classification was lost in the regenerated changelog because I'm guessing they were grouped by hand instead of labeled. Maybe [rst2myst](https://pypi.org/project/rst-to-myst/) would give a better starting point?

manics:
This is what I use for [converting the output of github-activity to RST](https://github.com/jupyterhub/repo2docker/pull/1194):
```
github-activity --target jupyterhub/repo2docker --since 2022.02.0  | sed -r \
  -e 's%\[#([0-9]+)\]\(https://github.com/jupyterhub/repo2docker/pull/[0-9]+\)%:pr:`\1`%' \
  -e 's%\(\[\@([A-Za-z0-9-]+)\]\(https://github.com/[A-Za-z0-9-]+\)\)%by :user:`\1`%'
```
Perhaps you can try reversing the sed commands?

minrk:
I don't even think we need to do that - the directives still exist and work, and rst2myst handles the syntax translation correctly.

consideRatio:
Thanks for working this @yuvipanda!!! I recall thinking I'd help drive the release effort in this repo once only to run into the "oh no... its a .rst changelog" and loosing momentum to get it done.

I'd rather retain the changelog as it but with syntax conversion than to re-generate it and adopt github-activity going onwards, but I figure its also okay to re-generate things but I'd appreciate being able to overview changes to the changelogs content if that is done.

I figure if a changelog is regenerated, we can have changes based on:
- PRs that was categorized manually rather than based on the labels
- PRs that was omitted because they were automated things that we perceived as too noisy, perhaps like weekly pre-commit updates

Btw for reference I wrote some notes on how I go about generating changelogs using github-activity in https://github.com/jupyterhub/team-compass/issues/563, it includes the procedure of generating a changelog and then going to the PRs not labelled to label then, and then re-generating the changelog etc to get them categorized.

minrk:
I pushed the rst2myst changelog.md here, which should be the simplest change. No manual edits.

consideRatio (approved):

PR #1283, opened 2023-06-08
Relax pins to major versions and refreeze, introduce explicit jupyter_server v1 pin

consideRatio:
I figured we should go for a refreeze, but I saw that we were pinning to the patch versions of various packages in [repo2docker/buildpacks/conda/environment.yml](https://github.com/jupyterhub/repo2docker/blob/main/repo2docker/buildpacks/conda/environment.yml) which is too conservative I think.

In this PR I unpin minor and patch versions in that file, and refreeze. I think this is critical for the sustainability of maintenance.

Practically, I think its very sustainable and relevant for us to review bumps in the major version of all of these separately, and okay to hold back on bumping a major version until we do. But I don't think its a good practice or sustainable to review and bump minor/patch versions manually in the same way.

If we retain the minor/patch version pining, my process before refreeze would be to bump these manually in a commit, practically working around having them pinned. If I wouldn't do that, I fear that we would only do so because someone explicitly requests a version bump to get some new feature, or that something has broken because of a transient dependency was updated that needs a newer version for compatibility.

---

R test failures stems from #1284 and ought to be unrelated as they also show up in main branch.


repo2docker/buildpacks/conda/environment.yml
```diff
@@ -4,9 +4,10 @@ dependencies:
...
-  - jupyterlab==3.4.8
-  - jupyterhub-singleuser==3.1.1
-  - notebook==6.4.12
+  - ipywidgets==8.*   # https://github.com/jupyter-widgets/ipywidgets
+  - jupyter-offlinenotebook==0.2.*   # https://github.com/manics/jupyter-offlinenotebook
+  - jupyter-resource-usage==0.7.*   # https://github.com/jupyter-server/jupyter-resource-usage
+  - jupyter_server==1.*   # https://github.com/jupyter-server/jupyter_server
+  - jupyterhub-singleuser==3.*   # https://github.com/jupyterhub/jupyterhub
+  - jupyterlab==3.*   # https://github.com/jupyterlab/jupyterlab
+  - notebook==6.*   # https://github.com/jupyter/notebook
```
consideRatio, line 13: These are teh changes to review in this PR. We go from various minor/patch versions to the latest available in the same major version.

jupyter-server was pinned explicitly, as it was indirectly pinned by jupyterlab 3.4 which when upgraded to jupyterlab 3.5+ wasn't indirectly pinned any more, and that broke tests we had and would been a significant change we want to do intentionally in a separate PR.

consideRatio: R test failures shows up in the main branch as well, and tracking that is done now in https://github.com/jupyterhub/repo2docker/issues/1284. I believe them to be entirely unrelated from this PRs changes.

consideRatio:
Thank you @yuvipanda!!!!!!

PR #1285, opened 2023-06-09
Microsoft killed MRAN, stop relying on it

### LINKED ISSUE #1284: R tests broken - mran.microsoft.com retired
Sometime between between May 31 and Jun 4, our test suite [started failing for the R buildpack](https://github.com/jupyterhub/repo2docker/actions/runs/5168625028/jobs/9310336319#step:6:94).

```
No snapshot found for 2016-01-03 or 7 days prior in mran.microsoft.com
```

I figure its a regression caused by mran.microsoft.com no longer providing snapshots that old.

Or ooops, based on https://mran.microsoft.com/timemachine, its worse, its being shut down.

image

The banner links to [this announcement](https://techcommunity.microsoft.com/t5/azure-sql-blog/microsoft-r-application-network-retirement/ba-p/3707161).

> After January 31, 2023, we will no longer maintain the [CRAN Time Machine snapshots](https://cran.microsoft.com/snapshot/). Existing snapshots will be available until June 30, 2023.

> **_Recommended Action:_**
>You should plan to migrate from the use of Microsoft R Application Network website and CRAN Time Machine to avoid any interruptions. The CRAN Time Machine functionality can be replaced by using the miniCRAN R package as [documented](https://learn.microsoft.com/en-us/sql/machine-learning/package-management/create-a-local-package-repository-using-minicran?view=sql-server-ver16). The use of “[Explore Packages](https://mran.microsoft.com/packages)” functionality from MRAN can be replaced by [CRAN Task Views](https://cran.r-project.org/web/views/).

minrk:
- remove cutoff for picking RSPM, always use RSPM
- remove now-redundant r4.0-mran test
- shift base R 3.6 snapshot date to after December 2018, which seems to be the first snapshot with IRKernel available
- Provide an error message when people have too old snapshot dates

closes #1284


yuvipanda:
Or maybe the earliest version of IRKernel available in posit does not support 3.6?

yuvipanda:
Looking at https://packagemanager.posit.co/client/#/repos/2/packages/IRkernel, the earliest snapshot of IRkernel was 2018 november, so am trying to set the failed snapshot to that date to see what happens.

If this passes, I think we can add a conditional that does the following:

1. If the snapshot date is < 2018-11-17, we install IRKernel from a snapshot that's set to 2018-11-17, as that is when IRkernel was first available
2. If snapshot date is < 2017-10-10, we just error out, as posit package manager does not have snapshots before that.

yuvipanda:
Alternatively, we can just error out if snapshot date is < 2018-11-17.

yuvipanda:
The version of IRKernel from that snapshot is definitely compatible, as I see in the `DESCRIPTION` file `Depends: R (>= 3.2.0)`

minrk:
@yuvipanda thanks for fixing! Go ahead and merge if you think it looks okay.

yuvipanda:
@minrk what do you think of just erroring out if the requested snapshot date is too old?

minrk:
If we know it won't work and we can make the error clearer, that sounds fine to me.

minrk:
Since the failure case doesn't change, it's an improvement of the communication _when_ it fails, I'm also okay with merging this as-is and improving the error message in a separate PR.

yuvipanda:
@minrk I've now added a useful error message and a unit test.

PR #1287, opened 2023-06-13
Upgrade base image from to Ubuntu 22.04 from 18.04

### LINKED ISSUE #1258: Update  to a more recent version : repo2docker/buildpacks/base.py l:16 FROM buildpack-deps:bionic
### Proposed change

Switch to a more recent default buildpack than bionic (5Years old)

### Alternative options

For compatibility/legacy  reason, Allow selection of the buildpack

Proposed change would impact the default buildpack, so i guess a lot of peoples ?

Easy to add, hard to test.

### Who can do this work?

Simple change, it s only a few charcters to change.

Then there's a need to test it
- Not sure if the project has non regression tests, in that case it's simple
- If there's no non regression test, then need to find relevant test cases.

yuvipanda:
18.04 was EOL May 2023, with only paid support now available (https://ubuntu.com/blog/18-04-end-of-standard-support).

Let's bump the default base image to 22.04. This might break some repos that have apt.txt based packages, but I don't really think there's a way around that. If specific package pinning is needed, users unfortunately need to use a Dockerfile.

Closes #1258

### DIFF
--- a/docs/source/howto/base_image.md
+++ b/docs/source/howto/base_image.md
@@ -27,7 +27,7 @@ There are **no guarantees that repositories will behave the same way as other re
 For example these are two scenarios that would make your repositories non-reproducible:
 
 - **Your base image is different from `Ubuntu:jammy`.**
-  If you change the base image in a way that is different from repo2docker's default (the Ubuntu `bionic` image), then repositories that **you** build with repo2docker may be significantly different from those that **other** instances of repo2docker build (e.g., those from [`mybinder.org`](https://mybinder.org)).
+  If you change the base image in a way that is different from repo2docker's default (the Ubuntu `jammy` image), then repositories that **you** build with repo2docker may be significantly different from those that **other** instances of repo2docker build (e.g., those from [`mybinder.org`](https://mybinder.org)).
 - **Your base image changes over time.**
   If you choose a base image that changes its composition over time (e.g., an image provided by some other community), then it may cause repositories build with your base image to change in unpredictable ways.
   We recommend choosing a base image that you know to be stable and trustworthy.
--- a/docs/source/howto/base_image.md
+++ b/docs/source/howto/base_image.md
@@ -7,7 +7,7 @@ To do so, use the `base_image` traitlet when invoking `repo2docker`.
 Note that this is not configurable by individual repositories, it is configured when you invoke the `repo2docker` command.
 
 ```{note}
-By default repo2docker builds on top of the `buildpack-deps:bionic` base image, an Ubuntu-based image.
+By default repo2docker builds on top of the `buildpack-deps:jammy` base image, an Ubuntu-based image.
 ```
 
 ## Requirements for your base image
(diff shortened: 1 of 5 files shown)


manics:
How about we make an announcement https://discourse.jupyter.org/tag/announcement before merging/deploying this to mybinder.org, to manage expectations?

consideRatio (approved):
LGTM, it would be good to get a notice in the changelog as well under the unreleased section about this breaking change.

I've added a label about it breaking for now

rgaiacs:
> I wonder if we can and should announce this via mybinder.org directly

It is a great idea.

manics:
> I wonder if we can and should announce this via mybinder.org directly instead of discourse if its to be announced

How about both? We're using Discourse as the support forum and questions about binder are often handled by non-jupyterhub community members, so I think it's polite to give everyone advance warning.

minrk:
Both sound great. I can post to discourse.

consideRatio:
I opened https://github.com/jupyterhub/mybinder.org-deploy/issues/2663 about not using dev releases in mybinder.org, allowing us to link out to a repo2docker's changelog etc in announcements independently if they are on made on mybinder.org or discouse, where we anyhow would need to describe breaking changes.

minrk:
Announcement [on the forum](https://discourse.jupyter.org/t/repo2docker-and-mybinder-org-upgrading-base-image-to-ubuntu-22-04/19865)

minrk:
and banner for mybinder.org: https://github.com/jupyterhub/mybinder.org-deploy/pull/2665

yuvipanda:
Conversation in https://github.com/jupyterhub/mybinder.org-deploy/issues/2663 has come to a conclusion to continue current practice. I think there's been enough lead time now! Anyone wanna hit the merge button? :)

docs/source/howto/base_image.md
```diff
@@ -26,7 +26,7 @@ Changing the base image may have an impact on the reproducibility of repositorie
 There are **no guarantees that repositories will behave the same way as other repo2docker builds if you change the base image**.
 For example these are two scenarios that would make your repositories non-reproducible:
 
-- **Your base image is different from `Ubuntu:bionic`.**
+- **Your base image is different from `Ubuntu:jammy`.**
   If you change the base image in a way that is different from repo2docker's default (the Ubuntu `bionic` image), then repositories that **you** build with repo2docker may be significantly different from those that **other** instances of repo2docker build (e.g., those from [`mybinder.org`](https://mybinder.org)).
```
consideRatio, line 30:
```suggestion
  If you change the base image in a way that is different from repo2docker's default (the Ubuntu `jammy` image), then repositories that **you** build with repo2docker may be significantly different from those that **other** instances of repo2docker build (e.g., those from [`mybinder.org`](https://mybinder.org)).
```

consideRatio (approved):
I locally merged this on my computer and searched for 18.04, 20.04, bionic, focal, and found no outdated references besides one now fixed - this LGTM and seems ready for merge!

manics:
I've posted one last warning https://discourse.jupyter.org/t/repo2docker-and-mybinder-org-upgrading-base-image-to-ubuntu-22-04/19865/2

vishwa2710:
Hi all! Wondering if this is going to get merged anytime soon?

ShalokShalom:
I just visted the site the first time, and wondered why you would announce, in August 23, that you plan to update the images to Ubuntu 22.04. It would be sad if this bad impression continues beyond your actual deployment.

consideRatio:
@yuvipanda @minrk @manics I went for a merge here. I verified that the tests still passed against the latest main branch.

yuvipanda:
Yay, thanks @consideRatio :)

PR #1288, opened 2023-06-13
Update version of R available

yuvipanda:
Fixes https://github.com/jupyterhub/repo2docker/issues/1278


minrk:
Docs failures seem to be a very recent bug in RTD. Last successful build was 30 minutes earlier, the build for this branch seems to have ignored our readthedocs.yaml (built with Python 3.7, then failed an import, clearly ignoring dependencies and the selection of 3.10).

yuvipanda:
@minrk yeah, the error seems fixed now.

PR #1293, opened 2023-07-08
Fix rstudio-build selection

yamaton:
Fix https://github.com/jupyterhub/repo2docker/issues/1290. 

Note that `apt-cache search libssl3` returns the exit code 0 even if there is no match.


yamaton:
@yuvipanda It's a single-line fix so would you take a look?

yuvipanda:
Ah damn! Thanks for catching this, @yamaton!

PR #1302, opened 2023-07-24
Update to JupyterLab 4 and Notebook 7

jtpio:
Fixes https://github.com/jupyterhub/repo2docker/issues/1095

### Changes

- [x] Update the base environment to JupyterLab 4 and Notebook 7
- [x] Update to Jupyter Server 2
- [x] Do not update Python 3.7 in `freeze.py` because these new versions dropped support for Python 3.7
- [x] Update to `jupyter-nbclassic-nbextension` commands
- [x] Update `jupyter-resource-usage` to 1.0


jtpio:
Looks like https://github.com/jupyterhub/nbgitpuller/pull/240 should help with this.

yuvipanda:
@jtpio think you'll have any time to push this forward?

jtpio:
It was paused because of the two failing tests. Maybe they will pass after a fresh rebase.

yuvipanda:
Hmm, the failures look real:

```
Jupyter command `jupyter-nbextension` not found.
```

Thoughts on splitting up the upgrade to jupyterlab 4 from notebook 7? I think the jupyterlab 4 change is uncontroversial and easy. But for notebook 7, we have to figure out what to do about nbclassic, and it's a bit more of a breaking change.

jtpio:
Right doing the JupyterLab 4 update separately would be fine.

I just pushed a commit to fix a reference to `jupyter-nbextension`. The update to Notebook 7 would indeed break the use of `jupyter-nbextension` because it is now available as `jupyter-nbclassic-nbextension` instead.

jtpio:
> Right doing the JupyterLab 4 update separately would be fine.

Opened https://github.com/jupyterhub/repo2docker/pull/1340.

PR #1310, opened 2023-10-12
r: Don't strictly pin version of rsession-proxy

yuvipanda:
If memory serves me right, the specific version semantics were here because at some point we were installing different versions of rsession-proxy for different versions of R. We don't do that anymore, so we should just unpin it and let the pip solver install the appropriate version.

Hopefully fixes https://discourse.jupyter.org/t/r-conda-example-repo-isnt-working-any-more/21739/3

### DIFF
--- a/repo2docker/buildpacks/_r_base.py
+++ b/repo2docker/buildpacks/_r_base.py
@@ -11,7 +11,6 @@ def rstudio_base_scripts(r_version):
 
     # Shiny server (not the package!) seems to be the same version for all R versions
     shiny_server_url = "https://download3.rstudio.org/ubuntu-14.04/x86_64/shiny-server-1.5.17.973-amd64.deb"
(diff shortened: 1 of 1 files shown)


manics:
I think we're pinning everything else that's installed by default in r2d. Is there anyway we can move this to a requirements.txt file that can be bumped by dependabot (or the freeze.py script, though dependabot is easier)?

yuvipanda:
@manics I've instead just bumped the version number to latest to minimize the amount of changes we need to (hopefully?) fix the current issue. Given I'm trying to focus on the frontend on binderhub, I don't think I currently have time to move things to a requirements.txt. I hope this is acceptable instead.

manics (approved):
Thanks! I'll open a separate issue about auto-bumping these r packages

yuvipanda:
Thanks, @manics!

PR #1325, opened 2023-12-20
Implement parameters to pass custom ignore file

### LINKED ISSUE #1323: Add an option to exclude files or folders to complement dockerignore/containerignore
### Proposed change

#1205 implements the support for `.dockerignore` and `.containerignore` files however this requires for people to explicitly add these files to their project.

The goal of this new option would be for example for BinderHub admins to be able to have  a default set of files and/or folders that they want to systematically ignore. For example, the .git folder which might be large and not of interest.

### Alternative options

Ensure all projects have a `.dockerignore` or `.containerignore` file

Currently I see it it mainly for the BinderHub admins.

This should not be a big task as part of the logic is already implemented in #1205 and the option values would basically be appended to the list.

### Who can do this work?

I can implement that.

sgaist:
Handling of the `.dockerignore` and `.containerignore` files is implemented in #1205, however, these files require people to prepare their projects with some knowledge of how they work and thus knowledge of docker/podman. The command parameters added in this PR allows the person responsible to build the image to provide an ignore file so that classic projects do not need one.

This can be useful for example on BinderHub. The administrator can provide sensible defaults if there is no ignore file present in the repository being built or force the use of a standard one.

The implementation provide three strategies:
- Merge
- Ours
- Theirs

Where merge, as the name suggest will put together the content of the two files, ours will force the use of the file passed in parameter and theirs will use the file present in the repository if present otherwise the one passed to the command.

Fixes #1323
Follow up off #1205


manics:
This means these files behave differently from all other configuration files. Do you think these are important enough that repo2docker should support special overrides and merging behaviour? Should something similar be considered for other files e.g. so an admin can set the default python/R version if it's not specified in a repo, or install some extra default packages like UI extensions or other utilities for local use? Should we add some callable hooks instead that allow full modification of the build context before building (and perhaps similar hooks at other stages)?

sgaist:
You are bringing very good questions.

In my original idea I had essentially BinderHub in mind and the optimization of the image building process. For example, avoid generating a big context with files considered useless and improve cache use (in the case of git cloning, no two clones are the same because the content of `.git` differs).

Your other suggestions sound interesting as well and the hooks would likely be a better fit or to keep things simple maybe a pre-build script ?

There's one thing though, the `.dockerignore`/`.containerignore` files influence what goes inside the build context while other configuration file would influence what is deployed within the image so they do not work at the same level in the build process.

manics:
The hook would be a pre-build hook called whilst constructing the build context (maybe at the start, just before creating the tar?). It would there have the full ability to add/remove/change files, whether that deleting unwanted files, adding a default version if not set, or modifying environment.yml.

Since the hook is a callable it would be defined in a repo2docker config file rather than a repo2docker command line arg, but in this particular case you need to pass the custom ignore file anyway that's not a problem.

sgaist:
I am not sure to follow you properly.

Do you have something like this in mind ?

`repo2docker.conf`:
```
pre-build: 
  hook: /path/to/extra_ignore_hook.py
  name: extra_ignore
  args: /path/to/extra_ignore_file
```

?

PR #1329, opened 2024-01-27
Update versioneer

TimoRoth:
Installing repo2docker on an up to date python version fails with various errors in these files.
This PR is simply the result of running "versioneer install" with the latest versioneer version.


manics:
It looks like pre-commit is causing problems. Can you try installing and running it locally?
```
pip install pre-commit
pre-commit run -a
git add -u
git commit ...
```

TimoRoth:
This somehow made pre-commit happy, after it commited a few more changes via the Workflow, but now all the tests are failing with that looks an unrelated error on first glance.

manics:
Thanks! Given that the failure is unrelated, the tests were passing, and it's going to break our main branch regardless, I'll merge this.

PR #1330, opened 2024-01-28
Support pytest=8

manics:
pytest 8 breaks our tests:
```
ERROR tests/base/node - pytest.PytestRemovedIn8Warning: The (fspath: py.path.local) argument to LocalRepo is deprecated. Please use the (path: pathlib.Path) argument instead.
See https://docs.pytest.org/en/latest/deprecations.html#fspath-argument-for-node-constructors-replaced-with-pathlib-path
```

Reported in https://github.com/jupyterhub/repo2docker/pull/1329#issuecomment-1913608729

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -2,6 +2,6 @@ build
 conda-lock
 pre-commit
 pytest-cov
-pytest>=7,<9
+pytest>=7
 pyyaml
 requests_mock
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -2,6 +2,6 @@ build
 conda-lock
 pre-commit
 pytest-cov
-pytest>=4.6,<8
+pytest>=7,<9
 pyyaml
 requests_mock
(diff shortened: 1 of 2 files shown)


dev-requirements.txt
```diff
@@ -2,6 +2,6 @@ build
 conda-lock
 pre-commit
 pytest-cov
-pytest>=4.6
+pytest>=7,<9
```
minrk, line 5: preemptively blocking pytest 9? Are there known planned incompatibilities to deal with?

minrk (approved):

dev-requirements.txt
manics: Nothing known, I'm just fed up of dealing with random breakages! Do you want me to remove the upper limit?

yuvipanda: Ya, I think we should remove the limit as otherwise we may get stuck here. In general I think all upper limit pins should come with a time expiry.

Otherwise happy to merge this.

manics, line 5:
```suggestion
pytest>=7
```

yuvipanda:
The test failure is real but unrelated:

> 1: package ‘IRkernel’ is not available (for R version 3.6.3) 

So I'm going to merge this.

PR #1332, opened 2024-02-03
Dockerfile: bump alpine from 3.17 to 3.19 and Python 3.10 to 3.11

yuvipanda:
Brings us to latest alpine, and to a newer python

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # syntax = docker/dockerfile:1.3
-ARG ALPINE_VERSION=3.17
+ARG ALPINE_VERSION=3.19
 FROM alpine:${ALPINE_VERSION}
 
 RUN apk add --no-cache git python3 python3-dev py3-pip py3-setuptools build-base
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,12 +5,17 @@ FROM alpine:${ALPINE_VERSION}
 RUN apk add --no-cache git python3 python3-dev py3-pip py3-setuptools build-base
 
 # build wheels in first image
+ENV VIRTUAL_ENV /opt/venv
+ENV PATH ${VIRTUAL_ENV}/bin:${PATH}
+
+RUN python3 -m venv ${VIRTUAL_ENV}
+
 ADD . /tmp/src
 RUN cd /tmp/src && git clean -xfd && git status
 RUN mkdir /tmp/wheelhouse \
  && cd /tmp/wheelhouse \
- && pip3 install wheel \
- && pip3 wheel --no-cache-dir /tmp/src \
+ && pip install wheel \
+ && pip wheel --no-cache-dir /tmp/src \
  && ls -l /tmp/wheelhouse
 
 FROM alpine:${ALPINE_VERSION}
@@ -18,13 +23,18 @@ FROM alpine:${ALPINE_VERSION}
 # install python, git, bash, mercurial
 RUN apk add --no-cache git git-lfs python3 py3-pip py3-setuptools bash docker mercurial
 
+ENV VIRTUAL_ENV /opt/venv
+ENV PATH ${VIRTUAL_ENV}/bin:${PATH}
+
+RUN python3 -m venv ${VIRTUAL_ENV}
+
 # install hg-evolve (Mercurial extensions)
-RUN pip3 install hg-evolve --user --no-cache-dir
+RUN pip install hg-evolve --user --no-cache-dir
 
 # install repo2docker
 COPY --from=0 /tmp/wheelhouse /tmp/wheelhouse
-RUN pip3 install --no-cache-dir --ignore-installed --no-deps /tmp/wheelhouse/*.whl \
- && pip3 list
+RUN pip install --no-cache-dir --ignore-installed --no-deps /tmp/wheelhouse/*.whl \
+ && pip list
 
 # add git-credential helper
 COPY ./docker/git-credential-env /usr/local/bin/git-credential-env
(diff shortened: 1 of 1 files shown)


manics:
What are the benefits of using a virtualenv inside the image?

yuvipanda:
@manics it is mandatory now, it errors out if you don't use one (https://peps.python.org/pep-0668/).

You can see the full config in https://github.com/jupyterhub/repo2docker/actions/runs/7768480733/job/21186449405

```
 #15 [stage-1 3/9] RUN pip3 install hg-evolve --user --no-cache-dir
#15 1.427 error: externally-managed-environment
#15 1.427 
#15 1.427 × This environment is externally managed
#15 1.427 ╰─> 
#15 1.427     The system-wide python installation should be maintained using the system
#15 1.427     package manager (apk) only.
#15 1.427     
#15 1.427     If the package in question is not packaged already (and hence installable via
#15 1.427     "apk add py3-somepackage"), please consider installing it inside a virtual
#15 1.427     environment, e.g.:
#15 1.427     
#15 1.427     python3 -m venv /path/to/venv
#15 1.427     . /path/to/venv/bin/activate
#15 1.427     pip install mypackage
#15 1.427     
#15 1.427     To exit the virtual environment, run:
#15 1.427     
#15 1.427     deactivate
#15 1.427     
#15 1.427     The virtual environment is not deleted, and can be re-entered by re-sourcing
#15 1.427     the activate file.
#15 1.427     
#15 1.427     To automatically manage virtual environments, consider using pipx (from the
#15 1.427     pipx package).
#15 1.427 
#15 1.427 note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
#15 1.427 hint: See PEP 668 for the detailed specification.
#15 ERROR: process "/bin/sh -c pip3 install hg-evolve --user --no-cache-dir" did not complete successfully: exit code: 1
```

Dockerfile
```diff
@@ -1,30 +1,40 @@
 # syntax = docker/dockerfile:1.3
-ARG ALPINE_VERSION=3.17
+ARG ALPINE_VERSION=3.19
 FROM alpine:${ALPINE_VERSION}
 
 RUN apk add --no-cache git python3 python3-dev py3-pip py3-setuptools build-base
 
 # build wheels in first image
```
consideRatio, line 7:
```suggestion
# build wheels in a build stage
```

Dockerfile
```diff
@@ -1,30 +1,40 @@
...
+ && pip wheel --no-cache-dir /tmp/src \
  && ls -l /tmp/wheelhouse
 
 FROM alpine:${ALPINE_VERSION}
 
 # install python, git, bash, mercurial
 RUN apk add --no-cache git git-lfs python3 py3-pip py3-setuptools bash docker mercurial
 
+ENV VIRTUAL_ENV /opt/venv
+ENV PATH ${VIRTUAL_ENV}/bin:${PATH}
```
consideRatio, line 27:
```suggestion
ENV VIRTUAL_ENV=/opt/venv
ENV PATH=${VIRTUAL_ENV}/bin:${PATH}
```

Outdated syntax, may be removed, see https://docs.docker.com/reference/dockerfile/#env.

Dockerfile
```diff
@@ -1,30 +1,40 @@
...
-ARG ALPINE_VERSION=3.17
+ARG ALPINE_VERSION=3.19
 FROM alpine:${ALPINE_VERSION}
 
 RUN apk add --no-cache git python3 python3-dev py3-pip py3-setuptools build-base
 
-# build wheels in first image
+# build wheels in a build stage
+ENV VIRTUAL_ENV /opt/venv
+ENV PATH ${VIRTUAL_ENV}/bin:${PATH}
```
consideRatio, line 9:
```suggestion
ENV VIRTUAL_ENV=/opt/venv
ENV PATH=${VIRTUAL_ENV}/bin:${PATH}
```

Dockerfile
```diff
@@ -1,30 +1,40 @@
 # syntax = docker/dockerfile:1.3
-ARG ALPINE_VERSION=3.17
+ARG ALPINE_VERSION=3.19
 FROM alpine:${ALPINE_VERSION}
 
 RUN apk add --no-cache git python3 python3-dev py3-pip py3-setuptools build-base
 
-# build wheels in first image
+# build wheels in a build stage
+ENV VIRTUAL_ENV=/opt/venv
```
consideRatio, line 8:
```suggestion
ARG VIRTUAL_ENV=/opt/venv
```

Dockerfile
```diff
@@ -1,30 +1,40 @@
...
+ && pip install wheel \
+ && pip wheel --no-cache-dir /tmp/src \
  && ls -l /tmp/wheelhouse
 
 FROM alpine:${ALPINE_VERSION}
 
 # install python, git, bash, mercurial
 RUN apk add --no-cache git git-lfs python3 py3-pip py3-setuptools bash docker mercurial
 
+ENV VIRTUAL_ENV=/opt/venv
```
consideRatio, line 26:
```suggestion
ARG VIRTUAL_ENV=/opt/venv
```

consideRatio:
Merging this resolves repo2docker-action https://github.com/jupyterhub/repo2docker-action/issues/113.

I pushed commits to:
- tweak comment unrelated to PR
- use modern ENV syntax with `=` signs between key/values
- use ARG instead of ENV when not needed to provide an env var only during docker build as compared to also during docker run

consideRatio (approved):
If tests pass, this LGTM.

consideRatio:
I marked this as breaking for possible highlight its a possible breaking change for users of the image

PR #1335, opened 2024-02-19
[WIP] Add meca content provider

stevejpurves:
This PR adds a content provider for MECA Bundles.

[MECA stands for Manuscript Exchange Common Approach](https://www.niso.org/standards-committees/meca), a NISO standard for transfer of manuscripts between manuscript systems.

The provider was written during in 2023 as part of the [AGU's NotebooksNow project](https://data.agu.org/notebooks-now/). The aim of that project is to make notebook based computational research papers a formal part of the scientific record, and has drawn on inputs from a number of working groups and stakeholder organisations. 

MECA was identified as a way to consistently package files from a git repository, paired with JATS xml, that would enable a set of notebooks, code, REES environment specification, into an single file that could be accepted and archived in existing publishing systems.

Support to produce MECA files has been included in Myst Markdown and Quatro author toolchains, enabling researchers to export to this format and the first AGU journal supporting the format is going live in 2024.

By adding provider as a standard provider in repo2docker/binderhub, we'll enable authors and researchers to reproduce research articles that include this bundle.

This provider is intended to be paired with the BinderHub Repo Provider being added in this PR:

### Operation

The provider receives a parsed url to a Meca bundle, which is a zip file. The zipfile is fetched to a local tmp folder and unpacked. Given a `manifest.xml` with expected tags to a `article-source-directory`, that identifies the path to use for docker creation. (we expect a REES compatible environment to be present. If this information is missing, we try to build from the root folder.

### DIFF
--- a/repo2docker/contentproviders/meca.py
+++ b/repo2docker/contentproviders/meca.py
@@ -1,0 +1,130 @@
+from .base import ContentProvider
+from requests import Session
+import os
+from hashlib import md5
+from os import path
+import tempfile
+import shutil
+import xml.etree.ElementTree as ET
+from zipfile import ZipFile, is_zipfile
+from urllib.parse import urlparse, urlunparse
+
+def get_hashed_slug(url, changes_with_content):
+    """Return a unique slug that is invariant to query parameters in the url"""
+    parsed_url = urlparse(url)
+    stripped_url = urlunparse(
+        (parsed_url.scheme, parsed_url.netloc, parsed_url.path, "", "", "")
+    )
+
+    return "meca-" + md5(f"{stripped_url}-{changes_with_content}".encode()).hexdigest()
+
+def fetch_zipfile(session, url, dst_dir):
+    resp = session.get(url, headers={"accept": "application/zip"}, stream=True)
+    resp.raise_for_status()
+
+    dst_filename = path.join(dst_dir, "meca.zip")
+    with open(dst_filename, "wb") as dst:
+        for chunk in resp.iter_content(chunk_size=128):
+            dst.write(chunk)
+
+    return dst_filename
+
+
+def handle_items(_, item):
+    print(item)
+
+
+def extract_validate_and_identify_bundle(zip_filename, dst_dir):
+    if not os.path.exists(zip_filename):
+        raise RuntimeError("Download MECA bundle not found")
+
+    if not is_zipfile(zip_filename):
+        raise RuntimeError("MECA bundle is not a zip file")
+
+    with ZipFile(zip_filename, "r") as zip_ref:
+        zip_ref.extractall(dst_dir)
+
+    try:
+        manifest = path.join(dst_dir, "manifest.xml")
+        if not os.path.exists(manifest):
+            raise RuntimeError("MECA bundle is missing manifest.xml")
+        article_source_dir = "bundle/"
+
+        tree = ET.parse(manifest)
+        root = tree.getroot()
+
+        bundle_instance = root.findall(
+            "{*}item[@item-type='article-source-directory']/{*}instance"
+        )
+        for attr in bundle_instance[0].attrib:
+            if attr.endswith("href"):
+                article_source_dir = bundle_instance[0].get(attr)
+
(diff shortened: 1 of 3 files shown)


repo2docker/contentproviders/meca.py
```diff
@@ -0,0 +1,130 @@
...
+        for chunk in resp.iter_content(chunk_size=128):
+            dst.write(chunk)
+
+    return dst_filename
+
+
+def handle_items(_, item):
+    print(item)
+
+
```
sgaist, line 36: This is unused
```suggestion
```

repo2docker/contentproviders/meca.py
```diff
@@ -0,0 +1,130 @@
...
+def extract_validate_and_identify_bundle(zip_filename, dst_dir):
+    if not os.path.exists(zip_filename):
+        raise RuntimeError("Download MECA bundle not found")
```
sgaist, line 39:
```suggestion
        raise RuntimeError("Downloaded MECA bundle not found")
```

repo2docker/contentproviders/meca.py
```diff
@@ -0,0 +1,130 @@
...
+        self.session.headers.update(
+            {
+                "user-agent": f"repo2docker MECA",
+            }
+        )
+
+    def detect(self, spec, ref=None, extra_args=None):
+        """`spec` contains a faux protocol of meca+http[s] for detection purposes
+        and we assume `spec` trusted as a reachable MECA bundle from an allowed origin
+        (binderhub RepoProvider class already checking for this).
```
sgaist, line 83:
```suggestion
        (binderhub RepoProvider class is already checking for this).
```

repo2docker/contentproviders/meca.py
```diff
@@ -0,0 +1,130 @@
...
+        An other HEAD check in made here in order to get the content-length header
+        """
+        parsed = urlparse(spec)
+        if not parsed.scheme.endswith("+meca"):
+            return None
+        parsed = parsed._replace(scheme=parsed.scheme[:-5])
+        url = urlunparse(parsed)
+
+        r = self.session.head(url)
+        changes_with_content = r.headers.get("ETag") or r.headers.get("Content-Length")
```
sgaist, line 94:
```suggestion
        headers = self.session.head(url).headers
        changes_with_content = headers.get("ETag") or headers.get("Content-Length")
```

repo2docker/contentproviders/meca.py
```diff
@@ -0,0 +1,130 @@
...
+from hashlib import md5
+from os import path
+import tempfile
+import shutil
+import xml.etree.ElementTree as ET
+from zipfile import ZipFile, is_zipfile
+from urllib.parse import urlparse, urlunparse
+
+def get_hashed_slug(url, changes_with_content):
+    """Return a unique slug that is invariant to query parameters in the url"""
```
sgaist, line 13:
```suggestion
    """Returns a unique slug that is invariant to query parameters in the url"""
```

sgaist (changes requested):
The implementation functionally wise looks fine.

There are some minor suggestions.

I would also recommend to run `pre-commit`, it will cleanup the imports as well as some whites pace related things.

stevejpurves:
thanks @sgaist 
I'm continuing to test locally this week and hope to pop this PR out of "draft" be the week end. I'm hitting some edge cases when:

- [] url specs contain multiple query params, these can be stripped after the first parameter
- [] the provider works when used in conjunction with binderhub but errors out when used from the commandline

yuvipanda:
Hi @stevejpurves :) I just merged another new provider (CKAN), so wanted to check in to see if there's anything I can do to help this move forward!

stevejpurves:
I've brought this up to date and also made some changes to allow a server to be launched from a local something.meca.zip file. Basically using a naming convention as way to initially detect a meca bundle specifically.

I guess that could be changed to check *any* local file for the expected contents, before returning true, but maybe this is fine for now.

I still intend to resolve:

- [ ] url specs contain multiple query params, these can be stripped after the first parameter
- [ ] the provider works when used in conjunction with binderhub but errors out when used from the command line

PR #1336, opened 2024-02-20
[MRG] Add CKAN content provider

u10313335:
This PR add support for [CKAN](https://ckan.org/) datasets as content provider for repo2docker, e.g.:

```
repo2docker https://demo.ckan.org/dataset/sample-dataset-1
```

It has been tested on some CKAN repositories such as:

- data.gov (https://catalog.data.gov/dataset/fdic-failed-bank-list)
- Canada Open Government Portal (https://open.canada.ca/data/en/dataset/d8f85d91-7dec-4fd1-8055-483b77225d8b)
- NHM Data Portal (https://data.nhm.ac.uk/dataset/samuel-dale)
- SEED in NSW (https://datasets.seed.nsw.gov.au/dataset/corridors-cessnock-lga)

We also adopt this PR in our Binder service (as the `CKAN dataset` repository): https://binder.depositar.io/, e.g.:

- https://binder.depositar.io/v2/ckan/https://data.depositar.io/dataset/binder-example-sea-turtle-sightings-in-taiwan

### DIFF
--- a/repo2docker/contentproviders/ckan.py
+++ b/repo2docker/contentproviders/ckan.py
@@ -1,6 +1,6 @@
 from datetime import datetime, timedelta, timezone
 from os import path
-from urllib.parse import urlparse
+from urllib.parse import parse_qs, urlparse
 
 from requests import Session
 
@@ -43,23 +43,33 @@ class CKAN(ContentProvider):
         if not parsed_url.netloc:
             return None
 
-        url_parts = parsed_url.path.split("/")
-        if url_parts[-2] == "dataset":
-            self.dataset_id = url_parts[-1]
+        url_parts_1 = parsed_url.path.split("/history/")
+        url_parts_2 = url_parts_1[0].split("/")
+        if url_parts_2[-2] == "dataset":
+            self.dataset_id = url_parts_2[-1]
         else:
             return None
 
         api_url_path = "/api/3/action/"
         api_url = parsed_url._replace(
-            path="/".join(url_parts[:-2]) + api_url_path
+            path="/".join(url_parts_2[:-2]) + api_url_path, query=""
         ).geturl()
 
         status_show_url = f"{api_url}status_show"
         resp = self.urlopen(status_show_url)
         if resp.status_code == 200:
+
+            # handle the activites
+            activity_id = None
+            if parse_qs(parsed_url.query).get("activity_id") is not None:
+                activity_id = parse_qs(parsed_url.query).get("activity_id")[0]
+            if len(url_parts_1) == 2:
+                activity_id = url_parts_1[-1]
+
             self.version = self._fetch_version(api_url)
             return {
                 "dataset_id": self.dataset_id,
+                "activity_id": activity_id,
                 "api_url": api_url,
                 "version": self.version,
             }
@@ -69,11 +79,21 @@ class CKAN(ContentProvider):
     def fetch(self, spec, output_dir, yield_output=False):
         """Fetch a CKAN dataset."""
         dataset_id = spec["dataset_id"]
+        activity_id = spec["activity_id"]
 
         yield f"Fetching CKAN dataset {dataset_id}.\n"
-        package_show_url = f"{spec['api_url']}package_show?id={dataset_id}"
+
+        # handle the activites
+        if activity_id:
+            fetch_url = (
+                f"{spec['api_url']}activity_data_show?"
+                f"id={activity_id}&object_type=package"
+            )
+        else:
+            fetch_url = f"{spec['api_url']}package_show?id={dataset_id}"
+
         resp = self.urlopen(
-            package_show_url,
+            fetch_url,
             headers={"accept": "application/json"},
         )
 
--- a/repo2docker/contentproviders/ckan.py
+++ b/repo2docker/contentproviders/ckan.py
@@ -105,6 +105,8 @@ class CKAN(ContentProvider):
 
         for resource in resources:
             file_url = resource["url"]
+            if file_url == "":
+                continue
             fname = file_url.rsplit("/", maxsplit=1)[-1]
             if fname == "":
                 fname = resource["id"]
--- a/repo2docker/contentproviders/ckan.py
+++ b/repo2docker/contentproviders/ckan.py
@@ -1,6 +1,6 @@
 from datetime import datetime, timedelta, timezone
 from os import path
-from urllib.parse import parse_qs, urlparse, urlencode
+from urllib.parse import parse_qs, urlencode, urlparse
 
 from requests import Session
 
@@ -89,11 +89,13 @@ class CKAN(ContentProvider):
 
         # handle the activites
         if activity_id:
-            fetch_url = (
-                f"{spec['api_url']}activity_data_show?" + urlencode({"id": activity_id, "object_type": "package"})
+            fetch_url = f"{spec['api_url']}activity_data_show?" + urlencode(
+                {"id": activity_id, "object_type": "package"}
             )
         else:
-            fetch_url = f"{spec['api_url']}package_show?" + urlencode({"id": dataset_id})
+            fetch_url = f"{spec['api_url']}package_show?" + urlencode(
+                {"id": dataset_id}
+            )
 
         resp = self.urlopen(
             fetch_url,
--- a/repo2docker/contentproviders/ckan.py
+++ b/repo2docker/contentproviders/ckan.py
@@ -50,7 +50,9 @@ class CKAN(ContentProvider):
             return None
 
         api_url_path = "/api/3/action/"
-        api_url = parsed_url._replace(path=api_url_path).geturl()
+        api_url = parsed_url._replace(
+            path="/".join(url_parts[:-2]) + api_url_path
+        ).geturl()
 
         status_show_url = f"{api_url}status_show"
         resp = self.urlopen(status_show_url)
--- a/repo2docker/contentproviders/ckan.py
+++ b/repo2docker/contentproviders/ckan.py
@@ -1,4 +1,3 @@
-import re
 from datetime import datetime, timedelta, timezone
 from os import path
 from urllib.parse import urlparse
@@ -37,7 +36,6 @@ class CKAN(ContentProvider):
         return self.session.get(url, **kwargs)
 
     urlopen = _request
-    url_regex = r"/dataset/[a-z0-9_\\-]*$"
 
     def detect(self, source, ref=None, extra_args=None):
         """Trigger this provider for things that resolve to a CKAN dataset."""
@@ -45,14 +43,18 @@ class CKAN(ContentProvider):
         if not parsed_url.netloc:
             return None
 
-        api_url = parsed_url._replace(
-            path=re.sub(self.url_regex, "/api/3/action/", parsed_url.path)
-        ).geturl()
+        url_parts = parsed_url.path.split("/")
+        if url_parts[-2] == "dataset":
+            self.dataset_id = url_parts[-1]
+        else:
+            return None
+
+        api_url_path = "/api/3/action/"
+        api_url = parsed_url._replace(path=api_url_path).geturl()
 
         status_show_url = f"{api_url}status_show"
         resp = self.urlopen(status_show_url)
         if resp.status_code == 200:
-            self.dataset_id = parsed_url.path.rsplit("/", maxsplit=1)[1]
             self.version = self._fetch_version(api_url)
             return {
                 "dataset_id": self.dataset_id,
--- a/repo2docker/contentproviders/ckan.py
+++ b/repo2docker/contentproviders/ckan.py
@@ -1,6 +1,6 @@
 from datetime import datetime, timedelta, timezone
 from os import path
-from urllib.parse import parse_qs, urlparse
+from urllib.parse import parse_qs, urlparse, urlencode
 
 from requests import Session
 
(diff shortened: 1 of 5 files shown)


repo2docker/contentproviders/ckan.py
```diff
@@ -0,0 +1,101 @@
...
+    url_regex = r"/dataset/[a-z0-9_\\-]*$"
+
+    def detect(self, source, ref=None, extra_args=None):
+        """Trigger this provider for things that resolve to a CKAN dataset."""
+        parsed_url = urlparse(source)
+        if not parsed_url.netloc:
+            return None
+
+        api_url = parsed_url._replace(
+            path=re.sub(self.url_regex, "/api/3/action/", parsed_url.path)
```
yuvipanda, line 49: Instead of using a regex here, can we write out the splitting logic more directly? Something like (haven't tested this, but you get the gist):

```python
url_parts = parsed_url.path.split('/')
if url_parts[-2] == 'dataset':
   dataset_id = url_parts[-1]
else:
   return None
api_url_path = f'/api/3/action/{dataset_id}'
api_url = parsed_url.replace(path=api_url_path)
```

yuvipanda: This also lets us re-use `dataset_id` down below

tests/unit/contentproviders/test_ckan.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+]
+
+
+@pytest.mark.parametrize("test_input, expected", test_hosts)
+def test_detect_ckan(test_input, expected):
+    assert CKAN().detect(test_input[0]) == expected
+
+    # Don't trigger the CKAN content provider
+    assert CKAN().detect("/some/path/here") is None
```
yuvipanda, line 29: Since these are not using the parameterized information, these should be their own test.

tests/unit/contentproviders/test_ckan.py
```diff
@@ -0,0 +1,58 @@
...
+        {
+            "dataset_id": "sample-dataset-1",
+            "api_url": "http://demo.ckan.org/api/3/action/",
+            "version": "1707387710",
+        },
+    )
+]
+
+
+@pytest.mark.parametrize("test_input, expected", test_hosts)
```
yuvipanda, line 24: This is currently failing with:

```
>       assert CKAN().detect(test_input[0]) == expected
E       AssertionError: assert {'api_url': '... '1709043532'} == {'api_url': '... '1707387710'}
E         Omitting 2 identical items, use -vv to show
E         Differing items:
E         {'version': '1709043532'} != {'version': '1707387710'}
E         Use -v to get more diff
```

Is there any way to lock this down some more?

yuvipanda (changes requested):
Apologies for the delay in reviewing! Thank you for the PR :) I've left a couple of style comments that should be fairly easy fixes.

The one question for me is wether the unit should be 'dataset' or 'dataset at a particular version'. This is illustrated by the current unit test failing because the version of the package has changed. The equivalent of sorts in git land is 'git repo url + hash', where the hash is counted as the 'ref'. If no 'ref' is provided, we default to using the latest (or whatever `HEAD`) points to. Does CKAN need something similar?

u10313335:
> The one question for me is wether the unit should be 'dataset' or 'dataset at a particular version'. This is illustrated by the current unit test failing because the version of the package has changed. The equivalent of sorts in git land is 'git repo url + hash', where the hash is counted as the 'ref'. If no 'ref' is provided, we default to using the latest (or whatever `HEAD`) points to. Does CKAN need something similar?

Thank you @yuvipanda  for your review! As a newcomer to Project Jupyter, it really encouraged me.

Regard to the question referring to the unit, there are indeed "activities" that work like the versions for datasets. For example:

https://demo.ckan.org/dataset/activity/sample-dataset-1

However, activities are often unavailable as they can be [hidden by admins](https://docs.ckan.org/en/2.10/maintaining/configuration.html#ckan-auth-public-activity-stream-detail). In fact, the above demo site of CKAN also restricts the access to activities. Another example is the [data.gov](https://catalog.data.gov/dataset/activity/electric-vehicle-population-data).

Therefore, from my point of view, it's much safer to take the 'dataset' rather than the 'dataset at a particular version' as the unit for the CKAN provider.

repo2docker/contentproviders/ckan.py
u10313335: Thanks. I have made changes according to your suggestion.

tests/unit/contentproviders/test_ckan.py
u10313335: Yes, they should. Fixed.

tests/unit/contentproviders/test_ckan.py
u10313335: Oh, I forget that the demo site of CKAN will be rebuild occasionally. Instead I wrote a mock response for the `test_detect_ckan` test.

manics:
> However, activities are often unavailable as they can be [hidden by admins](https://docs.ckan.org/en/2.10/maintaining/configuration.html#ckan-auth-public-activity-stream-detail). In fact, the above demo site of CKAN also restricts the access to activities. Another example is the [data.gov](https://catalog.data.gov/dataset/activity/electric-vehicle-population-data).

Does this just disable listing previous versions, or does it completely prevent access to previous versions? Do you know why admins chose to prevent this, since they're reducing the utility of their CKAN repository as a reproducible data source?

u10313335:
> > However, activities are often unavailable as they can be [hidden by admins](https://docs.ckan.org/en/2.10/maintaining/configuration.html#ckan-auth-public-activity-stream-detail). In fact, the above demo site of CKAN also restricts the access to activities. Another example is the [data.gov](https://catalog.data.gov/dataset/activity/electric-vehicle-population-data).
> 
> Does this just disable listing previous versions, or does it completely prevent access to previous versions? Do you know why admins chose to prevent this, since they're reducing the utility of their CKAN repository as a reproducible data source?

When the `ckan.auth.public_activity_stream_detail` is set to True, it will completely block **both list and access** to previous versions for the public (without login). The reason was explained in the [changelog](https://docs.ckan.org/en/2.10/changelog.html#v-2-9-0-2020-08-05):

> A full history of dataset changes is now displayed in the Activity Stream to admins, and optionally to the public. By default this is enabled for new installs, but disabled for sites which upgrade (just in case the history is sensitive).

AFAIK, most of sites were upgraded from the old versions of CKAN.

Or I think we could still support the versions (urls with activity_id), e.g.:

`https://demo.ckan.org/dataset/sample-dataset-1?activity_id=0da6b0a5-9fe5-4c34-9545-9079903a8449`

And go to the latest one when the version is missing, e.g.:

`https://demo.ckan.org/dataset/sample-dataset-1`

manics:
> I think we could still support the versions (urls with activity_id), e.g.:
> https://demo.ckan.org/dataset/sample-dataset-1?activity_id=0da6b0a5-9fe5-4c34-9545-9079903a8449
> And go to the latest one when the version is missing

That sounds reasonable to me- support a link to a fixed version where possible, fallback to the latest if not. Are there any disadvantages to doing this?

u10313335:
> That sounds reasonable to me- support a link to a fixed version where possible, fallback to the latest if not. Are there any disadvantages to doing this?

I don't see any disadvantages in this moment. I will try to handle this in the following days.

u10313335:
The content provider has updated to support the versions ("activities" in CKAN) and fallback to the latest. There are two types of URL patterns to the activities:

1. https://SITE_URL/dataset/DATASET_NAME?activity_id=ACTIVITY_ID (for CKAN version 2.9), e.g., https://demo.ckan.org/dataset/sample-dataset-1?activity_id=0da6b0a5-9fe5-4c34-9545-9079903a8449.

2. https://SITE_URL/dataset/DATASET_NAME/history/ACTIVITY_ID (for CKAN 2.10). I can't find any CKAN 2.10 sites which opens up access to activities for the public (we are still using the CKAN 2.9...) — sorry about that. But it has been tested with an internal CKAN 2.10 site.

Also note that the `metadata_modified` timestamp is still used to determine the "versions" in the repo2docker since the `metadata_modified` is available across all CKAN versions.

u10313335:
Hi @yuvipanda @manics, if there is any need for further improvement, please let me know. Thanks!

yuvipanda:
@u10313335 apologies for the delayed response. I think this is good to go, but I wanted to cleanup some of the URL handling a little bit. I can't seem to push to your branch, nor make a PR to your branch for some reason, so I've pushed my changes to https://github.com/yuvipanda/repo2docker/tree/ckan. Also  here's a patch file, which can be applied via `git am` if needed:

```
From 49613412583def1330cd8b7d6e42664728367945 Mon Sep 17 00:00:00 2001
From: YuviPanda <yuvipanda@gmail.com>
Date: Mon, 10 Jun 2024 15:49:40 -0700
Subject: [PATCH 1/2] Use urlencode to construct query strings

---
 repo2docker/contentproviders/ckan.py | 7 +++----
 1 file changed, 3 insertions(+), 4 deletions(-)

diff --git a/repo2docker/contentproviders/ckan.py b/repo2docker/contentproviders/ckan.py
index 9131b01..c1e31a5 100644
--- a/repo2docker/contentproviders/ckan.py
+++ b/repo2docker/contentproviders/ckan.py
@@ -1,6 +1,6 @@
 from datetime import datetime, timedelta, timezone
 from os import path
-from urllib.parse import parse_qs, urlparse
+from urllib.parse import parse_qs, urlparse, urlencode
 
 from requests import Session
 
@@ -86,11 +86,10 @@ class CKAN(ContentProvider):
         # handle the activites
         if activity_id:
             fetch_url = (
-                f"{spec['api_url']}activity_data_show?"
-                f"id={activity_id}&object_type=package"
+                f"{spec['api_url']}activity_data_show?" + urlencode({"id": activity_id, "object_type": "package"})
             )
         else:
-            fetch_url = f"{spec['api_url']}package_show?id={dataset_id}"
+            fetch_url = f"{spec['api_url']}package_show?" + urlencode({"id": dataset_id})
 
         resp = self.urlopen(
             fetch_url,
-- 
2.43.2


From 45fb7b908527630370b4ff1a99c89c5afafea2dc Mon Sep 17 00:00:00 2001
From: YuviPanda <yuvipanda@gmail.com>
Date: Mon, 10 Jun 2024 17:13:22 -0700
Subject: [PATCH 2/2] Cleanup URL parsing mechanisms

---
 repo2docker/contentproviders/ckan.py | 26 +++++++++++++++-----------
 1 file changed, 15 insertions(+), 11 deletions(-)

diff --git a/repo2docker/contentproviders/ckan.py b/repo2docker/contentproviders/ckan.py
index c1e31a5..9c74afd 100644
--- a/repo2docker/contentproviders/ckan.py
+++ b/repo2docker/contentproviders/ckan.py
@@ -43,28 +43,32 @@ class CKAN(ContentProvider):
         if not parsed_url.netloc:
             return None
 
-        url_parts_1 = parsed_url.path.split("/history/")
-        url_parts_2 = url_parts_1[0].split("/")
-        if url_parts_2[-2] == "dataset":
-            self.dataset_id = url_parts_2[-1]
-        else:
+        if "/dataset/" not in parsed_url.path:
+            # Not actually a dataset
             return None
 
-        api_url_path = "/api/3/action/"
+        # CKAN may be under a URL prefix, and we should accomodate that
+        url_prefix, dataset_url = parsed_url.path.split("/dataset/")
+
+        dataset_url_parts = dataset_url.split("/")
+        self.dataset_id = dataset_url_parts[0]
+
         api_url = parsed_url._replace(
-            path="/".join(url_parts_2[:-2]) + api_url_path, query=""
+            path=f"{url_prefix}/api/3/action/", query=""
         ).geturl()
 
         status_show_url = f"{api_url}status_show"
         resp = self.urlopen(status_show_url)
         if resp.status_code == 200:
 
-            # handle the activites
+            # Activity ID may be present either as a query parameter, activity_id
+            # or as part of the URL, under `/history/<activity-id>`. If `/history/`
+            # is present, that takes precedence over `activity_id`
             activity_id = None
-            if parse_qs(parsed_url.query).get("activity_id") is not None:
+            if "history" in dataset_url_parts:
+                activity_id = dataset_url_parts[dataset_url_parts.index("history") + 1]
+            elif parse_qs(parsed_url.query).get("activity_id") is not None:
                 activity_id = parse_qs(parsed_url.query).get("activity_id")[0]
-            if len(url_parts_1) == 2:
-                activity_id = url_parts_1[-1]
 
             self.version = self._fetch_version(api_url)
             return {
-- 
2.43.2

```

If these changes look good to you, apply those and I'll merge it? If not you can also select the 'allow edits from maintainers' button and I'll push these in too.

u10313335:
Thanks @yuvipanda again for the reviews and suggestions! I have applied your commits.

yuvipanda:
I've somehow missed your comment - sorry about that, @u10313335.

Tests are now unfortunately failing due to https://github.com/jupyterhub/repo2docker/issues/1354. I'm running all these tests locally - if they pass, I'll merge this PR (as it's already been waiting so long!) before figuring out how to fix that.

Thank you for your patience.

yuvipanda:
I was able to successfully run these tests locally, so I'm going to merge this one. I'll try to come back to the associated binderhub PR soon as well.

Thank you so much for your patience as we worked through this, @u10313335! Hope to continue to see PRs from you, and I think our reviewing speeds should be better now.

PR #1338, opened 2024-03-24
Dockerfile: update alpine from 3.17 to 3.18, and Python 3.10 to 3.11

consideRatio:
This seems to resolve https://github.com/jupyterhub/repo2docker-action/issues/113 which seem to stem from an issue in alpine:3.17 that has arisen - maybe it will get fixed over time - but I think this could also do the trick.


consideRatio:
Closing in favor of #1332

PR #1341, opened 2024-03-28
Changelog for 2024.03.0

manics:
We're about to make some breaking changes including https://github.com/jupyterhub/repo2docker/pull/1340 so lets get a release out first.


yuvipanda (approved):
NICE! Merge and release whenever you feel like.

PR #1347, opened 2024-05-21
add Python 3.12

### LINKED ISSUE #1225: Python 3.11 and beyond
### Proposed change

Support Python 3.11, and perhaps automatically support future Python releases?

Python 3.11.0 was released in [October 2022](https://peps.python.org/pep-0664/#schedule) but unfortunately I can’t yet use it in mybinder.org. 

While it looks like you only need to run a simple script to add the conda environment file for new Python versions, you’ll need to do this manually [once every 12 months](https://peps.python.org/pep-0602/). Is there any way this could be automated a bit or have the environment generated *on the fly* by periodically checking for newer Python versions on conda forge? 

### Alternative options
File a request for every new 3.x Python release, yearly. 

Anyone wanting to use features in newer Python releases.

 Adding 3.11: update conda/freeze.py to add 3.11 to the defaults, generate the environment files,  commit.

Adding automated updates based on conda forge version info: hard to say, but a few days perhaps, depending on how well understood the GitHub Action environment is and how you’d commit updated environment files.

### Who can do this work?

The base 3.11 work: most anyone, if the #1175 PR is typical. 

Automatic detection of new Python versions: requires conda forge API familiarity plus experience with GH actions to automate commits or PRs.

minrk:
just ran `python3 repo2docker/buildpacks/conda/freeze.py 3.12` and added 3.12 to the refreeze list

closes #1225


manics:
The latest version of requests breaks docker-py, causing all tests to fail:
https://github.com/docker/docker-py/issues/3256

manics:
`tests/venv/binder-dir` and `tests/venv/start/postBuild` are failing for an unrelated reason (also failing on main):
` KeyError: 'The Python module ipyleaflet is not a valid nbextension, it is missing the _jupyter_nbextension_paths() method.'`

minrk:
ipyleaflet 0.19 refactored some things, and `ipyleaflet` is no longer a valid extension package (jupyter_leaflet is). I'm not sure if this is intentional, but this command is still in their installation docs, so I'm guessing it is a mistake.

This `nbextension` step has not been required since notebook 5.3, so these tests (especially postBuild) aren't really testing that they do anything.

manics (approved):

PR #1348, opened 2024-05-26
update postBuild tests without ipyleaflet

manics:
`tests/venv/binder-dir/binder` and `tests/venv/start/postBuild` are currently failing with ipyleaflet 0.19:
```
KeyError: 'The Python module ipyleaflet is not a valid nbextension, it is missing the `_jupyter_nbextension_paths()` method.'
```

### DIFF
--- a/tests/venv/binder-dir/binder/postBuild
+++ b/tests/venv/binder-dir/binder/postBuild
@@ -1,3 +1,2 @@
 #!/bin/bash
-# FIXME: This doesn't work with ipyleaflet >= 0.19
-jupyter nbextension enable --py --sys-prefix ipyleaflet
+date
--- a/tests/venv/binder-dir/binder/postBuild
+++ b/tests/venv/binder-dir/binder/postBuild
@@ -1,2 +1,2 @@
 #!/bin/bash
-date
+echo 'Done!' > $HOME/postbuild.txt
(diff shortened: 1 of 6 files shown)


tests/venv/binder-dir/binder/postBuild
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-jupyter nbextension enable --py --sys-prefix ipyleaflet
+date
```
minrk, line 2: I think for this test we need a postBuild that has an effect we can test for, so we are verifying that the script executed

manics: Good point, I've changed this to create a file that can be checked in verify

manics:
I've created an issue for the remaining failing test (Julia) https://github.com/jupyterhub/repo2docker/issues/1349

tests/venv/binder-dir/verify
```diff
@@ -4,3 +4,5 @@ which gfortran
 test -z $(pip list | grep scipy)
 jupyter nbextension list
 jupyter nbextension list | grep 'jupyter-leaflet' | grep enabled
+
+grep 'Done!' $HOME/postbuild.txt
```
minrk, line 8: Makes sense! I think this change is needed in start/verify as well

PR #1355, opened 2024-06-28
Docker engine: replace ContainerConfig with Config

manics:
`ContainerConfig` has been removed, https://docs.docker.com/engine/deprecated/#container-and-containerconfig-fields-in-image-inspect says to use `Config`.

AFAIK the presence or absence of this field depends on the Docker host daemon, not docker-py.

Closes https://github.com/jupyterhub/repo2docker/issues/1354


yuvipanda:
Can confirm that I see `Config.WorkingDir`

yuvipanda:
I'll make a release shortly

PR #1365, opened 2024-08-16
exclude defaults channel by default

minrk:
make sure we're always using conda-forge by default

it's unlikely any packages came from defaults for users in some years (due to channel priority, this would only happen for packages not on conda-forge, and would often result in a broken env anyway), but this ensures defaults channel is opt-in, as it should be for the Anaconda ToS.

### DIFF
--- a/repo2docker/buildpacks/conda/install-base-env.bash
+++ b/repo2docker/buildpacks/conda/install-base-env.bash
@@ -30,7 +30,6 @@ export PATH="${PWD}/bin:$PATH"
 cat <<EOT >> ${CONDA_DIR}/.condarc
 channels:
   - conda-forge
-  - defaults
 auto_update_conda: false
 show_channel_urls: true
 update_dependencies: false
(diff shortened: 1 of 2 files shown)


rgaiacs:
What do you think of repo2docker to fail if [user's provided `environment.yml`](https://repo2docker.readthedocs.io/en/latest/config_files.html#environment-yml-install-a-conda-environment) uses default Anaconda repository? Maybe have an option for this behaviour for the special case of mybinder.org.

minrk:
I don't believe it is up to us. We could block egress to repo.anaconda.com, I suppose, but I don't think we need to. A user may install from the defaults channel and use it on mybinder.org if their use is permitted. mybinder.org doesn't enter into it, just like they can do on any cloud provider.

rgaiacs:
Looks good to me.

rgaiacs:
Renato Alves mentioned in the German RSE list that

> At the moment 3 channels are firewall blocked across all EMBL sites: defaults, main and r. All these channels are under the same terms of service.

@minrk can we update this PR to include all three channels? `defaults`, `main` and `r`.

minrk:
@rgaiacs I'm not sure what you mean. We had 2 channels (`conda-forge`, `defaults`), now there is 1 (`conda-forge`). `main` and `r` are not in the available channels list in config.

rgaiacs:
@minrk looks good. I need more coffee. :-D

minrk:
Because this removes an unused repodata request for the `defaults` channel, this should speed up installs ever so slightly.

PR #1376, opened 2024-10-28
julia_project.py: fix Pkg REPL api warning

fonsp:
Repo2docker currently prints this warning when using a Julia project:

```
Warning: The Pkg REPL mode is intended for interactive use only, and should not be used from scripts. It is recommended to use the functional API instead.
└ @ Pkg.REPLMode /srv/julia/share/julia/stdlib/v1.10/Pkg/src/REPLMode/REPLMode.jl:382
[0m  at 12:24:22
```

This PR fixes that. I also used `Pkg.instantiate()` instead of `Pkg.precompile()` because it makes a bit more sense in this context: we want to guarantee that packages are instantiated (which precompiles at the end), not just that they are precompiled: https://pkgdocs.julialang.org/v1/api/#Pkg.instantiate

I believe that `resolve` already instantiates (and precompiles), but this is not defined in spec so it is good to add the `instantiate` call.

### DIFF
--- a/repo2docker/buildpacks/julia/julia_project.py
+++ b/repo2docker/buildpacks/julia/julia_project.py
@@ -177,7 +177,7 @@ class JuliaProjectTomlBuildPack(PythonBuildPack):
                 "${NB_USER}",
                 r"""
             JULIA_PROJECT="" julia -e "using Pkg; Pkg.add(\"IJulia\"); using IJulia; installkernel(\"Julia\", \"--project={project}\");" && \
-            julia --project={project} -e 'using Pkg; Pkg.instantiate(); Pkg.resolve(); pkg"precompile"'
+            julia --project={project} -e 'using Pkg; Pkg.instantiate(); Pkg.resolve(); Pkg.instantiate()'
             """.format(
                     project=self.project_dir
                 ),


GeorgianaElena (approved):

agoose77:
Hi @fonsp, thanks for your contribution!

@GeorgianaElena and I took a look at this PR today, and confirmed that the call to `instantiate()` eliminates the console warning.

As we were unfamiliar with the history of these lines of code, we investigated whether the `instantiate(); resolve(); instantiate()` chain was strictly necessary, and started to explore this. As we time-boxed our work on this PR, we ultimately decided that further exploring was out of scope here. And we will follow up with an issue.

So, approving & merging this PR!

fonsp:
Thanks for the response! 

I believe https://github.com/jupyterhub/repo2docker/issues/1383 is nice! It should save about 1sec from the startup time. From my Julia experience (not a Pkg developer but heavy user): The extra `resolve` call is not necessary, unless the repo contains an improperly resolved Manifest (**very rare**, might happen when a repository contains a subfolder that is added as dependency, and something changed in the subfolder without the user updating their main Manifest). If so, there are some cases that will get fixed by `instantiate`, and others that will fail and show a nice error message in the binder build log, which is fine! (It will encourage people to just fix their Manifest.)

While in the Julia mood, maybe you could take a look at https://github.com/jupyterhub/repo2docker/issues/1375? This should help with stability and reproducibility of Julia projects!

PR #1384, opened 2024-12-08
Use self.log.warning instead of warnings.warn

yuvipanda:
warnings.warn sidesteps all the json logging bits we do, and repo2docker will produce non-json output that binderhub then struggles to consume.

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -16,7 +16,6 @@ import shutil
 import sys
 import tempfile
 import time
-import warnings
 from urllib.parse import urlparse
 
 import entrypoints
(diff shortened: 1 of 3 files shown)


yuvipanda:
Hmm, so the errors aren't actually coming from this - they're coming from somewhere else. Debugging...

yuvipanda:
I *think* this should fix the tests!

manics:
Can you also remove the `import warnings`? I'm surprised this isn't picked up by one of the linters.

yuvipanda:
@manics yea i noticed yesterday we don't actually have a flake8 linter here and that means there's actually a ton of unused imports! As well as things like #1392

manics (approved):

PR #1388, opened 2024-12-14
Resolve DOI more cleanly

yuvipanda:
Using doi.org, we only care to find out *where* the doi is pointing to. We don't need to go fetch the contents of that page fully.

By doing this, we:

1. Make doi resolution much faster, as we now make only 1 request per doi resolution than two!
2. Handle circumstances where dataverse (or other providers) restrict requests from datacenter IPs for subsets of URLs that serve HTML, rather than API requests (https://jupyter.zulipchat.com/#narrow/channel/103349-ask-anything/topic/Binder.20Dataverse.20error)

### DIFF
--- a/repo2docker/contentproviders/doi.py
+++ b/repo2docker/contentproviders/doi.py
@@ -62,7 +62,7 @@ class DoiProvider(ContentProvider):
                 # default Git provider as this leads to a misleading error.
                 self.log.error(f"DOI {doi} does not resolve: {e}")
                 raise
-            return resp.headers['Location']
+            return resp.headers["Location"]
         else:
             # Just return what is actulally just a URL
             return doi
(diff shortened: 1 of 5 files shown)


repo2docker/contentproviders/doi.py
```diff
@@ -49,7 +49,9 @@ def doi2url(self, doi):
             doi = normalize_doi(doi)
 
             try:
-                resp = self._request(f"https://doi.org/{doi}")
+                # We don't need to fetch the *contents* of the page the doi resolves to -
+                # only need to know what it redirects to.
+                resp = self._request(f"https://doi.org/{doi}", allow_redirects=False)
```
manics, line 54: Are there any rules on whether a DOI can redirect multiple times?

yuvipanda: @manics great question that led me down a rabbithole! I'm now convinced this approach is not robust enough, and will open a different series of PRs - starting with https://github.com/jupyterhub/repo2docker/pull/1389, and https://github.com/jupyterhub/repo2docker/pull/1390 is a WIP

yuvipanda:
Taking a different approach in https://github.com/jupyterhub/repo2docker/pull/1390

PR #1389, opened 2024-12-16
Stop mocking dataverse contentprovider test

yuvipanda:
As I was debugging https://github.com/jupyterhub/repo2docker/pull/1388, I realized that PR actually broke the dataverse provider, but the existing test was mocking so much that we didn't actually catch it!

IMO, since the point of contentproviders is to integrate with external content providers, they should be integration tests so we can catch issues with them more easily. Integration tests would have caught https://jupyter.zulipchat.com/#narrow/channel/103349-ask-anything/topic/Binder.20Dataverse.20error more cleanly than how it happened for example.

It also immediately showed me that some of the results we were using to test were not accurate, and I've updated the test fixtures to match what we actually expect them to return.

I will try to see if we can use a demo or test instance for the fetch test though, so we don't screw up download stats even more for the existing test doi we use.


yuvipanda:
@pdurbin is there a different DOI we can use for the 'fetch' test that doesn't screw up your stats even more? :)

tests/contentproviders/test_dataverse.py
```diff
@@ -0,0 +1,67 @@
+import hashlib
+import os
+from tempfile import TemporaryDirectory
+
+import pytest
+
+from repo2docker.contentproviders import Dataverse
+
+test_dv = Dataverse()
+harvard_dv = next(_ for _ in test_dv.hosts if _["name"] == "Harvard Dataverse")
```
manics, line 10: Can you add a comment explaining what this iterates through, and what we end up comparing in the test? Are the names `Harvard Dataverse`/`CIMMYT Research Data` considered immutable properties of repository?

yuvipanda: @manics I'll do so, although I didn't add these - they were already there in the existing file.

yuvipanda: @manics I've incorporated this into https://github.com/jupyterhub/repo2docker/pull/1390, and simply removed using "host" this way completely.

yuvipanda:
Closing in favor of https://github.com/jupyterhub/repo2docker/pull/1390

PR #1393, opened 2024-12-21
Only use requests, not a mix of requests & urlopen

yuvipanda:
Sort-of a follow up to https://github.com/jupyterhub/repo2docker/pull/993

We were:

1. In some cases, directly using requests
2. In some cases, directly using the stdlib's urlopen
3. In some cases, had a method named urlopen that simply passed things through to requests

This is unnecessarily confusing, and seems to primarily be done for the benefit of mocking the network calls. However, as described in the recently merged https://github.com/jupyterhub/repo2docker/pull/1390, I don't think mocking is appropriate here as it means we don't actually catch problems.

This PR mostly focuses on getting unifying to only using requests directly with as little indirection as possible. If any tests were directly using mocks here, they will be replaced with something that is testing things more directly as appropriate

### DIFF
--- a/repo2docker/contentproviders/ckan.py
+++ b/repo2docker/contentproviders/ckan.py
@@ -25,7 +25,7 @@ class CKAN(ContentProvider):
         Borrowed from the Hydroshare provider.
         """
         package_show_url = f"{api_url}package_show?id={self.dataset_id}"
-        resp = self.urlopen(package_show_url).json()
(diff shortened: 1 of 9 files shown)


rgaiacs:
@yuvipanda what is the status of this? https://github.com/jupyterhub/repo2docker/pull/1488 was also related with hydroshare.

PR #1397, opened 2025-01-21
Repo2Docker outdated for CaltechDATA

RohanBhattaraiNP:
Updated the API calling for properly working.


rgaiacs:
Hi @RohanBhattaraiNP,

Thanks for the pull request.

The test that we have for CaltechDATA is working, see https://github.com/jupyterhub/repo2docker/blob/cb2675769379490cbb4b8885030c068cfe65e80f/tests/unit/contentproviders/test_zenodo.py#L40-L47

Could you provide more details regarding this pull request? Thanks!

PR #1400, opened 2025-01-27
Upgrade Shiny server to 2024.12

rgaiacs:


manics (approved):
Tests are passing but I don't know how thorough they are- have you tested this yourself?

rgaiacs:
> have you tested this yourself?

I will test with some friends that use R and report back.

manics:
I found some time to do a basic test of this and it seems fine. Though more testing from real users is always useful!

PR #1395, opened 2025-01-15
BUG: COPY --chmod=0755 /usr/local/bin/{python3-login,repo2docker-entrypoint for rootless podman

westurner:
BUG: COPY `--chmod=0755` /usr/local/bin/{python3-login,repo2docker-entrypoint for rootless podman

Command this is building with (though repo2podman doesn't yet support volumes)

```sh
repo2docker --engine=podman -E -P --debug -e NB_USER=jovyan -e NB_UID=1000 --user-name=jovyan .
```

PR #1402, opened 2025-01-29
Shell out to `docker buildx build` to build images

yuvipanda:
pydocker uses the HTTP API (equivalent to `docker build`) which has been dead for ages. `docker buildx` (the interface to BuildKit) is what we *should* be using, but alas pydocker does not currently support it and I suspect it never will (https://github.com/docker/docker-py/issues/2230).

This PR simply shells out, as I think that's the way.

I'm going to list some intentional choices I've made here.

1. Intentionally replace the build functionality of the existing `DockerEngine` than make a new one. The old way is [deprecated](https://github.com/docker/buildx/issues/2297) and we should not support it.
2. Continue using `dockerpy` for all other non-build operations. It works fine and I don't see a need to change it right now.
3. Simply extract the tarball we construct elsewhere. This is kinda double work, but we can treat this as a possible optimization pathway in the future.
4. Currently ignore `container_limits` (because buildx doesn't support it, it should be set at the level of the builder instead)

## Questions to answer

- What do we do about container_limits here?
- We probably need to deprecate / throw errors on some existing traitlets as they won't really have any effect here. What traitlets are those?
- We should treat this as a breaking change and make a release after, right?

## Traitlets to deprecate

- `Repo2Docker.build_memory_limit` - this must be instead set up by the builder

Fixes https://github.com/jupyterhub/repo2docker/issues/875

### DIFF
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -141,7 +141,7 @@ def get_argparser():
     argparser.add_argument(
         "--build-memory-limit",
         # Removed argument, but we still want to support printing an error message if this is passed
-        help=argparse.SUPPRESS
+        help=argparse.SUPPRESS,
     )
 
     argparser.add_argument(
@@ -437,7 +437,10 @@ def make_r2d(argv=None):
     if args.build_memory_limit:
         # We no longer support build_memory_limit, it must be set in the builder instance
         print("--build-memory-limit is no longer supported", file=sys.stderr)
-        print("Use `docker buildx create` to create a custom builder with appropriate memory limits instead", file=sys.stderr)
+        print(
+            "Use `docker buildx create` to create a custom builder with appropriate memory limits instead",
+            file=sys.stderr,
+        )
         sys.exit(-1)
 
     if args.environment and not r2d.run:
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -438,7 +438,7 @@ def make_r2d(argv=None):
         # We no longer support build_memory_limit, it must be set in the builder instance
         print("--build-memory-limit is no longer supported", file=sys.stderr)
         print(
-            "Use `docker buildx create` to create a custom builder with appropriate memory limits instead",
+            "Check your build engine documentation to set memory limits. Use `docker buildx create` if using the default builder to create a custom builder with appropriate memory limits",
             file=sys.stderr,
         )
         sys.exit(-1)
--- a/repo2docker/__main__.py
+++ b/repo2docker/__main__.py
@@ -140,7 +140,8 @@ def get_argparser():
 
     argparser.add_argument(
         "--build-memory-limit",
-        help="Total Memory that can be used by the docker build process",
+        # Removed argument, but we still want to support printing an error message if this is passed
+        help=argparse.SUPPRESS
     )
 
     argparser.add_argument(
@@ -434,12 +435,10 @@ def make_r2d(argv=None):
         sys.exit(1)
 
     if args.build_memory_limit:
-        # if the string only contains numerals we assume it should be an int
-        # and specifies a size in bytes
-        if args.build_memory_limit.isnumeric():
-            r2d.build_memory_limit = int(args.build_memory_limit)
-        else:
-            r2d.build_memory_limit = args.build_memory_limit
+        # We no longer support build_memory_limit, it must be set in the builder instance
+        print("--build-memory-limit is no longer supported", file=sys.stderr)
+        print("Use `docker buildx create` to create a custom builder with appropriate memory limits instead", file=sys.stderr)
+        sys.exit(-1)
 
     if args.environment and not r2d.run:
         print("To specify environment variables, you also need to run " "the container")
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -165,12 +165,19 @@ class Repo2Docker(Application):
     build_memory_limit = ByteSpecification(
         0,
         help="""
-        Total memory that can be used by the docker image building process.
+        Unsupported.
 
-        Set to 0 for no limits.
+        When using docker, please use `docker buildx create` to create a new buildkit
+        builder with appropriate limits instead.
         """,
         config=True,
     )
+
+    @observe("build_memory_limit")
+    def build_memory_limit_changed(self, change):
+        print("Setting build_memory_limit is not supported", file=sys.stderr)
+        print("Use `docker buildx create` to create a custom builder with appropriate memory limits instead", file=sys.stderr)
+        sys.exit(-1)
 
     volumes = Dict(
         {},
@@ -856,6 +863,7 @@ class Repo2Docker(Application):
                     for l in picked_buildpack.build(
                         docker_client,
                         self.output_image_spec,
+                        # This is deprecated, but passing it anyway to not break backwards compatibility
                         self.build_memory_limit,
                         build_args,
                         self.cache_from,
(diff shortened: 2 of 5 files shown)


yuvipanda:
This isn't complete, putting it up here to see how many tests fail.

minrk (commented):
Nice! This will also unblock using `COPY --chmod` as in #1395

Re: `extra_build_args`, presumably we do want to keep an equivalent passthrough for extra args to pass to `buildx build` on the command-line (a list this time)?

repo2docker/docker.py
```diff
@@ -94,22 +99,43 @@ def build(
...
-            labels=labels,
-            platform=platform,
-            **kwargs,
-        )
+        if not shutil.which("docker"):
+            raise RuntimeError("The docker commandline client must be installed")
+        args = ["docker", "buildx", "build", "--progress", "plain", "--load"]
+        if buildargs:
+            for k, v in buildargs.items():
+                args += ["--build-arg", f"{k}={v}"]
```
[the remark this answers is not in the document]
yuvipanda: @minrk I think it's better to keep it through this way, so we can also print a 'here is the buildx command that was used' debug line. buildx has `--secret` that we don't support yet but we could for passing actual secrets.

yuvipanda:
@minrk I added `extra_buildx_build_args` in what feels like the right place for it. Will need to figure out a test.

repo2docker/docker.py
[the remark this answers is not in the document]
minrk: Makes sense!

repo2docker/docker.py
```diff
@@ -69,6 +74,14 @@ class DockerEngine(ContainerEngine):
         config=True,
     )
 
+    extra_buildx_build_args = List(
+        [],
```
minrk, line 78: If you want to specify the type of items in the list, give a single TraitType instance here (not list):
```suggestion
        Unicode(),
```

minrk:
> Will need to figure out a test.

[iidfile](https://docs.docker.com/reference/cli/docker/buildx/build/#:~:text=iidfile) looks like it might be a good arg to test with, since it just creates a file to check. Or `--check` appears to skip the build entirely, so should be quick.

minrk:
Here's a small test for extra buildx args with `--check` that goes really quick since it doesn't actually build the image:

```
python3 -m repo2docker --no-run '--DockerEngine.extra_buildx_build_args=["--check"]' .
```

which outputs (among other things):

> Check complete

tests/unit/test_app.py
```diff
@@ -78,21 +78,6 @@ def test_local_dir_image_name(repo_with_content):
     )
```
minrk, line 80: here's a test to exercise the new buildx args:

```suggestion

def test_extra_buildx_build_args(repo_with_content):
    upstream, sha1 = repo_with_content
    argv = ['--DockerEngine.extra_buildx_build_args=--check', upstream]
    app = make_r2d(argv)
    with patch("repo2docker.docker.execute_cmd") as execute_cmd:
        app.build()
        
    args, kwargs = execute_cmd.call_args
    cmd = args[0]
    assert cmd[:3] == ["docker", "buildx", "build"]
    # make sure it's inserted before the end
    assert "--check" in cmd[:-1]


```

repo2docker/docker.py
yuvipanda: Done!

yuvipanda:
I think there's an extra linebreak injected in every line now, which i'll figure a way out of

repo2docker/docker.py
```diff
@@ -69,6 +74,14 @@ class DockerEngine(ContainerEngine):
         config=True,
     )
 
+    extra_buildx_build_args = List(
```
manics, line 77: Is it worth pushing this up a level to `ContainerEngine` under a more generic name? It shouldn't be a breaking change since it'll be ignored by default.

yuvipanda:
I was wrong about `Repo2Docker.extra_build_args - as this is purely about pydocker's build method that we no longer use` - that's actually BUILD ARGS haha

yuvipanda:
Also, instead of having a separate `builder` traitlet, we can pass a different dockerx builder if needed with the generic extra args.

I have made this intentionally fail if --build-memory-limit is passed, which I think is the right call here. It clearly requires action from the admin, as their memory usage will no longer be limited.

repo2docker/docker.py
```diff
@@ -94,22 +99,43 @@ def build(
...
-            custom_context=custom_context,
-            decode=True,
-            dockerfile=dockerfile,
-            fileobj=fileobj,
-            path=path,
-            labels=labels,
-            platform=platform,
-            **kwargs,
-        )
+        if not shutil.which("docker"):
```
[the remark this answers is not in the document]
yuvipanda: Yeah, we'll have to see how it goes. But buildkit does give us a lot more knobs here - https://docs.docker.com/build/cache/backends/.

yuvipanda:
I actually don't think we need to do any newline mangling here. I think this is good to go!

@manics @minrk lmk if you think this needs anything else :) When we make a CHANGELOG, i can add a BREAKING entry

minrk (approved):
Nice! This is ok to go for me

repo2docker/app.py
```diff
@@ -165,13 +165,23 @@ def _default_log_level(self):
...
+        builder with appropriate limits instead.
         """,
         config=True,
     )
 
+    @observe("build_memory_limit")
+    def build_memory_limit_changed(self, change):
+        print("Setting build_memory_limit is not supported", file=sys.stderr)
+        print(
+            "Use `docker buildx create` to create a custom builder with appropriate memory limits instead",
```
manics, line 180:
```suggestion
            "Check your build engine documentation to set memory limits. Use `docker buildx create` if using the default builder to create a custom builder with appropriate memory limits",
```

manics:
> I was wrong about Repo2Docker.extra_build_args - as this is purely about pydocker's build method that we no longer use - that's actually BUILD ARGS haha

If `Repo2Docker.extra_build_args` is already passed as [`build_args`](https://github.com/jupyterhub/repo2docker/blob/b0cfd45b14b1ac36ab0485d5ac74ab729a99492d/repo2docker/app.py#L837) do we still need `extra_buildx_build_args`?

yuvipanda:
@manics those are for build args that are declared via `ARG` in dockerfile, rather than additional arguments to the build process! so yeah i do think we need that.

repo2docker/__main__.py
```diff
@@ -434,12 +435,13 @@ def make_r2d(argv=None):
...
-        # if the string only contains numerals we assume it should be an int
-        # and specifies a size in bytes
-        if args.build_memory_limit.isnumeric():
-            r2d.build_memory_limit = int(args.build_memory_limit)
-        else:
-            r2d.build_memory_limit = args.build_memory_limit
+        # We no longer support build_memory_limit, it must be set in the builder instance
+        print("--build-memory-limit is no longer supported", file=sys.stderr)
+        print(
+            "Use `docker buildx create` to create a custom builder with appropriate memory limits instead",
```
manics, line 441:
```suggestion
            Check your build engine documentation to set memory limits. Use `docker buildx create` if using the default builder to create a custom builder with appropriate memory limits,
```
I didn't notice this message was duplicated- otherwise this looks fine to merge!

yuvipanda:
Exciting. I'll merge and test / deploy ideally tomorrow.

PR #1403, opened 2025-01-30
add triplets for r versions 4.3, 4.4

minrk:
otherwise r-4.3 won't work

perhaps we should error when a minor version isn't found in the dict (or use `x.y.0`), because installing R will fail with a 404 trying to download .deb if it's not a 3-part version.


repo2docker/buildpacks/r.py
```diff
@@ -97,6 +99,9 @@ def r_version(self):
                 # available. Users can however explicitly specify the full version to get something specific
                 if r_version in version_map:
                     r_version = version_map[r_version]
+                elif len(r_version.split(".")) == 1:
```
manics, line 102:
```suggestion
                elif len(r_version.split(".")) == 2:
```
Shouldn't this be 2?

minrk:
Thanks, now fixed and tested

yuvipanda:
ty @minrk

PR #1406, opened 2025-02-09
Stop using datetime.utcnow() to fix warning

yuvipanda:
```
/Users/yuvipanda/code/repo2docker/./repo2docker/buildpacks/conda/freeze.py:101: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
```

### DIFF
--- a/repo2docker/buildpacks/conda/freeze.py
+++ b/repo2docker/buildpacks/conda/freeze.py
@@ -12,7 +12,7 @@ python freeze.py [3.8]
 import os
 import pathlib
 from argparse import ArgumentParser
-from datetime import datetime
+from datetime import datetime, UTC
 from subprocess import check_call
 
 from ruamel.yaml import YAML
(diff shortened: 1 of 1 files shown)


repo2docker/buildpacks/conda/freeze.py
```diff
@@ -12,7 +12,7 @@
 import os
 import pathlib
 from argparse import ArgumentParser
-from datetime import datetime
+from datetime import UTC, datetime
```
minrk, line 15: This is `datetime.timezone.UTC`. And `timezone.UTC` is a class attribute, so not directly importable which means you have to do something like

```suggestion
from datetime import datetime, timezone
UTC = timezone.UTC
```

if you want UTC as a top-level variable.

PR #1410, opened 2025-02-12
Add basic UI Playwright tests

manics:
Adds a Playwright test to make it easier to review changes to RStudio and RShiny. Currently we only check they can be installed, we don't check they can be run.


manics:
I've added another `external-verify` step, but it appears as `tests/ui/browser/verify::verify`
```
tests/ui/browser/verify::build PASSED                                    [ 25%]
tests/ui/browser/verify::verify PASSED                                   [ 50%]
tests/ui/browser/verify::verify PASSED                                   [ 50%]
/opt/hostedtoolcache/Python/3.13.1/x64/lib/python3.13/site-packages/coverage/control.py:892: CoverageWarning: No data was collected. (no-data-collected)
  self._warn("No data was collected.", slug="no-data-collected")
tests/ui/browser/verify::check-tmp PASSED                                [ 75%]
```

Is there an easy way to have `::external-verify` shown instead?

tests/conftest.py
```diff
@@ -246,6 +259,17 @@ def collect(self):
         args.append(str(self.path.parent))
         yield Repo2DockerTest.from_parent(self, name="build", args=args)
 
+        # If external-verify exists it should be run on the host
+        external_verify_script = self.path.parent / "external-verify"
+        if external_verify_script.exists():
+            yield Repo2DockerTest.from_parent(
+                self,
+                name=self.path.name,
```
minrk, line 267: This sets the test name for the external-verify (currently 'verify'):

```suggestion
                name=external_verify_script.name,
```

minrk (approved):

PR #1411, opened 2025-02-12
Require Python 3.9 to run repo2docker

manics:
We're making a breaking release:
https://github.com/jupyterhub/repo2docker/pull/1402
so we can use the opportunity to add other breaking changes. This only affects the Python version required to run repo2docker, it doesn't change the Python versions supported in environments.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -29,7 +29,7 @@ on:
 
 jobs:
   docker:
-    runs-on: ubuntu-22.04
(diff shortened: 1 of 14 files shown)


yuvipanda (approved):

yuvipanda:
@manics actually, given python 3.8 is *already* EOL, can we bump this to 3.9?

minrk:
👍 to setting 3.9

manics:
Done!

PR #1413, opened 2025-02-14
Simulate json output from docker buildx build

yuvipanda:
We still use the docker python api to push images, and to run. So we need to set string_output to false, and simulate the json output on buildx. Alternatively we could completely get rid of the python API, but that's a much bigger lift (especially around registry credentials) that I don't want us to do right now.

Without this, push progress is reported as raw bytes in the binderhub UI

### DIFF
--- a/repo2docker/docker.py
+++ b/repo2docker/docker.py
@@ -58,7 +58,7 @@ class DockerEngine(ContainerEngine):
     https://docker-py.readthedocs.io/en/4.2.0/api.html#module-docker.api.build
     """
 
-    string_output = True
+    string_output = False
 
     extra_init_args = Dict(
         {},
@@ -141,12 +141,18 @@ class DockerEngine(ContainerEngine):
 
                 args += [d]
 
-                yield from execute_cmd(args, True)
+                for line in execute_cmd(args, True):
+                    # Simulate structured JSON output from buildx build, since we
+                    # do get structured json output from pushing and running
+                    yield {"stream": line}
         else:
             # Assume 'path' is passed in
             args += [path]
 
-            yield from execute_cmd(args, True)
+            for line in execute_cmd(args, True):
+                # Simulate structured JSON output from buildx build, since we
+                # do get structured json output from pushing and running
+                yield {"stream": line}
 
     def images(self):
         images = self._apiclient.images()


minrk:
Is there a simple test that would have caught this?

I was able to test manually by running a local registry:

```
docker run --rm 127.0.0.1:9000:5000 registry
```

and then pushing to it:

```
python3 -m repo2docker --no-run --json-logs --image-name=localhost:9000/test/image .
```

before:

```
{"message": "Reusing existing image (localhost:9000/test/image), not building."}
{"message": "b'{\"status\":\"The push refers to repository [localhost:9000/test/image]\"}\\r\\n'", "phase": "pushing"}
{"message": "b'{\"status\":\"Preparing\",\"progressDetail\":{},\"id\":\"9ffc2f2e8fb8\"}\\r\\n{\"status\":\"Preparing\",\"progressDetail\":{},\"id\":\"d12bcb450636\"}\\r\\n{\"status\":\"Preparing\",\"progressDetail\":{},\"id\":\"c20414e1266b\"}\\r\\n{\"status\":\"Preparing\",\"progressDetail\":{},\"id\":\"6838097029a0\"}\\r\\n{\"status\":\"Preparing\",\"progressDetail\":{},\"id\":\"3ec51339a507\"}\\r\\n'", "phase": "pushing"}
```

after:

```

but I'm not sure how to test this practically without launching a local registry for the test.

**Review by minrk — APPROVED** [review]

File: repo2docker/docker.py [hunk 1]
```diff
@@ -141,12 +141,18 @@ def build(
 
                 args += [d]
 
-                yield from execute_cmd(args, True)
+                for line in execute_cmd(args, True):
+                    # Simulate structured JSON output from buildx build, since we
+                    # do get structured json output from pushing and running
+                    yield {"stream": line}
```
manics, line 147: Is `line` guaranteed to never contain characters that need to be escaped?

yuvipanda: @manics not that I know of! It's also fundamentally the same process as what we were getting from the python client, just through subprocess instead.

yuvipanda: I'll try out a bunch and keep an eye on it though!

yuvipanda:
image

Seems to work fine with this!

PR #1417, opened 2025-02-20
Bump base image used to latest LTS

yuvipanda:
Moves from 22.04 to 24.04

### DIFF
--- a/repo2docker/app.py
+++ b/repo2docker/app.py
@@ -459,7 +459,7 @@ class Repo2Docker(Application):
     )
 
     base_image = Unicode(
-        "docker.io/library/buildpack-deps:jammy",
+        "docker.io/library/buildpack-deps:24.04",
         config=True,
         help="""
         Base image to use when building docker images.


yuvipanda:
@manics thoughts on rolling this into the release?

manics (approved):
Sounds fine to me. Can you add a note to https://repo2docker.readthedocs.io/en/latest/howto/breaking_changes.html even if it's just to mention the same guide applies? Could you also post on https://discourse.jupyter.org/ so that anyone providing support for users knows there's a potential breaking change?

weiji14:
Couple of comments:
1. Do we need to handle https://github.com/jupyterhub/repo2docker/issues/1346 somehow? E.g. by overriding the default 'ubuntu' user with 'jovyan'? Or is that for downstream repos to handle?
2. There are still a few references to 'Jammy Jellyfish (22.04)' in the repo (e.g. from this [search](https://github.com/search?q=repo%3Ajupyterhub%2Frepo2docker%20Jammy&type=code)), might need to update them to 'Noble Numba (24.04)'.

minrk:
I applied @manics proposal and added a test for it. Since @GeorgianaElena released 2025.12, I think we are all set to merge this one now.

PR #1419, opened 2025-02-25
maint: Adapt mamba activation

jjerphan:
In regards of [the depreciation and future removal of `etc/profile.d/mamba.sh` from `libmamba`'s distribution](https://github.com/mamba-org/mamba/pull/3788/files#diff-21512f6a332b63c284b7d1ac549d08d5b827c56c5882aaa677ede79e8fcd65c4R1-R3).

### DIFF
--- a/repo2docker/buildpacks/conda/activate-conda.sh
+++ b/repo2docker/buildpacks/conda/activate-conda.sh
@@ -1,11 +1,24 @@
-# enable conda and activate the notebook environment
-set -e
+set -ex
+
+# Setup conda
+CONDA_PROFILE="${CONDA_DIR}/etc/profile.d/conda.sh"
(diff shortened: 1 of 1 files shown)


minrk (approved):
Makes sense, I suppose, if the nice, simple consistency with conda is being removed. This makes mamba more complicated and less convenient to use, so I don't understand the reasoning behind removing the feature in favor of more complex requirements. No reason is given in the [motivating issue](https://github.com/mamba-org/mamba/issues/2524).

jjerphan:
Let's discuss it over https://github.com/mamba-org/mamba/issues/2524 (I think we need to get more feedback on this).

yuvipanda:
I've received multiple user reports of this being an issue, so merging this for now.

PR #1421, opened 2025-02-26
Switch to using CLI for everything except running the container

yuvipanda:
This PR originally was to try to convert just `push` to using the CLI, but expanded to
make some changes to how we were using docker-py for anything other than running
containers. Running containers should also be switched to using the CLI, but in a separate
PR.

1. Remove `find_images` method in the ContainerEngine interface, as it was
    only being used inefficiently to iterate through all images to check if a single
    image exists. Instead we now use inspect_image (which can return a None
    if the image does not exist). This is a breaking change for other container engines
2. Remove the separate `push` method from the ContainerEngine interface. This is a 
    breaking change.
3. Add `push` and `load` params to `build`. `push` will automatically push the resulting
    image to the registry, while `load` will ensure it's loaded onto the local image store and be
    ready to run. `push` never is called without a build in our codebase, so this is fine. It allows us
    to not load the image into the local image store when *only* pushing, as is the case with binderhub.
    `load` is only set if the image is to be run by repo2docker.
4. Use the CLI for  pushing images.
5. Change mocked tests of find_image to 'real' tests
6. Add a proper push test, including running an 'insecure' registry with a separate DIND to push to.
7. Validate that `registry_credentials` is in the right form

For authentication, we preserve existing authentication information handling behavior:

1. If registry_credentials are present, they are used but not leaked
    on to existing ~/.docker/config
2. If registry_credentials are not present, they are not used
3. Regardless of registry_credentials being present, we still will use
    existing authentication info in DOCKER_CONFIG

This authentication behavior is tested as well!

## Removed functionality

This does remove the functionality of testing at start time if docker api is accessible, and failing if it is not. This is because we now have two ways to access the docker api - cli and python, and I don't want to repeatedly do it for both (plus it was causing the registry test to fail, even though that only used the cli). I propose we let this be, and add it back once we get rid of dockerpy completely.

## TODO

- [x] Use stdin to pass password to docker login
- [x] Fix tests
- [x] Validate registry_credentials

Fixes https://github.com/jupyterhub/repo2docker/issues/1414


repo2docker/docker.py
```diff
@@ -162,10 +160,42 @@ def inspect_image(self, image):
...
+        with tempfile.TemporaryDirectory() as d:
+            new_dc_path = Path(d) / "config.json"
+            if dc_path.exists():
+                # If there is an existing DOCKER_CONFIG, copy it to new location so we inherit
+                # whatever configuration the user has already set
+                shutil.copy2(dc_path, new_dc_path)
+
+            env = os.environ.copy()
+            subprocess.check_call(
+                # FIXME: This should be using --password-stdin instead
```
manics, line 179: It might be easier to manipulate the `config.json` file directly:
```python
json.dumps({"auths":{"registry.host":{"auth":b64encode(f"{user}:{token}".encode()).decode()}}})
```

yuvipanda: @manics I did consider that, but then we'll have to merge the existing ~/.docker/config file with the new setup carefully. I would rather just outsource it to `docker login` like this instead.

yuvipanda:
The failing tests are because of the mock for the existing push. This is also why we couldn't catch https://github.com/jupyterhub/repo2docker/pull/1413

yuvipanda:
So, I'm just going to run a docker registry and test. Unfortunately while docker push works fine with registries on http, docker login does not without us changing docker daemon config to allow `insecure-registries`. So just for this test, I'm going to have to run a custom dind as well to test :D but totally doable though.

yuvipanda:
ok, moving `run` to CLI is a lot more work haha so i'm going to hold off on that, and try to scope this down.

yuvipanda:
I'm going to scope this back to just trying to implement docker push via the CLI again :)

yuvipanda:
@minrk @manics ok, this is ready for review now! \o/

minrk (approved):
makes sense. What update does repo2podman need to be compatible? Support `push`, `load` kwargs in build and that's it?

yuvipanda:
This has been deployed to mybinder.org and has been working fine!

It's still exporting to the local container image store by default because that's the behavior of the default buildkit backend.

PR #1428, opened 2025-06-24
Move `runtime.txt` parsing into base class

manics:
There's a discussion somewhere (I can't remember where) about using https://github.com/astrofrog/pypi-timemachine or similar so Python packages can be installed as if `pip` were run in the past, similar to R. If we do this we should define the cutoff date for Python in the same way as for R, which is to use `runtime.txt`.

Something to consider is we haven't formally defined what `runtime.txt` contains.

Our docs state it should be one of:
- `name-version-yyyy-mm-dd`: (`r-<RVERSION>-<YYYY>-<MM>-<DD>`)
- `name-version`: (`python-x.y`)
https://github.com/jupyterhub/repo2docker/blob/e795060aec3555a6203ed79c3676765fb3f3b850/docs/source/config_files.rst?plain=1#L201-L207

Our tests say it can also be:
- name-yyyy-mm-dd
https://github.com/jupyterhub/repo2docker/blob/e795060aec3555a6203ed79c3676765fb3f3b850/tests/unit/test_r.py#L38

This PR handles all three formats. I've made a breaking change and redefined the existing `.runtime` property instead of creating a new one to avoid leaving an unused `.runtime` property that may cause confusion in the future.

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased breaking changes
+
+`RBuildPack.runtime` previously returned the contents of `runtime.txt` as a string.
+It has been replaced by `BuildPack.runtime` which returns a tuple `(name, version, date)`.
+
 ## 2024.07.0
 
 ([full changelog](https://github.com/jupyterhub/repo2docker/compare/2024.03.0...2024.07.0))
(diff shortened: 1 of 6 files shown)


docs/source/changelog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased breaking changes
```
rgaiacs, line 3: I believe this will require a rebase given that we made a release in August 2025, see https://github.com/jupyterhub/repo2docker/releases/tag/2025.08.0.

minrk (approved):
Nice!

rgaiacs:
@manics do you want to press the green button?

PR #1432, opened 2025-07-09
Add --vanilla to R call

rgaiacs:
Closes https://github.com/jupyterhub/repo2docker/issues/1431


manics (approved):
It's been a month and no-one's raised any concrete objection on https://github.com/jupyterhub/repo2docker/issues/1431

PR #1436, opened 2025-08-08
Add note on root users and make postbuild actions more discoverable

choldgraf:
This does two things:

- Adds a page about postbuild and start workflows, so it's a bit more discoverable.
- Adds a note about root privileges and suggestions for what to do.

cc @chuckwondo

### DIFF
--- a/docs/source/use/actions-and-scripts.md
+++ b/docs/source/use/actions-and-scripts.md
@@ -3,6 +3,7 @@
 In addition to installing using [standard configuration files](#config-files), you can run your own custom actions and commands as part of the build or user session launch process.
 
 (commands-jovyan)=
+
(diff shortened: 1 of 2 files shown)


rgaiacs (approved):

docs/source/use/actions-and-scripts.md
```diff
@@ -0,0 +1,39 @@
+# Run scripts, commands, and actions as part of environment building
+
+In addition to installing using [standard configuration files](#config-files), you can run your own custom actions and commands as part of the build or user session launch process.
```
yuvipanda, line 3:
```suggestion
In addition to installing packages using [standard configuration files](#config-files), you can run your own custom actions and commands as part of the build or user session launch process.
```

PR #1444, opened 2025-08-16
Support pyproject.toml for pip install in Python buildpack

rgaiacs:
Related to https://github.com/jupyterhub/repo2docker/issues/1427

~This is not yet ready for review!~

The code is based on the [Pipfile buildpack](https://github.com/jupyterhub/repo2docker/blob/main/repo2docker/buildpacks/pipfile/__init__.py). We probably want to refactor some portions to avoid code duplication.


rgaiacs:
All tests are passing after support to Python < 3.11 was dropped.

rgaiacs:
I rebased this pull request and I made minor changes to reduce the code duplication. It need to be reviewed again.

repo2docker/buildpacks/pyproject/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+    def detect(self):
+        """Check if current repo should be built with the pyproject.toml buildpack."""
+        # first make sure python is not explicitly unwanted
+        name, _, _ = self.runtime
+        if name != "python":
+            return False
+
+        pyproject_file = self.binder_path("pyproject.toml")
+
+        return os.path.exists(pyproject_file)
```
minrk, line 135: As discussed [here](https://github.com/jupyterhub/repo2docker/issues/1427#issuecomment-2909112717),  we should not consider the presence of `pyproject.toml` sufficient. We should return False if pyproject.toml does not contain both `project` and `build-system.requires`.

repo2docker/buildpacks/pyproject/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+        if os.path.exists(pyproject_file):
+            with open(pyproject_file, "rb") as _pyproject_file:
+                pyproject_toml = tomllib.load(_pyproject_file)
+
+            if (
+                ("project" in pyproject_toml)
+                and ("build-system" in pyproject_toml)
+                and ("requires" in pyproject_toml["build-system"])
+            ):
+                return os.path.exists(pyproject_file)
```
minrk: this condition is always true at this point, we can avoid checking it again:

```suggestion
                return True
```

minrk:
I'm not sure I quite follow - The `detect` you have checks for the presence `build-system` and `project`, which is what we should do. This same condition should be applied to whether to invoke `pip install`.

We could isolate this into a single `_is_python_package` method which checks `setup.py` present _OR_ `pyproject.toml` contains `build-system`, and use that where we need this information, which it looks like is in two places.

This is because `pyproject.toml` can be used to configure tools (ruff, black, pytest, etc.) for repos that are _not_ packages. So it's not so much a safety net as it is an _accuracy_ of the check: `pyproject.toml` alone is insufficient information to decide if `pip install` should be called, you have to check its contents to see if it's a package or another kind of project directory.

repo2docker/buildpacks/python/__init__.py
```diff
@@ -34,6 +40,35 @@ def python_version(self):
...
+                    match = re.compile(r"\d+(\.\d+)*").match(
+                        raw_pyproject_minimum_version
+                    )
+                    if match:
+                        pyproject_minimum_version = match.group()
+                        pyproject_minimum_version_info = (
+                            pyproject_minimum_version.split(".")
+                        )
+
+                        if (py_version_info[0] < pyproject_minimum_version_info[0]) or (
```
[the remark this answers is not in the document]
rgaiacs: Thanks for the feedback. I'm finishing my day and I'm on holiday the rest of the week. I can pick this on Monday.

repo2docker/buildpacks/python/__init__.py
```diff
@@ -17,23 +26,42 @@ def python_version(self):
 
         name, version, _ = self.runtime
 
-        if name != "python" or not version:
-            # Either not specified, or not a Python runtime (e.g. R, which subclasses this)
+        if name is not None and name != "python":
+            # Either not a Python runtime (e.g. R, which subclasses this)
```
minrk:
```suggestion
            # Runtime specified, but not Python (e.g. R, which subclasses this)
```

repo2docker/buildpacks/python/__init__.py
```diff
@@ -17,23 +26,42 @@ def python_version(self):
...
+                pyproject = tomllib.load(_pyproject_file)
+
+            if "project" in pyproject and "requires-python" in pyproject["project"]:
+                # This is the minumum version!
+                # https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires
+                pyproject_python_specifier = SpecifierSet(
+                    pyproject["project"]["requires-python"]
+                )
+
+                if runtime_version not in pyproject_python_specifier:
```
minrk: If the default version doesn't satisfy, we should probably pick the lowest version available rather than error, right? But only if runtime.txt is _not_ specified?


```python
# pseudocode
if runtime_version not in pyproject_python_specifier:
    if version:
        raise ValueError(
                        f"Python version {version} in runtime.txt does not satisfy {pyproject_python_specifier} in pyproject.toml."
                    )
    else:
        # version not specified in runtime.txt, pick the lowest available satisfying version
        for python in sorted(self.major_pythons):
            if python in pyproject_python_specifier:
                runtime_version = python
                break
        else:
            raise ValueError(f"Unable to find a supported Python version satisfying {pyproject_python_specifier}")
```

repo2docker/buildpacks/python/__init__.py
```diff
@@ -17,23 +26,42 @@ def python_version(self):
...
             return self._python_version
 
-        py_version_info = version.split(".")
-        py_version = ""
-        if len(py_version_info) == 1:
-            py_version = self.major_pythons[py_version_info[0]]
+        if name is None or version is None:
+            self._python_version = self.major_pythons["3"]
+            runtime_version = Version(self.major_pythons["3"])
+            self.log.warning(
```
minrk: If we decide to go with my suggestion below, we should perhaps only show this warning if we don't end up picking a version based on pyproject.toml.

PR #1456, opened 2025-08-22
Fail earlier when user try to install future version of R

rgaiacs:
Closes https://github.com/jupyterhub/repo2docker/issues/1455


rgaiacs:
> What's the situation that leads us here? Can we distinguish between runtime.txt being absent vs having the version and missing the date?

If `runtime.txt` is absent, the values will be `None` because of https://github.com/jupyterhub/repo2docker/blob/176674867eab9c96d70f2ca2e4d0df80afcafdcf/repo2docker/buildpacks/base.py#L777-L778

From https://github.com/jupyterhub/repo2docker/blob/176674867eab9c96d70f2ca2e4d0df80afcafdcf/repo2docker/buildpacks/base.py#L793-L797 my understanding is that the data is mandatory but the date might be invalid and raise an exception.

rgaiacs:
Can this be merged? Or should we talk about it during https://github.com/jupyterhub/team-compass/issues/814?

PR #1458, opened 2025-08-25
build: migrate from setup.py to pyproject.toml

lixiname:
- Replace legacy setup.py with standardized pyproject.toml
- Follow PEP 517 (build system) and PEP 621 (metadata)
- Use setuptools as build backend with modern configuration
- add pyproject.toml file、test_setuptools.py  file (a test scripts of toml)、Modify  __init__.py  file
This improves dependency management, IDE support, and aligns with current Python packaging standards.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,80 @@
+[build-system]
+requires = ["setuptools>=61.0", "wheel", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "jupyter-repo2docker"
+dynamic = ["version"]
+description = "Repo2docker: Turn code repositories into Jupyter enabled Docker Images"
+readme = "README.md"
+license = {text = "BSD"}
+authors = [
+    {name = "Project Jupyter Contributors", email = "jupyter@googlegroups.com"}
+]
+maintainers = [
+    {name = "Project Jupyter Contributors", email = "jupyter@googlegroups.com"}
+]
+keywords = ["reproducible", "science", "environments", "docker"]
+classifiers = [
+    "Development Status :: 4 - Beta",
+    "Environment :: Console",
+    "Intended Audience :: Developers",
+    "Intended Audience :: System Administrators",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -0,0 +1,80 @@
+[build-system]
+requires = ["setuptools>=61.0", "wheel", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "jupyter-repo2docker"
+dynamic = ["version"]
+description = "Repo2docker: Turn code repositories into Jupyter enabled Docker Images"
```
rgaiacs, line 8: Replace `Repo2docker` with `repo2docker`

rgaiacs (changes requested):
This looks a great start. All the tests are failing, probably because of the new test. Can change in the GitHub actions configuration be change to increase the version of Python to be used?

pyproject.toml
```diff
@@ -0,0 +1,69 @@
...
+    "Intended Audience :: Developers",
+    "Intended Audience :: System Administrators",
+    "Intended Audience :: Science/Research",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Topic :: Scientific/Engineering",
+    "Topic :: Software Development",
+    "Topic :: System :: Systems Administration",
+]
+requires-python = ">=3.9"
```
rgaiacs: Should we change, **in a separate pull request**, the CI job to run the tests with Python 3.9 so that we don't introduce code that only run in Python 3.10 or higher?

minrk: We [have that already](https://github.com/jupyterhub/repo2docker/blob/2456a4dc917f36248b6128907299570e94f87e7f/.github/workflows/test.yml#L72-L75).

But we should probably also require 3.10, soon.

minrk:
latest run succeeded in 12 minutes. We should probably have a per-test limit, though, to avoid this problem costing us 6 hours of CI time

PR #1460, opened 2025-08-25
add support for R 4.5.1

shaneknapp:
this should resolve https://github.com/jupyterhub/repo2docker/issues/1459

please let me know if there's anything else i need to do for this PR.  thanks in advance!


yuvipanda:
Can you change https://github.com/jupyterhub/repo2docker/blob/main/tests/r/r4.4-rspm/runtime.txt to explicitly test R 4.5? Rename the dir + modify runtime.txt I think should do?

PR #1464, opened 2025-09-01
Always call `ContainerEngine.build()` when `push`ing

manics:
This is so that if the image is built but not pushed it will still be pushed

The default `DockerEngine.build()` will always rebuild, it's not possible to only push.

Closes https://github.com/jupyterhub/repo2docker/issues/1463


rgaiacs (approved):
Looks good to me.

PR #1475, opened 2025-10-15
Add Python 3.14

rgaiacs:
Closes https://github.com/jupyterhub/repo2docker/issues/1474

~Please, don't merge it yet. It is missing a step that will be faster when I run tomorrow.~

This pull request includes changes created with

```bash
cd repo2docker/buildpacks/conda
python freeze.py
```


rgaiacs:
All tests pass. My understanding is that use of Python 3.14 is opt-in.

minrk:
Awesome! Great that this worked smoothly and so quickly.

PR #1479, opened 2025-10-22
fetch dataverse installations at runtime

minrk:
rather than a statically encoded file not updated very often

This is the same URL fetched on the official https://dataverse.org/installations page

allows override in case URLs change


minrk:
came up working on #1458 because the previous update used a custom setuptools entrypoint, which we are going to lose soon. Part of the reason the URL is overrideable is because the URL our entrypoint used doesn't resolve anymore.

PR #1481, opened 2025-10-22
require Python 3.10

### LINKED ISSUE #1480: Drop support for Python 3.9
Because Python 3.9 will reached end of support in October 2025, see https://www.python.org/downloads/ and [PEP 596](https://www.python.org/dev/peps/pep-0596).

Image

During JupyterHub and JupyterBook Collab Cafe on [21 October](https://github.com/jupyterhub/team-compass/issues/831), we briefly mentioned that we should make a new release for `repo2docker`. Maybe we should make a last release with support for Python 3.9 before drop support.

minrk:
remove unneeded args to pre-commit flags, which should read version targets from pyproject.toml

avoiding args in pre-commit means invoking the same tools directly doesn't produce different results

closes #1480

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,8 +32,8 @@ repos:
       - id: isort
 
   # Autoformat: markdown
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v4.0.0-alpha.8
+  - repo: https://github.com/rbubley/mirrors-prettier
+    rev: v3.6.2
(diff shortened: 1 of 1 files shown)


manics:
Since you're modifying pre-commit I think this is a good time to replace prettier with `rbubley/mirrors-prettier@v3.6.2`
https://github.com/jupyterhub/team-compass/issues/793

minrk:
@manics good call, done.

PR #1489, opened 2025-11-24
Don't error out if there's no disk space left when writing the logs

GeorgianaElena:
## Context

In https://github.com/2i2c-org/infrastructure/pull/7077 we investigated how we could allow user servers to start even if the respective users have reached their storage quota, i.e. they cannot use more space from the disk where their home directories are mounted to.

The results were that user images built with `repo2docker` action that didn't have a Dockerfile where unable to start because writing the logs failed:
```
Defaulted container "notebook" out of: notebook, block-cloud-metadata (init), block-nfs-access (init)
Activating profile: /srv/conda/etc/profile.d/conda.sh
/srv/conda/envs/notebook/lib/python3.11/subprocess.py:1016: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  self.stdout = io.open(c2pread, 'rb', bufsize)
INFO: Executing provided command
  File "/usr/local/bin/repo2docker-entrypoint", line 114, in <module>
    main()
  File "/usr/local/bin/repo2docker-entrypoint", line 95, in main
    tee(chunk)
  File "/usr/local/bin/repo2docker-entrypoint", line 81, in tee
    f.flush()
OSError: [Errno 28] No space left on device
```

## Motivation
For context, the motivation is that If they cannot start servers when they reach the quota, they won't be able to delete unneeded files and would need to be unblocked by someone with higher privileges to either bump their quota, or delete unwanted files. 

## What changes

This PR mostly does the same thing, except that if the logs cannot be written to `REPO_DIR` or `cwd` because of lack of space, it will end up writing the logs to `stdout` + `tmp` location or just `stdout` without raising any errors that would stop the servers from starting.

### DIFF
--- a/repo2docker/buildpacks/repo2docker-entrypoint
+++ b/repo2docker/buildpacks/repo2docker-entrypoint
@@ -45,7 +45,7 @@ def main():
             break
     # raise Exception if log_file could not be set
     if log_file is None:
-        raise Exception("Could not open '.jupyter-server-log.txt' log file " )
+        raise Exception("Could not open '.jupyter-server-log.txt' log file")
 
     # build the command
     # like `exec "$@"`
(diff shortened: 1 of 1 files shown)


yuvipanda:
> The secondary tee-to-file is nonessential functionality, so it seems okay to lose those messages rather than keep trying to find a destination file on every log chunk.

Generally +1 to this, it's ok to treat this as non-essential functionality and just fail

PR #1527, opened 2026-04-17
try to fix dependabot for Dockerfile

minrk:
- exclude repo2docker itself from requirements.txt because dependabot messes with paths (see #1522 and related PRs)
- group dependabot PRs
- apply safer cooldown default

### DIFF
--- a/scripts/refreeze
+++ b/scripts/refreeze
@@ -32,6 +32,7 @@ def refreeze():
     """
 
     alpine_version = get_alpine_version()
+
     run(
         [
             "docker",
@@ -48,7 +49,8 @@ def refreeze():
         python3 -m venv /tmp/venv
         source /tmp/venv/bin/activate
         pip install pip-tools
-        pip-compile
+        # exclude repo2docker itself because dependabot messes it up
+        pip-compile --unsafe-package=jupyter-repo2docker
         """,
         ]
     )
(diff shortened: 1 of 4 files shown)


scripts/refreeze
```diff
@@ -48,7 +49,8 @@ def refreeze():
         python3 -m venv /tmp/venv
         source /tmp/venv/bin/activate
         pip install pip-tools
-        pip-compile
+        # exclude repo2docker itself because dependabot messes it up
+        pip-compile --upgrade --unsafe-package=jupyter-repo2docker
```
manics, line 53:
```suggestion
        pip-compile --upgrade --no-allow-unsafe --unsafe-package=jupyter-repo2docker
```
The behaviour will change at some point https://pip-tools.readthedocs.io/en/latest/cli/pip-compile/
>      --allow-unsafe / --no-allow-unsafe
>                                  Pin packages considered unsafe: distribute,
>                                  pip, setuptools.
>                                  
>                                  WARNING: Future versions of pip-tools will
>                                  enable this behavior by default. Use --no-
>                                  allow-unsafe to keep the old behavior. It is
>                                  recommended to pass the --allow-unsafe now
>                                  to adapt to the upcoming change.

manics (approved):

PR #1528, opened 2026-04-17
R: update shiny-proxy to 1.4

minrk:
hopefully closes #1520

PR #1530, opened 2026-04-17
R: use `remotes::local_install` instead of deprecated `devtools::install_local`

minrk:
`devtools::install_local` doesn't work anymore due to missing dependency on `remotes`, and is deprecated since 2.5.0.

We are not using an independently pinned versions of devtools, despite indications in the code that we appear to intend to. I'm not sure if this loss of pinning is intentional, since I'm not an R expert.

I also don't know the possible implications of this change for repos with older versions. This is just the command the deprecation message pointed me to.

### DIFF
--- a/repo2docker/buildpacks/_r_base.py
+++ b/repo2docker/buildpacks/_r_base.py
@@ -12,7 +12,7 @@ def rstudio_base_scripts(r_version):
 
     # Shiny server (not the package!) seems to be the same version for all R versions
     shiny_server_url = "https://download3.rstudio.org/ubuntu-18.04/x86_64/shiny-server-1.5.22.1017-amd64.deb"
-    shiny_proxy_version = "1.3"
+    shiny_proxy_version = "1.4"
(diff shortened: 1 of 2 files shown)


minrk:
I found that we have two choices:

1. use [`pak::install_local()`](https://pak.r-lib.org/reference/get-started.html?q=install_local#install-local-package) which seems to unavoidably leave cache and tmp files for us to cleanup by hand, but appears to be the recommended 'new way' to do it, or
2. use [`remotes::install_local()`](https://remotes.r-lib.org/reference/install_local.html) which seems lighter weight, and is the missing dependency for the deprecated `devtools::local_install` that has stopped working, and doesn't seem to leave extra files around.

minrk:
huh, cleaning out the cache seems to result in the package not being available with `pak`, so I guess `remotes` is the simpler choice.

minrk:
this (including #1528) fixes the R tests, #1529 fixes unit tests, with both I think we're all green again.

manics (approved):

PR #1539, opened 2026-05-05
Use "docker push" instead of "docker build --push"

### LINKED ISSUE #1536: podman build doesn't support --load or --push
### Bug description

When trying to build using podman with `--push`, you get an error `unknown flag: --push`.  Since commit 07e97ee5 repo2docker tries to use `podman buildx build --push` but this isn't supported in podman. Reverting this commit allows podman to get a bit further (there are still other issues I'm trying to work out).

If podman is no longer supported, instead the documentation should be updated.

rgaiacs:
This improves `repo2docker` compatibility with Podman.

Closes https://github.com/jupyterhub/repo2docker/issues/1536

@minrk This PR needs a little more work.

- [x] `docker push` needs the correct name of the container image

cc @dylex

### DIFF
--- a/repo2docker/docker.py
+++ b/repo2docker/docker.py
@@ -164,9 +164,6 @@ class DockerEngine(ContainerEngine):
                 )
             args.append("--load")
 
-        if push:
-            args.append("--push")
-
         if buildargs:
             for k, v in buildargs.items():
                 args += ["--build-arg", f"{k}={v}"]
@@ -207,6 +204,9 @@ class DockerEngine(ContainerEngine):
                 args += [path]
 
                 yield from execute_cmd(args, True)
+
+            if push:
+                yield from execute_cmd([self.container_cli, "push", tag], True)
 
     def inspect_image(self, image):
         """
--- a/repo2docker/docker.py
+++ b/repo2docker/docker.py
@@ -189,8 +189,6 @@ class DockerEngine(ContainerEngine):
         args += self.extra_buildx_build_args
 
         with ExitStack() as stack:
-            if self.registry_credentials:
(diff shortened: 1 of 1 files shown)


dylex:
This looks reasonable and works for me as-is in a simple test.

The `docker_login` call could now be moved down and made conditional on `push` as well.

(The old `Repo2Docker.push_image` method used `client.push` and a lot of handling for progress reporting, but that's probably not necessary.)

repo2docker/docker.py
```diff
@@ -164,9 +164,6 @@ def build(
                 )
             args.append("--load")
 
-        if push:
-            args.append("--push")
```
minrk, line 168: Since we're using `docker buildx build`, I thought we needed `--push` *or* `--load` in the build, or it won't go anywhere? Has that changed? I could be misremembering, but [this suggests it at least _was_ the case](https://stackoverflow.com/questions/74707530/docker-buildx-fails-to-show-result-in-image-list)

I think we might need to check for podman here and do this only if it's not actually docker.

rgaiacs:
> I thought we needed --push or --load in the build

```bash
$ docker image ls | grep repo2docker | wc -l
0
```

Given a minimal `Dockerfile` like

```
FROM alpine:3.23.5

RUN apk add --no-cache python3
```

After run

```bash
$ docker buildx build --tag repo2docker .
```

we observe

```bash
$ docker image ls | grep repo2docker | wc -l
1
```

In my environment, I have Docker version 29.5.3, build d1c06ef.

The [mentioned question in Stack Overflow](https://stackoverflow.com/q/74707530) is from 2022. I believe that Docker received a patch to fix the reported issue.

minrk:
Thanks for confirming!

PR #1543, opened 2026-05-18
roll back to mamba 2.5.0

minrk:
2.6.0/2.6.1 seems to introduce some solver regressions leading to unsolvable and/or much slower solves in repo2docker use cases that work better with 2.5 (repo2docker solves are all upgrades of a base env rather than single-env solve&install). 

our existing downgrade test is much slower (15 minutes vs 4), and https://github.com/ecmwf/notebook-examples doesn't seem to complete (https://github.com/jupyterhub/mybinder.org-deploy/issues/3803) which works fine on 2.5.0.

https://github.com/mamba-org/mamba/pull/4270 may fix this, but we should check runtime for our downgrade test and https://github.com/ecmwf/notebook-examples when we try 2.6.2

cc @SylvainCorlay @jjerphan

### DIFF
--- a/repo2docker/buildpacks/conda/install-base-env.bash
+++ b/repo2docker/buildpacks/conda/install-base-env.bash
@@ -5,10 +5,12 @@ set -ex
 
 cd $(dirname $0)
 
-export MAMBA_VERSION="2.6.1"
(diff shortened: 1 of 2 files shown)


minrk:
conda tests completed in 14 minutes, down from 24 with 2.6.1

PR #1544, opened 2026-05-19
Quote LABEL key/value in Dockerfile template

### LINKED ISSUE #123: do not remove container after stop
Is there a way to not remove the container after it is stopped?

gistrec:
## Summary

Adds quoting on the LABEL key/value interpolation in `repo2docker/buildpacks/base.py` so that newlines, double-quotes, or shell metacharacters in operator-supplied label values (`--Repo2Docker.labels` / `Repo2Docker(labels=...)`) don't break the generated Dockerfile.

The template used to render:

```jinja
LABEL {{k}}="{{v}}"
```

with no escaping. A realistic break-the-build case is a CI pipeline that passes `git log -1 --pretty=%B` into a label — a multi-line commit message produces a Dockerfile that no longer parses.

This is not a vulnerability — repo2docker trusts the operator who configures the CLI flag — but the quoting cleanup was suggested by @minrk in the discussion of [GHSA-pv57-6gv8-cc38](https://github.com/jupyterhub/repo2docker/security/advisories/GHSA-pv57-6gv8-cc38):

> "no vulnerabilities here, but quoting on labels should be improved. A PR would be welcome adding shlex.quote to labels"

## Approach

The first revision of this PR used `shlex.quote`, as suggested. On closer inspection that turned out to be the wrong tool, so this now uses Dockerfile quoting instead. `shlex.quote` produces *shell* quoting, but a LABEL directive is never parsed by a shell — it goes through the Dockerfile parser and then the builder's own lexer, which has different rules. Three concrete problems:

1. **Backslashes were doubled.** `shlex.quote` single-quotes anything containing a metacharacter, and inside Dockerfile single quotes *no escape processing happens at all* (`processSingleQuote` has no escape branch). So the `\` → `\\` escaping applied before quoting reached the image verbatim: a value of `windows\path` landed as `windows\\path`.
2. **`$VAR` expansion was silently turned off.** LABEL is in `replaceEnvAllowed`, so `LABEL k="$FOO"` expands today. Single-quoting suppresses that — a quiet behaviour change for any value containing a dollar sign.
3. **Keys could stop being keys.** `shlex.quote` returns "safe" strings unquoted, so a key like `--flag` rendered as `LABEL --flag=...`, which buildx rejects as an unknown flag.

Instead, both key and value are now encoded as a Dockerfile double-quoted word:

```python
def _dockerfile_quote(value):
    if not isinstance(value, str):
        value = str(value)
    escaped = (
        value.replace("\\", "\\\\")
        .replace('"', '\\"')
        .replace("$", "\\$")
        .replace("\n", "\\n")
        .replace("\r", "\\r")
    )
    return f'"{escaped}"'
```

Per `processDoubleQuote` in BuildKit's `frontend/dockerfile/shell/lex.go` — and the byte-for-byte identical function in the Docker 17.09 classic builder, which the template comment still targets — that lexer unescapes **only** `\`, `"` and `$`:

```go
case '"', '$', sw.escapeToken:
    // These chars can be escaped, all other \'s are left as-is
```

So escaping exactly those three characters is necessary and sufficient for an arbitrary value to round-trip unchanged.

Rendered output for a typical label is unremarkable:

```dockerfile
LABEL "org.opencontainers.image.source"="https://github.com/foo/bar"
```

## Known limitations

- **Newlines are lossy.** A Dockerfile word cannot carry a newline: the parser is line-oriented, `\n` is not an escape sequence, and heredocs aren't permitted on LABEL. Newlines are flattened to a literal backslash-n. That is still the point of the PR — a multi-line commit message lands as one token instead of injecting a `RUN` — but the value is not byte-identical. The only way to preserve a real newline would be to pass labels out of band via the build API, as `DockerBuildPack` already does (`labels=self.get_labels()` in `buildpacks/docker.py`). Happy to do that as a follow-up if you'd prefer it to escaping.
- **A key containing `=` is unrepresentable.** `parseNameVal` splits on the first `=` of the *raw, still-quoted* word, so no amount of quoting helps. Previously this corrupted silently (key `a`, value `b=...`); now it fails loudly at build time.
- The encoding assumes the default `\` escape token; the template sets no `# escape=` directive.

## Changes

- `repo2docker/buildpacks/base.py`
  - new `_dockerfile_quote` helper (replaces `_quote_dockerfile_token`; the `shlex` import is gone again)
  - `render()` constructs a `jinja2.Environment` with a `dockerfile_quote` filter wired in (replacing `jinja2.Template(TEMPLATE)`)
  - the LABEL line in `TEMPLATE` applies it to both key and value
- `tests/unit/test_buildpack.py` — parametrised regression test that now asserts the **decoded** value, not just the absence of an injected directive. It includes a reference implementation of `processDoubleQuote` so the expected semantics are written down next to the assertions. Cases: newline injection, `" ; RUN echo pwned ; LABEL x="`, embedded double quotes, backtick + `$VAR`, Windows paths, trailing backslash, empty string, non-ASCII.
- `tests/unit/test_labels.py` — expectations updated for the new `LABEL "k"="v"` shape (keys are quoted too, so a key containing whitespace can't start a directive of its own).

## Test plan

- [x] `pytest tests/unit/test_buildpack.py tests/unit/test_labels.py` → 28 passed
- [x] Full `tests/unit` run: failure set is identical to the pre-PR baseline, i.e. this branch introduces no new failures
- [x] Fuzzed 20k random values (quotes, backslashes, `$`, backticks, tabs, CR, newlines, non-ASCII) through the **real BuildKit parser** via the `dockerfile` Python bindings, decoding with the lexer rules above — 20000/20000 round-trip exactly, modulo the documented newline flattening
- [x] `pyupgrade` / `isort` / `black` clean

Refs: https://github.com/jupyterhub/repo2docker/security/advisories/GHSA-pv57-6gv8-cc38

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -4,7 +4,6 @@ import io
 import logging
 import os
 import re
-import shlex
 import string
 import sys
 import tarfile
@@ -16,21 +15,36 @@ import jinja2
 from docker.utils.build import exclude_paths
 
 
-def _quote_dockerfile_token(value):
-    """Quote a label key/value for inclusion in a Dockerfile LABEL directive.
+def _dockerfile_quote(value):
+    """Encode a string as a double-quoted word for a Dockerfile LABEL directive.
 
-    Wraps :func:`shlex.quote` with an explicit escape of backslash, ``\\n`` and
-    ``\\r`` so the result occupies a single Dockerfile directive line.
-    Dockerfile single-quoted strings (which :func:`shlex.quote` produces for
-    values containing shell metacharacters) do not span newlines the way shell
-    does, so a label value containing a literal newline -- e.g., a multi-line
+    A LABEL line is processed in two passes: the parser splits it into words,
+    keeping quotes and escapes verbatim, and the shell lexer then strips the
+    quotes and resolves the escapes. Inside double quotes that lexer unescapes
+    only ``\\``, ``"`` and ``$`` (every other backslash is left alone) and
+    expands ``$VAR``, so escaping exactly those three characters is what makes
+    an arbitrary value survive unchanged.
+
+    A Dockerfile word cannot carry a newline at all: the parser is line
+    oriented and ``\\n`` is not an escape sequence. Newlines are therefore
+    flattened to a literal backslash-n, so that a multi-line value -- e.g. a
     ``git commit`` message piped through ``--Repo2Docker.labels`` in CI --
-    would otherwise break out of the LABEL directive.
+    lands as a single token instead of breaking out of the LABEL directive and
+    injecting further directives.
+
+    Note this assumes the default ``\\`` escape token; the template does not
+    set a ``# escape=`` parser directive.
     """
     if not isinstance(value, str):
         value = str(value)
-    escaped = value.replace("\\", "\\\\").replace("\n", "\\n").replace("\r", "\\r")
-    return shlex.quote(escaped)
+    escaped = (
+        value.replace("\\", "\\\\")
+        .replace('"', '\\"')
+        .replace("$", "\\$")
+        .replace("\n", "\\n")
+        .replace("\r", "\\r")
+    )
+    return f'"{escaped}"'
 
 
 # Only use syntax features supported by Docker 17.09
@@ -197,7 +211,7 @@ COPY --chown={{ user }}:{{ user }} src/ ${REPO_DIR}/
 # Put these at the end, since we don't want to rebuild everything
 # when these change! Did I mention I hate Dockerfile cache semantics?
 {% for k, v in labels|dictsort %}
-LABEL {{ k|sh_quote }}={{ v|sh_quote }}
+LABEL {{ k|dockerfile_quote }}={{ v|dockerfile_quote }}
 {%- endfor %}
 
 # We always want containers to run as non-root
@@ -512,7 +526,7 @@ class BuildPack:
         build_args = build_args or {}
 
         env = jinja2.Environment()
-        env.filters["sh_quote"] = _quote_dockerfile_token
+        env.filters["dockerfile_quote"] = _dockerfile_quote
         t = env.from_string(TEMPLATE)
 
         build_script_directives = []
--- a/repo2docker/buildpacks/base.py
+++ b/repo2docker/buildpacks/base.py
@@ -4,6 +4,7 @@ import io
 import logging
 import os
 import re
+import shlex
 import string
 import sys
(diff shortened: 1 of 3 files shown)


gistrec:
Rebooting this one — CI is green again (25/25).

Two things changed since the last review round.

**CI was red because of an existing test, and fixing it properly meant dropping `shlex.quote`.** `test_labels.py` asserts `LABEL first_label="firstlabel"`, and `shlex.quote` leaves "safe" strings unquoted, so the rendered line became `LABEL first_label=firstlabel`. Rather than just relax that assertion, I looked at why the shape changed — and `shlex.quote` turned out to be the wrong tool for this, despite being what the advisory thread suggested. A LABEL is never parsed by a shell:

1. `shlex.quote` single-quotes anything with a metacharacter, and inside Dockerfile single quotes there is no escape processing at all — so the `\` → `\\` escaping done before quoting reached the image verbatim. `windows\path` became `windows\\path`.
2. Single quotes also suppress `$VAR` expansion, which LABEL does perform today (`command.Label` is in `replaceEnvAllowed`) — a silent behaviour change.
3. `shlex.quote("--flag")` returns it bare, so a key like that rendered as `LABEL --flag=...` → buildx `unknown flag`.

So both key and value are now encoded as a Dockerfile double-quoted word, escaping exactly `\`, `"` and `$` — which is what `processDoubleQuote` in `frontend/dockerfile/shell/lex.go` unescapes, and nothing else:

```go
case '"', '$', sw.escapeToken:
    // These chars can be escaped, all other \'s are left as-is
```

The same function is byte-for-byte identical in the Docker 17.09 classic builder, so the template's "only 17.09 syntax" comment still holds.

**The tests now check the value, not just the absence of injection.** They decode the rendered token back with those lexer rules and compare to the input, so a mangled-but-safe value fails the test. I also fuzzed 20k random values (quotes, backslashes, `$`, backticks, tabs, CR, newlines, non-ASCII) through the real BuildKit parser via the `dockerfile` Python bindings: 20000/20000 round-trip exactly, apart from newlines.

Newlines are the one irreducible loss — a Dockerfile word can't carry one, so they're flattened to a literal `\n`. That still achieves the original goal (a multi-line commit message lands as one token instead of injecting a `RUN`), but if you'd rather have exact values, the alternative is to pass labels out of band via the build API the way `DockerBuildPack` already does — `BaseImage.build()` doesn't currently pass `labels=` to `client.build()`. Happy to switch to that instead if you prefer; it would delete this escaping problem rather than solve it.

PR #1546, opened 2026-06-02
Test mamba 2.7.0.alpha0

JohanMabille:


minrk:
Seems to have fixed the issue that motivated downgrading in #1543 

Slowest test (downgrade):

- current (2.5): 170s
- this PR (2.7): 160s
- reverted (2.6): 965s

overal conda tests down from 14:39 to 13:20

so I think we should be safe to upgrade to 2.7 when it's ready.

jjerphan:
Here we go: https://github.com/jupyterhub/repo2docker/pull/1547.

PR #1565, opened 2026-08-12
Add an extended buildpack

yijunge-ucb:
To address https://github.com/jupyterhub/roadmap/issues/2


repo2docker/buildpacks/extended/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+from functools import lru_cache
+
+from ..base import BuildPack
+
+# Matches an explicit `oci-image:<image>[:<tag>]` reference in runtime.txt.
+# The image reference itself is registry-agnostic: Docker Hub, GCR, GAR,
+# ECR, ACR, or a self-hosted registry are all just opaque strings after the
+# prefix, so there's no need to special-case any particular registry's
+# hostname format here.
+_OCI_IMAGE_RE = re.compile(r"^oci-image:\s*(?P<image>.+?)\s*$")
```
minrk, line 14: Maybe this should be an oci URL, rather than a special prefix?

```
oci://registry/image:tag
```

minrk (commented):
Nice!

To get the tests running, can you add `extended` to the matrix [here](https://github.com/jupyterhub/repo2docker/blob/808ae6bb888511b5464d9405abcef8f26e38bd1b/.github/workflows/test.yml#L58)?

I expect some of these tests are going to take a long time.

While it is useful that it looks like they are real-world use cases, it might help to have somewhat simpler test cases that don't install so many things. In particular, large numbers of R packages can take a really long time to install in tests, when we really only need to verify that `install.R` got picked up.

Can you also add some documentation [here](https://github.com/jupyterhub/repo2docker/blob/808ae6bb888511b5464d9405abcef8f26e38bd1b/docs/source/configuration/system.md#runtimetxt---specifying-runtimes) to cover the new `runtime.txt` option with some examples?

My only implementation issue is the amount of duplication. I anticipate that being a possible issue, but maybe something to be solved in a future refactor PR, rather than here.

PR #1567, opened 2026-08-19
add support for R 4.6

### LINKED ISSUE #1566: Add support for R 4.6
### Proposed change

Add support for R 4.6.x for repo2docker builds.

Hub users using R. ref https://github.com/NASA-IMPACT/science-support/issues/48

Should be a low lift. Making a PR now.

sunu:
Fixes https://github.com/jupyterhub/repo2docker/issues/1566.

ref https://github.com/NASA-IMPACT/science-support/issues/48#issuecomment-5338367378 cc @wildintellect


repo2docker/buildpacks/r.py
sunu, line 78: Should this be updated to a newer version too?

rgaiacs: No. There were some previous discussion to avoid change the default R version very often.

rgaiacs (approved):
It looks good to me. Thanks for the contribution.

repo2docker/buildpacks/r.py
wildintellect: How often was too often? in the R world major.minor.patch is a actually a big deal on the minor part. Minor is about [once per year ](https://cran.r-project.org/bin/windows/base/old/) and often has really important changes.

Also 4.3 had a major security issue, and should be actively discouraged (though I can understand wanting some backwards compatibility support)

rgaiacs: Maybe @yuvipanda or @minrk can provide you with an answer.

manics: For Python we say
> Starting in 2023, the default version of Python used when Python version is unspecified will be updated more often. Python itself releases a new version every year now, and repo2docker will follow, with the default Python version generally trailing the latest stable version of Python itself by 1-2 versions.
>
> If you choose not to specify a Python version, your repository is guaranteed to stop working, eventually. We strongly recommend specifying a Python version (in environment.yml, runtime.txt, Pipfile, etc.)

https://repo2docker.readthedocs.io/en/latest/howto/languages/#supported-versions-of-python

I think we could do the same for R? If a repo is going to break due to a change in the unpinned R version it's better for it to happen sooner when there's a chance the author is still around. @sunu Maybe start a new issue so we're not blocking this PR?

PR #1571, opened 2026-09-15
RStudio 2026.09.0-174

### LINKED ISSUE #1570: Unable to start R Studio due lack of privilege
This was first reported to me by @yfiua.

### Bug description

At the time of writing, mybinder.org is unable to start R Studio. When attempting to start R Studio from the shell by calling

```bash
rstudio-server start
```

the following error is reported to the user

```
2026-09-09T06:08:44.292152Z [rserver] ERROR Attempt to run server as user 'rstudio-server' (uid 999) from account 'jovyan' (uid 1000) without privilege, which is required to run as a different uid; LOGGED FROM: virtual rstudio::core::ProgramStatus rstudio::server::Options::read(int, char* const*, std::ostream&) src/cpp/server/ServerOptions.cpp:341
```

#### Expected behaviour

R Studio to start without problems.

#### Actual behaviour

```
$ rstudio-server start
2026-09-09T06:08:43.180834Z [rserver] INFO Reading rserver configuration from '/etc/rstudio/rserver.conf' using default config dir
Warning: --test-config is deprecated and will be removed in a future release; use --check-config instead.
[PASS] Config file /etc/rstudio/rserver.conf: no unrecognized options found
[PASS] server-data-dir: /var/run/rstudio-server exists
[PASS] R installation: found R 4.4.0
2026-09-09T06:08:44.291268Z [rserver] INFO Reading rserver configuration from '/etc/rstudio/rserver.conf' using default config dir
2026-09-09T06:08:44.292152Z [rserver] ERROR Attempt to run server as user 'rstudio-server' (uid 999) from account 'jovyan' (uid 1000) without privilege, which is required to run as a different uid; LOGGED FROM: virtual rstudio::core::ProgramStatus rstudio::server::Options::read(int, char* const*, std::ostream&) src/cpp/server/ServerOptions.cpp:341
```

### How to reproduce

1. Visit https://mybinder.org
2. Fill the GitHub repository field with https://github.com/gesiscss/tool_valitext
3. Click the "launch" buttom

manics:
This hopefully closes https://github.com/jupyterhub/repo2docker/issues/1570

Note I tried upgrading jupyter-rsession-proxy to 2.5.0 but that led to a redirect loop so I've left it unchanged at 2.3.0 - will investigate separately and open an issue


tests/ui/browser/runtime.txt
```diff
@@ -0,0 +1 @@
+r-4.5.1-2025-06-13
```
manics, line 1: I've switched from environment.yaml to runtime.txt. Should we actually have two tests, since we can't test the UI in any other way?

rgaiacs (approved):
The changes looks good to me.

But I did **not** understand what was the cause of the problem. Was it something in R Studio?

manics:
Yes, I think the first upgrade to RStudio included this bug https://github.com/rstudio/rstudio/issues/18255 though I've no idea why it only occurred with some repos.

